PR #10, opened 2025-05-27
ci: Enable initial unit tests and remove auto-formatting in PRs

chtruong814:
Enable initial unit tests and remove auto-formatting in PRs

* Add unit test placeholder
* Update copy-pr-bot to require `/ok to test` to run the CI job
* Remove the auto-formatting in favor of just checking it at linting

### DIFF
--- a/.github/copy-pr-bot.yaml
+++ b/.github/copy-pr-bot.yaml
@@ -26,4 +26,4 @@ additional_vetters:
   - hemildesai
   - maanug-nv
 auto_sync_draft: false
-auto_sync_ready: false
+auto_sync_ready: true


.github/copy-pr-bot.yaml
```diff
@@ -26,4 +26,4 @@ additional_vetters:
   - hemildesai
   - maanug-nv
 auto_sync_draft: false
-auto_sync_ready: true
+auto_sync_ready: false
```
ko3n1g: Is this on purpose?

.github/workflows/cicd-main.yml
```diff
@@ -191,15 +188,13 @@ jobs:
...
-          HAS_LABEL: ${{ github.event.label.name == 'Run CICD' }}
-          IS_SCHEDULED: ${{ github.event_name == 'schedule' }}
         run: |
           # Get workflow run details and check job conclusions
           LATEST_ATTEMPT=$(gh run view $RUN_ID --json jobs -q '[.jobs[] | select(.conclusion != null) | .conclusion] | last')
           NUM_FAILED=$(gh run view $RUN_ID --json jobs -q '[.jobs[] | select(.conclusion == "failure") | .name] | length')
           NUM_CANCELLED=$(gh run view $RUN_ID --json jobs -q '[.jobs[] | select(.conclusion == "cancelled") | .name] | length')
 
-          if [[ $NUM_FAILED -eq 0 && $NUM_CANCELLED -eq 0 && ("$HAS_LABEL" == "true" || "$IS_SCHEDULED" == "true") ]]; then
+          if [[ $NUM_FAILED -eq 0 && $NUM_CANCELLED -eq 0 ]]; then
```
ko3n1g, line 197: Just a headsup this might not work currently. IIRC, `gh cli` doesn't have access to a repo with the anonymous `github.token` token

.github/copy-pr-bot.yaml
chtruong814: So I removed the `Run CICD` label need for now and I thought we were thinking better to use `/ok to test`?  Or do you think we should just let any commit start running the CI?  Open to either tbh.  Early enough that we can wait to see if it's a pain without it.

.github/workflows/cicd-main.yml
chtruong814: It would just silently fail?  We have the `PAT` in the repo as a secret.  Maybe we just use that?

PR #22, opened 2025-05-30
Remove `LEGACY` checkpoint support to simplify checkpointing code

ananthsub:
NeMo users should be fully on distributed checkpoints. If we need to bring back loading legacy checkpoints from MLM users, we can resurrect the codepath for it.


hemildesai:
Is this ready? Will be good to get this in for the release

ananthsub:
I'll clean this up for review today

maanug-nv:
nit: should we remove this fn ?: https://github.com/NVIDIA-NeMo/Megatron-Hub/blob/main/src/megatron/hub/utils/common_utils.py#L129-L138

PR #23, opened 2025-05-30
Move to uv and pyproject.toml

hemildesai:
Fixes https://github.com/NVIDIA-NeMo/NeMo-LM/issues/13

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,15 @@ dependencies = [
     "transformers>=4.51.3",
     "wandb>=0.19.10",
     "megatron-core",
+    "six>=1.17.0",
+    "regex>=2024.11.6",
+    "pyyaml>=6.0.2",
+    "einops>=0.8.1",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -59,6 +55,25 @@ classifiers = [
...
+    "packaging>=25.0",
+    "tensorboard>=2.19.0",
+    "torch>=2.7.0",
+    "transformers>=4.51.3",
+    "wandb>=0.19.10",
+    "megatron-core",
+]
+
+[tool.uv.sources]
+megatron-core = { git = "https://github.com/NVIDIA/Megatron-LM.git" }
```
ko3n1g, line 81: Do we need this for NeMo-Run? Isn't the pypi wheel also good?

hemildesai: Nope, the pypi wheel is sufficient for NeMo-Run.

PR #30, opened 2025-06-02
Add Megatron comm overlap and mixed precision config

hemildesai:
Fixes https://github.com/NVIDIA-NeMo/NeMo-LM/issues/1

### DIFF
--- a/tests/unit_tests/training/test_config.py
+++ b/tests/unit_tests/training/test_config.py
@@ -63,6 +63,7 @@ def create_test_gpt_config(**kwargs: Any) -> GPTConfig:
         "hidden_size": 128,
         "num_attention_heads": 4,
         "seq_length": 512,
+        "apply_rope_fusion": False,
     }
     defaults.update(kwargs)
     return GPTConfig(**defaults)
@@ -75,6 +76,7 @@ def create_test_t5_config(**kwargs: Any) -> T5Config:
         "hidden_size": 128,
         "num_attention_heads": 4,
         "seq_length": 512,
+        "apply_rope_fusion": False,
     }
     defaults.update(kwargs)
     return T5Config(**defaults)


nemo_lm/training/comm_overlap.py
```diff
@@ -0,0 +1,551 @@
...
+    qkv_fprop=RingExchangeOverlapCfg(aggregate=False),
+    proj_dgrad=RingExchangeOverlapCfg(aggregate=False),
+    fc1_fprop=RingExchangeOverlapCfg(aggregate=False),
+    fc2_dgrad=RingExchangeOverlapCfg(aggregate=False),
+    proj_fprop=RingExchangeOverlapCfg(num_sm=1, set_sm_margin=True),
+    fc2_fprop=RingExchangeOverlapCfg(num_sm=1, set_sm_margin=True),
+)
+
+# llama3 70b LoRA
+userbuffers_fp8_h100_h8192_tp2_mbs1_seqlen4096_lora = TransformerLayerTPOverlapCfg(
```
ananthsub, line 201: is it preferable for these to sit alongside the model configs?

hemildesai: I feel like its better to have it in a centralized place so that the configs can be shared across models more easily. Wdyt?

nemo_lm/training/comm_overlap.py
```diff
@@ -0,0 +1,551 @@
...
+    overlap_param_gather_with_optimizer_step: bool = None
+    align_param_gather: bool = None
+    bucket_size: int = None
+    # Pipeline bubble overlap
+    defer_embedding_wgrad_compute: bool = None
+    wgrad_deferral_limit: int = None
+
+
+@dataclass(kw_only=True)
+class MegatronCommOverlapConfig:
```
ananthsub: could you add unit tests for this config?

nemo_lm/training/mixed_precision.py
```diff
@@ -0,0 +1,182 @@
...
+    initial_loss_scale: float = (None,)
+    min_loss_scale: float = (None,)
+    loss_scale_window: float = (None,)
+    hysteresis: float = (None,)
+    num_layers_at_start_in_bf16: int = 0
+    num_layers_at_end_in_bf16: int = 0
+
+
+@dataclass(kw_only=True)
+class MegatronMixedPrecisionConfig:
```
ananthsub: is this not redundant with the dtype config? why do we need two classes for this?

also why does this need to be included in the config container? my concern is that if a user imports a config container via a recipe as in https://github.com/NVIDIA-NeMo/NeMo-LM/pull/31, then mutates these values, then we need to explicitly synchronize the fields again

nemo_lm/training/mixed_precision.py
```diff
@@ -0,0 +1,182 @@
...
+
+    Args:
+        dtype_config: Source of dtype settings
+        config: Config object to update
+
+    Returns:
+        Updated config object
+    """
+    if hasattr(config, "__io__"):
+        config.__io__ = update_config_with_dtype_overrides(dtype_config, config.__io__)
```
ananthsub: do we still need to check for this?

nemo_lm/training/mixed_precision.py
hemildesai: > is this not redundant with the dtype config? why do we need two classes for this?

Updated in https://github.com/NVIDIA-NeMo/NeMo-LM/pull/30/commits/e203b6210394a918f709480b5a2c82c3fbbdd5ad

> then we need to explicitly synchronize the fields again

This is in the `validate` method instead of `__post_init__` in ConfigContainer which is only called inside `megatron_pretrain`

nemo_lm/training/mixed_precision.py
hemildesai: Removed in https://github.com/NVIDIA-NeMo/NeMo-LM/pull/30/commits/e203b6210394a918f709480b5a2c82c3fbbdd5ad

nemo_lm/training/mixed_precision.py
hemildesai: I could also move it to a separate `apply_overlaps` and `apply_mixed_precision` methods to make it more explicit.

nemo_lm/training/comm_overlap.py
```diff
@@ -0,0 +1,551 @@
...
+                    os.environ.pop("CUDA_DEVICE_MAX_CONNECTIONS")
+                logging.info("Unset CUDA_DEVICE_MAX_CONNECTIONS")
+
+    def setup(
+        self,
+        model_config: GPTConfig | T5Config,
+        optimizer_config: OptimizerConfig,
+        ddp_config: DistributedDataParallelConfig,
+    ) -> None:
+        """Apply configs set in comm_overlap_cfg on trainer config."""
```
ananthsub: nit: docstring mentions trainer config but this is likely carried over from nemo2

nemo_lm/training/mixed_precision.py
ananthsub: i'm in favor of the separate functions for applying the optimizations. this way there's only one setting in the corresponding configs that these map to. we can provide the performance optimized functions in the recipes similar to https://github.com/NVIDIA/NeMo/blob/97b9fed8e0f1b1613040c329b159ea865c537c81/nemo/collections/llm/recipes/llama3_8b.py#L202

hemildesai: Updated in https://github.com/NVIDIA-NeMo/NeMo-LM/pull/30/commits/38157a6051083de54efe3b8bc2740b30d8a85865

nemo_lm/training/comm_overlap.py
hemildesai: Removed for now.

nemo_lm/training/comm_overlap.py
hemildesai: Done.

nemo_lm/training/comm_overlap.py
```diff
@@ -0,0 +1,550 @@
...
+    overlap_param_gather_with_optimizer_step: bool = None
+    align_param_gather: bool = None
+    bucket_size: int = None
+    # Pipeline bubble overlap
+    defer_embedding_wgrad_compute: bool = None
+    wgrad_deferral_limit: int = None
+
+
+@dataclass(kw_only=True)
+class MegatronCommOverlapConfig:
```
ananthsub: supernit: is the `Megatron` prefix here and for mixed precision still needed? is the motivation to keep the name to help with migration?

hemildesai: Yeah I can remove the `Megatron` prefix. Shouldn't affect migration.

nemo_lm/training/config.py
```diff
@@ -721,3 +721,9 @@ def validate(self) -> None:
             self.scheduler.lr_warmup_steps = self.scheduler.lr_warmup_fraction * self.scheduler.lr_decay_iters
         else:
             self.scheduler.lr_warmup_steps = self.scheduler.lr_warmup_iters * self.train.global_batch_size
+
+        # Profiling
+        if self.profiling is not None:
+            assert self.profiling.use_pytorch_profiler != self.profiling.use_nsys_profiler, (
+                "Exactly one of pytorch or nsys profiler should be enabled, not both or neither"
+            )
```
ananthsub: i'd removed this assumption in https://github.com/NVIDIA-NeMo/NeMo-LM/pull/19
this way a user can have the profiling config instantiated, but choose to quickly toggle either the nsys or pytorch profiler with fewer overrides

PR #31, opened 2025-06-02
Add llama3 8b recipe and example pretrain script with yaml overrides

ananthsub:
Run example script to sanity test that config merges are applied as expected
```
python examples/llm/llama3_8b/pretrain_llama3_8b_nemo_run_script.py train.train_iters=100
```

### DIFF
--- a/nemo_lm/training/pretrain.py
+++ b/nemo_lm/training/pretrain.py
@@ -47,7 +47,7 @@ def megatron_pretrain(
     """
     config.validate()
 
-    dataset_provider = get_dataset_provider(config.dataset_config)
+    dataset_provider = get_dataset_provider(config.dataset)
 
     setup_output = setup(config, dataset_provider)
     state = setup_output.state
@@ -75,7 +75,7 @@ def megatron_pretrain(
             )
 
         barrier_and_log("after training is done")
-        ckpt_config = config.checkpoint_config
+        ckpt_config = config.checkpoint
         if ckpt_config.save and state.train_state.step != 0 and ckpt_config.save_interval != 0:
             save_checkpoint(
                 state,


examples/llm/llama3_8b/pretrain_llama3_8b.py
```diff
@@ -0,0 +1,106 @@
...
+    args, cli_overrides = parse_cli_args()
+
+    logger.info("Nemo-LM Llama3 8B Pretraining Script with YAML & CLI Overrides")
+    logger.info("------------------------------------------------------------------")
+
+    # Load base configuration from the recipe as a Python dataclass
+    cfg: ConfigContainer = pretrain_config()
+
+    # Convert the initial Python dataclass to an OmegaConf DictConfig for merging
+    merged_omega_conf, excluded_callables = safe_create_omegaconf_with_preservation(cfg)
```
hemildesai: Can't you use `cfg.to_dict()` here?

nemo_lm/recipes/llm/llama3_8b.py
```diff
@@ -0,0 +1,192 @@
...
+        ),
+        ddp_config=DistributedDataParallelConfig(
+            check_for_nan_in_grad=True,
+            grad_reduce_in_fp32=True,
+            overlap_grad_reduce=True,
+            overlap_param_gather=True,
+            average_in_collective=True,
+            use_distributed_optimizer=True,
+        ),
+        dataset_config=GPTDatasetConfig(
```
hemildesai: Can you use something like
```python
# Dataset config
if mock or data_paths is None:
    mock = True
    blend = [1.0]
    blend_per_split = [[1.0], [1.0], [1.0]]
    split = "1,1,1"
else:
    mock = False
    blend, blend_per_split = get_blend_and_blend_per_split(data_paths=data_paths)
    split = "9999,8,2"
```

PR #32, opened 2025-06-03
Clean up requirements

maanug-nv:
- Remove dependencies this repo no longer uses
- Add new dependencies
- Corresponding changes to `setup.py`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ install_requires = req_file("requirements.txt")
 extras_require = {
     # User packages
     'test': req_file("requirements_test.txt"),
-    'docs': req_file("requirements_docs.txt")
+    'docs': req_file("requirements_docs.txt"),
 }
 
 


requirements/requirements_test.txt
```diff
@@ -1,17 +1,6 @@
 black~=24.3
-click>=8.1
 coverage
-flake8
 isort>5.1.0,<6.0.0
-parameterized
-pylint
```
chtruong814, line 7: The lint commands are failing because we're removing pylint and flake8 here.  If we want to move to ruff, probably best to swap out flake8, pylint, and black at the same time of adding ruff.  Or did you not intend to remove flake8 and pylint?

maanug-nv: i didn't realize CI will install from this file. I will rever removal of pylint and flake8

maanug-nv: i can move to ruff in a future PR

maanug-nv:
moving changes to #23

PR #35, opened 2025-06-03
Dependency update

maanug-nv:
- Update deps and add dep groups
- Use uv to install deps in CI


.github/workflows/code-linting.yml
```diff
@@ -25,9 +25,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: "0.7.10"
```
chtruong814, line 31: I'm ok to do this in a case like this.

PR #43, opened 2025-06-04
[peft] Port walking utils for peft integration

ananthsub:
This is primarily for peft integration. This ports https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/fn/base.py and associated tests over to nemo lm

### DIFF
--- a/nemo_lm/peft/walk_utils.py
+++ b/nemo_lm/peft/walk_utils.py
@@ -25,7 +25,7 @@ Examples:
         >>> def add_tag(module, name=None, **kwargs):
         ...     module.tag = f"transformed_{name}"
         ...     return module
-        >>> 
+        >>>
         >>> model = nn.Sequential(nn.Linear(10, 5), nn.ReLU())
         >>> transformed = walk(model, add_tag)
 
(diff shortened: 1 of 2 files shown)


nemo_lm/utils/walking_utils.py
hemildesai, line 1: Can you rename this `peft_utils.py`?

ananthsub: I'd added this outside of peft since they're general functions. https://github.com/NVIDIA-NeMo/NeMo-LM/pull/44 adds some of the peft-specific layers/utils, so i've created a new directory to contain them. wdyt?

hemildesai: Is there any utility rn for this except for peft?

ananthsub: not that I can think of. the FNMixin is included on LightningModules in NeMo, but AFAICT these utils would only be used for peft right now

hemildesai: Then I would be in favor of keeping them in the peft folder or renaming to `peft_utils` for now. We can move to a common place/name if it gets used somewhere else too.

PR #44, opened 2025-06-04
[peft] Port base adapter wrapper and lora utils

ananthsub:
Port AdapterWrapper from https://github.com/NVIDIA/NeMo/blob/0af8b7df793ad7538f149ad9bcb8c2cae5134c1a/nemo/lightning/pytorch/callbacks/peft.py#L330-L444

LoRA layers from here: https://github.com/NVIDIA/NeMo/blob/c0fd982b5e6e7d5b296a940a91b52534404262c1/nemo/collections/llm/peft/lora.py#L1-L355

utils from https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/peft/utils.py

within this new repo, `ParallelLinearAdapter` no longer depends on `AdapterModuleUtil `in NeMo1. instead of getting the activation registry from `AdapterModuleUtil`, the registry is added to this class directly

### DIFF
--- a/tests/unit_tests/peft/test_utils.py
+++ b/tests/unit_tests/peft/test_utils.py
@@ -240,7 +240,7 @@ class TestUtilityFunctions:
         assert torch.equal(unpadded, original)
 
 
-@patch("nemo_lm.peft.utils.parallel_state")
+@patch("megatron.hub.peft.utils.parallel_state")
 class TestAll2AllCommunication:
     """Test All2All communication functions."""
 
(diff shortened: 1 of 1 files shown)


maanug-nv:
do we need `adapter_unfreeze()` in the ParallelLinearAdapter?

ananthsub:
> do we need `adapter_unfreeze()` in the ParallelLinearAdapter?

AFAICT, `unfreeze_enabled_adapters()` and `unfreeze_enabled_adapters()` from `AdapterModuleUtil` are not used on the nemo2 path

nemo_lm/peft/utils.py
```diff
@@ -0,0 +1,623 @@
...
+            out_features,
+            config=model_parallel_config,
+            bias=False,
+            gather_output=lin_out_gather_output,
+            init_method=self._get_init_fn(row_init_method),
+        )
+
+        if dropout > 0.0:
+            if dropout_recompute:
+                import thunder
```
maanug-nv: Should this be added as a dependency?

ananthsub: i think this was recently added for the automodel path. i'll remove this to avoid adding another dependency. if it's relevant for mcore, we can add this back

PR #56, opened 2025-06-09
Example of NeMo Run Partial recipe integration 

ananthsub:
This enables the existing developer path of doing config overrides in Python and launching with NeMo Run. Users would update the ConfigContainer dataclass directly, then call the `get_partial_fn`, which will internally handle serialization gaps and return a run.Partial object that can be launched with nemo run executors. 

This complements the existing example which runs with nemo.run Script mode using omegaconf overrides.

Testing:
1. Added unit tests for conversion of ConfigContainer to format that can be launched with nemo-run
```
pytest tests/unit_tests/recipes/utils/test_nemo_run_utils.py
```

2. Run the example in dryrun mode
```
 python examples/llm/llama3_8b/pretrain_llama3_8b_nemo_run_partial.py --dryrun
```

Lockfile updated after adding `nemo_run` as an optional dependency for recipes:
```
 uv add --optional recipes nemo_run
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,7 +106,7 @@ eval = [
     "pytest-httpserver"
 ]
 recipes = [
-    "nemo_run",
+    "nemo-run",
 ]
 
 [dependency-groups]
--- a/uv.lock
+++ b/uv.lock
@@ -379,6 +379,64 @@ wheels = [
 ]
 
 [[package]]
+name = "bcrypt"
+version = "4.3.0"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/bb/5d/6d7433e0f3cd46ce0b43cd65e1db465ea024dbb8216fb2404e919c2ad77b/bcrypt-4.3.0.tar.gz", hash = "sha256:3a3fd2204178b6d2adcf09cb4f6426ffef54762577a7c9b54c159008cb288c18", size = 25697, upload-time = "2025-02-28T01:24:09.174Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/bf/2c/3d44e853d1fe969d229bd58d39ae6902b3d924af0e2b5a60d17d4b809ded/bcrypt-4.3.0-cp313-cp313t-macosx_10_12_universal2.whl", hash = "sha256:f01e060f14b6b57bbb72fc5b4a83ac21c443c9a2ee708e04a10e9192f90a6281", size = 483719, upload-time = "2025-02-28T01:22:34.539Z" },
+    { url = "https://files.pythonhosted.org/packages/a1/e2/58ff6e2a22eca2e2cff5370ae56dba29d70b1ea6fc08ee9115c3ae367795/bcrypt-4.3.0-cp313-cp313t-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:c5eeac541cefd0bb887a371ef73c62c3cd78535e4887b310626036a7c0a817bb", size = 272001, upload-time = "2025-02-28T01:22:38.078Z" },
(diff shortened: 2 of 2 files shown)


nemo_lm/recipes/utils/nemo_run_utils.py
```diff
@@ -0,0 +1,146 @@
...
+import nemo_run as run
+
+from nemo_lm.training.config import ConfigContainer
+from nemo_lm.training.pretrain import megatron_pretrain
+
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+def get_pretrain_fn(config: ConfigContainer, forward_step_func: Callable) -> run.Partial:
```
hemildesai: Can we rename this to get_partial_fn and take the target as an arg?

nemo_lm/training/initialize.py
```diff
@@ -383,23 +380,6 @@ def _initialize_distributed(
                 )
 
 
-def _compile_dataset_helpers() -> None:
```
hemildesai: Why are we removing this?

nemo_lm/recipes/utils/nemo_run_utils.py
ananthsub: both pretrain/finetune will take the exact same args, so it shouldn't be a concern. however i'm not sure if later we might add other functions which accept different arguments, which would complicate which arguments to pass here

nemo_lm/training/initialize.py
ananthsub: mcore should compile the helpers [during install](https://github.com/NVIDIA/Megatron-LM/blob/main/setup.py#L108-L120), so is there a need to repeat it here? i can split this out to its own PR though

hemildesai: This is still helpful to have. There may be cases where user mounts a git clone of Megatron-LM in the container image when on the fly compilation may be needed.

nemo_lm/recipes/utils/nemo_run_utils.py
hemildesai: Let's tackle that when we add more frontend API functions.

pyproject.toml
```diff
@@ -76,10 +76,12 @@ dependencies = [
     "nvidia-modelopt[torch]>=0.25.0,<=0.29.0 ; sys_platform != 'darwin'",
     "nvidia-resiliency-ext>=0.3.0,<1.0.0 ; sys_platform != 'darwin'",
     "hydra-core>1.3,<=1.3.2",
+    "nemo_run",
```
maanug-nv: can this be put under 'dev' ? iirc we decided nemo run should not be explicitly required

pyproject.toml
```diff
@@ -107,6 +108,31 @@ dev = [
     "ruff>=0.9.9",
     "mypy>=1.8.0",
 ]
+recipes = [
+    "nemo_run",
+]
+all = [
```
hemildesai: I don't think we need all, theres uv option to select all groups

pyproject.toml
```diff
@@ -107,6 +108,31 @@ dev = [
     "ruff>=0.9.9",
     "mypy>=1.8.0",
 ]
+recipes = [
```
hemildesai: Can you add this to optional dependencies instead of dependency group?

docker/Dockerfile.ci
```diff
@@ -24,12 +24,18 @@ COPY src /workspace/src
...
 RUN pip install --no-cache-dir --upgrade --force-reinstall --no-deps \
-    git+https://github.com/NVIDIA/Megatron-LM.git@${MCORE_COMMIT} \
-    git+https://github.com/NVIDIA/NeMo-RUN.git@${NEMO_RUN_COMMIT}
+    git+https://github.com/NVIDIA/Megatron-LM.git@${MCORE_COMMIT}
+
+RUN pip install --no-cache-dir --upgrade --force-reinstall \
+    git+https://github.com/NVIDIA/NeMo-Run.git@${NEMO_RUN_COMMIT}
+
+# Install optional dependencies
+RUN pip install --no-cache-dir -e .[recipes,test,dev,docs]
```
ananthsub: @hemildesai @chtruong814 could you take a look at this for handling nemo run as an optional dependency for testing?

hemildesai: I think this will change with https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/72 but we can go ahead for now

hemildesai: Missing end of file

PR #62, opened 2025-06-10
Rename `nemo_lm` to set up namespace package `megatron.hub`

ananthsub:
- Set up as namespace package for `megatron.hub` , using https://github.com/NVIDIA/Megatron-Energon as an example
- Update `pyproject.toml` for new name/setup
- Update src/, examples/ tests/, docs/


maanug-nv:
Please make sure to update name in README as well

PR #64, opened 2025-06-11
Cleanup NeMo-LM references

ananthsub:
Update references now that repo is renamed to megatron-hub


maanug-nv:
can we have description in pyproject.toml match README title?

PR #65, opened 2025-06-11
Add nemo run plugins

### LINKED ISSUE #28: Plugins support for nemo run launching
Compatibility with https://github.com/NVIDIA/NeMo/blob/main/nemo/lightning/run/plugins.py

hemildesai:
Closes #28

### DIFF
--- a/src/megatron/hub/recipes/run_plugins.py
+++ b/src/megatron/hub/recipes/run_plugins.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import os
 from dataclasses import dataclass
 from typing import Optional
@@ -23,6 +24,8 @@ from megatron.hub.training.config import (
     ProfilingConfig,
 )
 
+
+logger: logging.Logger = logging.getLogger(__name__)
 
 # This file contains plugins based on NeMo-Run's run.Plugin API.
 # Plugins operate both on a configured task and an executor at the same time, and are specific to NeMo-Run.
@@ -56,10 +59,10 @@ class PreemptionPlugin(run.Plugin):
                 task.args.append(
                     f"train.exit_signal_handler_for_dataloader={str(self.enable_exit_handler_for_data_loader)}"
                 )
-                print(
+                logger.info(
                     f"{self.__class__.__name__} added CLI override: train.exit_signal_handler={str(self.enable_exit_handler)}"
                 )
-                print(
+                logger.info(
                     f"{self.__class__.__name__} added CLI override: train.exit_signal_handler_for_dataloader={str(self.enable_exit_handler_for_data_loader)}"
                 )
         else:
@@ -71,7 +74,7 @@ class PreemptionPlugin(run.Plugin):
         # Apply signal configuration for both task types when using SlurmExecutor
         if isinstance(executor, run.SlurmExecutor):
             # Sends a SIGTERM self.preempt_time seconds before hitting time limit
-            print(
+            logger.info(
                 f"{self.__class__.__name__} will send a SIGTERM {self.preempt_time} seconds before the job's time limit for your Slurm executor."
             )
             executor.signal = f"TERM@{self.preempt_time}"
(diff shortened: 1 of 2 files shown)


src/megatron/hub/recipes/run_plugins.py
```diff
@@ -0,0 +1,395 @@
...
+        """Set up the preemption plugin."""
+        if isinstance(task, run.Script):
+            # For run.Script, append CLI overrides to the script arguments
+            if self.enable_exit_handler:
+                task.args.append("train.exit_signal_handler=true")
+                print(f"{self.__class__.__name__} added CLI override: train.exit_signal_handler=true")
+        else:
+            # Enable exit signal handler in training config
+            if self.enable_exit_handler and hasattr(task, "train"):
+                task.train.exit_signal_handler = True
```
ananthsub: should the plugin also have options to set these options? 
https://github.com/NVIDIA-NeMo/Megatron-Hub/blob/c1934e9a0660fd753373caa2104e329380117c8b/src/megatron/hub/training/config.py#L326-L330

src/megatron/hub/recipes/run_plugins.py
```diff
@@ -0,0 +1,395 @@
...
+            if not hasattr(task, "ft"):
+                task.ft = FaultToleranceConfig()
+
+            task.ft.enable_ft_package = self.enable_ft_package
+            task.ft.calc_ft_timeouts = self.calc_ft_timeouts
+
+            # Check if nsys profiling is enabled and warn if so
+            if hasattr(task, "profiling") and task.profiling and task.profiling.use_nsys_profiler:
+                print("Warning: Nsys not supported with the FaultTolerancePlugin.")
+                task.profiling.use_nsys_profiler = False
```
ananthsub: should this check also be added to the configcontainer validation, since we can't catch this for the script case? it'd also be good to add this restriction to the docstring

src/megatron/hub/recipes/run_plugins.py
```diff
@@ -0,0 +1,395 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from dataclasses import dataclass
+from typing import Optional
+
+import nemo_run as run
```
ananthsub, line 20: does this depend on https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/56 for handling nemo_run as an optional dependency for recipes?

src/megatron/hub/recipes/run_plugins.py
```diff
@@ -0,0 +1,395 @@
...
+    def setup(self, task: run.Partial | run.Script, executor: run.Executor):
+        """Set up the preemption plugin."""
+        if isinstance(task, run.Script):
+            # For run.Script, append CLI overrides to the script arguments
+            if self.enable_exit_handler:
+                task.args.append("train.exit_signal_handler=true")
+                print(f"{self.__class__.__name__} added CLI override: train.exit_signal_handler=true")
+        else:
+            # Enable exit signal handler in training config
+            if self.enable_exit_handler and hasattr(task, "train"):
```
ananthsub: is the `task` the `ConfigContainer` ? from https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/56 the `megatron_pretrain` is what's wrapped with run.Partial, so does this need to access `task.config` first to get to the ConfigContainer?

hemildesai:
>How did you want to test these plugins in CI? AFAICT testing the nemo run plugins in nemo resorted to using end to end tests, but since these plugins override or specify overrides to the script or set env vars, do we need full e2e tests for these?

I think unit tests for individual plugins and end to end tests in the CI with the plugins enabled are both required.

src/megatron/hub/recipes/run_plugins.py
```diff
@@ -0,0 +1,395 @@
...
+            'nvtx' and 'cuda' events will be traced.
+        record_shapes (bool): Whether to record tensor shapes. Default is False.
+    """
+
+    profile_step_start: int
+    profile_step_end: int
+    profile_ranks: Optional[list[int]] = None
+    nsys_trace: Optional[list[str]] = None
+    record_shapes: bool = False
+
```
ananthsub, line 166: this will now need the gpu metrics option from https://github.com/NVIDIA/NeMo/pull/13882

src/megatron/hub/recipes/run_plugins.py
hemildesai: Yes, but let's add that in a separate PR.

src/megatron/hub/recipes/run_plugins.py
```diff
@@ -0,0 +1,417 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import os
+from dataclasses import dataclass
+from typing import Optional
+
+import nemo_run as run
```
ananthsub, line 20: since nemo_run is an optional dependency for recipes, does this need a conditional import?

PR #67, opened 2025-06-11
Add NVRx straggler detector

ananthsub:
Fixes https://github.com/NVIDIA-NeMo/Megatron-Hub/issues/25

- Add manager class to replicate logic in Lightning Callback
- Add configuration for ConfigContainer to enable using this class
- Update train loop to call detector
- Add unit and e2e tests with detector enabled

### DIFF
--- a/src/megatron/hub/training/train.py
+++ b/src/megatron/hub/training/train.py
@@ -381,21 +381,6 @@ def train(
             should_toggle_forward_pre_hook,
         )
 
-        if should_exit:
-            # Straggler detected - save checkpoint before exiting
-            print_rank_0("Exiting due to straggler detection. Saving checkpoint...")
-            if config.checkpoint.save is not None:
-                save_checkpoint_and_time(
-                    global_state,
-                    model,
-                    optimizer,
-                    scheduler,
-                    num_floating_point_operations_so_far,
-                    checkpointing_context,
-                    train_data_iterator=train_data_iterator,
-                )
-            break
-
         # Checkpoint and decide whether to exit.
         should_exit = checkpoint_and_decide_exit(
             global_state,


src/megatron/hub/training/config.py
```diff
@@ -670,6 +724,7 @@ class ConfigContainer(Container):
     dist: DistributedInitConfig = field(default_factory=DistributedInitConfig)
     ft: Optional[FaultToleranceConfig] = None
     straggler: Optional[StragglerDetectionConfig] = None
+    nvrx_straggler: Optional[NVRxStragglerDetectionConfig] = None
```
hemildesai, line 732: It seems confusing to have two straggler configs? Can we unify them and add an option like `source="nvrx"`

ananthsub: i agree, it's confusing to have two like this. 

i didn't combine them primarily to avoid confusion over which settings are needed for the existing mcore straggler vs the nvrx one (otherwise needing to prefix all fields for the nvrx straggler with `nvrx_`), and also, if for some reason, someone wanted to use both simultaneously (maybe to compare mcore vs nvrx straggler detectors within the same run?)

hemildesai: Makes sense.

src/megatron/hub/training/nvrx_straggler.py
```diff
@@ -0,0 +1,322 @@
...
+            return train_step_func
+
+        if self.wrapped_function is not None:
+            self.logger.warning("Train step function already wrapped. Skipping.")
+            return train_step_func
+
+        try:
+            # Create a wrapper object with train_step method for nvidia-resiliency-ext
+            # TODO: See if NVRx can support functions directly without needing them attached to a class
+            class TrainStepWrapper:
```
maanug-nv, line 96: Maybe I'm not understanding how this wrapper works. But would this work with the RerunStateMachine/SDC rerun functionality? Or would they interfere with one another?

src/megatron/hub/training/train.py
```diff
@@ -366,6 +381,23 @@ def train(
             should_toggle_forward_pre_hook,
         )
 
+        # Check NVRx straggler detection
+        should_exit = check_nvrx_straggler_detection(global_state.nvrx_straggler_manager)
+        if should_exit:
```
maanug-nv: I think it would be more consistent to move this check to `checkpoint_and_decide_exit()` . What do you think?

maanug-nv:
Just sharing my thoughts, this would of course be a large ask that we likely don't have b/w for right now. I don't think it makes sense to maintain two modules that serve the same purpose (MCore StragglerDetector and NVRx StragglerDetector). Obviously, it's not solely our decision what to do about this, but if it was, I would suggest merging any functionality from MCore's version that isn't present in NVRx into NVRx's Straggler Detector, and deprecating MCore's.

ananthsub:
> Just sharing my thoughts, this would of course be a large ask that we likely don't have b/w for right now. I don't think it makes sense to maintain two modules that serve the same purpose (MCore StragglerDetector and NVRx StragglerDetector). Obviously, it's not solely our decision what to do about this, but if it was, I would suggest merging any functionality from MCore's version that isn't present in NVRx into NVRx's Straggler Detector, and deprecating MCore's.

A challenge for merging is that MLM will treat NVRx as an optional dependency. So the MCore straggler detector will likely still exist with functionality independent or replicated from NVRx. But let's raise this as a discussion point for mcore after

src/megatron/hub/training/train.py
ananthsub: that makes sense to me

src/megatron/hub/training/train.py
```diff
@@ -366,6 +381,21 @@ def train(
             should_toggle_forward_pre_hook,
         )
 
+        if should_exit:
+            # Straggler detected - save checkpoint before exiting
```
maanug-nv: shouldn't this be removed now that the check is in `checkpoint_and_decide_exit()`

tests/functional_tests/test_nvrx_straggler.py
```diff
@@ -0,0 +1,326 @@
...
+                        log_content = f.read()
+
+                # Check for NVRx straggler detection activity
+                has_gpu_perf_logs = "GPU relative performance" in log_content
+                has_rank_scores = any(keyword in log_content for keyword in ["Rank=", "Score="])
+                has_straggler_detection = "straggler" in log_content.lower()
+                has_nvidia_resiliency = "nvidia_resiliency" in log_content.lower()
+
+                # Assert that NVRx is actually working
+                assert has_gpu_perf_logs or has_rank_scores or has_straggler_detection or has_nvidia_resiliency, (
```
maanug-nv, line 285: this is great btw, thanks for checking the logs

PR #70, opened 2025-06-11
chore: Update codeowners

ko3n1g:


.github/CODEOWNERS
```diff
@@ -1,7 +1,7 @@
...
-/setup.py @ko3n1g @chtruong814 @thomasdhc @pablo-garay
-pyproject.toml @ko3n1g @chtruong814 @thomasdhc @pablo-garay
-requirements/ @ko3n1g @chtruong814 @thomasdhc @pablo-garay
+.github/ @nvidia-nemo/automation
+docker/ @nvidia-nemo/automation
+.pylintrc.* @nvidia-nemo/automation
+.flake8.* @nvidia-nemo/automation
+/setup.py @nvidia-nemo/automation
+pyproject.toml @nvidia-nemo/automation
+requirements/ @nvidia-nemo/automation
```
ananthsub: `requirements/` doesn't exist 


do you want to add automation as a codeowner for `ruff.toml` and https://github.com/NVIDIA-NeMo/Megatron-Hub/blob/ddfbc090e8e4cf4f2876a2919ed14fdcdb570930/src/megatron/hub/package_info.py too ?

ko3n1g: Thanks for catching that. 

Is there a preference for the `ruff.toml` file? We could also define that within `pyproject.toml`.

cc @hemildesai

PR #72, opened 2025-06-12
ci: Switch to UV

ko3n1g:
I'm adding some dependencies from old Nemo because export-deploy requires these. Please let me know if we should undo this. 

I started to disconnect export-deploy from old Nemo's dependencies and moved required dependencies to Megatron-Hub.

### DIFF
--- a/docker/manifest.json
+++ b/docker/manifest.json
@@ -2,7 +2,7 @@
     "vcs-dependencies": {
         "megatron-lm": {
             "repo": "https://github.com/NVIDIA/Megatron-LM.git",
-            "ref": "ad36348f397451dbc4a4b4adb5e40055c89fec2f"
+            "ref": "9eebd5173cbd6dea5e6a324d418b669ac000989e"
         },
         "export-deploy": {
             "repo": "https://github.com/NVIDIA-NeMo/NeMo-Export-Deploy.git",


.github/actions/test-template/action.yml
```diff
@@ -173,7 +173,7 @@ runs:
 
         echo "$cmd" | tee "retry_job.sh"
         bash retry_job.sh
-        docker exec -t nemo_container_${{ github.run_id }} bash -c 'pip install -e .'
+        docker exec -t nemo_container_${{ github.run_id }} bash -c 'uv pip install --no-deps -e .'
```
thomasdhc, line 176: Why did you have to update to no-deps here?

docker/Dockerfile.ci
```diff
@@ -15,34 +15,41 @@ FROM nvcr.io/nvidia/pytorch:25.05-py3
...
+ENV UV_PROJECT_ENVIRONMENT=/opt/venv
+ENV PATH="$UV_PROJECT_ENVIRONMENT/bin:$PATH"
+ENV UV_LINK_MODE=copy
+
 # Install all
 WORKDIR /workspace
 
+RUN bash -ex <<"EOF"
+    uv venv ${UV_PROJECT_ENVIRONMENT} --system-site-packages
+    source ${UV_PROJECT_ENVIRONMENT}/bin/activate
```
thomasdhc, line 30: Does it need to be activated? Anytime we install we'll use uv command right?

.github/actions/test-template/action.yml
ko3n1g: Our installation in the dockerfile is: first project dependencies, then overrides.

Later when we clone and add the code, we do an editable install to make sure PYTHONPATHs are set. But if we were to install with project dependencies, we might loose our overrides again.

docker/Dockerfile.ci
ko3n1g: good point, maybe line 30 is actually not necessary. All command that go through `uv` should respect `UV_PROJECT_ENVIRONMENT`'s venv. Only if we were to use directly python or pip without uv, we'd want to source the environment before to make sure we use the right binary

.github/actions/test-template/action.yml
thomasdhc: Gotcha

pyproject.toml
```diff
@@ -76,12 +76,28 @@ dependencies = [
...
     "hydra-core>1.3,<=1.3.2",
+    "wget",
+    "h5py",
+    "ijson",
+    "boto3",
+    "megatron-energon",
+    "qwen_vl_utils",
+    "sacrebleu",
+    "rouge_score",
+    "decord"
```
chtruong814: Why do we need these additional dependencies?

PR #76, opened 2025-06-13
Port over perf tuning doc

maanug-nv:
Copy [performance tuning guide](https://docs.nvidia.com/nemo-framework/user-guide/latest/performance/performance-guide.html#) from nemofw user guide. 

Todo:

- [x] replace all mentions of 'NeMo'/'NeMo Framework' with Megatron ?
- [x] update pointers to config options (e.g.  `recipe.trainer.use_distributed_optimizer=true`)

### DIFF
--- a/docs/performance-guide.md
+++ b/docs/performance-guide.md
@@ -2,7 +2,13 @@
 
 Megatron-Hub provides a wide range of features for performant and memory-efficient LLM training on GPUs, and comes pre-configured with optimal settings. However, factors such as model architecture, hyperparameters, GPU count, and GPU type can affect the available options, and additional tuning may be necessary to achieve optimal performance. This document explores the factors that affect training performance, highlights common issues, and outlines techniques for performance tuning that lead to higher MFU (Model FLOPS Utilization) and TCO.
 
+```{Note}
 This guide makes references to several configuration settings. These settings will be referenced relative to the the config class that contains them, e.g. `OptimizerConfig.lr`. Please see <project:apidocs/index.rst> for more details on configuration settings.
+```
+
+```{Note}
+This guide references several configuration settings from `TransformerConfig`. Please apply these to the appropriate ModelProvider for your model, e.g. `GPTModelProvider`, as the `ConfigContainer` does not accept a raw `TransformerConfig`.
+```
 
 ## Low Precision Training
 


docs/performance-guide.md
```diff
@@ -0,0 +1,488 @@
...
+   > 2. `recipe.model.config.cross_entropy_loss_fusion=true`
+   > 3. `recipe.model.config.gradient_accumulation_fusion=true`
+   > 4. `recipe.model.config.bias_activation_fusion=true`
+   > 5. `recipe.model.config.bias_dropout_fusion=true`
+   > 6. `recipe.model.config.apply_rope_fusion=true`
+
+2. NeMo offers different Flash Attention options, which can be chosen by environment
+
+   > 1. FlashAttention2 (default): `NVTE_FLASH_ATT=1`
+   > 2. cuDNN fused attention: `NVTE_FLASH_ATT=0, NVTE_FUSED_ATT=1`
```
ananthsub:
```suggestion
   > 1. FlashAttention2 (default): `NVTE_FLASH_ATTN=1`
   > 2. cuDNN fused attention: `NVTE_FLASH_ATTN=0, NVTE_FUSED_ATTN=1`
```

docs/performance-guide.md
```diff
@@ -0,0 +1,488 @@
...
+
+1. Dataset preparation
+
+   > 1. Fine-tuning datasets with shorter sequences of variable length can be packed into longer sequences, up to a set maximum length, for best efficiency.
+
+2. To use this feature, the microbatch size must be set to 1. In place of increasing the micro batch size, the maximum sequence length can be increased, which will effectively increase the number of individual sequences per packed sequence.
+
+3. Enabled with:
+
+   > 1. `recipe.data.packed_sequence_specs.packed_sequence_size=<max sequence length>`
```
ananthsub: @hemildesai do you know why the fine-tuning dataset config and builder don't use the packed sequence specs dataclass? https://github.com/NVIDIA-NeMo/Megatron-Hub/blob/57895c6bb13c488850d6532487d5659d1a837de3/src/megatron/hub/data/datasets/packed_sequence.py#L114

unlike this? https://github.com/NVIDIA/NeMo/blob/41a3862785a0d854621ee0a245780c846b80dcd2/nemo/collections/llm/gpt/data/fine_tuning.py#L76

hemildesai: They should use it. If not, its probably a bug.

ananthsub: https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/86 to expose this in the config vs a dict of options

docs/performance-guide.md
```diff
@@ -0,0 +1,500 @@
...
+
+5. Keep parameters in FP8 at FP8 training
+
+   > 1. In FP8 training, after optimizer step execution, we can keep the parameters in FP8. Compared to the baseline that keeps the intermediate weight values in BF16, FP8 parameters lower memory usage and improve communication performance. The below knob enables keeping the parameters in FP8.
+   >
+   >    > 1. `MixedPrecisionConfig.fp8_param_gather=True`
+
+## Operator Fusion
+
+1. All operator fusions are enabled by default in NeMo-Run scripts. You can control specific fusion behaviors using the following configuration knobs:
```
ananthsub: the NeMo-Run reference is unclear to me from megatron hub. users might be confused where these are enabled

docs/performance-guide.md
```diff
@@ -0,0 +1,500 @@
...
+   >
+   > 2. Megatron-Hub uses the distributed optimizer as the default method for data-parallel training. It shards master parameters and optimizer states across data-parallel ranks, reducing model state memory usage without increasing communication overhead compared to traditional data-parallel training.
+   >
+   >    > 1. `OptimizerConfig.use_distributed_optimizer=true`
+
+2. Per-tensor Sharding (Tensor-parallel or Context-parallel mappings)
+
+   > 1. Tensor parallelism (TP) is the primary recommendation when a model exceeds GPU memory capacity under data-parallel mapping. However, since it involves higher communication overhead, the tensor-parallel size should ideally be confined to the high-bandwidth intra-node network (NVLink domain).
+   >
+   >    > 1. `TransformerConfig.tensor_model_parallel_size=<int>`
```
ananthsub, line 40: from the config container, it won't be obvious to users where TransformerConfig settings live. should we reference the ModelProvider instead for these settings?

docs/performance-guide.md
```diff
@@ -0,0 +1,500 @@
...
+   > 2. `GPTProvider.cross_entropy_loss_fusion=true`
+   > 3. `GPTProvider.gradient_accumulation_fusion=true`
+   > 4. `TransformerConfig.bias_activation_fusion=true`
+   > 5. `TransformerConfig.bias_dropout_fusion=true`
+   > 6. `TransformerConfig.apply_rope_fusion=true`
+
+2. Megatron-Hub offers different Flash Attention options, which can be chosen by environment
+
+   > 1. FlashAttention2 (default): `NVTE_FLASH_ATTN=1`
+   > 2. cuDNN fused attention: `NVTE_FLASH_ATTN=0, NVTE_FUSED_ATTN=1`
```
ananthsub: should this be set through the `TransformerConfig.attention_backend` instead?

docs/performance-guide.md
maanug-nv: Good point. I think that is referring to these [scripts](https://github.com/NVIDIA/NeMo/blob/main/scripts/performance/llm/pretrain_llama4_e16.py#L87-L91). Since we don't have performance scripts currently, I think I will just remove this first sentence? 
I think it makes more sense for these fusions to be added to the performance mode of the recipe, if they always improve perf (we still plan to add performance mode, correct?)

docs/performance-guide.md
maanug-nv: yes good point, updated

docs/performance-guide.md
ananthsub: > (we still plan to add performance mode, correct?)

yes, definitely

PR #79, opened 2025-06-13
Adding training and datasets dependency-groups

marcromeyn:
In order to split off the training-related code, this small PR reworks the dependencies inside `pyproject.toml`.

This is the first branch discussed in: [From NeMo-LM to Megatron-Hub: A Proposal](https://docs.google.com/document/d/1h3ZMLYh2PhF2f4zAhUGYNW_di5D12dVM_kZ7OmLd8us/edit?usp=sharing)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,7 @@ override-dependencies = [
 
 [tool.uv.sources]
 nemo-toolkit = { git = "https://github.com/NVIDIA/NeMo.git", rev = "5e63ca3c1631bb156187705f9c4cf643a67f9c05" }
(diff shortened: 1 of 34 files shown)


pyproject.toml
```diff
@@ -88,6 +80,23 @@ requires = ["setuptools>=64", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [dependency-groups]
+training = [
```
hemildesai, line 103: I think you want these as optional dependencies.

marcromeyn: Good point, fixed now

PR #84, opened 2025-06-14
Add workflow for building docs

maanug-nv:
This is to ensure no PRs are merged that will break the docs build.

### DIFF
--- a/docker/Dockerfile.ci
+++ b/docker/Dockerfile.ci
@@ -47,7 +47,7 @@ RUN --mount=type=secret,id=GH_TOKEN bash -ex <<"EOF"
     
     uv sync --link-mode copy --locked --all-groups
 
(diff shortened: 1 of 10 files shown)


.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          submodules: 'recursive'
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+
+      - name: Sphinx build
+        run: |
+          cd docs
+          uv run --only-group docs sphinx-build --fail-on-warning --builder html . _build/html
```
chtruong814, line 38: Thanks for this. The job is still failing because of `--fail-on-warning`.  Could you resolve that?

maanug-nv: That's desired behavior. I've asked @ananthsub to fix source of the warning

PR #94, opened 2025-06-16
Integrate peft into training loop

ananthsub:
Fixes https://github.com/NVIDIA-NeMo/Megatron-Hub/issues/27

Changes in this PR that have been split out for review:
- Llama config fixes: https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/127
- peft test changes to use the pre wrap hook: https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/126
- importer fixes: https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/128

1. Add a new `finetune` entry point similar to `pretrain` to add extra assertions for SFT/PEFT that a checkpoint path must exist for finetuning. The implementation then calls the existing pretrain function as the training loop itself is the same, with slight tweaks to the initialization, checkpoint loading, and checkpoint saving logic.

2. Add `PEFT` to the `ConfigContainer` as an optional field

3. Refactor model initialization in `setup.py` for PEFT: the base model is read from the `pretrained_checkpoint` setting in the `CheckpointConfig`. Note: for now, `pretrained_checkpoint` is expected to be a megatron checkpoint imported prior to beginning finetuning. This uses the pre-wrap hook from the `ModelProvider` in order to initialize the base model, load the pretrained weights in, and apply the PEFT transformation, after which the distributed wrap occurs. This matches the NeMo behavior here:  https://github.com/NVIDIA/NeMo/blob/0af8b7df793ad7538f149ad9bcb8c2cae5134c1a/nemo/lightning/pytorch/callbacks/peft.py#L217-L220 , https://github.com/NVIDIA/NeMo/blob/0af8b7df793ad7538f149ad9bcb8c2cae5134c1a/nemo/lightning/pytorch/callbacks/peft.py#L244-L245

5. Pretrained model checkpoint loading: I split out the main loading logic to work directly if given a checkpoint path (`_load_checkpoint_from_path`). Currently `load_checkpoint` conditionally selects the `load` directory if set in the config, and falls back to `pretrained_checkpoint` if `load` is unavailable, setting the `finetune` flag in the checkpoint config accordingly. This doesn't work for PEFT since PEFT loads form both `pretrained_checkpoint` and `load` for the pretrained weights and adapter states, respectively, and `pretrained_checkpoint` must be loaded first.

6. For loading the adapter states, we re-use the `load_checkpoint` function : we use the config + `load_dir` to detect if we are resuming from a training checkpoint that contains the adapter states only. If so, we use the adapter filter key to select the adapter state dict keys to request when loading the checkpoint. This matches the NeMo behavior here: https://github.com/NVIDIA/NeMo/blob/0af8b7df793ad7538f149ad9bcb8c2cae5134c1a/nemo/lightning/pytorch/callbacks/peft.py#L239-L242

7. The rest of checkpoint loading works as before for optimizer / scheduler / RNG states

8. When saving checkpoints, we apply the adapter filter logic as is done here for the model states to reduce the checkpoint size: https://github.com/NVIDIA/NeMo/blob/5f86c1bd760684de80da6c779d89ea4a013a6c91/nemo/lightning/pytorch/callbacks/peft.py#L500

Added unit + integration + e2e tests to confirm saving + resuming works as expected with the same perf/loss as NeMo: https://wandb.ai/nvidia/megatron-hub-custom-loop-peft


src/megatron/hub/training/checkpointing.py
```diff
@@ -1005,12 +1055,13 @@ def load_checkpoint(
...
     if not skip_load_to_model_and_opt:
+        load_strict = False if is_peft_resume else strict
         if len(model) == 1:
-            model[0].load_state_dict(state_dict["model"], strict=strict)
+            model[0].load_state_dict(state_dict["model"], strict=load_strict)
         else:
             for i in range(len(model)):
                 mpu.set_virtual_pipeline_model_parallel_rank(i)
-                model[i].load_state_dict(state_dict["model%d" % i], strict=strict)
+                model[i].load_state_dict(state_dict["model%d" % i], strict=load_strict)
```
ananthsub: NeMo reference: https://github.com/NVIDIA/NeMo/blob/0af8b7df793ad7538f149ad9bcb8c2cae5134c1a/nemo/lightning/pytorch/callbacks/peft.py#L261C1-L262C1

src/megatron/hub/training/checkpointing.py
```diff
@@ -982,6 +1014,24 @@ def load_checkpoint(
...
+        and load_dir == cfg.checkpoint.load
+        and load_dir != cfg.checkpoint.pretrained_checkpoint
+        and not cfg.checkpoint.finetune
+    )
+
+    if is_peft_resume:
+        load_kwargs["sharded_state_dict"] = apply_peft_adapter_filter_to_state_dict(
+            load_kwargs["sharded_state_dict"], cfg.peft
+        )
+
```
ananthsub, line 1039: NeMo reference: https://github.com/NVIDIA/NeMo/blob/0af8b7df793ad7538f149ad9bcb8c2cae5134c1a/nemo/lightning/pytorch/callbacks/peft.py#L239-L242

src/megatron/hub/training/checkpointing.py
```diff
@@ -514,6 +515,10 @@ def save_checkpoint(
         rerun_state=rerun_state,
     )
 
+    # Apply PEFT filtering to save adapter-only checkpoints
+    if cfg.peft is not None:
+        state_dict = apply_peft_adapter_filter_to_state_dict(state_dict, cfg.peft)
```
ananthsub, line 520: NeMo reference: https://github.com/NVIDIA/NeMo/blob/0af8b7df793ad7538f149ad9bcb8c2cae5134c1a/nemo/lightning/pytorch/callbacks/peft.py#L500

src/megatron/hub/training/converters/llama.py
```diff
@@ -307,14 +307,14 @@ def hf_config(self) -> "HFLlamaConfig":
         return self._hf_config
 
     @property
-    def tron_config(self) -> LlamaConfig:
+    def tron_config(self) -> LlamaModelProvider:
```
marcromeyn: Should we rename this property given the new name of the library?

src/megatron/hub/training/setup.py
```diff
@@ -161,6 +167,13 @@ def setup(
 
     # Model, optimizer, and learning rate.
     timers("model-and-optimizer-setup", log_level=0).start(barrier=True)
+
+    # Register PEFT pre-wrap hook if PEFT is configured
+    if cfg.peft is not None:
```
marcromeyn, line 178: I would prefer this to move into `get_peft_provider` which is importable from `megatron.hub.peft`. It can take in the model-provider and peft-config.

src/megatron/hub/training/setup.py
```diff
@@ -264,3 +284,96 @@ def _update_model_config_funcs(
             model_config.param_sync_func = model_config.param_sync_func[0]
     model_config.finalize_model_grads_func = finalize_model_grads
     model_config.grad_scale_func = optimizer.scale_loss
+
+
+def _create_peft_pre_wrap_hook(cfg: ConfigContainer, state: GlobalState) -> Callable[[list[MegatronModule]], list[MegatronModule]]:
```
marcromeyn, line 295: Is there a way to generalize this, so it doesn't depend on `ConfigContainer` so it can move into the peft submodule? This seems useful outside of training as well. I would prefer that everything around applying peft & loading adapters is part of the peft module.

src/megatron/hub/models/llama/llama_provider.py
```diff
@@ -48,6 +48,13 @@ class LlamaModelProvider(GPTModelProvider):
     attention_dropout: float = 0.0
     hidden_dropout: float = 0.0
     share_embeddings_and_output_weights: bool = False
+    # Fusions
+    bias_activation_fusion: bool = True
```
marcromeyn: We can't enable all fusions by default.

src/megatron/hub/training/setup.py
ananthsub: The main challenge I see with this is the risk of maintaining 2 distinct checkpoint loading paths that might diverge, one inside of PEFT and one inside training/checkpointing. This could be worked around, considering that the peft prewrap hook should deal with only model states and could use mcore dist checkpointing APIs directly, but it's still a concern. I'd like to take this refactoring on in a separate PR to unblock the rest of this integration though.

src/megatron/hub/training/setup.py
ananthsub: if importing from peft, then that means peft takes a dependency on training/checkpointing (or otherwise duplicates the model checkpoint loading code)

this PR assumes the user imports the base model checkpoint and saves it in the mhub checkpoint format, and then sets this path as the `pretrained_checkpoint` in the checkpoint config to use along with peft

src/megatron/hub/training/setup.py
```diff
@@ -202,6 +214,14 @@ def setup(
         timers("load-checkpoint").stop(barrier=True)
         timers.log(["load-checkpoint"])
 
+    _update_model_config_funcs(
```
maanug-nv, line 223: Why does this need to move? I don't think these functions would be affected by checkpoint load?

ananthsub: > Why does this need to move? I don't think these functions would be affected by checkpoint load?

it doesn't, it's unrelated to peft. it doesn't need to move. i was moving this closer to where MLM has it https://github.com/NVIDIA/Megatron-LM/blob/0600a3caf02cdd196e37b3245ff10270c341a5ca/megatron/training/training.py#L2099-L2118

but it's not needed for peft at all

maanug-nv: Ah ok. Fine to keep the change, I just wanted to make sure I wasn't misunderstanding anything. Looking at MLM again, looks like that's not included in any timed section, so it's more accurate now.

PR #98, opened 2025-06-17
ci: Refactor test setup

ko3n1g:


tests/functional_tests/L2_Launch_training.sh
```diff
@@ -17,5 +17,5 @@ set -xeuo pipefail # Exit immediately if a command exits with a non-zero status
 
 export CUDA_VISIBLE_DEVICES="0,1"
 
-python -m torch.distributed.run --nproc_per_node=2 --nnodes=1 -m coverage run --data-file=/workspace/.coverage --source=/workspace/ --parallel-mode -m pytest tests/functional_tests/test_pretrain.py -v -s -x -m "not pleasefixme" --tb=short
+python -m torch.distributed.run --nproc_per_node=2 --nnodes=1 -m pytest -o log_cli=true -o log_cli_level=INFO -v -s -x -m "not pleasefixme" --tb=short tests/functional_tests/training
```
ananthsub: n00b question: earlier these were written with torchrun but changed in https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/72/files#diff-0afdd6e6c94953c9ad30bed771c21d3eab79bd313825afcf93c59a98e0386183

was that because of uv? or something else?

ananthsub: also n00b question: why don't we want individual scripts per test file like https://github.com/NVIDIA/NeMo/blob/main/tests/functional_tests/L2_NeMo_2_llama3_fault_tolerance_plugin.sh ? versus launching all tests under the same folder?

ko3n1g: Yeah so I made weird experiences with calling `torchrun`. I might need to re-validate it, but I had some scenarios where site-packages weren't found presumably because `torchrun` is defined in pre-installed dist-packages. `python` however is in site-packages (managed by UV) and from where we definitely have access to both site and dist packages

ko3n1g: We'd like to suggest a different setup to old Nemo. Old Nemo had shell scripts had we mounted into GitHub action jobs. Two disadvantages:

* Engineers need to add tests to two places (maybe even three): Python, Shell, and GitHub workflow
* For CI is it difficult to manage the test suite since we don't have any tool to filter, select or exclude tests

We would like to suggest to broadly move to Pytest for both unit and e2e tests. Ideally there would only be a single shell script that orchestrates torchrun/pytest. But due to the nature of Eval with their client-server architecture, their setup is a bit different so they have their own shell script.

I refactored the straggler and pretrain tests so that they fit into this format.

Let me know what you think and feel free to share your concerns

PR #101, opened 2025-06-17
ci: Add install test

ko3n1g:


ko3n1g:
> Do you think we can move this workflow to the CI templates repo? Ok to merge here for now.

Yeah my mind was also circling around that. I'll need to see first how much is common between repos, but there should be some denominator

PR #102, opened 2025-06-17
Fix eval tests

athitten:


ko3n1g:
Thanks for sharing! Let's keep this one in draft, I'd like to fix it from #98

tests/functional_tests/evaluation/test_evaluation.py
```diff
@@ -83,9 +86,8 @@ def test_gsm8k_evaluation(self):
             logger.info("Evaluation completed.")
 
         finally:
-            deploy_proc.send_signal(signal.SIGINT)
+            subprocess.run(["pkill", "-9", "python"], check=False)
```
athitten, line 89: FYI this worked well for me in terminating the gsm8k test, but I am unsure if its a good idea to add this in a CI setup. I will leave that to you.

ko3n1g: What concern do you have in mind? Do you see a better way? I agree it's very bruteforce, but I'm not aware of unwanted side-effects yet

athitten: Oh sorry missed your message @ko3n1g.  I can't think of a better way. The concern I have with this is that won't it lead to killing other active python processes ? I don't know how our CI works, if there are other tests running asynchronously in the same env or they would be async but in a diff env where executing `pkill -9 python` won't lead to killing other python tests. If its the latter then we should be okay to have `pkill -9 python`

Also FYI @marta-sd since we might need this in the new repo.

ko3n1g: Ah gotcha, no this cannot spill over to other tests so it should be safe

athitten: Great, thank you for confirming @ko3n1g ! Will add this change to the new Eval repo

PR #105, opened 2025-06-18
chore: Guard optional dependencies

ko3n1g:


src/megatron/hub/training/nvrx_straggler.py
```diff
@@ -16,10 +16,18 @@
...
 
-import nvidia_resiliency_ext.straggler as straggler
 import torch
 
 from megatron.hub.training.config import NVRxStragglerDetectionConfig
+from megatron.hub.utils.import_utils import MISSING_NVRX_MSG
+
+
+try:
+    import nvidia_resiliency_ext.straggler as straggler
```
ananthsub, line 26: why do we need this guard when nvidia-resiliency-ext is listed as a dependency?

ko3n1g: We have a marker to exclude nvrx from darwin, but we expect also macOS users to pip install megatron-hub

src/megatron/hub/recipes/run_plugins.py
```diff
@@ -14,15 +14,27 @@
...
+
+try:
+    import nemo_run as run
+
+    HAVE_NEMO_RUN = True
+except (ImportError, ModuleNotFoundError):
+    from unittest.mock import MagicMock
+
+    run = MagicMock()
+    run.Plugin = types.new_class("Plugin", ())
```
marcromeyn: Why is this needed? It might be better to change all the run-related type-hints to strings.

src/megatron/hub/recipes/run_plugins.py
```diff
@@ -51,6 +63,9 @@ class PreemptionPlugin(run.Plugin):
     enable_exit_handler_for_data_loader: bool = False
 
     def setup(self, task: run.Partial | run.Script, executor: run.Executor):
+        if not HAVE_NEMO_RUN:
```
marcromeyn, line 66: I wonder if it's better to introduce a class-level decorator that automatically does this check on each method and raises the error if nemo-run is not installed.

src/megatron/hub/recipes/run_plugins.py
ko3n1g: `run.Plugin` is not a type hint but a class that inherit from. I don't see a better solution. An alternative would be to hide those classes (`PreemptionPlugin` as an example) behind `if HAVE_RUN`, but then the whole class doesn't exist anymore which potentially breaks imports of it from other modules. I think that's actually worse

src/megatron/hub/recipes/run_plugins.py
ko3n1g: It's an option, but it's also a bit hacky. These are dataclasses, they don't offer an init to the user. Surely we can hack ourselves into the init, but in the current from everthing is quite explicit and easy to understand (at the cost of redundancy). Personally, I prefer explicity/redundant over implicit code. DRY is not always helpful imho

PR #108, opened 2025-06-18
[Refactor] Model providers

marcromeyn:


src/megatron/hub/models/llama/recipes/llama3_8b.py
```diff
@@ -0,0 +1,286 @@
...
+from megatron.hub.training.config import (
+    CheckpointConfig,
+    ConfigContainer,
+    GPTDatasetConfig,
+    LoggerConfig,
+    RNGConfig,
+    SchedulerConfig,
+    TokenizerConfig,
+    TrainingConfig,
+)
```
ananthsub: I'd still strongly prefer to keep recipes separate from models/ for now.

1. this is to ensure that `models/` does not depend on `training` while the training loop + configs are in megatron hub.
2. `recipes/` at the top-level is easier for users to find + use vs being nested in `models/<individual model>/recipes`

wdyt?

examples/llm/llama3_8b/pretrain_llama3_8b.py
```diff
@@ -57,10 +57,10 @@
 
 from omegaconf import OmegaConf
 
-from megatron.hub.models.utils import forward_step
 from megatron.hub.recipes.llm.llama3_8b import pretrain_config
 from megatron.hub.training.config import ConfigContainer
 from megatron.hub.training.pretrain import megatron_pretrain
+from megatron.hub.training.step import forward_step
```
ananthsub: i'd prefer to split this rename/move out into its own PR since it can go in right away without being blocked by the model providers

src/megatron/hub/common/model_provider_mixin.py
```diff
@@ -0,0 +1,180 @@
...
+        parallel_state.initialize_model_parallel(
+            tensor_model_parallel_size=getattr(self, "tensor_model_parallel_size", 1),
+            pipeline_model_parallel_size=getattr(self, "pipeline_model_parallel_size", 1),
+            virtual_pipeline_model_parallel_size=getattr(self, "virtual_pipeline_model_parallel_size", None),
+            context_parallel_size=getattr(self, "context_parallel_size", 1) or 1,
+            expert_model_parallel_size=getattr(self, "expert_model_parallel_size", 1) or 1,
+            **kwargs,
+        )
+        if seed is not None:
+            model_parallel_cuda_manual_seed(seed)
```
ananthsub: do other args for seeding need to be exposed?

src/megatron/hub/common/model_provider_mixin.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def __call__(self, **kwargs: Unpack["GetModelKwargs"]) -> list[ModelT]:
+        return self.get_model(**kwargs)
+
+    @property
+    def meta_model(self) -> list[ModelT]:
+        return self(wrap_with_ddp=False, init_model_with_meta_device=True)
```
ananthsub: why is this needed? could you add a docstring? i didn't see it described in the doc

src/megatron/hub/models/gpt_provider.py
```diff
@@ -0,0 +1,371 @@
...
+    # Fusions
+    masked_softmax_fusion: bool = field(default_factory=fusions.can_enable_masked_softmax_fusion)
+    cross_entropy_loss_fusion: bool = True  # Generally beneficial, no specific dependencies
+    gradient_accumulation_fusion: bool = field(default_factory=fusions.can_enable_gradient_accumulation_fusion)
+    bias_activation_fusion: bool = False  # Disabled by default as it can interfere with certain architectures
+    persist_layer_norm: bool = True  # Generally beneficial for performance
+    bias_dropout_fusion: bool = field(default_factory=fusions.can_enable_bias_dropout_fusion)
+    apply_rope_fusion: bool = field(default_factory=fusions.can_enable_apply_rope_fusion)
+
+    model_transform: Callable[[list[MegatronModule]], list[MegatronModule]] | None = None
```
ananthsub: i'd prefer to add `model_transform` later only when it's used

src/megatron/hub/common/model_provider_mixin.py
```diff
@@ -0,0 +1,180 @@
...
+    Megatron models with proper initialization and distributed data parallel
+    support. Subclasses must implement the `provide` method to return the
+    specific model instance.
+    """
+
+    CONFIG_NAME = "mhub_model.json"
+    DEFAULT_CONFIG_FORMAT = "json"
+
+    @abc.abstractmethod
+    def provide(self, pre_process=None, post_process=None) -> ModelT:
```
ananthsub, line 55: please add typehints

src/megatron/hub/common/model_provider_mixin.py
```diff
@@ -0,0 +1,180 @@
...
+        pass
+
+    def get_model(
```
ananthsub: it's confusing for me to read multiple `get_model` definitions where the `get_model` method calls the `get_model` function below. maybe the function can be `get_model_from_provider` or something to distinguish it that way

src/megatron/hub/common/model_provider_mixin.py
```diff
@@ -0,0 +1,180 @@
...
+            virtual_pipeline_model_parallel_size=getattr(self, "virtual_pipeline_model_parallel_size", None),
+            context_parallel_size=getattr(self, "context_parallel_size", 1) or 1,
+            expert_model_parallel_size=getattr(self, "expert_model_parallel_size", 1) or 1,
+            **kwargs,
+        )
+        if seed is not None:
+            model_parallel_cuda_manual_seed(seed)
+
+    def __call__(self, **kwargs: Unpack["GetModelKwargs"]) -> list[ModelT]:
+        return self.get_model(**kwargs)
```
ananthsub: when is this expected to be used vs. get_model? for composition?

src/megatron/hub/common/model_provider_mixin.py
```diff
@@ -0,0 +1,180 @@
...
+    """Mixin class for providing Megatron model instances.
+
+    This abstract base class provides functionality to create and configure
+    Megatron models with proper initialization and distributed data parallel
+    support. Subclasses must implement the `provide` method to return the
+    specific model instance.
+    """
+
+    CONFIG_NAME = "mhub_model.json"
+    DEFAULT_CONFIG_FORMAT = "json"
```
ananthsub, line 52: since the checkpoints generated here default to saving as yaml, does it make sense to have that be the default here? or was json set to match huggingface?

src/megatron/hub/common/model_provider_mixin.py
```diff
@@ -0,0 +1,180 @@
...
+            fp16=fp16,
+            bf16=bf16,
+            use_torch_fsdp2=use_torch_fsdp2,
+            wrap_with_ddp=wrap_with_ddp,
+            data_parallel_random_init=data_parallel_random_init,
+            use_cpu_initialization=use_cpu_initialization,
+            init_model_with_meta_device=init_model_with_meta_device,
+        )
+
+    def initialize_model_parallel(self, seed: int | None = None, **kwargs) -> None:
```
ananthsub: n00b question: why is this added now? i see how it will be needed for hypercomm grid changes in the future to specify the mesh, but for training shouldn't we rely on `initialize.py` still? my concern is having that branch and being unclear where parallel state is being initialized

src/megatron/hub/common/model_provider_mixin.py
```diff
@@ -0,0 +1,180 @@
...
+            **kwargs,
+        )
+
+    def save_pretrained(
+        self,
+        save_directory,
+        config_format: str | None = None,
+        config_name: str | None = None,
+        **kwargs,
+    ):
```
ananthsub: please add docstrings/types so these render nicely in the docs page

src/megatron/hub/converters/common.py
```diff
@@ -313,7 +329,7 @@ def hf_config(self) -> "PretrainedConfig":
 
     @property
     @abstractmethod
-    def tron_config(self) -> GPTConfig | T5Config:
+    def tron_config(self) -> GPTModelProvider | T5ModelProvider:
```
ananthsub, line 332: have you tested the converters with the providers being added? does it still work?

src/megatron/hub/core/models/model_provider.py
```diff
@@ -0,0 +1,315 @@
...
+    if wrap_with_ddp:
+        model = _ddp_wrap(
+            model,
+            use_torch_fsdp2,
+            data_parallel_random_init,
+            ddp_config,
+            overlap_param_gather_with_optimizer_step,
+        )
+
+    return model
```
ananthsub: will a user still be able to initialize just the base model without distributed without needing to write a custom model provider? for instance, that makes the peft application very easy to read in a few lines of code vs. needing to define a custom peft provider that abstracts this

src/megatron/hub/core/utils/fusions.py
```diff
@@ -0,0 +1,167 @@
...
+can be enabled based on the current environment and dependencies.
+"""
+
+import logging
+import os
+from typing import TYPE_CHECKING
+
+
+if TYPE_CHECKING:
+    from megatron.hub.models.gpt_provider import GPTModelProvider
```
ananthsub: this create a circular dependency between fusions and the gpt provider. why isn't `validate_rope_fusion_compatibility` split out to its own file under `models/ `

ananthsub:
@chtruong814 @ko3n1g we're missing the copyright check in CI as well

src/megatron/hub/core/utils/fusions.py
marcromeyn: We actually only need `TransformerConfig`, I will change it. I put it in core since I believe it should be moved to mcore.

src/megatron/hub/core/models/model_provider.py
marcromeyn: I don't fully understand the question? You can always just call: `provider.provide()`. I was thinking that for PEFT no custom model-provider is needed, just a function that takes in one and adds a transform.

src/megatron/hub/converters/common.py
marcromeyn: I don't think anything has changed from the perspective of the converter.

src/megatron/hub/common/model_provider_mixin.py
marcromeyn: Yeah in training this would be unused. It's just to make creating a model easier for single-GPU, just to inspect things. So in a bridge you can do: `bridge.to_model()` which then will call this function under-the-hood.

src/megatron/hub/models/llama/recipes/llama3_8b.py
marcromeyn: I don't have a strong opinion on this. Would you imagine the recipes folder to be completely flat? Because inside the models dir, we would have a dir per model. If we were to do the same inside the recipes, isn't it kind of the same structure as to put it inside models directly?

src/megatron/hub/common/model_provider_mixin.py
marcromeyn: It's to be able to inspect the model structure without any memory. I added a doc-string now

src/megatron/hub/training/step.py
```diff
@@ -211,7 +211,7 @@ def get_batch(
     return batch.values()
 
 
-def forward_step(state: GlobalState, data_iterator: Iterable, model: GPTModel) -> tuple[torch.Tensor, partial]:
+def forward_step(state: GlobalState, data_iterator: Iterable, model: GPTModelProvider) -> tuple[torch.Tensor, partial]:
```
ananthsub: the model's typehint shouldn't be updated

src/megatron/hub/training/step.py
```diff
@@ -18,9 +18,9 @@
```
ananthsub: rename file as `gpt_step` ?

src/megatron/hub/models/llama/recipes/llama3_8b.py
ananthsub: it's similar, but we'll also have recipe-specific utilities that need to sit in a common place. i don't want to put these under `training/` as these could be optional dependencies like `nemo_run`. it also makes the dependencies clearer to me at least that `training/` depends on `models/` and `recipes/` depends on both `models/` and `training/` but `models/` is as independent as possible

src/megatron/hub/common/model_provider_mixin.py
ananthsub: bump about this - the existing nemoFW codebase is hard to read partially because of similar or identically named functions across multiple classes

src/megatron/hub/utils/import_utils.py
```diff
@@ -221,6 +221,9 @@ def __delete__(cls, *args, **kwargs):
     def __len__(cls):
         raise UnavailableError(cls._msg)
 
+    def __getitem__(cls, item):
+        raise UnavailableError(cls._msg)
+
```
ananthsub, line 226: (not blocking) why did these need changes?

marcromeyn: I was using a uv-managed env inside my workstation. The env didn't have transformer-engine, which made some of the tests fail. This change allows `hasattr` and therefore allowed the tests not to raise an exception.

src/megatron/hub/common/model_provider_mixin.py
```diff
@@ -0,0 +1,258 @@
...
+            fp16=fp16,
+            bf16=bf16,
+            use_torch_fsdp2=use_torch_fsdp2,
+            wrap_with_ddp=wrap_with_ddp,
+            data_parallel_random_init=data_parallel_random_init,
+            use_cpu_initialization=use_cpu_initialization,
+            init_model_with_meta_device=init_model_with_meta_device,
+        )
+
+    def initialize_model_parallel(
```
maanug-nv, line 151: why does this need to duplicate what's happening in `hub.training.initialize._initialize_distributed` ? Why can't we reuse that?

src/megatron/hub/core/models/model_provider.py
ananthsub: for peft with NeMo, the code flow is:

1. apply PEFT transforms with model parallel initialization disabled: https://github.com/NVIDIA/NeMo/blob/06cf0b663f05ec1f8ca9acae087f23e49a5ff31a/nemo/lightning/pytorch/callbacks/peft.py#L219-L220
2. then setup model parallel: https://github.com/NVIDIA/NeMo/blob/06cf0b663f05ec1f8ca9acae087f23e49a5ff31a/nemo/lightning/pytorch/callbacks/peft.py#L246

which includes this logic before the DDP wrap occurs: https://github.com/NVIDIA/NeMo/blob/06cf0b663f05ec1f8ca9acae087f23e49a5ff31a/nemo/lightning/megatron_parallel.py#L544-L581

that's the way the logic is split up currently in mhub: https://github.com/NVIDIA-NeMo/Megatron-Hub/blob/3eb70906c1d907d2168a61953a23f33cbe22c9fa/src/megatron/hub/models/__init__.py#L158-L187

I'm not sure what the expected calling pattern is now with the model provider. Initializing the model through the provider with `wrap_with_ddp=False` before applying the PEFT transforms doesn't keep the same semantics as the allocation on GPU and wrapping with `Float16Module` occurs before PEFT is applied, unlike NeMo where this should happen after PEFT is applied

src/megatron/hub/common/model_provider_mixin.py
marcromeyn: That's future work, for now we want to keep the models separate from the training-related functionality. This method is not used inside the training-loop.

src/megatron/hub/core/models/model_provider.py
marcromeyn: This is exactly why a `ModelProvider` can overwrite parts of the `provide_model` method. The main mechanism I have in mind for this are hooks (similar to how `nn.Module` does it).  This would mean a user could call `register_pre_wrap_hook` and/or `register_post_wrap_hook` on a model-provider. For PEFT the user-facing API I have in mind is:

```
from megatron.hub.peft import get_peft_model_provider
provider =  AutoBridge.from_pretrained("Qwen/Qwen2.5-3B-Instruct").to_provider()
peft_provider = get_peft_provider(provider, Lora(...))
```

`get_peft_model_provider` would just register the peft-transform as pre-wrap-hook (appending after the loading the base-model weights registered by the bridge) and then it could just post-wrap-hook to load the adapter weights if needed. So there's no need for a PEFT-specific model-provider, just a function (like PEFT from HF) that calls `register_pre_wrap_hook` on the model-provider passed in.

PR #109, opened 2025-06-18
Port GPT full TE layer autocast spec from NeMo

ananthsub:
Port from NeMo: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/models/language_modeling/megatron/gpt_full_te_layer_autocast_spec.py

To remove dependency from nemo for this spec

Changes from NeMo:
1. inlining `ApexGuardDefaults` and `torch_dtype_from_precision` into this file
2. remove `HAVE_MEGATRON_CORE` import guards

### DIFF
--- a/src/megatron/hub/models/gpt_provider.py
+++ b/src/megatron/hub/models/gpt_provider.py
@@ -56,6 +56,7 @@ def transformer_engine_layer_spec(config: "GPTModelProvider") -> ModuleSpec:
         fp8=bool(config.num_moe_experts and (config.fp8 is not None)),
     )
 
+
 def transformer_engine_full_layer_spec(config: "GPTModelProvider") -> ModuleSpec:
     """Create a full Transformer Engine layer specification with autocast support.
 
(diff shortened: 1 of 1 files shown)


src/megatron/hub/models/gpt_full_te_layer_autocast_spec.py
```diff
@@ -0,0 +1,366 @@
...
+        return hidden_states, context
+
+    def _get_layer_offset(self):
+        pipeline_rank = parallel_state.get_pipeline_model_parallel_rank()
+
+        num_layers_per_pipeline_rank = (
+            self.config.num_layers // parallel_state.get_pipeline_model_parallel_world_size()
+        )
+
+        assert self.config.virtual_pipeline_model_parallel_size is None, (
```
maanug-nv: I don't think this assert would apply for us? Not sure why this spec would only be used for Nemo 1.0 either, as it is exposed in the Nemo 2.0 path.

maanug-nv:
Unit tests in a follow-up ?

PR #111, opened 2025-06-18
MockGPTDatasetConfig

maanug-nv:
Create a MockGPTDatasetConfig helper class for users to know which fields must not be set to use mock data.


src/megatron/hub/training/config.py
```diff
@@ -141,6 +141,14 @@ def __post_init__(self) -> None:
         assert self.eod_mask_loss is not None
 
 
+@dataclass
+class MockGPTDatasetConfig(GPTDatasetConfig):
```
ananthsub, line 145: this check for the data builder registry needs to use isinstance instead of type to support subclassing like this:
https://github.com/NVIDIA-NeMo/Megatron-Hub/blob/20ab6bff79e405994c3ac3d36839b48642d098c1/src/megatron/hub/data/utils.py#L144-L162

PR #119, opened 2025-06-24
Align adapter shape with base linear output shape

ananthsub:
Sync with NeMo: https://github.com/NVIDIA/NeMo/pull/14009


cuichenx:
There was a TE bug, so 14009 and this PR are not needed. 
See details in issue !2419

PR #120, opened 2025-06-24
Use lazy init for import and export as part of conversion

hemildesai:
Replicates https://github.com/NVIDIA/NeMo/pull/14008/commits/8414cf78b8b700e884514612464baf21e394e727

### DIFF
--- a/src/megatron/hub/converters/common.py
+++ b/src/megatron/hub/converters/common.py
@@ -46,6 +46,57 @@ logger = logging.getLogger(__name__)
 HF_ASSETS_DIR = "hf_assets"
 
 
+def materialize_meta_tensors(obj: Any) -> Any:
+    """Recursively replace *meta* tensors with real CPU tensors.
+
+    This is handy when models are initialized on the *meta* device.
+    Such tensors lack storage and need to be converted before any real computation or serialization.
+
+    The function walks through common container types (``dict``, ``list``, ``tuple``)
+    and converts any tensor residing on the *meta* device to a freshly allocated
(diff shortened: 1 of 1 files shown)


src/megatron/hub/converters/common.py
```diff
@@ -46,6 +46,57 @@
 HF_ASSETS_DIR = "hf_assets"
 
 
+def materialize_meta_tensors(obj: Any) -> Any:
```
ananthsub, line 49: could you explain why it's needed / how the initializing meta tensors resolves the issue?

hemildesai: This is needed because when loading the mcore state dict from the checkpoint, we get errors of the form "Could not load sharded tensor into meta device".

hemildesai: And we don't care about actual initialization of the weights (which is slow), so we just replace meta tensors with empty tensors on cpu.

src/megatron/hub/converters/common.py
```diff
@@ -46,6 +46,57 @@
...
+
+    The function walks through common container types (``dict``, ``list``, ``tuple``)
+    and converts any tensor residing on the *meta* device to a freshly allocated
+    ``torch.empty`` tensor on the CPU that keeps the same ``shape`` and ``dtype``.
+
+    Args:
+        obj: Arbitrary python object that may contain ``torch.Tensor`` instances.
+
+    Returns
+    -------
```
maanug-nv, line 63: please use google style

PR #124, opened 2025-06-26
ci: Buildless runs

ko3n1g:


docker/Dockerfile.ci
```diff
@@ -11,40 +11,11 @@
...
-
-ARG UV_VERSION=0.7.2
-ENV PATH="/root/.local/bin:$PATH"
-RUN curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | sh
-ENV UV_PROJECT_ENVIRONMENT=/opt/venv
-ENV PATH="$UV_PROJECT_ENVIRONMENT/bin:$PATH"
-ENV UV_LINK_MODE=copy
-
-# Install all
+FROM nvidia/cuda:12.8.1-cudnn-devel-ubuntu24.04
```
maanug-nv, line 15: can we use `cudnn-runtime-ubuntu24.04` ?

ko3n1g: Looks like this doesn't work: https://github.com/NVIDIA-NeMo/Megatron-Hub/actions/runs/15909558410/job/44873243583

CONTRIBUTING.md
```diff
@@ -57,35 +57,49 @@
...
 Adding a new dependency will update UV's lock-file. Please check this into your branch:
 
 ```bash
 git add uv.lock pyproject.toml
-git commit -m "build: Adding dependencies"
+git commit -s -m "build: Adding dependencies"
 git push
 ```
+
+## Linting and Formatting
```
marcromeyn, line 90: What about adding a `Makefile` or `justfile` (like energon does [source](https://github.com/NVIDIA/Megatron-Energon/blob/develop/justfile)) for stuff like this?

docker/common/install.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
marcromeyn, line 1: Would it be possible to use this script as well for development of the library outside of a docker-image? Would be nice if we could offer something like `make/just venv` which creates the env from scratch.

ko3n1g: Outside of a docker-image increases complexity a lot since we don't know what we're dealing with anymore. Could be macOS, could be Windows, could be an outdated ubuntu version. We also don't know which (or even if) cuda is installed. Nothing speaks against bisecting this installer and to pick what you need for environment. But for now I would recommend to install the linting tools into your workstation with `uv sync --only-group dev` and to run tests via a container

CONTRIBUTING.md
ko3n1g: Sure, nothing against that if users find it useful. But let's approach that in a follow-up PR. I'd like to get this speed-up in asap

PR #125, opened 2025-06-26
Adding bridge to the Hub for 2-way HF-binding

marcromeyn:


src/megatron/hub/bridge/auto_bridge.py
```diff
@@ -0,0 +1,288 @@
...
+
+        Raises:
+            ValueError: If no suitable bridge is found
+        """
+        # Try each bridge in order
+        for bridge_cls in _BRIDGES:
+            if hasattr(bridge_cls, "supports") and bridge_cls.supports(config):
+                # Found a supporting bridge - check if it has the required method
+                if hasattr(bridge_cls, method_name):
+                    return bridge_cls
```
ananthsub: why not decorate the protocol with runtime checkable and then check with isinstance here?

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -0,0 +1,657 @@
...
+from megatron.hub.models.gpt_provider import GPTModelProvider
+
+
+MegatronModelT = TypeVar("ModelT", bound=MegatronModule)
+DataclassT = TypeVar("DataclassT")
+
+
+class CausalLMBridge(Generic[MegatronModelT]):
+    """
+    Bridge for converting Causal Language Models between HuggingFace and Megatron formats.
```
ananthsub, line 42: if it's huggingface specific, should this mention HF in the name somewhere, similar to the HF importer/exporter names nemo used?

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -0,0 +1,657 @@
...
+            >>> # Load weights from a different checkpoint
+            >>> bridge.load_weights(megatron_model, "./finetuned_model")
+        """
+        if hf_path is None:
+            if not isinstance(self.hf_pretrained, PreTrainedCausalLM):
+                raise ValueError("hf_path is required when hf_pretrained is not a PreTrainedCausalLM instance")
+            pre_trained = self.hf_pretrained
+        else:
+            pre_trained = PreTrainedCausalLM.from_pretrained(hf_path)
+        self._model_bridge.load_weights(model, pre_trained)
```
ananthsub, line 265: where is `self._model_bridge` initialized?

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -0,0 +1,657 @@
...
+            >>> # Save without progress bar (useful in scripts)
+            >>> bridge.save_weights(megatron_model, "./weights", show_progress=False)
+
+        Note:
+            - This method is collective and must be called by all ranks
+            - Uses safetensors format for efficient loading and security
+            - Automatically handles model sharding for large models
+            - The saved weights can be loaded with HuggingFace's from_pretrained
+        """
+        torch.distributed.barrier()
```
ananthsub: do you need any guards/assertions that distributed is already initialized?

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -0,0 +1,657 @@
...
+        return self._model_bridge.bridge_state_to_megatron(self.hf_pretrained)
+
+    def to_model(
+        self,
+        load_weights: bool = True,
+        hf_path: str | Path | None = None,
+        **kwargs: Unpack[GetModelKwargs],
+    ) -> list[MegatronModelT]:
+        provider = self.to_provider(load_weights, hf_path)
+        return provider(**kwargs)
```
ananthsub, line 435: why offer `to_model` here instead of forcing users to always go through the provider? is it just for convenience?

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -0,0 +1,657 @@
...
+
+        provider: ModelProviderMixin = self._model_bridge.provider_bridge(self.hf_pretrained)
+
+        if load_weights:
+            if hf_path is None:
+                provider.register_pre_wrap_hook(partial(self._model_bridge.load_state_from_hf, self.hf_pretrained))
+            else:
+                # Load from specified path
+                pre_trained = PreTrainedCausalLM.from_pretrained(hf_path)
+                provider.register_pre_wrap_hook(partial(self._model_bridge.load_state_from_hf, pre_trained))
```
ananthsub, line 485: do you expect this to be integrated into the training flow to allow users to specify the HF model as the pretrained checkpoint? or will this remain separate and users should still run the importer before starting finetuning?

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -0,0 +1,657 @@
...
+
+        return provider
+
+    @property
+    def transformer_config(self) -> TransformerConfig:
+        _model_provider = self.to_megatron(load_weights=False)
```
ananthsub, line 491: where is `to_megatron` implemented in this class?

src/megatron/hub/bridge/weight_bridge.py
```diff
@@ -0,0 +1,1763 @@
...
+    def pp_rank(self) -> int:
+        """Get current pipeline parallel rank."""
+        return mpu.get_pipeline_model_parallel_rank()
+
+    @property
+    def pp_group(self) -> Optional[torch.distributed.ProcessGroup]:
+        """Get pipeline parallel process group."""
+        if self.pp_size > 1:
+            return mpu.get_pipeline_model_parallel_group()
+        return None
```
ananthsub, line 439: @yaoyu-33 how would this work with orthotope changes? in the future would we need to explicitly pass the groups?

examples/bridge/2_way_hf_binding.py
```diff
@@ -0,0 +1,92 @@
...
+    model_name = hf_model_id.split("/")[-1]
+    if output_dir:
+        save_path = os.path.join(output_dir, model_name)
+    else:
+        save_path = model_name
+
+    bridge = CausalLMBridge.from_pretrained(hf_model_id)
+
+    # Formatting
+    console = Console()
```
yaoyu-33: can we move this console table gen to a utils and keep this script clean?

examples/bridge/generate_from_hf.py
```diff
@@ -0,0 +1,85 @@
...
+        cur_attention_mask = torch.ones_like(cur_input_ids, dtype=torch.bool)
+
+    # Decode the generated token sequence
+    generated_text = tokenizer.decode(generated_tokens)
+    print(f"Generated text:\n{generated_text}")
+    return generated_text
+
+
+if __name__ == "__main__":
+    model = bridge.to_model(wrap_with_ddp=False)
```
yaoyu-33: maybe call model mcore_model
and move hf_model init and tokenizer init here as well.
Then gen seq will just be a utils to compare 2 model's seq gen.

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -0,0 +1,657 @@
...
+            if not isinstance(self.hf_pretrained, PreTrainedCausalLM):
+                raise ValueError("hf_path is required when hf_pretrained is not a PreTrainedCausalLM instance")
+            pre_trained = self.hf_pretrained
+        else:
+            pre_trained = PreTrainedCausalLM.from_pretrained(hf_path)
+        self._model_bridge.load_weights(model, pre_trained)
+
+        return model
+
+    @overload
```
yaoyu-33, line 269: do we need the overload typing check here?

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -0,0 +1,657 @@
...
+        See Also:
+            GPTModelProvider: The provider class for creating models
+            load_weights: Method to load weights into existing models
+        """
+
+        provider: ModelProviderMixin = self._model_bridge.provider_bridge(self.hf_pretrained)
+
+        if load_weights:
+            if hf_path is None:
+                provider.register_pre_wrap_hook(partial(self._model_bridge.load_state_from_hf, self.hf_pretrained))
```
yaoyu-33, line 481: still feel a bit confused about the bridge / provider / model here. Provider is actually MCore config container?

src/megatron/hub/bridge/model_bridge.py
```diff
@@ -0,0 +1,1013 @@
...
+
+    param_name: str
+    weight: torch.Tensor
+
+
+class HFSaveTask(NamedTuple):
+    """Task for saving HuggingFace model weights with pipeline parallel metadata."""
+
+    pp_rank: int
+    vpp_rank: Optional[int]
```
yaoyu-33: plz rename vpp_rank to vp_stage to keep consistent with mcore.

src/megatron/hub/bridge/weight_bridge.py
yaoyu-33: yes - so I am thinking can we make `gather_from_tp_ranks` etc. to be just functions not a class method, it can take tp_group, pp_group as args.  and just use group.size() and group.rank() to get necessary values.

src/megatron/hub/bridge/state_bridge.py
```diff
@@ -0,0 +1,263 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+from typing import List, Optional, Tuple
+
+from megatron.hub.bridge.weight_bridge import MegatronWeightBridge
+
+
+class MegatronStateBridge:
```
yaoyu-33, line 21: feels like state bridge is a confusing name, I can understand weight bridge and model bridge. but I am not sure what state bridge do. Is it just a weight bridges combiner? - if so maybe rename the class and move it to model_bridge.py or weight_bridge.py

src/megatron/hub/bridge/weight_bridge.py
yaoyu-33: i dont think we need `pp_rank` etc. these are all going to change soon.

src/megatron/hub/bridge/state_bridge.py
marcromeyn: It bridges the state-dict, a pytorch `nn.Module` has a state_dict that contains weights. So that's why I named it state-bridge.

src/megatron/hub/bridge/causal_bridge.py
marcromeyn: It's the object that produces a model. So it contains all the arguments for the `__init__` of the model, so `TransformerConfig` + init-args. Next to that it contains the method `provide_model` so it can customize how the model is produced. That extra functionality means it's not strictly a config, it contains behavior as well.

src/megatron/hub/bridge/causal_bridge.py
marcromeyn: The public API for users if: `to_provider` & `to_model`.

src/megatron/hub/bridge/causal_bridge.py
marcromeyn: I think to improve the UX, it would be great it we can integrate this into the training-loop. I was under the impression that that would already work? Since you could do `AutoBridge.from_pretrained(...).to_provider()` and pass it to the `ConfigContainer`, or wouldn't that work?

src/megatron/hub/bridge/causal_bridge.py
marcromeyn: It was a request by the dev-tech team.

src/megatron/hub/bridge/causal_bridge.py
marcromeyn: It's a property: [source](https://github.com/NVIDIA-NeMo/Megatron-Hub/blob/fe453647a22a0539c77465259c879cee43319b43/src/megatron/hub/bridge/causal_bridge.py#L492-L494)

src/megatron/hub/bridge/causal_bridge.py
marcromeyn: Right now we only have HF as source, but in the future that could change. The model-bridge that gets resolved is flexible enough for that. So for now I was thinking it makes more sense to just name it by the type of model.

src/megatron/hub/bridge/causal_bridge.py
marcromeyn: It's just for type-hints, the bridge is implemented in a generic way.

examples/bridge/generate_from_hf.py
```diff
@@ -0,0 +1,84 @@
...
+    generated_text = tokenizer.decode(generated_tokens)
+    print(f"Generated text:\n{generated_text}")
+    return generated_text
+
+
+if __name__ == "__main__":
+    megatron_model = bridge.to_model(wrap_with_ddp=False)
+
+    prompt = "Hello, how are you?"
+    generate_sequence(prompt, megatron_model, HF_MODEL_ID)
```
ISEEKYAN: the third argument is max_token, it is a typo to pass HF_MODEL_ID

src/megatron/hub/models/llama/llama_causal_bridge.py
```diff
@@ -0,0 +1,128 @@
...
+from megatron.hub.bridge.weight_bridge import (
+    GatedMLPWeightBridge,
+    QKVWeightBridge,
+    TPAwareWeightBridge,
+)
+from megatron.hub.models.llama.llama_provider import Llama31ModelProvider, LlamaModelProvider
+
+
+@MegatronModelBridge.impl(source=LlamaForCausalLM, target=GPTModel)
+class MegatronCausalLlamaBridge(MegatronModelBridge):
```
ISEEKYAN: the name is a little confusing given that the file is `llama_caluse_bridge.py` and the class is `MegatronCausalLlamaBridge`

src/megatron/hub/__init__.py
```diff
@@ -11,3 +11,16 @@
...
+from megatron.hub.bridge.auto_bridge import AutoBridge
+from megatron.hub.bridge.causal_bridge import CausalLMBridge
+from megatron.hub.models import *  # noqa: F403
+
+
+__all__ = [
+    "AutoBridge",
+    "CausalLMBridge",
+    *_models.__all__,
+]
```
ananthsub, line 26: why not have these imported from `megatron.hub.bridge` ? by default we won't have training or recipe imports here

examples/training/llama/pretrain_llama3_8b_nemo_run_script.py
ananthsub, line 1: the existing examples should be moved under `recipes` not `training`

src/megatron/hub/bridge/README.md
```diff
@@ -0,0 +1,324 @@
...
+bridge = CausalLMBridge.from_pretrained("meta-llama/Llama-3.2-1B")
+provider = bridge.to_megatron()
+
+# The provider is lazy - configure parallelism before creating models
+provider.tensor_model_parallel_size = 8
+provider.pipeline_model_parallel_size = 2
+
+# Create the distributed model
+model = provider(wrap_with_ddp=False)
+```
```
ananthsub: nit: confusing that the comment says create the distributed model though `wrap_with_ddp=False`

src/megatron/hub/bridge/causal_bridge.py
ananthsub: there's a few collisions with checkpoint loading:

1. `pretrained_checkpoint` handling: the training explicitly checks the `pretrained_checkpoint` field in the `CheckpointConfig` to determine the base weights to load. i've extended this for peft to use the pre-wrap hook to load from the checkpoint. however, if the provider is generated from the auto bridge that includes this hook, then the training code will do a double load since the two don't know about each other. however, we can't assume that all providers will be generated from the bridge that comes with the hook installed. and since the hook is opaque, i don't think it's safe to infer that if there's a pre-wrap hook already registered, then training should skip loading the pretrained checkpoint.

2. outside peft, we should skip loading the base weights entirely if there's already a checkpoint generated from training to resume from, assuming the user configures `load` in the CheckpointConfig. in that case we shouldn't run the pre-wrap hook here to load weights

edit: there won't be a collision as long as `load_weights` is properly set. for peft, `load_weights` can always be set for the base model from HF, but for SFT we should check if an existing training checkpoint exists first

src/megatron/hub/bridge/hf_pretrained/vlm.py
```diff
@@ -0,0 +1,618 @@
...
+from transformers.generation.utils import GenerateOutput
+
+from megatron.hub.bridge.hf_pretrained.base import PreTrainedBase
+
+
+# Python 3.12+ supports PEP 692 (TypedDict Unpack)
+if sys.version_info >= (3, 12):
+    pass
+else:
+    pass
```
ananthsub: dead code

src/megatron/hub/bridge/weight_bridge.py
```diff
@@ -0,0 +1,1763 @@
...
+import torch.distributed
+import torch.nn as nn
+from megatron.core import mpu
+from megatron.core.transformer.transformer_config import TransformerConfig
+
+
+WeightType = TypeVar("WeightType", torch.Tensor, Dict[str, torch.Tensor])
+
+
+class MegatronWeightBridge(ABC, Generic[WeightType]):
```
ananthsub, line 29: since we need to port models from nemo to mhub with the bridge support, do you have a guide specifically comparing the existing nemo HF importer/exporters to using the megatron weight bridge? ideally we want this process to be as mechanical as possible. the current converters offer this by mostly following the same implementation

e.g. https://github.com/NVIDIA/NeMo/blob/d7b87b1481c4205de3d5c9ff8c7b6644bbe0ecd1/nemo/collections/llm/gpt/model/qwen2.py#L197-L395

src/megatron/hub/__init__.py
marcromeyn: It's more because HF-transformers is flat like this, so I would prefer if we do the same for models & bridge in megatron-hub.

src/megatron/hub/bridge/state_bridge.py
yaoyu-33: can we call it state_dict_bridge maybe?

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -0,0 +1,664 @@
...
+            config: HuggingFace model config object
+
+        Returns:
+            True if this bridge can handle the model, False otherwise
+        """
+        architectures = getattr(config, "architectures", [])
+        if not architectures:
+            return False
+
+        return any(arch.endswith("ForCausalLM") for arch in architectures)
```
yaoyu-33, line 121: I feel this might be a bit hacky to identify causallm vs. vlm

yaoyu-33: but we can update in the future. Maybe add a todo here

src/megatron/hub/bridge/state_bridge.py
yaoyu-33: and weight bridge is actually param bridge? like bridge for each param

src/megatron/hub/models/llama/llama_causal_bridge.py
```diff
@@ -0,0 +1,128 @@
...
+from megatron.hub.bridge.weight_bridge import (
+    GatedMLPWeightBridge,
+    QKVWeightBridge,
+    TPAwareWeightBridge,
+)
+from megatron.hub.models.llama.llama_provider import Llama31ModelProvider, LlamaModelProvider
+
+
+@MegatronModelBridge.impl(source=LlamaForCausalLM, target=GPTModel)
+class LlamaCausalBridge(MegatronModelBridge):
```
yaoyu-33, line 33: why do we need Causal in the name?

src/megatron/hub/models/llama/llama_causal_bridge.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Megatron Hub Bridge for Llama Causal LM.
```
yaoyu-33, line 35: need example usage here

src/megatron/hub/models/llama/llama_causal_bridge.py
marcromeyn: Because we could have a llama embedding model etc as well

examples/bridge/README.md
ananthsub, line 1: remove the empty readme?

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -0,0 +1,665 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
ananthsub, line 1: nit: filename and class don't match
should this be lm_causal_bridge.py?

PR #127, opened 2025-06-26
Fix Llama model provider configs

ananthsub:
Fix llama model configs after https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/108

Current error when loading from a pretrained checkpoint for llama3 because of the config mismatch / missing inheritance
```
megatron.core.dist_checkpointing.core.CheckpointingException: Global shape mismatch for loaded (torch.Size([32, 6144, 4096])) and expected ((32, 12288, 4096)) tensor for key decoder.layers.self_attention.linear_qkv.weight`
```

Resyncing with settings from nemo solves the issue: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/gpt/model/llama.py

### DIFF
--- a/tests/unit_tests/models/llama/test_llama_provider.py
+++ b/tests/unit_tests/models/llama/test_llama_provider.py
@@ -208,12 +208,20 @@ class TestLlama3QueryGroupsInheritance:
 
     def test_llama3_base_provider_has_correct_num_query_groups(self):
         """Test that Llama3ModelProvider has num_query_groups=8."""
-        provider = Llama3ModelProvider()
+        provider = Llama3ModelProvider(
+            num_layers=32,
+            hidden_size=4096,
+            num_attention_heads=32,
+        )
         assert provider.num_query_groups == 8
 
     def test_llama31_provider_inherits_num_query_groups(self):
         """Test that Llama31ModelProvider inherits num_query_groups=8 from Llama3ModelProvider."""
-        provider = Llama31ModelProvider()
+        provider = Llama31ModelProvider(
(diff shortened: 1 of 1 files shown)


src/megatron/hub/models/llama/llama_provider.py
```diff
@@ -105,13 +113,27 @@ class Llama3ModelProvider(LlamaModelProvider):
...
+    hidden_dropout: float = 0.0
+    attention_dropout: float = 0.0
+    normalization: str = "RMSNorm"
     init_method_std: float = 0.01
     layernorm_epsilon: float = 1.0e-05
+    add_bias_linear: bool = False
+    activation_func: Callable = F.silu
+    gated_linear_unit: bool = True
+    # Fusions
+    bias_activation_fusion: bool = True
```
maanug-nv, line 125: why do we duplicate these fusions , should already be coming from the `LlamaModelProvider` ? Just for documenting?

ananthsub: i've added these again primarily to make comparisons between nemo's llama configs easier (these fields are repeated there too): https://github.com/NVIDIA/NeMo/blob/578ce5da5ee63e8b72b7a31ee2edf1288e14a427/nemo/collections/llm/gpt/model/llama.py#L130-L156

maanug-nv: fair enough. we can remove at a later point if we want to make sure these settings for fusions always match up for all classes that inherit from `LlamaModelProvider`

PR #184, opened 2025-07-03
Rename WeightBridge to Mapping

yaoyu-33:
Some renaming to increase the clarity. Also including one small fix to llama causal bridge which fixes the layernorm param name.

Rename WeightBridge to Mapping
Rename `to` to `hf_param`

### DIFF
--- a/src/megatron/hub/bridge/param_mapping.py
+++ b/src/megatron/hub/bridge/param_mapping.py
@@ -22,6 +22,7 @@ import torch.nn as nn
 from megatron.core import mpu
 from megatron.core.transformer.transformer_config import TransformerConfig
 
+
 WeightType = TypeVar("WeightType", torch.Tensor, Dict[str, torch.Tensor])
 
 


src/megatron/hub/bridge/model_bridge.py
```diff
@@ -223,7 +223,7 @@ def provider_bridge(self, hf_pretrained) -> LlamaModelProvider:
 
             def state_bridge(self) -> MegatronStateBridge:
                 return MegatronStateBridge(
-                    TPAwareWeightBridge(
+                    TPAwareMapping(
                         megatron="embedding.word_embeddings.weight",
                         to="model.embed_tokens.weight"
```
marcromeyn: A bunch of the doc-strings are still showing the old names

yaoyu-33: cool I will double check again.

src/megatron/hub/bridge/README.md
```diff
@@ -52,16 +52,16 @@ The bridge framework uses a layered architecture with clear separation of concer
...
+│ (MegatronStateBridge)│   (MegatronParamMapping)         │
 └─────────────────────────────────────────────────────────┘
 ```
 
 ### Core Components
 
 1. **MegatronModelBridge**: High-level orchestrator that coordinates the conversion process
 2. **MegatronStateBridge**: Registry of parameter name mappings between formats
-3. **MegatronWeightBridge**: Handles weight transformations and distributed communication
+3. **MegatronParamMapping**: Handles weight transformations and distributed communication
```
ananthsub, line 64: now it seems like the names are flipped: the Mapping layer is MegatronStateBridge while the Transformation Layer is MegatronParamMapping.

and in the description, MegatronStateBridge is responsible for parameter name mappings, while MegatronParamMapping handles weight transforms..

yaoyu-33: The naming confusion is resolved after rename as "MegatronStateBridge" as well.

src/megatron/hub/bridge/param_mapping.py
```diff
@@ -462,12 +462,12 @@ def from_megatron(
         if megatron_weight is None:
             return {}
 
-        return {self.to: megatron_weight}
+        return {str(self.hf_param): megatron_weight}
```
ananthsub: what guarantees `self.hf_param` is a `str` and not `dict[str, str]` here?

src/megatron/hub/bridge/param_mapping.py
```diff
@@ -167,7 +167,7 @@ def broadcast_from_pp_rank(self, tensor: Optional[torch.Tensor]) -> Optional[tor
         """Broadcast a tensor from the pipeline-parallel rank that owns it.
 
         Broadcasts to **all** PP ranks. This mirrors the behaviour of
-        `broadcast_from_megatron_pp` in the original MBridge implementation and
+        `broadcast_from_megatron_pp` in the original MMapping implementation and
```
ananthsub: MMapping is odd to read. is there any reason to mention the original MBridge implementation at all?

src/megatron/hub/bridge/param_mapping.py
```diff
@@ -584,20 +584,20 @@ def from_megatron(
             return {}
 
         if self.tp_size == 1:
-            return {self.to: megatron_weight}
+            return {str(self.hf_param): megatron_weight}
```
ananthsub: same questions here

src/megatron/hub/bridge/param_mapping.py
```diff
@@ -22,17 +22,16 @@
 from megatron.core import mpu
 from megatron.core.transformer.transformer_config import TransformerConfig
 
-
 WeightType = TypeVar("WeightType", torch.Tensor, Dict[str, torch.Tensor])
```
ananthsub, line 26: for the lint error:
```suggestion

WeightType = TypeVar("WeightType", torch.Tensor, Dict[str, torch.Tensor])
```

PR #185, opened 2025-07-03
Add mixed precision config to recipes

hemildesai:
- precision ported from https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/recipes/llama3_8b.py#L138

### DIFF
--- a/src/megatron/hub/recipes/llama/llama3_70b_16k.py
+++ b/src/megatron/hub/recipes/llama/llama3_70b_16k.py
@@ -166,8 +166,6 @@ def pretrain_config(
         sequence_parallelism=sequence_parallelism,
     )
 
-    config.dataset.sequence_length = SEQUENCE_LENGTH_16K
-
     # Apply precision configuration again after overriding model
     if isinstance(precision_config, str):
         precision_config = get_mixed_precision_config(precision_config)


src/megatron/hub/recipes/llama/llama3_8b.py
```diff
@@ -200,4 +203,8 @@ def pretrain_config(
         rng=RNGConfig(seed=1234),
     )
 
+    if isinstance(precision_config, str):
+        precision_config = get_mixed_precision_config(precision_config)
+    precision_config.setup(cfg.model, cfg.optimizer, cfg.ddp)
```
ananthsub, line 209: you can also update the recipe unit test to confirm the mixed precision settings are now enabled across these configs

hemildesai: Added in https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/185/commits/a5461f398749f0332da4cb0b46869a87442d86b5

src/megatron/hub/recipes/llama/llama3_70b_16k.py
```diff
@@ -161,4 +166,11 @@ def pretrain_config(
         sequence_parallelism=sequence_parallelism,
     )
 
+    config.dataset.sequence_length = SEQUENCE_LENGTH_16K
```
ananthsub: seq length is set here already: https://github.com/NVIDIA-NeMo/Megatron-Hub/blob/09166382b0acf619e461fb99779f1149dbc25ba5/src/megatron/hub/recipes/llama/llama3_70b.py#L174

were you seeing issues with the seq length otherwise?

PR #188, opened 2025-07-04
State Bridge Naming Update

yaoyu-33:


src/megatron/hub/bridge/README.md
```diff
@@ -234,7 +234,7 @@ To add support for a new model architecture:
 
 3. **Define Weight Mappings**
    ```python
-   def state_bridge(self):
+   def mapping_registry(self):
```
marcromeyn, line 237: Let's remove registry here, just mappings.

src/megatron/hub/bridge/mapping_registry.py
```diff
@@ -13,30 +13,30 @@
...
 
 import re
-from typing import List, Optional, Tuple
+from typing import List, Optional
 
 from megatron.hub.bridge.param_mapping import MegatronParamMapping
 
 
-class MegatronStateBridge:
+class MegatronMappingRegistry:
```
marcromeyn, line 21: Since the mappings themselve have `megatron_param` and `hf_param`, I would prefer `MegatronParamMapping`

src/megatron/hub/bridge/README.md
yaoyu-33: rename to state_dict_mapping

PR #211, opened 2025-07-07
Add comm overlap configs with user buffers and default garbage collection

hemildesai:
ported pre-training performance optimizations from https://github.com/NVIDIA/NeMo/tree/main/nemo/collections/llm/recipes

### DIFF
--- a/src/megatron/hub/training/setup.py
+++ b/src/megatron/hub/training/setup.py
@@ -100,14 +100,15 @@ def setup(
         A SetupOutput named tuple containing the initialized state, model,
         optimizer, scheduler, data iterators, and checkpointing context.
     """
-    state = GlobalState()
-    state.cfg = cfg
     # TODO: Freeze state.cfg
 
     cfg.validate()
     # Apply communication overlap configuration if provided at the very beginning
     if cfg.comm_overlap is not None:
         cfg.comm_overlap.setup(cfg.model, cfg.optimizer, cfg.ddp)
+
+    state = GlobalState()
+    state.cfg = cfg
 
     # Initialize async checkpoint worker if enabled
     init_async_checkpoint_worker(state)


src/megatron/hub/recipes/llama/llama2_7b.py
```diff
@@ -93,6 +94,7 @@ def pretrain_config(
     lr_warmup_iters: int = 2000,
     # Precision recipe
     precision_config: str | MixedPrecisionConfig = "bf16_mixed",
+    comm_overlap_config: CommOverlapConfig | None = None,
```
maanug-nv, line 97: use Optional type for consistency?

src/megatron/hub/recipes/llama/llama2_7b.py
```diff
@@ -209,4 +215,10 @@ def pretrain_config(
         precision_config = get_mixed_precision_config(precision_config)
     precision_config.setup(cfg.model, cfg.optimizer, cfg.ddp)
 
+    if comm_overlap_config is None:
+        comm_overlap_config = CommOverlapConfig(
+            tp_comm_overlap=False,
+        )
+    comm_overlap_config.setup(cfg.model, cfg.optimizer, cfg.ddp)
```
maanug-nv: Can we do this in `training.setup.setup` ? If a user is not using a recipe, they will have to copy+paste these lines, right?

maanug-nv: I guess that means `CommOverlapConfig` would need to be part of the ConfigContainer. wdyt?

ananthsub: i'd prefer `CommOverlapConfig` to not be part of the ConfigContainer if no other part of the training loop will read from that config directly, similar to the precision config

src/megatron/hub/training/setup.py
```diff
@@ -104,6 +104,11 @@ def setup(
     state.cfg = cfg
     # TODO: Freeze state.cfg
 
+    cfg.validate()
+    # Apply communication overlap configuration if provided at the very beginning
+    if cfg.comm_overlap is not None:
+        cfg.comm_overlap.setup(cfg.model, cfg.optimizer, cfg.ddp)
+
```
ananthsub, line 109: I'd do the validation + comm overlap setup before initializing the state and attaching the config, just in case the state reads any fields from the config

hemildesai: Done in https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/211/commits/aaa6f2121f40e1ab2ec8fc1150dd2c0132f8b685

PR #222, opened 2025-07-09
[sync] Allow user to load from specified iteration

ananthsub:
Sync from 
https://github.com/NVIDIA/Megatron-LM/blob/e11d28592f19c122859be764b7afe7c208d9acc1/megatron/training/checkpointing.py#L944-L947


src/megatron/hub/training/checkpointing.py
```diff
@@ -1378,6 +1378,11 @@ def _load_base_checkpoint(
             train_state = read_train_state(tracker_filename)
             iteration = train_state.step
             # release = train_state.release
+
+    # Allow user to specify the loaded iteration.
```
hemildesai, line 1382: We should also skip reading the train state in this case and allow passing custom model config to support MLM checkpoints directly

ananthsub: @hemildesai could you clarify how the custom model config would be passed? i think this is an issue with tron and loading MLM checkpoints since we're missing the args

PR #225, opened 2025-07-09
MoE Token Drop parity

ananthsub:
Offer a functional version of the MoE Token drop callback in NeMo that updates the model provider settings

NeMo reference: https://github.com/NVIDIA/NeMo/blob/main/nemo/lightning/pytorch/callbacks/moe_token_drop.py

### DIFF
--- a/src/megatron/hub/training/utils/moe_token_drop.py
+++ b/src/megatron/hub/training/utils/moe_token_drop.py
@@ -12,18 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from megatron.hub.models.gpt_provider import GPTModelProvider
+from megatron.core.transformer.transformer_config import TransformerConfig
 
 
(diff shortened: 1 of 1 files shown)


src/megatron/hub/training/utils/moe_token_drop.py
```diff
@@ -0,0 +1,55 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from megatron.hub.models.gpt_provider import GPTModelProvider
+
+
+def apply_moe_token_drop(
+    model_provider: GPTModelProvider,
```
ananthsub: this can accept the TransformerConfig directly. it doesn't need to be the model provider

PR #228, opened 2025-07-10
Add functional tests for train + resume + finetuning from ckpt

ananthsub:
`tmp_path` will be different across processes when launched with torchrun/torch distributed run. this PR adds utils for synchronizing the paths before the training begins to avoid hangs when loading checkpoints

this PR also adds functional tests to go from pretrain -> save checkpoint -> finetune (sft/lora/dora)

### DIFF
--- a/tests/functional_tests/training/test_sft.py
+++ b/tests/functional_tests/training/test_sft.py
@@ -41,7 +41,6 @@ from tests.functional_tests.utils import broadcast_path, initialize_distributed
 @dataclass
 class Llama3ModelProvider145M(Llama3ModelProvider):
     rotary_base: int = 500_000
-    seq_length: int = 8192
     num_layers: int = 2
     hidden_size: int = 768
     ffn_hidden_size: int = 2688
@@ -133,7 +132,7 @@ class TestSupervisedFinetuning:
     ):
         """Create training configuration with customizable parameters."""
         return ConfigContainer(
-            model=Llama3ModelProvider145M(),
+            model=Llama3ModelProvider145M(seq_length=seq_length),
             train=TrainingConfig(
                 train_iters=train_iters,
                 eval_interval=5,


tests/functional_tests/utils.py
```diff
@@ -0,0 +1,57 @@
...
+    }
+    torch.distributed.init_process_group(**init_process_group_kwargs)
+    torch.distributed.barrier(device_ids=[get_local_rank_preinit()])
+
+
+def broadcast_path(path: str | Path) -> str:
+    """
+    Broadcast a path from rank 0 to all ranks.
+    """
+    if get_world_size_safe() == 1:
```
maanug-nv: nitpick, I don't think we should use our code as tools in our tests (when the test isn't targeting that code). Distributed init is required for this anyway

maanug-nv: ie we can use torch.distributed.get_rank and get_world_size

tests/functional_tests/training/test_pretrain.py
```diff
@@ -66,6 +68,17 @@ def test_pretrain_with_checkpoint(self, tmp_path):
         """
         Test end to end training with checkpoint functionality.
         """
+        initialize_distributed()
+        shared_base_dir = broadcast_path(tmp_path)
+
+        checkpoint_dir = os.path.join(shared_base_dir, "checkpoints")
+        tensorboard_dir = os.path.join(shared_base_dir, "tensorboard")
+
+        # Create subdirectories (all ranks can do this safely)
```
maanug-nv: I dont understand this comment, why is only rank 0 making these subdirs then?

src/megatron/hub/core/models/model_provider.py
```diff
@@ -94,6 +94,12 @@ def get_model(
         if _model is not None:
             model = _model
 
+    # Set tensor model parallel attributes if not set
+    # In case pre_wrap_hook augmented the model (e.g. adding PEFT adapters)
+    for model_module in model:
+        for param in model_module.parameters():
+            tensor_parallel.set_defaults_if_not_set_tensor_model_parallel_attributes(param)
+
```
ananthsub: matching https://github.com/NVIDIA/NeMo/blob/b0a39a92a134f0691fbdda31fbccdc99b2079897/nemo/lightning/megatron_parallel.py#L549-L550

src/megatron/hub/training/checkpointing.py
```diff
@@ -1049,7 +1049,9 @@ def _load_checkpoint_from_path(
     assert state.train_state.skipped_train_samples == 0
     assert state.train_state.consumed_valid_samples == 0
 
-    state.train_state = read_train_state(get_checkpoint_train_state_filename(checkpoint_name))
+    if not cfg.checkpoint.finetune:
+        state.train_state = read_train_state(get_checkpoint_train_state_filename(checkpoint_name))
+
```
ananthsub: @maanug-nv @hemildesai this fixes an issue from the tests for resuming with peft. load_checkpoint is called twice, meaning that the train state was being loaded twice. the second time, it failed on the assertions above. 

this also better matches MLM's handling: https://github.com/NVIDIA/Megatron-LM/blob/3a2a972d9ffd5cee534b761f0803ea1b07389d2f/megatron/training/checkpointing.py#L1416-L1425

PR #237, opened 2025-07-11
Add dummy default vocab size for null tokenizer in recipes

ananthsub:


marcromeyn:
Maybe just add a `VOCAB_SIZE` to `llama_provider.py` and use that across the board.

PR #238, opened 2025-07-11
Bridge config fixes

maanug-nv:
Changes needed to support models output by the bridge.

Case 1: With recipe and mock data (benchmarking case)
If `ModelProvider.vocab_size` is `None` , that value is copied from the tokenizer [here](https://github.com/NVIDIA-NeMo/Megatron-Hub/blob/c7f081bd841bd689cb5a5f46b72f4e98df9ceccd/src/megatron/hub/training/setup.py#L168). However this value is 1 greater than the vocab size provided to TokenizerConfig (see [here](https://github.com/NVIDIA-NeMo/Megatron-Hub/blob/c7f081bd841bd689cb5a5f46b72f4e98df9ceccd/src/megatron/hub/training/tokenizers/tokenizer.py#L1027)). This creates an issue when the vocab size **is** set in the ModelProvider, which is the case for the ModelProvider output by the bridge. In that case, the model vocab size is 1 less than the real tokenizer vocab size (including the EOD token id), so the embedding weight is short 1 row. Rather than adding a special case for this scenario (bridge model and NullTokenizer), it seems simpler to modify the NullTokenizer. We can simply let the last index be the EOD token instead of appending it. Since the NullTokenizer is only used with mock data, it shouldn't impact text generation.

Case 2: With real tokenizer configured by user
In this case, it makes sense to let the user decide which takes priority, so we can simply make an assertion. If they want the model vocab size to take priority, it should be copied to the tokenizer config. To let tokenizer vocab size take priority, they can just set model vocab size to `None`.

### DIFF
--- a/src/megatron/hub/training/config.py
+++ b/src/megatron/hub/training/config.py
@@ -785,9 +785,13 @@ class ConfigContainer(Container):
         if self.peft is not None:
             assert self.checkpoint.pretrained_checkpoint is not None, "PEFT requires a pretrained checkpoint path"
 
-        if isinstance(self.dataset, GPTDatasetConfig):
-            assert self.model.seq_length == self.dataset.sequence_length, (
-                f"Please ensure sequence length configuration in model config and "
-                f"dataset config match.\nSequence length in model config: {self.model.seq_length}, "
-                f"Sequence length in dataset config: {self.dataset.sequence_length}"
-            )
+        data_seq_length = (
+            self.dataset.seq_length
+            if isinstance(self.dataset, FinetuningDatasetConfig)
+            else self.dataset.sequence_length
+        )
+        assert self.model.seq_length == data_seq_length, (
+            f"Please ensure sequence length configuration in model config and "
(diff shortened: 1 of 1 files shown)


maanug-nv:
Updated this PR to include the assert for matching sequence length config, and remove `variable_seq_length` from the Llama bridge.

src/megatron/hub/training/config.py
```diff
@@ -784,3 +784,10 @@ def validate(self) -> None:
 
         if self.peft is not None:
             assert self.checkpoint.pretrained_checkpoint is not None, "PEFT requires a pretrained checkpoint path"
+
+        if isinstance(self.dataset, GPTDatasetConfig):
```
ananthsub: also add an assertion for finetuning? https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/05725c562b25b399d5ae2384fb59306318f72efa/src/megatron/hub/data/builders/finetuning_dataset.py#L56

maanug-nv: updated

PR #241, opened 2025-07-14
Add megatron checkpoint load and save api to Bridge

yaoyu-33:
This pr included few changes required for NeMo RL to work properly.

### DIFF
--- a/src/megatron/bridge/models/model_bridge.py
+++ b/src/megatron/bridge/models/model_bridge.py
@@ -446,17 +446,9 @@ class MegatronModelBridge(Generic[HFPreTrained, ModelProviderTarget, MegatronMod
         Raises:
             ValueError: If input parameters are invalid.
         """
-        if hf_pretrained is None:
-            raise ValueError("hf_pretrained cannot be None")
-
-        if megatron_model is None:
-            raise ValueError("megatron_model cannot be None")
 
         if not isinstance(megatron_model, list):
             megatron_model = [megatron_model]
-
-        if not megatron_model:
-            raise ValueError("megatron_model list cannot be empty")
 
         for task in self._build_plan_hf_to_megatron(hf_pretrained, megatron_model):
             hf_state_dict: Mapping[str, torch.Tensor] = hf_pretrained.state
@@ -532,17 +524,8 @@ class MegatronModelBridge(Generic[HFPreTrained, ModelProviderTarget, MegatronMod
             - DISTRIBUTE: Each rank yields its shard (experimental)
         """
 
-        if megatron_model is None:
-            raise ValueError("megatron_model cannot be None")
-
-        if hf_pretrained is None:
-            raise ValueError("hf_pretrained cannot be None")
-
         if not isinstance(megatron_model, list):
             megatron_model = [megatron_model]
-
-        if not megatron_model:
-            raise ValueError("megatron_model list cannot be empty")
 
         # Normalize mode parameter
         if isinstance(mode, str):


src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -420,6 +423,132 @@ def save_hf_weights(self, model, path: str | Path, show_progress: bool = True) -
...
+
+        Example:
+            >>> # Save model checkpoint after conversion
+            >>> bridge.save_megatron_model(megatron_model, "./megatron_checkpoint")
+
+            >>> # Save with custom checkpoint format
+            >>> bridge.save_megatron_model(
+            ...     megatron_model,
+            ...     "./megatron_checkpoint",
+            ...     ckpt_format="zarr"
```
ananthsub: let's not mention zarr in the docstring since it's deprecated

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -420,6 +423,132 @@ def save_hf_weights(self, model, path: str | Path, show_progress: bool = True) -
...
+            - The checkpoint format follows Megatron's standard structure for compatibility
+        """
+        try:
+            from megatron.hub.training.checkpointing import save_checkpoint
+            from megatron.hub.training.config import CheckpointConfig, ConfigContainer, LoggerConfig
+            from megatron.hub.training.state import GlobalState
+        except ImportError:
+            raise ImportError("megatron.hub.training is not installed.")
+        # Get model config from the first model instance
+        model_config = get_model_config(model[0])
```
ananthsub: why is this passing the mcore model config instead of the model provider?

yaoyu-33: i think we only have info to model instance atm, ideally provider and the cfg should be identical for model saving.

ananthsub: since we serialize the model config in the `run_config.yaml` saved as part of the checkpoint, it's possible for users to instantiate the model provider directly from the config there. if using this API, they'd save the wrong config, meaning initialization will be broken

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -420,6 +423,132 @@ def save_hf_weights(self, model, path: str | Path, show_progress: bool = True) -
         if torch.distributed.is_available() and torch.distributed.is_initialized():
             torch.distributed.barrier()
 
+    @overload
+    def save_megatron_model(self, model: list[MegatronModelT], path: str | Path) -> None:
+        ...
+
+    def save_megatron_model(self, model, path: str | Path, ckpt_format: str = "torch_dist") -> None:
```
marcromeyn: We have `save_hf_pretrained` so wouldn't `save_megatron_pretrained` make more sense as a name?

src/megatron/hub/bridge/causal_bridge.py
```diff
@@ -420,6 +423,132 @@ def save_hf_weights(self, model, path: str | Path, show_progress: bool = True) -
...
+        # Save the checkpoint
+        save_checkpoint(
+            state=state,
+            model=model,
+            optimizer=None,
+            opt_param_scheduler=None,
+            num_floating_point_operations_so_far=0,
+        )
+
+    def load_megatron_model(self, path: str | Path, **kwargs: Unpack[GetModelKwargs]) -> list[MegatronModelT]:
```
marcromeyn: The API was intended to work using: `CausalLMBridge.from_pretrained`, it should detect that the ckpt passed in is megatron-format and instantiate things from there. With `load_weights` indicating whether or not to load the weights.

PR #242, opened 2025-07-14
Raise error for checkpoint config for `async_save` enabled without `save` set

ananthsub:
Follow up to https://github.com/NVIDIA-NeMo/Megatron-Hub/pull/235

### DIFF
--- a/tests/unit_tests/training/test_config.py
+++ b/tests/unit_tests/training/test_config.py
@@ -796,13 +796,6 @@ class TestRerunConfigValidation:
             cfg.load_optim = True
             cfg.__post_init__()
 
-        with pytest.raises(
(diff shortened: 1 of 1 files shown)


src/megatron/hub/training/config.py
```diff
@@ -446,6 +449,11 @@ class CheckpointConfig:
     replication_factor: int = 2
     """Number of machines storing the replica of a given rank's data."""
 
+    def __post_init__(self) -> None:
+        """Post-initialization checks for checkpoint config."""
+        if not self.save and self.async_save:
+            logger.warning("async_save is set, but save is not set. This will result in no checkpoints being saved.")
```
hemildesai: Shouldn't this raise an error?

PR #243, opened 2025-07-15
[sync] Prep for asymmetric pipeline/mtp

ananthsub:
port asymmetric pipeline parallel changes from https://github.com/NVIDIA/NeMo/pull/14007


ZhiyuLi-Nvidia:
Thank you @ananthsub. Do you plan to add a test for asymmetric pipeline? 
Other than that LGTM!

PR #251, opened 2025-07-15
rename: megatron-hub -> megatron-bridge

ananthsub:
as title
renaming Megatron-Hub to Megatron-Bridge, updating imports from megatron.hub to megatron.bridge, and any other instances where megatron hub terminology is used.

I'm not changing `src/megatron/bridge/bridge` in this PR since there are several APIs within this module that follow the bridge naming. I'd prefer to discuss any specific naming changes there outside of this PR in order for this PR to be as mechanical as possible

### DIFF
--- a/tests/unit_tests/bridge/test_causal_bridge.py
+++ b/tests/unit_tests/bridge/test_causal_bridge.py
@@ -357,7 +357,9 @@ class TestCausalLMBridgeEdgeCases:
             bridge = CausalLMBridge(mock_hf_model)
 
             # Now patch the from_pretrained method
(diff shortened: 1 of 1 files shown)


src/megatron/bridge/package_info.py
```diff
@@ -24,12 +24,12 @@
 __shortversion__ = ".".join(map(str, VERSION[:3]))
 __version__ = ".".join(map(str, VERSION[:3])) + "".join(VERSION[3:])
 
-__package_name__ = "megatron.hub"
+__package_name__ = "megatron.bridge"
```
chtruong814: @ko3n1g I think package name should be megatron_bridge, right?

ko3n1g: yeah, `megatron_bridge`

PR #256, opened 2025-07-16
build: Make TE a required dependency

ko3n1g:


src/megatron/hub/training/comm_overlap.py
```diff
@@ -431,7 +423,7 @@ def _get_model_comm_overlap_cfgs(
             elif not model_cfg.sequence_parallel:
                 logging.warning("Disabling tensor parallel communication overlap due to sequence_parallel=False.")
                 self.user_comm_overlap_cfg.tp_comm_overlap = False
-            elif not HAVE_TE:
+            else:
```
maanug-nv: This change will always disable the overlap. Since TE is a hard dependency, we should instead delete this branch of the condition.

src/megatron/hub/peft/lora.py
```diff
@@ -122,7 +112,7 @@ def transform(self, module: nn.Module, name: Optional[str] = None, prefix: Optio
                     and module.quant_state.__class__ == bitsandbytes.functional.QuantState
                 ):
                     lora_cls = patch_linear_module
-                elif HAVE_TE and module.__class__ == te.Linear:
+                elif module.__class__ == te.Linear:
                     if TELinearAdapter is None:
```
maanug-nv: We can remove this error as well. TELinearAdapter should never be None now

src/megatron/hub/models/t5_provider.py
```diff
@@ -60,10 +56,7 @@ def local_layer_spec(encoder_config: "T5ModelProvider", decoder_config: "T5Model
 
 def default_layer_spec(encoder_config: "T5ModelProvider", decoder_config: "T5ModelProvider") -> ModuleSpec:
```
maanug-nv: I don't think we need this fn anymore.

tests/unit_tests/models/test_gpt_full_te_layer_autocast_spec.py
```diff
@@ -458,7 +417,6 @@ def test_get_gpt_full_te_layer_autocast_spec_different_num_layers(
 class TestWithoutTransformerEngine:
     """Test behavior when Transformer Engine is not available."""
 
-    @patch("megatron.hub.models.gpt_full_te_layer_autocast_spec.HAVE_TE", False)
     def test_get_gpt_full_te_layer_autocast_spec_without_te(self):
```
maanug-nv: This test also doesn't make sense anymore

tests/unit_tests/models/test_gpt_full_te_layer_autocast_spec.py
```diff
@@ -468,35 +426,32 @@ def test_get_gpt_full_te_layer_autocast_spec_without_te(self):
...
-                    "num_attention_heads": 8,
-                    "init_method": lambda x: x,
-                    "output_layer_init_method": lambda x: x,
-                    "hidden_dropout": 0.1,
-                    "attention_dropout": 0.1,
-                    "tp_size": 1,
-                    "params_dtype": torch.float32,
-                }
-                AutocastTransformerLayer(**basic_config)
+        with pytest.raises(AssertionError, match="AutocastTransformerLayer requires Transformer Engine"):
```
maanug-nv: this won't be raised when TE is present

maanug-nv: actually this assertion doesn't exist anymore

src/megatron/bridge/peft/lora.py
```diff
@@ -122,7 +111,7 @@ def transform(self, module: nn.Module, name: Optional[str] = None, prefix: Optio
                     and module.quant_state.__class__ == bitsandbytes.functional.QuantState
                 ):
                     lora_cls = patch_linear_module
-                elif HAVE_TE and module.__class__ == te.Linear:
+                elif module.__class__ == te.Linear:
                     if TELinearAdapter is None:
```
maanug-nv: We can remove this error, TELinearAdapter should never be None now

pyproject.toml
```diff
@@ -83,6 +83,7 @@ dependencies = [
     "nvidia-modelopt[torch]; sys_platform != 'darwin'",
     "nvidia-resiliency-ext; sys_platform != 'darwin'",
     "hydra-core>1.3,<=1.3.2",
+    "transformer-engine[pytorch]==2.4.0; sys_platform != 'darwin'"
```
chtruong814: Can we have a better range for this so that it doessn't override what is installed in the container?

ko3n1g: yeah, uprading to te 2.5 now. This PR was created before the wheel existed

ko3n1g: there is some urgency to merge this PR since we're wanting to have TE as a required dep

chtruong814: Please set it to some range that will use what's in the container.  In the container for now, we'll be installing at commits.  After this release, we need to see what their 2.6 release branch is and install from that commit.  So maybe have a min range here for now?

ko3n1g: Yeah, my idea was that after we've cut the branch, `main` would lift its upper bound to less than 2.7.

PR #258, opened 2025-07-16
Remove `core/` and `common/` directories, merge `bridge/` into `models/`

ananthsub:
`core/utils` -> top-level `utils/` directory
`core/models` -> `models/`
`common/model_provider_mixin.py` -> `models/model_provider_mixin.py`
`common/` -> `models/`
`bridge/` -> `models/` to get rid of the awkward `bridge.bridge` imports

### DIFF
--- a/src/megatron/bridge/models/causal_bridge.py
+++ b/src/megatron/bridge/models/causal_bridge.py
@@ -25,6 +25,7 @@ from transformers import AutoConfig
 from transformers.configuration_utils import PretrainedConfig
 from typing_extensions import Unpack
 
+from megatron.bridge.models import model_bridge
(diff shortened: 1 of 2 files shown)


examples/bridge/multi_gpu_hf.py
```diff
@@ -43,7 +43,7 @@
 from rich.table import Table
 
 from megatron.bridge import CausalLMBridge
-from megatron.bridge.common.decorators import torchrun_main
+from megatron.bridge.bridge.decorators import torchrun_main
```
hemildesai: Why is there a duplicate bridge?

hemildesai: We should remove the duplicate path and bring everything inside it to the top level.

PR #264, opened 2025-07-17
Add few api changes required on RL side

yaoyu-33:
1. allow None optimizer
2. allow load megatron model directly on gpu
3. add `external_gpu_device_mapping` to mcore init api
4. fix artifacts saving

### DIFF
--- a/tests/unit_tests/models/hf_pretrained/test_causal_lm.py
+++ b/tests/unit_tests/models/hf_pretrained/test_causal_lm.py
@@ -494,7 +494,6 @@ class TestPreTrainedCausalLMSaveLoad:
         mock_generation_config.save_pretrained.assert_called_once_with(save_dir)
 
 
-
 class TestPreTrainedCausalLMDeviceManagement:
     """Test device management methods."""
 


src/megatron/bridge/training/initialize.py
```diff
@@ -41,6 +41,7 @@ def initialize_megatron(
     skip_mpu_initialization: bool = False,
     get_embedding_ranks: Optional[Callable[[list[int], Optional[int]], list[int]]] = None,
     get_position_embedding_ranks: Optional[Callable[[list[int], Optional[int]], list[int]]] = None,
+    external_gpu_device_mapping: bool = False,
```
ananthsub: n00b question: why add this as an extra arg vs including it in the distributed init config here?  https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/bf8bfcd5ad1688d00b26b82a984b8f0cd51d31a7/src/megatron/bridge/training/config.py#L54-L56

src/megatron/bridge/training/setup.py
```diff
@@ -301,8 +301,9 @@ def _update_model_config_funcs(
         model_config.param_sync_func = [model_chunk.start_param_sync for model_chunk in model]
         if len(model) == 1:
             model_config.param_sync_func = model_config.param_sync_func[0]
-    model_config.finalize_model_grads_func = finalize_model_grads
-    model_config.grad_scale_func = optimizer.scale_loss
+    if optimizer is not None:
```
ananthsub, line 303: when is optimizer not None? at least it's not reflected in the typehint here

yaoyu-33: in RL when we try to load model without optimizer, the optimizer is allowed to be None

ananthsub: i see, it comes from here: https://github.com/NVIDIA-NeMo/RL/blob/51d8006ea1605ad705c5454c66d6987548fb4518/nemo_rl/models/policy/megatron_policy_worker.py#L197-L206

in that case, please update the typehint in L283 to be `Optional[MegatronOptimizer]`.
for later, we should optimize the setup to avoid initializing the optimizer if we don't need it, like evaluation only cases

src/megatron/bridge/training/initialize.py
yaoyu-33: maybe we could, I am not aware it's config class from bridge repo. I thought it was from mcore

yaoyu-33: updated to use dist conf here

PR #265, opened 2025-07-17
Remove `bitsandbytes` dependency from lora

ananthsub:
`bitsandbytes` isn't used for the mcore path - this is holdover from porting from NeMo. The mcore path should use nvfp4 from model opt instead


cuichenx:
if you're trying to remove automodel references, you can also remove `TELinearAdapter`

PR #272, opened 2025-07-18
Adding baichuan causal-bridge + model-providers

marcromeyn:
This PR contains some extra changes to support the model:
  - Added `PyTorchStateSource` class for lazy loading of .bin/.pt checkpoint files, matching SafeTensorsStateSource functionality
  - Enhanced `CausalLMBridge` with optional `bridge_class` parameter for models requiring `trust_remote_code=True`
  - Improved automatic checkpoint format detection in `PreTrainedBase` with fallback logic for both local and HuggingFace Hub models
  - Updated export functionality to handle manual bridges correctly when model architecture classes aren't in standard transformers

  These changes are fully backward compatible and extend bridge support to legacy PyTorch checkpoints and custom model architectures.

### DIFF
--- a/src/megatron/bridge/models/baichuan/baichuan_causal_bridge.py
+++ b/src/megatron/bridge/models/baichuan/baichuan_causal_bridge.py
@@ -1,0 +1,138 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+from megatron.core.models.gpt.gpt_model import GPTModel
+from transformers import AutoModelForCausalLM
+
+from megatron.bridge.models.baichuan.baichuan_mapping import BaichuanQKVMapping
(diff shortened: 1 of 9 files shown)


src/megatron/bridge/models/baichuan/baichuan_causal_bridge.py
```diff
@@ -0,0 +1,138 @@
...
+            seq_length=hf_config.max_position_embeddings,
+            position_embedding_type=position_embedding_type,
+            fp16=(self.dtype_from_hf(hf_config, default=torch.float32) == torch.float16),
+            bf16=(self.dtype_from_hf(hf_config, default=torch.float32) == torch.bfloat16),
+            params_dtype=self.dtype_from_hf(hf_config, default=torch.float32),
+            generation_config=hf_pretrained.generation_config,
+        )
+
+        provider.gradient_accumulation_fusion = False
+        provider.variable_seq_lengths = True
```
ananthsub, line 83: why are these set to true outside of the provider? @maanug-nv was previously running into errors with `variable_seq_lengths` set for the llama causal bridge htiting assertions for the moe dispatcher: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/238

src/megatron/bridge/models/baichuan/baichuan_provider.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    Config for Baichuan2 7B: https://huggingface.co/baichuan-inc/Baichuan2-7B-Base
+    """
+
+    num_layers: int = 32
+    hidden_size: int = 4096
+    ffn_hidden_size: int = 11008
+    num_attention_heads: int = 32
+    num_query_groups: int = 32  # No GQA in 7B model
+    vocab_size: int = 125696
```
ananthsub, line 55: n00b question: why do we set vocab size here vs letting it come from the tokenizer for training or HF config for the bridge?

PR #274, opened 2025-07-19
Update README now that TE is required

maanug-nv:
Given install is not necessarily as simple as 'pip install megatron-bridge', I have moved install instructions to the top, followed by the quickstart.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,9 +33,9 @@ docker run --rm -it -w /workdir -v $(pwd):/workdir \
 
 TransformerEngine is a required dependency for Megatron Bridge. To install on bare metal (without any container), the following system requirements need to be fulfilled:
 
-- PyTorch
-- CUDA
-- cuDNN
+- PyTorch >= 2.7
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,13 +16,51 @@ Megatron Bridge is an extension of NVIDIA's Megatron Core library that enables p
...
 
+### 📦 Bare metal install with TransformerEngine
+
+TransformerEngine is a required dependency for Megatron Bridge. To install on bare metal (without any container), the following system requirements need to be fulfilled:
+
+- Python 3.12
+- PyTorch 2.7
+- CUDA 12.8
+- cuDNN: 9.3
+- Ubuntu 24.04
```
ko3n1g: This will be outdated quickly. Can we point to the Nemo-FW container [release notes](https://docs.nvidia.com/nemo-framework/user-guide/latest/softwarecomponentversions.html#software-component-versions)?

PR #277, opened 2025-07-21
Fix bugs for TP/PP to work properly in conversions

yaoyu-33:
There was some issues with TP Mappings and PP Layer adjustment.
One of the major issue is the global param name which shows the actual layer number which cause issues in both way conversions.
Another issue we fixed is for tied embedding and output layer weights. It needs an extra comm to pass over the shared weights within embd group.

Update example scripts to be able to run tests with TP and PP


.github/workflows/cicd-main.yml
```diff
@@ -121,6 +121,20 @@ jobs:
...
+          - script: L2_Launch_training
+            timeout: 60
+          - script: L2_Multi_GPU_Conversion_TP
+            timeout: 30
+          - script: L2_Multi_GPU_Conversion_PP
+            timeout: 30
+          - script: L2_Generate_From_HF_TP
+            timeout: 30
+          - script: L2_Generate_From_HF_PP
+            timeout: 30
```
ananthsub, line 126: the tests were initially structured such that we wouldn't need an individual script per functional test. this was also done to streamline test times, but i'm not sure if that's an issue now @ko3n1g

examples/models/generate_from_hf.py
```diff
@@ -12,73 +12,236 @@
...
-    """Generate text sequence"""
-    input_ids = tokenizer.encode(prompt, return_tensors="pt")
-    input_ids = input_ids.cuda()
-    position_ids = torch.arange(input_ids.shape[1], device=input_ids.device).unsqueeze(0)
-    attention_mask = torch.ones_like(input_ids, dtype=torch.bool).to(input_ids.device)
+import torch
+import torch.distributed as dist
+from megatron.core import parallel_state
+from megatron.core.pipeline_parallel.schedules import get_forward_backward_func
+from nemo.utils.get_rank import get_last_rank
```
ananthsub: can't have this import. use `megatron.bridge.utils.common_utils.get_last_rank` instead

.github/workflows/cicd-main.yml
yaoyu-33: oh okay, I think we can figure out and update later

src/megatron/bridge/models/param_mapping.py
```diff
@@ -1058,42 +1068,129 @@ def __init__(self, megatron_param: str, gate: str, up: str):
...
+                    else:
+                        raise ValueError(
+                            f"Parameter {param_name} exists but has wrong type or dimensions (expected ndim == {expected_ndim}, got {target_param.ndim if isinstance(target_param, torch.Tensor) else 'not a tensor'})"
+                        )
+                else:
+                    raise ValueError(
+                        f"Parameter name suggests {param_name} but module {megatron_module} has no {param_name}"
+                    )
+
+        raise ValueError(f"Could not determine parameter type for {self.megatron_param}")
```
ananthsub, line 1097: this is mostly copy/paste with `ColumnParallelMapping._get_target_param_and_shape` - worth splitting out to its own helper function?

src/megatron/bridge/training/config.py
```diff
@@ -106,6 +106,13 @@ class DistributedInitConfig:
     overlapped with other computation kernels.
     """
 
+    external_gpu_device_mapping: bool = False
+    """If True, indicates that GPU device mapping has been externally managed
+    (e.g., via CUDA_VISIBLE_DEVICES environment variable). When True, uses device 0
+    instead of local rank for CUDA device selection. This is useful when launching
+    with external process managers that handle GPU visibility.
+    """
+
```
ananthsub: this was already merged in https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/264

yaoyu-33: after merging with main it's gone

src/megatron/bridge/models/param_mapping.py
yaoyu-33: i will check in later prs.

.github/workflows/cicd-main.yml
ko3n1g: I'm not happy about this change. I missed it unfortunately, but we shouldn't have tests that are too fine-grained like this. Especially the TP/PP axis we should have batched together. 

@yaoyu-33 : Can you please correct this with a soon follow up PR?

Issue #216, opened 2025-07-08
Refactor converters to use bridge


ananthsub:
Now that the bridge is merged, the converters should be refactored to avoid duplicate state mapping/config code: https://github.com/NVIDIA-NeMo/Megatron-Hub/blob/main/src/megatron/hub/training/converters/common.py

The importer will still be useful to convert HF models to the Megatron checkpoint format, given a model provider/bridge.
The exporter will still be useful to convert from Megatron checkpoint format to HF models. Currently the bridges do not have any dependency on training/ for the checkpoint format understanding.

maanug-nv:
After this is complete, I'd like to request we have a thorough guide in the docs on how to do 2 cases:
- convert just the HF model, eg for pretraining
- convert HF model and checkpoint format, eg for PEFT

PR #278, opened 2025-07-21
Add megatron save and load method to bridge

yaoyu-33:
Add megatron save and load method to bridge.
- we are able to save a megatron checkpoint
- we are able to load weights and model from megatron checkpoint

These will need training folder

Refactored the converter folder because many are no longer needed

Fixes https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/216

### DIFF
--- a/tests/unit_tests/models/test_causal_bridge.py
+++ b/tests/unit_tests/models/test_causal_bridge.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from unittest.mock import Mock, patch, mock_open
+from unittest.mock import Mock, mock_open, patch
 
 import pytest
 import torch
@@ -682,9 +682,9 @@ class TestCausalLMBridgeEdgeCases:
         with patch("megatron.bridge.training.model_load_save.load_megatron_model") as mock_load_megatron_model:
             with patch("megatron.bridge.utils.instantiate_utils.instantiate") as mock_instantiate:
                 with patch("yaml.safe_load") as mock_yaml_load:
-                    with patch("builtins.open", mock_open(read_data="model:\n  _target_: some.model")) as mock_file:
+                    with patch("builtins.open", mock_open(read_data="model:\n  _target_: some.model")):
                         from pathlib import Path
-                        
+
                         # Mock path.exists() to return True for the config file
                         with patch.object(Path, "exists") as mock_exists:
                             with patch.object(Path, "iterdir") as mock_iterdir:
@@ -694,7 +694,7 @@ class TestCausalLMBridgeEdgeCases:
 
                                 # Mock iterdir to return empty list (no iter_ folders)
                                 mock_iterdir.return_value = []
-                                
+
                                 # Mock exists to return True for config file
                                 mock_exists.return_value = True
 
@@ -717,9 +717,9 @@ class TestCausalLMBridgeEdgeCases:
         with patch("megatron.bridge.training.model_load_save.load_megatron_model") as mock_load_megatron_model:
             with patch("megatron.bridge.utils.instantiate_utils.instantiate") as mock_instantiate:
                 with patch("yaml.safe_load") as mock_yaml_load:
-                    with patch("builtins.open", mock_open(read_data="model:\n  _target_: some.model")) as mock_file:
+                    with patch("builtins.open", mock_open(read_data="model:\n  _target_: some.model")):
                         from pathlib import Path
-                        
+
                         # Create mock folder objects
                         mock_iter_folder_1 = Mock()
                         mock_iter_folder_1.is_dir.return_value = True
@@ -738,7 +738,7 @@ class TestCausalLMBridgeEdgeCases:
 
                                 # Mock iterdir to return the iter folders
                                 mock_iterdir.return_value = [mock_iter_folder_1, mock_iter_folder_2]
-                                
+
                                 # Mock exists to return True for config file
                                 mock_exists.return_value = True
 
@@ -760,13 +760,13 @@ class TestCausalLMBridgeEdgeCases:
         bridge = CausalLMBridge(mock_hf_model)
 
         from pathlib import Path
-        
+
         # Mock path.exists() to return False for the config file and iterdir() to return empty list
         with patch.object(Path, "exists") as mock_exists:
             with patch.object(Path, "iterdir") as mock_iterdir:
                 # Mock iterdir to return empty list (no iter_ folders)
                 mock_iterdir.return_value = []
-                
+
                 # Mock exists to return False for config file (missing)
                 mock_exists.return_value = False
 


src/megatron/bridge/models/causal_bridge.py
```diff
@@ -425,6 +428,124 @@ def save_hf_weights(self, model, path: str | Path, show_progress: bool = True) -
...
+
+        Note:
+            - This method is collective and must be called by all ranks
+            - The saved checkpoint can be loaded with Megatron's checkpoint loading utilities
+            - The checkpoint format follows Megatron's standard structure for compatibility
+        """
+        try:
+            from megatron.hub.training.checkpointing import save_checkpoint
+            from megatron.hub.training.config import CheckpointConfig, ConfigContainer, LoggerConfig
+            from megatron.hub.training.state import GlobalState
```
ananthsub: using old imports

src/megatron/bridge/models/causal_bridge.py
```diff
@@ -425,11 +428,93 @@ def save_hf_weights(self, model, path: str | Path, show_progress: bool = True) -
...
+
+        Note:
+            - This method is collective and must be called by all ranks
+            - The saved checkpoint can be loaded with Megatron's checkpoint loading utilities
+            - The checkpoint format follows Megatron's standard structure for compatibility
+        """
+        try:
+            from megatron.hub.training.model_io import save_megatron_model
+        except ImportError:
+            raise ImportError("megatron.hub.training is not installed.")
```
ananthsub: imports need to be updated

src/megatron/bridge/training/model_io.py
```diff
@@ -0,0 +1,282 @@
...
+        >>> # Save model checkpoint
+        >>> save_megatron_model(megatron_model, "./megatron_checkpoint")
+
+    Note:
+        - This method is collective and must be called by all ranks
+        - The saved checkpoint can be loaded with Megatron's checkpoint loading utilities
+        - The checkpoint format follows Megatron's standard structure for compatibility
+    """
+    # Get model config from the first model instance
+    model_config = get_model_config(model[0])
```
ananthsub: Edit: this isn't the case because get_model_config gets the provider config set during instantiation here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/7fa0b1ffcf9daafe37b941fa4f5cd0f67e880495/src/megatron/bridge/models/gpt_provider.py#L216

https://github.com/NVIDIA/Megatron-LM/blob/94bd476bd840c2fd4c3ebfc7448c2af220f4832b/megatron/core/utils.py#L495-L497

so we don't need to have users pass in the provider config explicitly

src/megatron/bridge/training/converters/common.py
ananthsub, line 1: to fully replace the functionality here, could you add scripts for importing a model to a megatron checkpoint and exporting a megatron checkpoint to HF?

src/megatron/bridge/models/causal_bridge.py
```diff
@@ -425,11 +428,93 @@ def save_hf_weights(self, model, path: str | Path, show_progress: bool = True) -
...
+            ...     wrap_with_ddp=False
+            ... )
+
+        Note:
+            - This method is collective and must be called by all ranks
+            - The checkpoint must have been saved with save_megatron_model
+            - The model architecture must match the bridge configuration
+        """
+        try:
+            from megatron.hub.core.utils.instantiate_utils import instantiate
```
ananthsub:
```suggestion
            from megatron.bridge.utils.instantiate_utils import instantiate
```

src/megatron/bridge/training/model_load_save.py
```diff
@@ -0,0 +1,282 @@
...
+import torch.distributed as dist
+from megatron.core import parallel_state
+from megatron.core.dist_checkpointing.strategies.torch import TorchDistLoadShardedStrategy
+from megatron.core.optimizer import OptimizerConfig
+from megatron.core.utils import get_model_config
+
+from megatron.hub.training.checkpointing import save_checkpoint
+from megatron.hub.training.config import CheckpointConfig, ConfigContainer, LoggerConfig
+from megatron.hub.training.state import GlobalState
+
```
```suggestion
from megatron.bridge.training.checkpointing import save_checkpoint
from megatron.bridge.training.config import CheckpointConfig, ConfigContainer, LoggerConfig
from megatron.bridge.training.state import GlobalState

```

src/megatron/bridge/training/model_load_save.py
```diff
@@ -0,0 +1,282 @@
...
+    if skip_temp_dist_context:
+        return _load_checkpoint()
+    else:
+        # Use appropriate backend based on initialization type
+        backend = "gloo" if use_cpu_init else "nccl"
+        with temporary_distributed_context(backend=backend):
+            return _load_checkpoint()
+
+
+def save_megatron_model(model, path: Union[str, Path], ckpt_format: str = "torch_dist") -> None:
```
```suggestion
def save_megatron_model(model: list[MegatronModule], path: Union[str, Path], ckpt_format: str = "torch_dist") -> None:
```

is there any need to customize the ckpt_format right now? we can always expose this if needed afterwards

src/megatron/bridge/training/model_load_save.py
```diff
@@ -0,0 +1,282 @@
...
+        >>> # Save model checkpoint
+        >>> save_megatron_model(megatron_model, "./megatron_checkpoint")
+
+    Note:
+        - This method is collective and must be called by all ranks
+        - The saved checkpoint can be loaded with Megatron's checkpoint loading utilities
+        - The checkpoint format follows Megatron's standard structure for compatibility
+    """
+    # Get model config from the first model instance
+    model_config = get_model_config(model[0])
```
ananthsub, line 205: as a sanity check, can you add validation that the model config is a model provider? https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/ea0c6ed469122c0907a69574e717e2212de70534/src/megatron/bridge/models/model_provider_mixin.py#L47

being able to instantiate from the checkpoint yaml won't work for megatron models instantiated outside of the provider

src/megatron/bridge/training/model_load_save.py
```diff
@@ -0,0 +1,282 @@
...
+    
+    try:
+        yield
+    finally:
+        if backend == "gloo":
+            parallel_state.destroy_model_parallel()
+            dist.destroy_process_group()
+
+
+def load_megatron_model(
```
ananthsub, line 118: @maanug-nv we can improve this in https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/290 so that these are unified, and we don't have 2 different entry points for loading model states

src/megatron/bridge/training/model_load_save.py
```diff
@@ -0,0 +1,282 @@
...
+        scheduler=None,
+        dataset=None,
+        logger=LoggerConfig(),
+        tokenizer=None,
+        checkpoint=CheckpointConfig(
+            async_save=False, 
+            save=str(path), 
+            save_optim=False, 
+            ckpt_format=ckpt_format
+        ),
```
ananthsub: set `save_rng=False` as well? https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/ea0c6ed469122c0907a69574e717e2212de70534/src/megatron/bridge/training/config.py#L357

src/megatron/bridge/models/causal_bridge.py
```diff
@@ -425,11 +428,93 @@ def save_hf_weights(self, model, path: str | Path, show_progress: bool = True) -
...
+            from megatron.bridge.training.model_load_save import load_megatron_model
+        except ImportError:
+            raise ImportError("megatron.bridge.training is not available.")
+
+        checkpoint_path = Path(path) / "iter_0000000"
```
ananthsub: should the user pass this path in directly? i don't think we should assume the iter 0 subdirectory here

yaoyu-33: I made a support for both

src/megatron/bridge/models/causal_bridge.py
```diff
@@ -425,11 +426,109 @@ def save_hf_weights(self, model, path: str | Path, show_progress: bool = True) -
         if torch.distributed.is_available() and torch.distributed.is_initialized():
             torch.distributed.barrier()
 
-    def save_megatron_model(self, model, path: str | Path, ckpt_format: str = "torch_dist") -> None:
-        raise NotImplementedError
+    def save_megatron_model(self, model, path: str | Path) -> None:
```
ananthsub:
```suggestion
    def save_megatron_model(self, model: list[MegatronModule], path: str | Path) -> None:
```

PR #280, opened 2025-07-22
add data unit tests

dimapihtar:
This PR adds unit tests for data code part.


tests/unit_tests/data/test_loaders.py
```diff
@@ -0,0 +1,72 @@
...
+# limitations under the License.
+
+import json
+
+from megatron.bridge.data.loaders import (
+    get_blend_and_blend_per_split,
+)
+
+
+DATA_PATH = "/opt/data/datasets/train/test_text_document"
```
ananthsub, line 22: is this meant only to run in CI / with the container?

dimapihtar: @ananthsub I don't think so. It should be mounted to the container but don't exist only in container.

@ko3n1g Could you confirm, please?

ko3n1g: This fixture downloads the dataset, both local and in CI: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/dpykhtar/data_unit_tests/tests/unit_tests/conftest.py#L40

Retrospectively, I think two things should be improved:

1. Download into `/tmp` instead of `/opt` 
2. Add a `yield` statement so that its blocking

But yes, this works local and in CI

@dimapihtar : Would you mind adding these changes? If you're busy, I can also raise a follow up PR

dimapihtar: > This fixture downloads the dataset, both local and in CI: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/dpykhtar/data_unit_tests/tests/unit_tests/conftest.py#L40
> 
> Retrospectively, I think two things should be improved:
> 
> 1. Download into `/tmp` instead of `/opt`
> 2. Add a `yield` statement so that its blocking
> 
> But yes, this works local and in CI
> 
> @dimapihtar : Would you mind adding these changes? If you're busy, I can also raise a follow up PR

@ko3n1g yes, could you create that PR, please? Thank you!

ko3n1g: Please merge `ko3n1g/ci/improve-datadownload` into this branch. You can add the argument `ensure_test_data` to your tests, it contains the path to the base directory of the dataset files. Please test this branch in your PR and if everything works as intended, approve this PR: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/310/files

Issue #281, opened 2025-07-22
Support Megatron-LM checkpoint conversion workflow for RL


terrykong:
In [NeMo RL](https://github.com/NVIDIA-NeMo/RL) we want to support native Megatron-LM checkpoints to avoid the user having to convert to a huggingface checkpoint to get started with training. Currently the workflow requires many conversions so we would like to propose a more direct path to avoid needing too many converters. As far as I can tell, this could be done if there was an API for these conversions:

* unidirectional megatron-lm config -> megatron-bridge config ("flat -> modular config" + validation)
* unidirectional megatron-lm config -> HF config (for vllm)

Open to suggestions if the team thinks there's other ways to go about this.

## [Current] NeMo RL workflow (HF as input)
```mermaid
%%{init: {'theme':'base'}}%%
flowchart LR
classDef ckpt fill:#fdd,stroke:#d62728,stroke-width:1px
classDef cfg fill:#dfd,stroke:#2ca02c,stroke-width:1px

subgraph "What user has"
  direction TB
    subgraph "Megatron-LM"
    direction TB
    A_CKPT((ckpt))
  end
end

subgraph "NeMo RL"
  direction TB
  subgraph "HF (input ckpt)"
    direction TB
    HF_CKPT((ckpt))
    HF_CFG((config))
  end
  subgraph "Megatron-Bridge"
    direction TB
    D_CKPT((ckpt))
    D_CFG((config))
  end
  V[VLLM]
end

A_CKPT -->|extract+convert| HF_CFG
A_CKPT -->|convert| HF_CKPT
HF_CKPT -->|convert| D_CKPT
HF_CFG -->|convert| D_CFG
HF_CFG -->|convert+configure| D_CKPT
HF_CFG -->|init| V

class A_CKPT ckpt
class HF_CKPT ckpt
class D_CKPT ckpt
class HF_CFG cfg
class D_CFG cfg
```

## [Proposed] NeMo RL workflow (megatron-lm as input)
```mermaid
%%{init: {'theme':'base'}}%%
flowchart LR
classDef ckpt fill:#fdd,stroke:#d62728,stroke-width:1px
classDef cfg fill:#dfd,stroke:#2ca02c,stroke-width:1px

subgraph "What user has"
  direction TB
    subgraph "Megatron-LM"
    direction TB
    A_CKPT((ckpt))
  end
end

subgraph "NeMo RL"
  direction TB
  subgraph "HF (input)"
[...]

PR #286, opened 2025-07-23
Fallback support to load megatron-lm checkpoints

ananthsub:
part of https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/281 as we work on merging checkpoint formats
this updates the load checkpoint path to handle checkpoint metadata generated by megatron-lm, which today don't have the run_config.yaml or the train state.pt available. 

these values are instead generated from the megatron-lm args in the state dict, or from the metadata tracker. this allows someone to resume training based on an MLM checkpoint

### DIFF
--- a/tests/unit_tests/training/test_checkpointing.py
+++ b/tests/unit_tests/training/test_checkpointing.py
@@ -1344,9 +1344,9 @@ class TestMegatronLMCompatibility:
         assert train_state.skipped_train_samples == 50
         assert train_state.consumed_valid_samples == 10000
         assert train_state.floating_point_operations_so_far == 5000000
-        assert train_state.do_train is True
-        assert train_state.do_valid is True
-        assert train_state.do_test is True
+        assert train_state.do_train is False
+        assert train_state.do_valid is False
+        assert train_state.do_test is False
 
         # Verify checkpoint version was set
         mock_set_version.assert_called_with(3.0)
@@ -1379,9 +1379,9 @@ class TestGetTrainStateFromStateDict:
         assert result.skipped_train_samples == 250
         assert result.consumed_valid_samples == 12000
         assert result.floating_point_operations_so_far == 7500000
-        assert result.do_train is True
-        assert result.do_valid is True
-        assert result.do_test is True
+        assert result.do_train is False
+        assert result.do_valid is False
+        assert result.do_test is False
 
     def test_get_train_state_missing_iteration(self):
         """Test creating TrainState when iteration is missing."""


src/megatron/bridge/training/checkpointing.py
```diff
@@ -283,6 +352,41 @@ def read_run_config(run_config_filename: str) -> dict[str, Any]:
     return config_obj[0]
 
 
+def _extract_megatron_lm_args_from_state_dict(state_dict: dict[str, Any]) -> dict[str, Any]:
+    """Extract and convert legacy Megatron-LM args from checkpoint state_dict to NeMo-LM config format.
```
maanug-nv:
```suggestion
    """Extract and convert legacy Megatron-LM args from checkpoint state_dict to Megatron Bridge config format.
```

src/megatron/bridge/training/checkpointing.py
```diff
@@ -1566,3 +1694,28 @@ def _build_sharded_state_dict_metadata(use_distributed_optimizer: bool, ckpt_ful
...
+        # Fallback if args not found
+        legacy_train_state.consumed_train_samples = 0
+        legacy_train_state.skipped_train_samples = 0
+        legacy_train_state.consumed_valid_samples = 0
+
+    # Extract floating point operations count from state_dict (like Megatron-LM does)
+    legacy_train_state.floating_point_operations_so_far = state_dict.get("num_floating_point_operations_so_far", 0)
+    legacy_train_state.do_train = True
+    legacy_train_state.do_valid = True
+    legacy_train_state.do_test = True
```
maanug-nv: These will be updated later in setup based on what data splits the user configured, correct?

ananthsub: it looks like these won't be updated if the flags are already set: 
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/3a77ad53e2d2360c7a6a527e3658e36902114fb7/src/megatron/bridge/data/loaders.py#L256-L258

i'm not sure what the use case is where `do_train` is set but there's no dataloader or iterations to use from it

maanug-nv: should we set them all to false here so that it's decided by these lines in `loaders.py` ?

ananthsub: - I think checkpointing should set these as False by default
- I also think the `loaders.py` should not rely on the existing train state values at all. what do you think?

maanug-nv: Oh I see. Yeah that makes sense. Agreed, doesn't really make sense to consider value of `do_train` from the checkpoint.

PR #292, opened 2025-07-24
Clean up some functions not getting used anytime soon

yaoyu-33:
After checking with few users the `order` arg and `mode` arg is not used any time soon, which could cause additional confusions. So we remove it for now.

### DIFF
--- a/src/megatron/bridge/models/causal_bridge.py
+++ b/src/megatron/bridge/models/causal_bridge.py
@@ -288,10 +288,6 @@ class CausalLMBridge(Generic[MegatronModelT]):
 
         Args:
             model: Megatron model instance or list of instances
-            order: Export order for weights
-                - "megatron": Follow Megatron's parameter order
-                - "hf": Follow HuggingFace state dict order
-                - "safetensors": Group by safetensors file, then by key
             cpu: Whether to move tensors to CPU before yielding
             show_progress: Display progress bar during export
 


src/megatron/bridge/models/model_bridge.py
```diff
@@ -555,38 +510,24 @@ def stream_weights_megatron_to_hf(
...
-                    # Each rank gets its shard (no gathering)
-                    # In this mode, we need to modify the behavior to skip gathering
-                    # This would require changes to the param mappings themselves
-                    # For now, we'll yield whatever shard this rank has
-                    if task.pp_rank == mpu.get_pipeline_model_parallel_rank() and weight is not None:
-                        # Return the local shard without gathering
-                        yield HFWeightTuple(
-                            task.param_name + f"_rank{mpu.get_tensor_model_parallel_rank()}",
-                            weight.cpu() if cpu else weight,
-                        )
```
ananthsub, line 589: the PR title makes it seem like these have no implementation as the reason for removal. is it that implementations beyond replicate are broken?

yaoyu-33: @ananth, first of all the existing implementation is broken. Even you set the configs like consolidate, dist etc. it's not working as expected.
I consulted CN Dev, this part/func is not needed for any time soon. Instead of fixing all the options, I clean up to one replicate option only.

src/megatron/bridge/models/causal_bridge.py
```diff
@@ -70,8 +69,7 @@ class CausalLMBridge(Generic[MegatronModelT]):
         >>> # Convert weights with custom settings
         >>> for name, weight in bridge.export_hf_weights(
         ...     megatron_model,
-        ...     order="safetensors",
-        ...     mode="consolidate"
+            ...     cpu=True
```
ananthsub:
```suggestion
        ...     cpu=True
```

src/megatron/bridge/models/causal_bridge.py
```diff
@@ -304,10 +294,6 @@ def export_hf_weights(
                 - "safetensors": Group by safetensors file, then by key
             cpu: Whether to move tensors to CPU before yielding
             show_progress: Display progress bar during export
-            mode: Weight distribution mode
-                - "consolidate": Gather to rank 0
-                - "replicate": All ranks get full tensors
-                - "distribute": Each rank keeps its shard
```
ananthsub, line 310: `order` also needs to be removed then

PR #295, opened 2025-07-24
Change TPAwareMapping to AutoMapping

yaoyu-33:
For all plain TP params, we should use AutoMapping for them. Only QKV, Gated and Moe maybe need their own implemenation.

AutoMapping will dispatch automatically.


examples/models/2_way_hf_binding.py
```diff
@@ -69,3 +70,6 @@ def main(hf_model_id: str = HF_MODEL_ID, output_dir: str = None) -> None:
 
     args = parser.parse_args()
     main(args.hf_model_id, args.output_dir)
+
+    if torch.distributed.is_initialized():
+        torch.distributed.destroy_process_group()
```
ananthsub, line 75: not sure if this will works with the functional tests grouped together running in one session

PR #307, opened 2025-07-28
docs: Add contrib guide for tests

ko3n1g:


CONTRIBUTING.md
```diff
@@ -41,6 +41,15 @@ docker run --rm -it -w /workdir -v $(pwd):/workdir \
...
 ```
 
+## 📝 Writing tests
+
+We use [pytest](https://docs.pytest.org/en/stable/) for writing both unit and functional tests.
+
+Unit tests aim to test functions in isolation. They generally do not depend on artifacts like Hugging Face checkpoints or larger datasets. Exception to this is a small toy dataset consisting of tokenizers.  
+Unit tests are stored at `tests/unit_tests`. Please add your test to an existing folder or create a new one if no one matches.
+
+Functional tests are integration tests that perform model training or operate on larger artifacts. We use pytest for writing these. In some cases, it might be desired to run your test (or parts of it) in a subprocess to avoid process contamination. We use `subprocess.Run` for this inside the pytest function. Please add your test into one of the predefined folders. If none of the folders matches semantically, please reach out to the automation team for consultation.
```
maanug-nv: How will an external contributor (outside of nvidia) know who the automation team is for this repo?

yaoyu-33: probably say @ko3n1g  in the comment?

ko3n1g: Thanks, I added our team slug and explained that they should tag us in the PR

PR #287, opened 2025-07-23
Remove CausalLM Bridge, Merge all functions into AutoBridge

yaoyu-33:
Remove the level of CausalLM Bridge. All functions are merged into AutoBridge.
Also removed some names related to causal

PR #315, opened 2025-07-30
Support loading model from MLM checkpoints

maanug-nv:
Modify `load_megatron_model()` to add support for loading the model from an MLM checkpoint.

TODO:
- Rebase on main after #314 is merged
- Update AutoBridge call of `load_megatron_model()` to remove model config instantiation, once #287 is merged (currently [here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/models/causal_bridge.py#L492-L497))

### DIFF
--- a/tests/unit_tests/training/test_model_load_save.py
+++ b/tests/unit_tests/training/test_model_load_save.py
@@ -233,12 +233,12 @@ class TestLoadMegatronModel:
         mock_instantiate.assert_called_once_with(mock_run_cfg_dict["model"])
         mock_cpu_context.assert_called_once()
         mock_model_cfg.provide.assert_called_once()
-        mock_load_weights.assert_called_once_with(ckpt_path, mock_model, return_state_dict=True)
+        mock_load_weights.assert_called_once_with(ckpt_path, [mock_model], return_state_dict=True)
         assert mock_model_cfg.params_dtype == torch.bfloat16
 
         result = load_megatron_model(ckpt_path, return_state_dict=False, use_cpu_init=True)
         assert result == mock_model
-        mock_load_weights.assert_called_with(ckpt_path, mock_model, return_state_dict=False)
+        mock_load_weights.assert_called_with(ckpt_path, [mock_model], return_state_dict=False)
 
     @pytest.mark.parametrize("model_type", ["gpt", "mamba", "resnet"])
     @patch("megatron.bridge.training.model_load_save.temporary_distributed_context")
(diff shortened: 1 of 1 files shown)


src/megatron/bridge/training/model_load_save.py
```diff
@@ -144,6 +143,39 @@ def load_megatron_model(
...
+        get_checkpoint_run_config_filename,
+        read_run_config,
+    )
+    from megatron.bridge.training.mlm_compat.arguments import _load_args_from_checkpoint, _transformer_config_from_args
+    from megatron.bridge.training.mlm_compat.model import _gpt_provider, _mamba_provider
+    from megatron.bridge.utils.instantiate_utils import instantiate
+
+    try:
+        run_config = read_run_config(get_checkpoint_run_config_filename(checkpoint_path))
+        mbridge_ckpt = True
```
ananthsub: why do you need the try/catch? why not explicitly check for the filepath existence before reading?

src/megatron/bridge/training/model_load_save.py
```diff
@@ -144,6 +143,39 @@ def load_megatron_model(
...
+            mlm_args = _load_args_from_checkpoint(checkpoint_path)
+            mbridge_ckpt = False
+        except AssertionError:
+            raise RuntimeError(f"Checkpoint at {checkpoint_path} is not in a supported format.")
+
+    if mbridge_ckpt:
+        model_cfg = instantiate(run_config["model"])
+    else:
+        model_cfg = _transformer_config_from_args(mlm_args)
+        assert model_type in ["gpt", "mamba"], f"model type {model_type} not supported."
```
```suggestion
        assert model_type in ("gpt", "mamba"), f"model type {model_type} not supported."
```

src/megatron/bridge/training/checkpointing.py
```diff
@@ -1011,6 +1011,8 @@ def _load_model_weights_from_checkpoint(
     if torch.distributed.is_initialized():
         torch.distributed.barrier()
 
+    return state_dict
```
ananthsub: update the return type for the function

src/megatron/bridge/training/model_load_save.py
maanug-nv: Initially made sense to me to let `read_run_config()` handle checking if file exists. But i guess there's no need, can just check with `os.path.exists()`. Updating

src/megatron/bridge/training/model_load_save.py
```diff
@@ -156,18 +189,16 @@ def _load_checkpoint():
...
             with megatron_cpu_init_context(model_cfg):
-                model = model_cfg.provide()
+                model = _call_model_provider(model_cfg)
         else:
-            model = model_cfg.provide()
+            model = _call_model_provider(model_cfg)
 
-        strategy = TorchDistLoadShardedStrategy()
-        state_dict = strategy.load(model.sharded_state_dict(), Path(dist_ckpt_folder))
+        state_dict = _load_model_weights_from_checkpoint(checkpoint_path, model)
```
ananthsub: should `_load_model_weights_from_checkpoint` accept `return_state_dict` as an argument? if so, we could skip calling load_state_dict there to preserve the existing behavior

maanug-nv: yeah i was thinking about this as well. was wondering how much time it takes for `_load_model_state_dict()` to run.

maanug-nv: doesn't hurt to add that i guess

PR #318, opened 2025-07-30
Fix CPU Init

yaoyu-33:
For CPU init to work properly it needs to update provider first to make sure the config.use_cpu_init set to True.

Also add skip init while it's going to load the HF weights later.

### DIFF
--- a/tests/unit_tests/models/test_model_provider_mixin.py
+++ b/tests/unit_tests/models/test_model_provider_mixin.py
@@ -50,7 +50,7 @@ def ddp_config():
     return DistributedDataParallelConfig()
 
 
-@patch("megatron.bridge.models.model_provider_mixin.get_model")
+@patch("megatron.bridge.models.model_instantiation.get_model")
 @patch("megatron.bridge.models.model_provider_mixin.parallel_state")
 @patch("megatron.bridge.models.model_provider_mixin.torch.distributed")
 def test_provide_distributed_model_with_hooks_as_args(
@@ -74,7 +74,7 @@ def test_provide_distributed_model_with_hooks_as_args(
     post_hook.assert_called_once_with(mock_model)
 
 
-@patch("megatron.bridge.models.model_provider_mixin.get_model")
+@patch("megatron.bridge.models.model_instantiation.get_model")
 @patch("megatron.bridge.models.model_provider_mixin.parallel_state")
 @patch("megatron.bridge.models.model_provider_mixin.torch.distributed")
 def test_provide_distributed_model_with_registered_hooks(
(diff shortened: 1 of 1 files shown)


src/megatron/bridge/models/model_provider_mixin.py
```diff
@@ -139,8 +138,11 @@ def provide_distributed_model(
         final_pre_wrap_hook = pre_wrap_hook or self.pre_wrap_hook
         final_post_wrap_hook = post_wrap_hook or self.post_wrap_hook
 
+        # Late import to avoid circular dependency
+        from megatron.bridge.models.model_creation import get_model
```
ananthsub: should we just merge these into a single module? especially now that `get_model` takes in the `ModelProviderMixin`

src/megatron/bridge/models/model_creation.py
```diff
@@ -285,39 +280,27 @@ class ModelProviderProtocol(Protocol):
...
     This protocol ensures that model provider classes implement the required
-    get_model method with the correct signature. It's used for type checking
+    provide method with the correct signature. It's used for type checking
     and documentation of the expected interface.
 
     Model providers should implement this protocol to create models compatible
     with Megatron-Core's distributed training infrastructure.
     """
 
-    def get_model(
```
ananthsub, line 295: this must have been broken already since get_model was renamed to `provide_distributed_model`

src/megatron/bridge/models/model_creation.py
```diff
@@ -43,7 +44,7 @@
 
 
 def get_model(
-    model_provider: Callable[[bool, bool, int], nn.Module],
+    model_provider: ModelProviderMixin,
```
ananthsub: (not blocking) this typing isn't exact because `ModelProviderMixin` doesn't have these fields. the subclasses extend from `TransformerConfig` and `ModelProviderMixin` independently. this could be addressed if we have a base/abstract class to do this which all provider implementations inherit from. the same issue exists with nemo2 configs where code relies on `GPTConfig | T5Config` for this typing. 

the motivation for this split design was to accommodate MLM's way of specifying the model provider functionally vs. through the config/mixin

src/megatron/bridge/models/model_provider_mixin.py
yaoyu-33: yeah I like the idea, any other place using it?

src/megatron/bridge/models/model_creation.py
yaoyu-33: removed protocol ...

src/megatron/bridge/models/model_creation.py
yaoyu-33: agree

src/megatron/bridge/models/model_instantiation.py
```diff
@@ -43,7 +44,7 @@
 
 
 def get_model(
-    model_provider: Callable[[bool, bool, int], nn.Module],
+    model_provider: ModelProviderMixin,
```
ananthsub, line 47: the callstack/dependencies don't really make sense anymore. `ModelProviderMixin.provide_distributed_model` calling `get_model` passing itself as an argument is too much indirection. 

I don't think we should have a separate `get_model` / `model_provider.py` module here. IMO this should all be merged with `model_provider_mixin` (maybe we can rename that as `model_provider`)

PR #344, opened 2025-08-04
Update README.md

snowmanwwg:


README.md
```diff
@@ -12,9 +12,9 @@
 
 ## Overview
 
-Megatron Bridge is an extension of NVIDIA's Megatron Core library that enables pretraining and finetuning HuggingFace models using a performant and scalable training loop, with features like model parallelisms and FP8 precision.
+Megatron Bridge is library under [NeMo Framework](https://github.com/NVIDIA-NeMo/.github/blob/main/profile/README.md) that leverages [megatron-core](https://github.com/NVIDIA/Megatron-LM/tree/main/megatron/core) and [transformer engine](https://github.com/NVIDIA/TransformerEngine) to provide state-of-the-art training throughput for popular 🤗Hugging Face models. By leveraging features such as model parallelisms, reduced precision and custom training loop, Megatron Bridge is designed for researchers and community developers who need to train large-scale models efficiently while maintaining flexibility for experimentation and customization.
```
ananthsub: n00b question: why link to the nemo github vs the nemo framework docs or the github organization?

euronymous-aithal: couple of comments:
 - is this a library ? can we just say training infrastructure ? 
 - we shouldn't give the impression that  its only meant for HF models  
 - mixed precision is the right wording

euronymous-aithal: i would also make it explicit that this is a "native Pytorch" only infrastructure just to explicitly distance ourselves from Lightning

README.md
```diff
@@ -97,23 +97,23 @@ torchrun --nproc-per-node=<num devices> /path/to/script.py
 
 ## 🚀 Key Features
```
euronymous-aithal, line 97: please provide explicit links for every bullet where appropriate

README.md
snowmanwwg: It is already linked to the github org readme :)  I plan to add more overviews and the repo diagram there, so all repos will link there for the context on "how is this associated with nemo FW.

snowmanwwg:
```suggestion
Megatron Bridge is a PyTorch native library under [NeMo Framework](https://github.com/NVIDIA-NeMo) that leverages [megatron-core](https://github.com/NVIDIA/Megatron-LM/tree/main/megatron/core) to provide state-of-the-art training throughput for top models. It enables researchers and community developers to do both pre and post training using a performant and scalable training loop, with features like model parallelisms and mixed precisions (FP8, BF16, FP4 etc.).  Megatron Bridge users can either leverage existing 🤗HuggingFace models or define their custom PyTorch model definitions for end-to-end workflows with flexibility.
```

README.md
```diff
@@ -12,9 +12,9 @@
...
 
-Megatron Bridge is designed for researchers and engineers who need to train large-scale models efficiently while maintaining flexibility for experimentation and customization.
+If you already have a 🤗Hugging Face checkpoint, Megatron Bridge provides [conversion tools](https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/src/megatron/bridge/models/conversion) to seamlessly transiton to megatron formats and start training right away! There are also pre-configured recipes and throughput optimized benchmark scripts for you to jumstart your training jobs.
```
snowmanwwg:
```suggestion
```

PR #356, opened 2025-08-06
add sft unit tests

dimapihtar:


tests/unit_tests/data/datasets/test_sft.py
```diff
@@ -0,0 +1,299 @@
...
+            pass
+
+    def test_build_samples_mapping(self, ensure_test_data):
+        dataset = get_gpt_sft(ensure_test_data)
+        dataset._build_samples_mapping()
+
+    def test_gpt_sft_dataset(self, ensure_test_data):
+        dataset = get_gpt_sft(ensure_test_data)
+
+        assert len(dataset) == 100
```
ananthsub: make this a constant? it's not immediately obvious to readers where the value comes from otherwise

Issue #353, opened 2025-08-05
Llama3-70B bf16 pre-training contains several loss spikes in the initial phase


adityavavreNVDA:
We notice lot of loss spikes when pre-training the Llama3-70B model with bfloat16 precision using megatron bridge. This also happens when using a different seed which suggests that it is not a data issue. Comparatively, launching the pre-training with NeMo 2.0 does not have as many loss spikes. The loss value also differs by ~0.2 at 2.8k steps. The loss curves can be found below:

<img width="2122" height="916" alt="Image" src="https://github.com/user-attachments/assets/b52ec895-c127-4472-b021-c1c45ffa29fe" />

The configuration files for both the NeMo 2.0 and megatron bridge runs are also attached. 

[mbridge_config.txt](https://github.com/user-attachments/files/21604877/mbridge_config.txt)

[nemo2_config.txt](https://github.com/user-attachments/files/21604883/nemo2_config.txt)

ericharper:
What's the plan for debugging this?

adityavavreNVDA:
> What's the plan for debugging this?

Right now comparing configs from the two runs and also with the 8B model run since that had no spikes. Once we match the configs we can launch another run to verify if any config diff causes this issue. @ananthsub can add his thoughts.

ananthsub:
our hypothesis is that the loss spike is on checkpoint resume. some differences to note:
- the nemo2 runs were done with the 25.04 nemo container, whereas the mbridge runs were with the 25.07rc5 container. 
- the megatron bridge runs use async saving while disabling the checkpoint persistent workers for writing. this is my suspicion for why there is a loss spike on resume. 

@adityavavreNVDA will re-run using synchronous saving for both on the same container versions to confirm

adityavavreNVDA:
New loss curves with synchronous ckpt saving: 

<img width="1077" height="474" alt="Image" src="https://github.com/user-attachments/assets/3757fcc9-9e5d-4046-9141-11dbf38ffdf7" />

PR #359, opened 2025-08-06
Update async save default for recipes

ananthsub:
- async save with persistent workers is consuming more memory, causing OOMs. update the default for recipes to use synchronous saving
- add assertions to checkpoint config that async save must be used with persistent workers 

Fixes https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/353


maanug-nv:
> async save with persistent workers is consuming more memory
just want to clarify whether this means it's consuming more and more memory each save (leak of some kind), or it's just a fixed amount more than synchronous

ananthsub:
> just want to clarify whether this means it's consuming more and more memory each save (leak of some kind), or it's just a fixed amount more than synchronous

it's a fixed amount of memory on global rank 0

PR #367, opened 2025-08-08
[DRAFT] Add GPT-OSS Support

yaoyu-33:


src/megatron/bridge/models/conversion/param_mapping.py
```diff
@@ -795,7 +795,13 @@ def hf_to_megatron(
         megatron_module: nn.Module,
     ) -> torch.Tensor:
         """Replicate weight to all TP ranks."""
-        target_device = megatron_module.weight.device
+        try:
+            target_device = megatron_module.weight.device
+        except AttributeError:
+            # the parameter may not be called "weight"
+            parameter_name = [k for k, v in megatron_module.named_parameters()][0]
+            target_device = getattr(megatron_module, parameter_name).device
```
ananthsub: `target_device = next(megatron_module.parameters()).device` to avoid the full list creation?

src/megatron/bridge/utils/common_utils.py
```diff
@@ -111,3 +113,60 @@ def print_rank_last(message: str) -> None:
...
+def use_dist_ckpt(ckpt_format: str) -> bool:
+    """Check if the checkpoint format indicates a distributed checkpoint.
+
+    Args:
+        ckpt_format: The checkpoint format string (e.g., "torch", "torch_dist").
+
+    Returns:
+        True if the format is not "torch", False otherwise.
+    """
+    return ckpt_format != "torch"
```
ananthsub: bad rebase? we shouldn't need have these anymore

cuichenx:
Ready for review @yaoyu-33 @ananthsub

src/megatron/bridge/models/conversion/param_mapping.py
```diff
@@ -954,7 +960,11 @@ def hf_to_megatron(
         megatron_module: nn.Module,
     ) -> torch.Tensor:
         """Replicate weight to all TP ranks."""
-        target_device = megatron_module.weight.device
+        try:
```
yaoyu-33: do not use try.
just use hasattr

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -421,10 +458,7 @@ def load_weights_hf_to_megatron(
             if task.megatron_module is None:
                 continue
             # 1) Fetch source tensor(s) from HF state dict
-            if isinstance(task.mapping.hf_param, str):
-                hf_weights = hf_state_dict[task.mapping.hf_param]
-            else:
-                hf_weights = {k: hf_state_dict[v] for k, v in task.mapping.hf_param.items()}
+            hf_weights = self.modify_loaded_hf_weight(task.mapping.hf_param, hf_state_dict)
```
yaoyu-33: maybe_modify

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -583,7 +617,9 @@ def stream_weights_megatron_to_hf(
         embeddings_are_tied = model_config.share_embeddings_and_output_weights
         for task in self._with_progress_tracking(megatron_to_hf_tasks, "Converting to HuggingFace", show_progress):
             converted_weights_dict = task.mapping.megatron_to_hf(task.param_weight, task.megatron_module)
-
+            converted_weights_dict = self.modify_converted_hf_weight(
```
yaoyu-33: call it maybe modify

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -819,15 +857,20 @@ def build_conversion_tasks(
...
-                if isinstance(mapping.hf_param, str):
-                    if mapping.hf_param not in hf_keys:
-                        logger.warning(f"WARNING: Can't find {mapping.hf_param} in hf_keys")
-                        continue
-                else:
-                    missing_params = [hf_param for hf_param in mapping.hf_param.values() if hf_param not in hf_keys]
-                    if missing_params:
-                        logger.warning(f"WARNING: Can't find the following HF parameters in hf_keys: {missing_params}")
-                        continue
+                if not mapping.allow_hf_name_mismatch:
```
yaoyu-33, line 863: what mapping needs this?

src/megatron/bridge/models/conversion/param_mapping.py
```diff
@@ -672,16 +674,24 @@ def gather_from_ep_ranks(
...
-        return {param_name: gathered_weights[i] for i, param_name in enumerate(gathered_expert_param_names)}
+        # this should be in the right order because of the all-gather
+        weights_dict = {}
+        for i, param_name in enumerate(gathered_expert_param_names):
+            if param_name in weights_dict:
+                weights_dict[param_name] = torch.cat(
+                    [weights_dict[param_name], gathered_weights[i].unsqueeze(0)], dim=0
+                )
+            else:
+                weights_dict[param_name] = gathered_weights[i].unsqueeze(0)
```
yaoyu-33, line 695: unsqueeze(0) needed? didnt see this need in qwen3 moe

src/megatron/bridge/models/conversion/model_bridge.py
cuichenx: The quantized expert layers need this.

I considered the following alternative designs

1. Specify both quantized key names and de-quantized key names (earlier commits on this PR)
    - Complication 1 : need to pass in a “quantized” flag to GPTOSSBridge.__init__ or GPTOSSBridge.mapping_registry to decide which key to use (quantized for import, de-quantized for export). This flag needs to be passallow_hf_name_mismatched down from the example export script all the way down to autobridge and modelbridge. This made the code very messy..
    - Complication 2: during export, there is a check to ensure the key names are identical as import, and only matching tensors are written to the exported checkpoint. I tried to overcome this by expanding the logic of the strict flag, but that also required passing strict throughput autobridge and modelbridge.
2. Specify only quantized key names (import names), and infer de-quantized key names during export
    - Complication: same as a2 above. Moreover, during export the name also differed from the Mapping class.
3. Specify only de-quantized key names (export names), and infer quantized key names during import
    - Complication: during import, the de-quantized keys are not found in the cached hf keys, even though it is found in the loaded keys (due to the maybe_modify_converted_hf_weight hook). So the solution is to bypass the assertion with a allow_hf_name_mismatch flag in the mapping class itself, signifying the mapping is using a modified hf key name.
    - Benefit: quantized key names can be easily inferred during the de-quantization step. An added benefit is that the same import code can support both quantized and de-quantized checkpoints.

src/megatron/bridge/models/conversion/param_mapping.py
cuichenx: you're right, this is fixed

cuichenx:
/ok to test [cf20767](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/367/commits/cf2076797e11db243ebf029bc36c0e31ccfaf234)

cuichenx:
/ok to test [d3619c5](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/367/commits/d3619c5fe937a8bcaf65413f6a9c42c6227d7dad)

PR #369, opened 2025-08-09
Add mamba model providers

maanug-nv:
- Base MambaModel ModelProvider class
- NemotronH ModelProvider classes


maanug-nv:
I think we should rename SSMProvider -> MambaProvider and move it under models/mamba. it's more consistent. The name of the provider should indicate which MCore model it will build.

src/megatron/bridge/models/mamba/nemotron_h.py
```diff
@@ -0,0 +1,58 @@
...
+    make_vocab_size_divisible_by: int = 128
+    activation_func: callable = lambda x: torch.pow(F.relu(x), 2)
+    tokenizer_library: str = "tiktoken"
+    tokenizer_name: str = "TiktokenTokenizer"
+    mapping_type: str = "nvidia-hybrid-nemotronh"
+    masked_softmax_fusion: bool = True
+    apply_query_key_layer_scaling: bool = False
+    persist_layer_norm: bool = True
+    attention_softmax_in_fp32: bool = False
+    vocab_size: int = 131072
```
maanug-nv: @JRD971000 do you know why this was hardcoded here in NeMo 2.0? For GPT in this repo, we first build the tokenizer (with padded vocab size) and then copy that to the model config. The same should work for Mamba models, correct?

src/megatron/bridge/models/ssm_provider.py
```diff
@@ -0,0 +1,125 @@
...
+            tokenizer: Tokenizer used with the model
+
+        Returns:
+            MCoreMambaModel: Configured Megatron Core Mamba model instance
+        """
+        mamba_stack_spec = self.mamba_stack_spec
+        if not isinstance(mamba_stack_spec, ModuleSpec):
+            mamba_stack_spec = mamba_stack_spec()
+
+        assert getattr(self, "virtual_pipeline_model_parallel_size", None) is None and vp_stage is None, (
```
maanug-nv, line 93: @JRD971000 any idea of this is still true?

src/megatron/bridge/models/mamba/mamba_provider.py
```diff
@@ -0,0 +1,219 @@
...
+            tokenizer: Tokenizer used with the model
+
+        Returns:
+            MCoreMambaModel: Configured Megatron Core Mamba model instance
+        """
+        mamba_stack_spec = self.mamba_stack_spec
+        if not isinstance(mamba_stack_spec, ModuleSpec):
+            mamba_stack_spec = mamba_stack_spec()
+
+        assert getattr(self, "virtual_pipeline_model_parallel_size", None) is None and vp_stage is None, (
```
maanug-nv, line 94: @JRD971000 any idea of this is still true?

JRD971000: Yeah we stil don't have VPP

src/megatron/bridge/models/mamba/nemotron_h.py
JRD971000: I vaguely remember there were some issues with the Tiktoken at the time. I would try the normal way again, in may have been resolved

src/megatron/bridge/models/mamba/mamba_provider.py
```diff
@@ -0,0 +1,219 @@
...
+    """Configuration for a 130M parameter Mamba model."""
+
+    hybrid_override_pattern: str = "M" * 24
+    num_layers: int = 24
+    seq_length: int = 2048
+    hidden_size: int = 768
+    mamba_num_groups: int = 1
+    ffn_hidden_size: int = 768
+    make_vocab_size_divisible_by: int = 16
+
```
ananthsub, line 140: to confirm, is `mapping_type` dropped because we'll have individual bridges for MambaProvider vs NVIDIAMamba vs NVIDIAMambaHybrid? 

and the existing hardcoded tokenizer settings will be represented in the recipes?

https://github.com/NVIDIA/NeMo/blob/3425e0ce74e17120a60684f8da1d5fe044a10d20/nemo/collections/llm/gpt/model/ssm.py#L862C5-L862C17

src/megatron/bridge/models/mamba/nemotron_h.py
```diff
@@ -0,0 +1,98 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
ananthsub: nit: also name this nemotron_h_provider.py?

ananthsub (commented):
LGTM, will let @JRD971000 confirm

src/megatron/bridge/models/mamba/mamba_provider.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+
+    # Model configuration
+    fp16_lm_cross_entropy: bool = False
+    parallel_output: bool = True
+    share_embeddings_and_output_weights: bool = False
+    params_dtype: torch.dtype = torch.bfloat16
+    fp16: bool = False
+    bf16: bool = True
+    mamba_num_groups: int = 8
```
ananthsub, line 50: missing `num_layers=2` default?

src/megatron/bridge/models/mamba/mamba_provider.py
```diff
@@ -0,0 +1,219 @@
...
+    seq_len_interpolation_factor: Optional[float] = None
+    apply_rope_fusion: bool = True
+    make_vocab_size_divisible_by: int = 128
+    gated_linear_unit: bool = False
+    normalization: str = "RMSNorm"
+    add_bias_linear: bool = False
+    hidden_dropout: float = 0.0
+    attention_dropout: float = 0.0
+    layernorm_epsilon: float = 1e-5
+    attention_backend: AttnBackend = AttnBackend.flash
```
ananthsub, line 69: i couldn't see how `get_attention_mask_from_fusion` is currently used https://github.com/NVIDIA/NeMo/blob/3425e0ce74e17120a60684f8da1d5fe044a10d20/nemo/collections/llm/gpt/model/ssm.py#L202C5-L202C35

@JRD971000 could you share more?

maanug-nv: seems like it's an [arg](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/gpt/data/core.py) for GPTSFTDataset.
We have the same [arg](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/data/datasets/sft.py#L211), maybe we just need to include it [here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/data/datasets/sft.py#L140)

src/megatron/bridge/models/mamba/mamba_provider.py
maanug-nv: i hadn't thought that far ahead actually, i just wasn't including it since not adding any bridge(s) in this PR.

src/megatron/bridge/models/mamba/mamba_provider.py
maanug-nv: not sure if this actually used in nemo 2.0, though or just carried over from nemo 1.0, @JRD971000 any idea?

maanug-nv: synced with @JRD971000 offline, doesn't seem like `get_attention_mask_from_fusion` belongs in model config so it's fine as is. Might want to change default setting in GPTSFTDataset, but we can address that in a separate PR.

ananthsub (approved):

PR #370, opened 2025-08-09
Add end to end test script for CI

ananthsub:
Provide a generic script for training, intended to be used by internal CI systems. Including this in the repo means it will be part of the container build by default. otherwise, the CI system would need to mount this script or build a separate dev container to include the testing script. 

This is not intended to be a generic API for external users to call, which is why it's under the `tests/` directory and not `examples/`

### DIFF
--- a/tests/end_to_end_tests/train_from_recipe.py
+++ b/tests/end_to_end_tests/train_from_recipe.py
@@ -1,0 +1,449 @@
+#!/usr/bin/env python3
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


tests/end_to_end_tests/train_from_recipe.py
```diff
@@ -0,0 +1,457 @@
...
+            profile_step_start=10,
+            profile_step_end=min(12, args.max_steps),
+        )
+
+    return config
+
+
+def main():
+    """Main entry point for the training script."""
+    parser = argparse.ArgumentParser(description="Megatron-Bridge Recipe Training Script")
```
maanug-nv, line 289: nit, can we put parser arguments and setup in a function

maanug-nv: specifically 324-399

maanug-nv (commented):
Can we briefly discuss what are the pros and cons of having this script live in the repo itself vs in internal CI ?

ananthsub:
> Can we briefly discuss what are the pros and cons of having this script live in the repo itself vs in internal CI ?

the aim of adding this to the repo is to avoid one-off container builds for CI that include a dedicated script for testing, and to avoid needing executor-specific mounting logic to run tests on different environments. the tradeoff is that a developer needs to make changes across multiple repos for new features, but this should be comparatively rarer vs adding configs that already fit into the existing arguments

maanug-nv (approved):

PR #381, opened 2025-08-12
Change default for attention mask creation

maanug-nv:
Since a fused attention kernel will be used in most cases (TE is required for Megatron Bridge), update `attention_mask_from_fusion` to avoid creating an attention mask in the batch that will not be used. Users can disable if needed. Applies to SFT only.


JRD971000 (approved):
lgtm

PR #387, opened 2025-08-13
Fix: Bridge Miscellaneous Fixes and API Updates (For Nemo-rl)

yaoyu-33:
This PR contains several miscellaneous fixes and API improvements for the Megatron-Hub bridge system, focusing on better error handling, configuration logic fixes, and API cleanup.

### 🔧 **API Improvements**

**1. Removed Confusing `__call__` Methods**
- **Files**: `src/megatron/bridge/models/model_provider.py`, `src/megatron/bridge/models/conversion/auto_bridge.py`
- **Commit**: `e611c67 - remove __call__ that's confusing`
- **Changes**:
  - Removed `__call__` method from `ModelProviderMixin` class
  - Removed `__call__` method from `AutoBridge` class
  - These methods were causing confusion about the proper API usage and were redundant with existing explicit methods

**2. Updated API for NeMo-RL Use**
- **Files**: Various bridge conversion modules
- **Commit**: `ea1d948 - update api for nemo-rl use`
- **Changes**:
  - Enhanced bridge APIs to better support NeMo-RL integration
  - Added support for pre-built conversion tasks in streaming methods
  - Added `conversion_tasks` parameter to `stream_weights_hf_to_megatron` method in `MegatronModelBridge`

### 🐛 **Bug Fixes**

**3. Fixed Configuration Communication Issues**
- **Files**: `src/megatron/bridge/models/conversion/param_mapping.py`
- **Commit**: `c58ec17 - remove non-picklable objects for config comm`
- **Changes**:
  - Fixed issue where non-picklable objects (`grad_sync_func`, `no_sync_func`) were causing failures during configuration broadcasting across pipeline parallel ranks
  - Added explicit cleanup of these attributes before broadcasting in `QKVMapping.megatron_to_hf()` method

**4. Fixed Hook Execution Logic**
- **Files**: `src/megatron/bridge/models/model_provider.py`
- **Commit**: `f7a005c - fix hook`
- **Changes**:
  - Fixed pre-wrap hook execution logic in `get_model()` function
  - Changed from executing hook on the entire model list to executing on individual model modules
  - Added support for both single callable hooks and list of callable hooks
  - Improved error handling with descriptive error messages

**5. Enhanced Error Messages**
- **Files**: `src/megatron/bridge/utils/instantiate_utils.py`
- **Commit**: `d9f4c8f - fix error msg`
- **Changes**:
  - Improved error message formatting in `instantiate_node()` function
  - Added missing `.{key}` component to make error messages more specific and helpful for debugging

**6. Fixed Data Sequence Length Logic**
- **Files**: `src/megatron/bridge/training/config.py`
- **Commit**: `ce95877 - fix data_seq_length logic`
- **Changes**:
  - Fixed sequence length validation logic in `ConfigContainer.validate()` method
  - Added proper handling for cases where dataset might be `None`
  - Improved the conditional logic for determining `data_seq_length` from different dataset types

### ⚡ **Code Quality Improvements**

**7. Import Cleanup**
- **Files**: `src/megatron/bridge/models/conversion/auto_bridge.py`, `src/megatron/bridge/models/conversion/model_bridge.py`
- **Changes**:
  - Removed unnecessary `TYPE_CHECKING` imports and simplified import structure
  - Fixed method name consistency (`_build_conversion_tasks` → `build_conversion_tasks`)
  - Added proper type imports for better IDE support and type checking

### Affected Components
- **Bridge Conversion System**: Core weight conversion and mapping logic
- **Model Provider API**: Model instantiation and configuration
- **Training Configuration**: Sequence length validation and dataset handling
- **Utility Functions**: Error handling and instantiation logic


src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -312,6 +305,7 @@ def export_hf_weights(
         model: list[MegatronModelT],
         cpu: bool = False,
         show_progress: bool = True,
+        conversion_tasks: Optional[List[WeightConversionTask]] = None,
```
ananthsub, line 308: conversion_tasks isn't passed through to `stream_weights_megatron_to_hf`

src/megatron/bridge/models/model_provider.py
```diff
@@ -460,14 +456,18 @@ def get_model(
     else:
         model = _create_model(model_provider, model_type)
 
-    if pre_wrap_hook:
-        _model = pre_wrap_hook(model)
-        if _model is not None:
-            model = _model
-
```
ananthsub: why does this need to be changed? this will break peft here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/2f18dff6971b977454de128cde338efcff338c33/src/megatron/bridge/training/setup.py#L326-L370

since this will now call load_checkpoint once per model chunk instead of once overall before applying peft over the chunks

src/megatron/bridge/training/config.py
```diff
@@ -825,11 +825,13 @@ def validate(self) -> None:
...
-            self.dataset.seq_length
-            if isinstance(self.dataset, FinetuningDatasetConfig)
-            else self.dataset.sequence_length
-        )
+        if isinstance(self.dataset, FinetuningDatasetConfig):
+            data_seq_length = self.dataset.sequence_length
+        elif self.dataset is not None:
+            data_seq_length = self.dataset.sequence_length
+        else:
+            data_seq_length = self.model.seq_length
```
ananthsub: why do we need this case to work around the assertion in L835?

yaoyu-33: this sets the correct fallback order. There are cases user are using custom dataset, and their dataset has not seq length.

src/megatron/bridge/models/model_provider.py
yaoyu-33: i will change in rl side then. I will just update to allow the hook to be a list of callables.

ananthsub: theoretically this shouldn't need API changes. the user can call `model_provider.register_pre_wrap_hook(...)` multiple times:
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/46c79c67b318f1aa73e279e3ab0c01d5c57f8b23/src/megatron/bridge/models/model_provider.py#L261C18-L261C33

which will be composed here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/46c79c67b318f1aa73e279e3ab0c01d5c57f8b23/src/megatron/bridge/models/model_provider.py#L217-L239

my main questions are:
- does RL need to operate on individual chunks, or can they process the list of chunks together? it feels like an individual chunk is meaningless without also knowing its vp stage
- is repeatedly calling `register_pre_wrap_hook` too inconvenient? we could also add `register_pre_wrap_hooks` to take in a list of hooks. but if it's more convenient to pass in a list of hooks directly, we can support it this way

src/megatron/bridge/training/config.py
ananthsub: is this the issue? https://github.com/NVIDIA-NeMo/RL/blob/e1f56c42ae175d3863ccaf4e21b7de7e9c46c2e1/nemo_rl/models/policy/megatron_policy_worker.py#L608

if so, we could also change the check to only apply if the dataset is not None, and get the sequence length explicitly based on whether it's a finetuning dataset or gpt dataset config

src/megatron/bridge/models/model_provider.py
```diff
@@ -105,7 +105,7 @@ def provide_distributed_model(
         data_parallel_random_init: bool = True,
         use_cpu_initialization: None | bool = False,
         init_model_with_meta_device: bool | None = None,
-        pre_wrap_hook: Callable[[list[MegatronModule]], list[MegatronModule]] | None = None,
+        pre_wrap_hook: Union[Callable[[list[MegatronModule]], list[MegatronModule]], list[Callable[[list[MegatronModule]], list[MegatronModule]]]] | None = None,
         post_wrap_hook: Callable[[list[MegatronModule]], list[MegatronModule]] | None = None,
```
ananthsub: do you need a list for post_wrap_hook too?

yaoyu-33: yeah, RL provide a list of hooks here

PR #389, opened 2025-08-13
Add Mamba and Nemotron-H pretrain recipes

maanug-nv:
part of https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/149

### DIFF
--- a/src/megatron/bridge/recipes/mamba/mamba2_130m.py
+++ b/src/megatron/bridge/recipes/mamba/mamba2_130m.py
@@ -198,6 +198,7 @@ def pretrain_config(
         checkpoint=CheckpointConfig(
             save_interval=2000,
             save=checkpoint_dir,
+            load=checkpoint_dir,
             ckpt_format="torch_dist",
             fully_parallel_load=True,
         ),
--- a/src/megatron/bridge/recipes/mamba/mamba2_1_3b.py
+++ b/src/megatron/bridge/recipes/mamba/mamba2_1_3b.py
@@ -198,6 +198,7 @@ def pretrain_config(
         checkpoint=CheckpointConfig(
             save_interval=2000,
             save=checkpoint_dir,
+            load=checkpoint_dir,
             ckpt_format="torch_dist",
             fully_parallel_load=True,
         ),
--- a/src/megatron/bridge/recipes/mamba/mamba2_2_7b.py
+++ b/src/megatron/bridge/recipes/mamba/mamba2_2_7b.py
@@ -198,6 +198,7 @@ def pretrain_config(
         checkpoint=CheckpointConfig(
             save_interval=2000,
             save=checkpoint_dir,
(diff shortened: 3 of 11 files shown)


src/megatron/bridge/recipes/mamba/mamba2_130m.py
```diff
@@ -0,0 +1,209 @@
...
+            num_workers=1,
+        ),
+        logger=LoggerConfig(
+            log_interval=10,
+            tensorboard_dir=tensorboard_dir,
+        ),
+        tokenizer=TokenizerConfig(tokenizer_type="HuggingFaceTokenizer", tokenizer_model="EleutherAI/gpt-neox-20b"),
+        checkpoint=CheckpointConfig(
+            save_interval=2000,
+            save=checkpoint_dir,
```
ananthsub, line 200: we should also set `load=checkpoint_dir` for the auto-resume if it exists functionality. this is a gap with the existing recipes

IMO we should have a `recipes/utils/pretrain_utils.py` for all of these (src/megatron/bridge/recipes/utils/pretrain_utils.py from https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/378 is starting to introduce)

otherwise it's really tedious making updates to the defaults for existing recipes like in https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/359

src/megatron/bridge/recipes/mamba/mamba2_1_3b.py
```diff
@@ -0,0 +1,209 @@
...
+            # Dataloader config parameters
+            data_sharding=True,
+            dataloader_type="single",
+            num_workers=1,
+        ),
+        logger=LoggerConfig(
+            log_interval=10,
+            tensorboard_dir=tensorboard_dir,
+        ),
+        tokenizer=TokenizerConfig(tokenizer_type="HuggingFaceTokenizer", tokenizer_model="EleutherAI/gpt-neox-20b"),
```
ananthsub, line 197: do we need to pass through `use_fast=True` too? https://github.com/NVIDIA/NeMo/blob/e503a6ef56680bb4342add71202fa2d8176d12ad/nemo/collections/llm/recipes/mamba2_1_3b.py#L47

maanug-nv: how would we do that here? kwargs are only taken in `build_tokenizer()`. 
Also `use_fast` is the default: https://github.com/huggingface/transformers/blob/main/src/transformers/models/auto/tokenization_auto.py#L1015

src/megatron/bridge/recipes/mamba/mamba2_130m.py
maanug-nv: i'll just add this explicitly for now (should be easy to do with bash loop + sed). Let's add 378 after CF ?

PR #390, opened 2025-08-13
Remove `__call__` from AutoBridge

ananthsub:
this API is carryover from merging the CausalLM bridge into the AutoBridge. however, it's not obvious at all that `__call__` should export to HF weights, so I'm removing this API to avoid confusion for users.


ananthsub:
handled in https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/387

PR #394, opened 2025-08-14
Gemma model provider + bridge

ananthsub:
- add model configs for gemma2b, gemma7b, codegemma2b, codegemma7b
- add bridge + conversion tests


ananthsub:
installation test failures are an unrelated issue from flashinfer

src/megatron/bridge/models/gemma/gemma_provider.py
```diff
@@ -0,0 +1,133 @@
...
+
+    num_layers: int = 28
+    hidden_size: int = 3072
+    num_attention_heads: int = 16
+    num_query_groups: int = 16
+    ffn_hidden_size: int = 24576
+
+
+@dataclass
+class CodeGemmaModelProvider2B(GemmaModelProvider2B):
```
yaoyu-33, line 115: All start from `GemmaModelProvider` plz, do not create more levels here. Same below

src/megatron/bridge/models/gemma/gemma_provider.py
```diff
@@ -0,0 +1,141 @@
...
+    activation_func: Callable = openai_gelu
+    gated_linear_unit: bool = True
+    position_embedding_type: str = "rope"
+    add_bias_linear: bool = False
+    seq_length: int = 8192
+    kv_channels: int = 256
+    attention_dropout: float = 0.0
+    hidden_dropout: float = 0.0
+    share_embeddings_and_output_weights: bool = True
+    # Note: different behavior compared to NeMo 1.0
```
suiyoubi, line 42: I wonder if we still needs this comment in mbridge since nemo1 is quite a long time ago

src/megatron/bridge/models/gemma/gemma_provider.py
```diff
@@ -0,0 +1,141 @@
...
+
+    num_layers: int = 28
+    hidden_size: int = 3072
+    num_attention_heads: int = 16
+    num_query_groups: int = 16
+    ffn_hidden_size: int = 24576
+
+
+@dataclass
+class CodeGemmaModelProvider2B(GemmaModelProvider):
```
suiyoubi: can this subclass the `GemmaModelProvider2B ` since the arch is identical

src/megatron/bridge/models/gemma/gemma_provider.py
```diff
@@ -0,0 +1,141 @@
...
+
+    num_layers: int = 18
+    hidden_size: int = 2048
+    num_attention_heads: int = 8
+    num_query_groups: int = 1
+    ffn_hidden_size: int = 16384
+
+
+@dataclass
+class CodeGemmaModelProvider7B(GemmaModelProvider):
```
suiyoubi: can this subclass the `GemmaModelProvider2B ` since the arch is identical

src/megatron/bridge/models/gemma/gemma_provider.py
ananthsub: @yaoyu-33 what do you think? i had this previously but the suggestion was to avoid the extra inheritance: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/394#discussion_r2298992780

src/megatron/bridge/models/gemma/gemma_provider.py
ananthsub: same as above, https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/394#discussion_r2298992780

PR #401, opened 2025-08-15
Add gpt3 175b pretrain recipe

ananthsub:


src/megatron/bridge/recipes/gpt/gpt3_175b.py
```diff
@@ -0,0 +1,228 @@
...
+        pipeline_parallelism_dtype=pipeline_parallelism_dtype,
+        virtual_pipeline_parallelism=virtual_pipeline_parallelism,
+        context_parallelism=context_parallelism,
+        sequence_parallelism=sequence_parallelism,
+    )
+
+    opt_config, scheduler = distributed_fused_adam_with_cosine_annealing(
+        lr_warmup_iters=lr_warmup_iters,
+        lr_decay_iters=train_iters,
+        max_lr=lr,
```
maanug-nv, line 154: in the nemo 2.0 adam optimizer recipe, min_lr is optional, and defaults to 0.1 * max_lr . ours does not do this. we can either add the same default behavior or manually set min_lr appropriately (so 0.9e-5 here).

src/megatron/bridge/recipes/gpt/gpt3_175b.py
```diff
@@ -0,0 +1,228 @@
...
+    )
+    opt_config.use_precision_config = False
```
maanug-nv: Is this supposed to be `use_precision_aware_optimizer` , I don't see a `use_precision_config` in OptimizerConfig?

tests/unit_tests/recipes/gpt/test_gpt3_175b.py
```diff
@@ -0,0 +1,252 @@
...
+    def test_pretrain_config_precision_configuration(self):
+        """Test precision configuration for GPT3 175B."""
+        config = pretrain_config()
+
+        # Should have precision config
+        assert config.mixed_precision is not None
+        # Grad reduce should be optimized for performance
+        assert config.mixed_precision.grad_reduce_in_fp32 is False
+
+    def test_pretrain_config_gpt3_specific_optimizations(self):
```
maanug-nv: This test is not adding any new assertions, just repeating things already asserted above. i think it can be removed.

PR #403, opened 2025-08-15
Malay/perf scripts

malay-nagda:


scripts/performance/llm/pretrain_llama3_8b.py
```diff
@@ -0,0 +1,77 @@
...
+        exp.add(
+            pretrain(
+                config=recipe(
+                    mock=True, global_batch_size=128, precision_config=precision_config),
+                forward_step_func=forward_step,
+            ),
+            executor=executor,
+            name = exp_name,
+            plugins=plugins,
+        )
```
ananthsub: if you want to launch directly like this, you'll need to wrap the pretrain function like this instead: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/92955f30e567212bccfc857bbcd2615bbb8bffcf/examples/recipes/llama/pretrain_llama3_8b_nemo_run_partial.py#L60-L62

alternatively, you can use the run.Script option: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/92955f30e567212bccfc857bbcd2615bbb8bffcf/examples/recipes/llama/pretrain_llama3_8b_nemo_run_script.py#L135-L146

malay-nagda: Thanks!

src/megatron/bridge/recipes/utils/to_config.py
```diff
@@ -0,0 +1,123 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
ananthsub: same with this, plus i want to avoid confusion with other configs mentiond in the repo.

src/megatron/bridge/recipes/utils/fdl_torch.py
```diff
@@ -0,0 +1,122 @@
+"""Fiddle extensions to handle PyTorch code more elegantly.
```
ananthsub: i'd prefer to include this all under nemo_run_utils.py - that way we can also handle the fiddle dependency to one place for users, especially those who don't want to depend on nemo-run. what do you think?

pyproject.toml
```diff
@@ -88,7 +88,7 @@ dependencies = [
     "megatron-core>=0.14.0a0,<0.15.0",
     "nvidia-modelopt[torch]>=0.33.0a0,<0.34.0; sys_platform != 'darwin'",
     "nvidia-resiliency-ext>=0.4.0a0,<0.5.0; sys_platform != 'darwin'",
-    "transformer-engine[pytorch]>=2.6.0a0,<2.8.0; sys_platform != 'darwin'"
+    # "transformer-engine[pytorch]>=2.6.0a0,<2.8.0; sys_platform != 'darwin'"
```
ananthsub: what issues were you having with installing TE?

malay-nagda: I use pyproject to install dependencies in a venv on login node- we cannot install TE there...

I will remove it when I open this PR for review....

src/megatron/bridge/recipes/utils/fdl_torch.py
malay-nagda: yup. makes sense. 
instead of having a giant monolithic file. can we have a `nemo_run_utils` dir with-

- nemo_run_utils.py (we can rename this or dirname)
- fdl_torch.py
- to_config.py
?

ananthsub: sure, that makes sense. still, we should guard each of the imports expected to come from nemo-run as we don't explicitly depend on fiddle otherwise for the repo

scripts/performance/llm/run_script.py
```diff
@@ -0,0 +1,188 @@
...
+
+    if args.model_name == "llama3" and args.model_size == "8b":
+        recipe = llama3_8b_pretrain_config(mock=True, precision_config=precision_config)
+    elif args.model_name == "llama3" and args.model_size == "70b":
+        recipe = llama3_70b_pretrain_config(mock=True, precision_config=precision_config)
+    elif args.model_name == "llama31" and args.model_size == "405b":
+        recipe = llama31_405b_pretrain_config(mock=True, precision_config=precision_config)
+    else:
+        raise ValueError(f"Model {args.model_name} {args.model_size} not supported")
+
```
ananthsub: nit: this style of import could make it more generic: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/eb25f11bd8a479a104a8eea8f82db059e9cfe31f/tests/end_to_end_tests/train_from_recipe.py#L378-L393

src/megatron/bridge/recipes/llama/llama31_405b.py
```diff
@@ -212,7 +212,7 @@ def pretrain_config(
             log_interval=10,
             tensorboard_dir=tensorboard_dir,
         ),
-        tokenizer=TokenizerConfig(tokenizer_type="NullTokenizer", vocab_size=DEFAULT_NULL_TOKENIZER_VOCAB_SIZE),
+        tokenizer=TokenizerConfig(tokenizer_type="NullTokenizer", vocab_size=vocab_size),
```
ananthsub, line 215: you'll need to update the recipe unit tests for this:

for example: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/eb25f11bd8a479a104a8eea8f82db059e9cfe31f/tests/unit_tests/recipes/llama/test_llama31_8b.py#L342-L347

malay-nagda: Done!

scripts/performance/llm/run_script.py
malay-nagda: makes sense. I'll add this in a follow-up PR.

PR #412, opened 2025-08-16
Add functional tests for recipes

ananthsub:
These are intended to be short end to end tests (train with mock data, eval, save checkpoint) to validate training doesn't crash. this should be configured only for models that can fit on the 2 GPUs available from github actions

### DIFF
--- a/tests/functional_tests/recipes/test_mamba_recipes.py
+++ b/tests/functional_tests/recipes/test_mamba_recipes.py
@@ -16,7 +16,6 @@
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


liding-nv:
some tests failed due to OOM, i.e.,
`
  FAILED tests/functional_tests/recipes/test_mamba_recipes.py::TestMambaRecipes::test_mamba_pretrain_recipes[pretrain_config-mamba2_1_3b] - torch.OutOfMemoryError: CUDA out of memory. 
`
i think it is okay to skip 3b model for the purpose of this test

liding-nv (approved):

PR #447, opened 2025-08-20
Refactor inference model load 

maanug-nv:
- Enables overriding parallelisms
- Enables vp

### DIFF
--- a/src/megatron/bridge/training/model_load_save.py
+++ b/src/megatron/bridge/training/model_load_save.py
@@ -270,7 +270,7 @@ def build_and_load_model(
             model = _call_model_provider(model_cfg)
 
         maybe_state_dict = _load_model_weights_from_checkpoint(
-            checkpoint_path, [model], return_state_dict=return_state_dict
+            checkpoint_path, model, return_state_dict=return_state_dict
         )
 
         if return_state_dict:
--- a/src/megatron/bridge/training/model_load_save.py
+++ b/src/megatron/bridge/training/model_load_save.py
@@ -210,7 +210,7 @@ def build_and_load_model(
     model_type: Optional[Literal["gpt", "mamba"]] = None,
     megatron_args: Optional[argparse.Namespace] = None,
     return_state_dict: bool = False,
-    use_cpu_init: bool = True,
+    use_cpu_init: bool = False,
     skip_temp_dist_context: Optional[bool] = None,
 ) -> Union[Any, dict[str, torch.Tensor]]:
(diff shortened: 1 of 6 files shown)


src/megatron/bridge/training/model_load_save.py
```diff
@@ -211,15 +200,58 @@ def load_megatron_model(
...
+    Args:
+        checkpoint_path: path to an MCore distributed checkpoint directory. Used only for
+                    model weights (e.g., /path/to/model/checkpoints/iter_0000001).
+        model_cfg: Model config from load_model_config(). Either a TransformerConfig or
+            a model provider (e.g. GPTModelProvider) depending on source of checkpoint.
+        model_type: If the checkpoint is from MegatronLM, the model type is required. Currently,
+            only GPT and Mamba models are supported.
+        megatron_args: If the checkpoint is from MegatronLM, this is required.
+        return_state_dict: If True, return the state dict instead of model instance. Default: False.
+        use_cpu_init: If True, use CPU initialization context for the model and Gloo backend.
```
pthombre, line 231: Can we set the default value of use_cpu_init to be False? 

Most usecases of this API will require GPU usage

maanug-nv: fair point. will change

src/megatron/bridge/training/model_load_save.py
```diff
@@ -157,49 +158,37 @@ def load_tokenizer(checkpoint_path: str) -> MegatronTokenizer:
     return build_tokenizer(cfg, vocab_size_divisor, tp_size)
 
 
-def load_megatron_model(
+def load_model_config(
```
yaoyu-33, line 161: plz update in AutoBridge as well? since the api changed here

maanug-nv: `load_megatron_model()` still [exists](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/447/files#diff-9651e8f8d9a1d7212ea707caad37f91824ef7773b48274f2e2ea506d1e6e3218R291). From AutoBridge's perspective, I don't think the API has changed.

This PR is splitting up `load_megatron_model()` logic, but keeping the API that is used by AutoBridge. Conversion tests are passing so I think it's still working.

pthombre:
Confirmed that MBridge model load is working with/without custom parallelism in the inference path.

MLM model load is currently working without parallelism, but running into errors with custom parallelism parameters.

pthombre:
Verified MLM loading with various parallelisms as well.

PR #455, opened 2025-08-22
fix: padded vocab size handling in setup

ananthsub:
1. If the user explicitly sets `vocab_size` on the model, this takes the top precedence. In this case, we validate that:
- the vocab size is greater than the tokenizer's vocab size
- the vocab size meets the padding/TP requirements for divisibility

2. Otherwise, if the model vocab size is not explicitly set, we use the padded vocab size generated when building the tokenizer. Currently, if the model vocab size is not set, we set the vocab size using the unpadded tokenizer vocab size. This results in never calling the padding logic [here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/fed237de248543d1e206904b89feca5a071d1964/src/megatron/bridge/models/gpt_provider.py#L201-L202). This case also matches the [MLM behavior](https://github.com/NVIDIA/Megatron-LM/blob/c6aab54e9f1021009d79946ca97bf3426d4d28d8/megatron/training/tokenizer/tokenizer.py#L108-L128)


src/megatron/bridge/training/utils/flop_utils.py
```diff
@@ -79,6 +80,13 @@ def num_floating_point_operations(cfg: ConfigContainer, batch_size: int) -> floa
             # Shared Experts.
             + ((shared_expert_ffn_hidden_size / cfg.model.hidden_size) * gated_linear_multiplier)
             # Logit.
-            + (cfg.tokenizer.padded_vocab_size / (2 * cfg.model.num_layers * cfg.model.hidden_size))
+            + (_get_padded_vocab_size(cfg) / (2 * cfg.model.num_layers * cfg.model.hidden_size))
         )
     )
+
+
+def _get_padded_vocab_size(cfg: ConfigContainer) -> int:
```
maanug-nv: nit: take cfg.model instead of cfg

maanug-nv: same for theoretical memory utls

src/megatron/bridge/training/model_load_save.py
```diff
@@ -250,6 +247,13 @@ def _call_model_provider(model_cfg):
         else:
             assert model_type in ("gpt", "mamba"), f"model type {model_type} not supported."
             assert megatron_args is not None, "megatron_args must be provided if the checkpoint is from MegatronLM."
+
+            # Re-calculate the padded vocab size based on the model config instead of the args from the checkpoint
+            megatron_args.padded_vocab_size = calculate_padded_vocab_size(
+                megatron_args.vocab_size,
+                model_cfg.make_vocab_size_divisible_by,
```
maanug-nv: I think this is not an attribute of TransformerConfig or ModelParallelConfig. So it needs to come from mlm args

PR #465, opened 2025-08-22
Use codecov placeholder if docs only for a PR

chtruong814:
Use codecov placeholder if docs only for a PR


ko3n1g (changes requested):
This was is problematic. In case codecov doesn’t run (I.e. when no Head commit is found), the skipped codecov mock check will still satisfy the PR and engineers can merge. 

Skipping codecov on docs-only is not trivial, but this implementation I had before and I removed it after learning that required status checks also satisfy on skipped jobs

ko3n1g:
Maybe this works better: https://community.codecov.com/t/i-would-like-to-skip-the-codecov-when-source-code-when-files-unrelated-to-the-tests-are-modified/4433

PR #466, opened 2025-08-22
Fix broken image from README

chtruong814:
Fix broken image from README.

The image was uploaded in a separate PR.


ko3n1g (changes requested):
This is problematic. In case codecov doesn’t run (I.e. when no Head commit is found), the skipped codecov mock check will still satisfy the PR and engineers can merge. 

Skipping codecov on docs-only is not trivial, but this implementation I had before and I removed it after learning that required status checks also satisfy on skipped jobs

ko3n1g (approved):

chtruong814:
This may not be necessary
https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/460/files

PR #467, opened 2025-08-22
Add symlink of repo image to docs root

chtruong814:
Add symlink of repo image to docs root

During a docs build, it includes the README, which references this image. However, it could not find it because it's looking relative to the docs directory now.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -27,9 +27,7 @@ jobs:
 
   build-docs:
     needs: [pre-flight]
-    if: |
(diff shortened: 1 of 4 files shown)


.github/workflows/build-docs.yml
```diff
@@ -27,9 +27,7 @@ jobs:
 
   build-docs:
     needs: [pre-flight]
-    if: |
-      !(needs.pre-flight.outputs.docs_only == 'true'
```
ko3n1g, line 31: Why this change?

ko3n1g (approved):
Nvm, thanks for fixing this

Issue #150, opened 2025-06-30
Llama Nemotron -Super 1.5 Support


ananthsub:
Support Llama Nemotron variants for pretrain/SFT/PEFT

- [x] Model providers (porting configs from nemo2):
- [x] Model bridge for HF conversion + functional tests
- [ ] Pretraining recipe: 
- [ ] Finetune recipe
- [ ] End to end testing in CI
- [ ] Updated docs

terrykong:
@ananthsub someone is asking for `nvidia/Llama-3_3-Nemotron-Super-49B-v1_5` and running into an issue using it from mbridge. looks like it was supported in nemo2, so afaiu, we just need converter support? https://docs.nvidia.com/nemo-framework/user-guide/latest/llms/llama_nemotron.html#nemo-2-0-finetuning-recipes

related: https://github.com/NVIDIA-NeMo/RL/issues/1134

ananthsub:
here's the PR for the Llama Nemotron converter: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/678

I'll create a separate issue for Llama Nemotron to keep this for Nemotron3/4 support

PR #485, opened 2025-08-25
Nemotron model provider

ananthsub:
Split out model provider from https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/378 for easier review

part of #150

### DIFF
--- a/src/megatron/bridge/models/nemotron/nemotron_provider.py
+++ b/src/megatron/bridge/models/nemotron/nemotron_provider.py
@@ -1,0 +1,142 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


src/megatron/bridge/models/nemotron/nemotron_provider.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+
+    num_layers: int = 32
+    seq_length: int = 4096
+    hidden_size: int = 3072
+    ffn_hidden_size: int = 9216
+    num_attention_heads: int = 24
+    num_query_groups: int = 8
+    kv_channels: Optional[int] = 128
+    init_method_std: float = 0.0134
```
suiyoubi, line 84: More like a general question for Mbridge: Do we need to put vocab_size in the `Provider` ?

ananthsub: vocab size doesn't need to be set in the provider config for training. if it's not set, the vocab size used is from the tokenizer. however, if the vocab size is explicitly set on the provider, then that is used (similar to nemo2)

code reference:
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/f37f883a8c53a959e6e0cc3c8b6bdd3261261e4d/src/megatron/bridge/training/setup.py#L385-L418

I'll be sure to address in the docs for training

suiyoubi: Sounds good thanks for the explaination. I'll aprove this PR 👍

PR #489, opened 2025-08-26
default to using 8 workers for dataloading

ananthsub:
the default for the dataloading config is to use 8 workers: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/eb25f11bd8a479a104a8eea8f82db059e9cfe31f/src/megatron/bridge/training/config.py#L151-L153

to match default workers used for mock data/recipes in nemo2: https://github.com/NVIDIA/NeMo/blob/449b7df1b88f7e6cbd8eb62a3300b60c8f565645/nemo/collections/llm/gpt/data/mock.py#L64

so we should remove the override set in these recipes

### DIFF
--- a/src/megatron/bridge/recipes/deepseek/deepseek_v2.py
+++ b/src/megatron/bridge/recipes/deepseek/deepseek_v2.py
@@ -218,6 +218,7 @@ def pretrain_config(
             # Dataloader config parameters
             data_sharding=True,
             dataloader_type="single",
+            num_workers=8,
         ),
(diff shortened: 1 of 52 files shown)


src/megatron/bridge/recipes/deepseek/deepseek_v2.py
```diff
@@ -218,7 +218,6 @@ def pretrain_config(
             # Dataloader config parameters
             data_sharding=True,
             dataloader_type="single",
-            num_workers=1,
```
malay-nagda, line 221: can we explicitly set to 8 here? that way we use this even if someone changes the default for whatever reason @ananthsub

malay-nagda (approved):

PR #496, opened 2025-08-27
Rename wrap_cast_to_fp32 => wrap_cast_model_output_to_fp32 and fix condition.

terrykong:
Mirroring change that was introduced here https://github.com/NVIDIA-NeMo/RL/pull/918

### DIFF
--- a/src/megatron/bridge/models/model_provider.py
+++ b/src/megatron/bridge/models/model_provider.py
@@ -435,6 +435,7 @@ def get_model(
         list[Callable[[list[MegatronModule]], list[MegatronModule]]],
     ]
     | None = None,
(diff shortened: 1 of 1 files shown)


src/megatron/bridge/models/model_provider.py
```diff
@@ -435,6 +435,7 @@ def get_model(
         list[Callable[[list[MegatronModule]], list[MegatronModule]]],
     ]
     | None = None,
+    wrap_cast_model_output_to_fp32: bool = True,
```
ananthsub, line 438: @terrykong unfortunately the code at the moment is more duplicated that it needs to be. this argument needs to be made passed from `provide_distributed_model`. it should also be added to `GetModelKwargs`

PR #499, opened 2025-08-27
feat: in process restart integration

ananthsub:
https://nvidia.github.io/nvidia-resiliency-ext/inprocess/index.html

enable low-overhead restarts inside the same process for distributed PyTorch workloads. This PR adds dedicated configuration for in-process restart, internal wrapping of the training entrypoint, explicit lifecycle hooks for persistent async checkpoint workers, and proper distributed init interactions.

Internally, pretrain conditionally wraps a new private `_pretrain` function when config.inprocess_restart.enabled is true, passing state, the NVRx wrapper, and a restart Store. the state is needed in order to have a reference to explicitly abort the persistent async workers as part of the termination sequence.

as part of this, the pretrain/setup API is refactored:
the config validation + runtime overrides are applied at the beginning of pretrain, instead of setup. this is so that we can initialize the global state object and pass it for when the new private function is wrapped.

the persistent async worker initialization is also slightly refactored: we add an explicit method to the global state class for initializing these workers rather than lazily creating the calls queue on its first access. this allows us to remove the init_async_checkpoint_worker function from checkpointing.py

### DIFF
--- a/src/megatron/bridge/training/pretrain.py
+++ b/src/megatron/bridge/training/pretrain.py
@@ -12,13 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Callable
+from typing import Callable, Optional
+
+import torch.distributed as dist
+
+
+try:
+    from nvidia_resiliency_ext.inprocess import CallWrapper
+except ImportError:
+    CallWrapper = type(None)
 
 from megatron.bridge.data.utils import get_dataset_provider
 from megatron.bridge.training.checkpointing import save_checkpoint
 from megatron.bridge.training.config import ConfigContainer
 from megatron.bridge.training.eval import evaluate_and_print_results
+from megatron.bridge.training.mixed_precision import get_mixed_precision_config
 from megatron.bridge.training.setup import setup
+from megatron.bridge.training.state import GlobalState
 from megatron.bridge.training.train import _finish_train, train
 from megatron.bridge.training.utils.log_utils import barrier_and_log
 from megatron.bridge.utils.common_utils import print_rank_0
@@ -30,7 +40,7 @@ def pretrain(
     config: ConfigContainer,
     forward_step_func: Callable,
 ) -> None:
-    """Main function to run the pretraining pipeline.
+    """Main function to run the training pipeline.
 
     Sets up the environment, model, optimizer, scheduler, and data iterators.
     Performs training, validation, and optionally testing based on the provided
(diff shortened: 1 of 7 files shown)


marcromeyn:
> as part of this, the pretrain/setup API is refactored:
the config validation + runtime overrides are applied at the beginning of pretrain, instead of setup. this is so that we can initialize the global state object and pass it for when the new private function is wrapped.

Should this be seen as a breaking change or is it hidden from the user?

examples/resiliency/inprocess_restart.py
```diff
@@ -0,0 +1,272 @@
+#!/usr/bin/env python3
```
marcromeyn: Are we testing this example?

ananthsub:
> > as part of this, the pretrain/setup API is refactored:
> > the config validation + runtime overrides are applied at the beginning of pretrain, instead of setup. this is so that we can initialize the global state object and pass it for when the new private function is wrapped.
> 
> Should this be seen as a breaking change or is it hidden from the user?

The changes are to the setup function, which are hidden from the user

examples/resiliency/inprocess_restart.py
ananthsub: no it's not run through CI, similar to https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/examples/recipes/llama

marcromeyn: Could we have an integration test for this functionality then?

src/megatron/bridge/training/state.py
```diff
@@ -304,6 +307,22 @@ def _set_signal_handler(self) -> None:
...
+        self._timers = None
+        self._train_state = None
+        self._tensorboard_logger = None
+        self._wandb_logger = None
+        self._energy_monitor = None
+        self._energy_monitor_created = False
+        self._signal_handler = None
+        self._straggler_timer = None
+        self._nvrx_straggler_manager = None
+        self._nvrx_straggler_created = False
```
maanug-nv, line 325: does `self._async_calls_queue` need to be cleaned up here? we wouldn't want to finish saving checkpoints (from the crashed run) after restarting, right?

maanug-nv: hadn't read through other changes. i can see it is destroyed in 'inprocess_restart.py' through an Abort object. can you point me to any documentation for that, trying to understand abort vs finalize etc.

src/megatron/bridge/training/pretrain.py
```diff
@@ -45,8 +49,50 @@ def pretrain(
...
+
+    Args:
+        state: Global training state containing the validated configuration and runtime objects
+        forward_step_func: Function that performs a single forward/backward step
+        store: Optional distributed Store used by in-process restart for coordination
+        inprocess_call_wrapper: Optional wrapper injected by nvrx to expose restart iteration
+    """
+    # Handle in-process restart store prefix
+    if inprocess_call_wrapper is not None:
+        iteration = inprocess_call_wrapper.iteration
```
maanug-nv: nit: can we use a different variable name, iteration already exists in this function, and i believe this one is not referring to train iters. maybe `restart_attempt` ?

maanug-nv:
This feature needs documentation. Totally fine to do that in a separate PR, but please create an issue to track.

tests/functional_tests/L2_Launch_training.sh
```diff
@@ -17,5 +17,37 @@ set -xeuo pipefail # Exit immediately if a command exits with a non-zero status
...
+    # Run fault injection test with ft_launcher
+    ft_launcher \
+      --rdzv_backend=c10d --rdzv_endpoint=127.0.0.1:29501 \
+      --nnodes=1 --nproc-per-node=2 \
+      --ft-param-rank_section_timeouts=setup:600,step:180,checkpointing:420 \
+      --ft-param-rank_out_of_section_timeout=300 \
+      --monitor-interval=5 --max-restarts=3 \
+      --ft-restart-policy=min-healthy \
+      -m coverage run --data-file=/workspace/.coverage --source=/workspace/ --parallel-mode -m pytest -o log_cli=true -o log_cli_level=INFO -v -s -x -m "not pleasefixme" --tb=short -rA \
+      tests/functional_tests/training/test_inprocess_restart.py::TestInProcessRestartIntegration::test_inprocess_restart_with_fault_injection
```
maanug-nv: why can't we just use 1 command with '`tests/functional_tests/training/test_inprocess_restart.py` ? That way we can also add more test_* functions without needing new commands in this script

src/megatron/bridge/training/state.py
ananthsub: you can see the code ordering here: https://nvidia.github.io/nvidia-resiliency-ext/inprocess/usage_guide.html#implementation-overview

abort is called before finalize, so the async calls queue and the persistent workers are terminated first in the abort section before the remainder of the state is reset in the finalize portion.

maanug-nv (approved):

PR #505, opened 2025-08-27
feat: Add Multi-Storage Client (MSC) support for SFT dataset

shunjiad:


ananthsub (approved):

ko3n1g (changes requested):
We already install MSC through Mcore, so we can drop it here

PR #519, opened 2025-08-29
Skip attention_mask when dataset.get_attention_mask_from_fusion is set

gautham-kollu:
Attention_mask is not required as some attn backends can generate them. Hence, skip passing it to avoid copying a huge tensor from host to device. 

Tested:
  Locally tested by setting on/off the config


src/megatron/bridge/training/config.py
```diff
@@ -164,6 +164,10 @@ class DataloaderConfig:
 @dataclass
 class GPTDatasetConfig(MCoreGPTDatasetConfig, DataloaderConfig):
     """Configuration specific to GPT datasets, inheriting from MCore and base DataloaderConfig."""
+    
+    get_attention_mask_from_fusion: bool = False
+    """If set, the dataset will not pass a None attention mask and the casual 
+    mask is autogenerated from the attn backend"""
```
erhoo82, line 176: `will not pass` or `will pass`?
Also, I think `casual mask` should be just `attention mask` because casual is an attention type?

src/megatron/bridge/training/gpt_step.py
```diff
@@ -64,12 +64,17 @@ def get_packed_seq_params(batch: dict[str, torch.Tensor]) -> PackedSeqParams:
...
+    data_iterator: Iterable,
+    use_mtp: bool = False,
+    get_attention_mask_from_fusion: bool = False,
+) -> dict[str, torch.Tensor]:
     """Get a batch of data from the iterator.
 
     Args:
         data_iterator: The data iterator to get the batch from.
         use_mtp: Whether Multi-Token Prediction layers are enabled.
+        get_attention_mask_from_fusion: If set, the dataset will not pass a None attention mask
```
erhoo82: `will not pass` to `will pass`?

gautham-kollu: Updated

src/megatron/bridge/training/config.py
gautham-kollu: Updated

erhoo82:
@gautham-kollu 
Can we change `get_attention_mask_from_fusion` to something like `skip_getting_attention_mask_from_dataset`?
I think getting from fusion sounds vague, and the main goal of this knob is skipping copying the attention mask tensor from the input sample.

Also, can we set this knob to True in all the pretrain recipes that use causal mask? I think that what we do in nemo2.

PR #525, opened 2025-08-30
Megatron-FSDP support for Megatron-Bridge

BoxiangW:


src/megatron/bridge/recipes/qwen/qwen3_30b_a3b.py
```diff
@@ -187,9 +189,10 @@ def pretrain_config(
             grad_reduce_in_fp32=True,
             overlap_grad_reduce=True,
             overlap_param_gather=True,
-            average_in_collective=True,  # Not supported for custom FSDP for now, need to be set to False if using FSDP
-            data_parallel_sharding_strategy="optim_grads_params",  # For custom FSDP only
+            average_in_collective=False,  # Not supported for Megatron FSDP for now, need to be set to False if using Megatron FSDP
+            data_parallel_sharding_strategy="optim_grads_params",  # For Megatron FSDP only
```
ananthsub, line 193: do we need this set now in each of the recipes as a default?

src/megatron/bridge/recipes/qwen/qwen3_8b.py
```diff
@@ -95,6 +96,8 @@ def pretrain_config(
     # Precision recipe
     precision_config: Optional[Union[MixedPrecisionConfig, str]] = "bf16_mixed",
     comm_overlap_config: Optional[CommOverlapConfig] = None,
+    # Checkpoint
+    ckpt_format: str = "torch_dist",
```
ananthsub: why is ckpt_format exposed here?

src/megatron/bridge/training/setup.py
```diff
@@ -191,8 +183,26 @@ def setup(
...
+        cfg.ddp.use_megatron_fsdp = True
+
+        if cfg.model.gradient_accumulation_fusion == True:
+            print_rank_0("Gradient accumulation fusion is not supported with Megatron FSDP, setting to False")
+            cfg.model.gradient_accumulation_fusion = False
+
+        if cfg.ddp.average_in_collective == True:
+            print_rank_0("average_in_collective is not supported with Megatron FSDP, setting to True")
+            cfg.ddp.average_in_collective = False
+
```
ananthsub: why not have this run in `ConfigContainer.validate` instead? we should consider the config as frozen beyond this point: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/685b95a5f2918cfee4a508f7ac12da8f206e29c4/src/megatron/bridge/training/setup.py#L111-L126

shouldn't this also set the `ckpt_format` to use pytorch DCP instead? is the mcore dist ckpt path fully compatible with megatron FSDP?

src/megatron/bridge/training/train.py
```diff
@@ -660,8 +661,9 @@ def disable_forward_pre_hook(model: list[DDP], param_sync: bool = True) -> None:
         param_sync: Whether to synchronize parameters across model chunks
     """
     for model_chunk in model:
-        assert isinstance(model_chunk, DDP)
-        model_chunk.disable_forward_pre_hook(param_sync=param_sync)
+        # only works for DDP, Megatron FSDP or FSDP2 does not have this method
+        if isinstance(model_chunk, DDP):
+            model_chunk.disable_forward_pre_hook(param_sync=param_sync)
```
ananthsub: we shouldn't need these changes here. we should change the condition in L188 to check that megatron FSDP is not used and (distributed optimizer + overlap param gather are used)

src/megatron/bridge/training/setup.py
BoxiangW: No, I will add ckpt_format="fsdp_dtensor" in a separate PR

src/megatron/bridge/training/train.py
BoxiangW: Thanks for the pointer

src/megatron/bridge/recipes/qwen/qwen3_8b.py
BoxiangW: This is for future use case where Megatron-FSDP will use 'fsdp_dtensor' format

src/megatron/bridge/recipes/qwen/qwen3_30b_a3b.py
BoxiangW: 'average_in_collective=False' is reverted, but why are we using average loss now?

src/megatron/bridge/training/setup.py
BoxiangW: Also I have moved the validation part to `ConfigContainer.validate()` now, thanks for the pointer

src/megatron/bridge/recipes/qwen/qwen3_30b_a3b.py
ananthsub: for nemo2 compatibility this was using the same configs as https://github.com/NVIDIA-NeMo/NeMo/blob/55027591facf4119d4eb4ff12f9f90cef56a5064/nemo/collections/llm/recipes/qwen3.py#L146-L147

src/megatron/bridge/training/config.py
```diff
@@ -769,10 +772,46 @@ def validate(self) -> None:
...
+            if self.checkpoint.ckpt_format == "fsdp_dtensor":
+                assert self.dist.use_megatron_fsdp and not self.dist.use_torch_fsdp2, (
+                    "fsdp_dtensor checkpoint format only supports Megatron FSDP"
+                )
+
+        # Megatron FSDP Config checks
+        if self.dist.use_megatron_fsdp:
+            assert self.dist.use_torch_fsdp2 is False, (
+                "use_megatron_fsdp and use_torch_fsdp2 cannot be True at the same time"
+            )
```
ananthsub: redundant iwth L775-777?

src/megatron/bridge/training/config.py
```diff
@@ -769,10 +772,46 @@ def validate(self) -> None:
...
+                    "Megatron FSDP only supports fsdp_dtensor checkpoint format"
+                )
+
+            if self.model.gradient_accumulation_fusion == True:
+                print_rank_0("Gradient accumulation fusion is not supported with Megatron FSDP, setting to False")
+                self.model.gradient_accumulation_fusion = False
+
+            if self.ddp.average_in_collective == True:
+                print_rank_0("average_in_collective is not supported with Megatron FSDP, setting to True")
+                self.ddp.average_in_collective = False
```
ananthsub: please add unit tests here for the config validation: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/527599c429713163135914f22c8314a845bdcae6/tests/unit_tests/training/test_config.py#L365

BoxiangW: Yep, just pushed some config unit tests

src/megatron/bridge/training/config.py
BoxiangW: yes, thanks for catching

src/megatron/bridge/models/gpt_provider.py
```diff
@@ -116,6 +116,9 @@ class GPTModelProvider(TransformerConfig, ModelProviderMixin[MCoreGPTModel]):
 
     generation_config: Optional[Any] = None
 
+    # Parallel backend
+    use_megatron_fsdp: bool = False
+
```
ananthsub: why is this needed on the provider?

src/megatron/bridge/training/config.py
```diff
@@ -773,10 +776,42 @@ def validate(self) -> None:
...
+            # only check if saving or loading
+            if self.checkpoint.ckpt_format == "fsdp_dtensor":
+                assert self.dist.use_megatron_fsdp and not self.dist.use_torch_fsdp2, (
+                    "fsdp_dtensor checkpoint format only supports Megatron FSDP"
+                )
+
+        # Megatron FSDP Config checks
+        if self.dist.use_megatron_fsdp:
+            # Set Megatron FSDP Configs
+            self.model.use_megatron_fsdp = True
```
ananthsub: why do we need to set this on the provider? it's not an option on the transformer config or model parallel config

src/megatron/bridge/training/config.py
```diff
@@ -773,10 +776,42 @@ def validate(self) -> None:
...
+                    "Megatron FSDP only supports fsdp_dtensor checkpoint format"
+                )
+
+            if self.model.gradient_accumulation_fusion == True:
+                print_rank_0("Gradient accumulation fusion is not supported with Megatron FSDP, setting to False")
+                self.model.gradient_accumulation_fusion = False
+
+            if self.ddp.average_in_collective == True:
+                print_rank_0("average_in_collective is not supported with Megatron FSDP, setting to True")
+                self.ddp.average_in_collective = False
```
ananthsub:
```suggestion
            if self.model.gradient_accumulation_fusion:
                print_rank_0("Gradient accumulation fusion is not supported with Megatron FSDP, setting to False")
                self.model.gradient_accumulation_fusion = False

            if self.ddp.average_in_collective:
                print_rank_0("average_in_collective is not supported with Megatron FSDP, setting to True")
                self.ddp.average_in_collective = False
```

src/megatron/bridge/training/config.py
```diff
@@ -77,6 +77,9 @@ class DistributedInitConfig:
     """If set to True, initialize_megatron() skips DDP initialization and returns function to complete it instead.
     Also turns on --use-cpu-initialization flag. This is for external DDP manager."""
 
+    use_megatron_fsdp: bool = False
+    """Use Megatron's Fully Sharded Data Parallel. Cannot be used together with use_torch_fsdp2."""
+
```
ananthsub, line 82: why do we need this flag on the distributed init config? couldn't we rely on this being set from DistributedDataParallelConfig , and then propagate to the optimizer config here?

BoxiangW: It should be alongside fsdp2 config in my opinion

src/megatron/bridge/training/config.py
BoxiangW: Fixed

src/megatron/bridge/training/config.py
BoxiangW: model.use_megatron_fsdp removed

src/megatron/bridge/models/gpt_provider.py
BoxiangW: removed

src/megatron/bridge/models/model_provider.py
```diff
@@ -515,7 +523,10 @@ def get_model(
     # GPU allocation.
     # For FSDP2, we don't allocate GPU memory here. We allocate GPU memory
     # in the fully_shard function of FSDP2 instead.
-    if not (use_torch_fsdp2 or model_config.use_cpu_initialization) and not model_config.init_model_with_meta_device:
+    if (
+        not (use_megatron_fsdp or use_torch_fsdp2 or model_config.use_cpu_initialization)
+        and not model_config.init_model_with_meta_device
+    ):
```
ananthsub: why do we need the `use_megatron_fsdp` check here?

BoxiangW: I think it is not needed, I haved fixed it based on https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/blob/main/megatron/training/training.py#L897-906

src/megatron/bridge/training/config.py
ananthsub: i see, the config validation explicitly checks if this flag is set either in the ddp config or the dist init config, and if any one is True, all the corresponding configs are synced to use megatron FSDP. given that all the recipes are updated to set this in the ddp config, to me its more obvious if the ddp config is treated as the source of truth. but we can resolve this in a follow up PR to unblock the initial integration

src/megatron/bridge/training/setup.py
```diff
@@ -228,7 +221,7 @@ def setup(
             optimizer,
             scheduler,
             checkpointing_context=checkpointing_context,
-            skip_load_to_model_and_opt=HAVE_FSDP2 and cfg.dist.use_torch_fsdp2,
+            skip_load_to_model_and_opt=cfg.dist.use_torch_fsdp2 or cfg.model.use_megatron_fsdp,
```
ananthsub: the tests are failing because this is no longer an attribute on the model provider
```suggestion
            skip_load_to_model_and_opt=cfg.dist.use_torch_fsdp2 or cfg.dist.use_megatron_fsdp,
```

PR #535, opened 2025-09-02
Reverse param mappings dict in bridges

maanug-nv:
In Bridge, we use megatron_param as the central reference for parameter lookup, and then use it to find the corresponding Hugging Face (HF) parameter. Given this flow, it makes more sense to define param_mapping from Megatron parameters to HF parameters.

### DIFF
--- a/src/megatron/bridge/models/llama/llama_bridge.py
+++ b/src/megatron/bridge/models/llama/llama_bridge.py
@@ -93,25 +93,25 @@ class LlamaBridge(MegatronModelBridge):
         }
 
         mapping_list = []
-        # Convert each dictionary entry to AutoMapping(hf_param, megatron_param)
+        # Convert each dictionary entry to AutoMapping(megatron_param, hf_param)
         for megatron_param, hf_param in param_mappings.items():
-            mapping_list.append(AutoMapping(hf_param=hf_param, megatron_param=megatron_param))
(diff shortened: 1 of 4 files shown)


src/megatron/bridge/models/llama/llama_bridge.py
```diff
@@ -83,18 +83,18 @@ def mapping_registry(self) -> MegatronMappingRegistry:
...
+            "decoder.layers.*.mlp.linear_fc1.layer_norm_weight": "model.layers.*.post_attention_layernorm.weight",
+            "decoder.layers.*.self_attention.linear_proj.weight": "model.layers.*.self_attn.o_proj.weight",
+            "decoder.layers.*.mlp.linear_fc2.weight": "model.layers.*.mlp.down_proj.weight",
         }
 
         mapping_list = []
         # Convert each dictionary entry to AutoMapping(hf_param, megatron_param)
-        for hf_param, megatron_param in param_mappings.items():
+        for megatron_param, hf_param in param_mappings.items():
             mapping_list.append(AutoMapping(hf_param=hf_param, megatron_param=megatron_param))
```
yaoyu-33: reverse these 2 as well though order doesnt matter just make it consistent. similarly for qkv mapping, put magatron_param first then q,k,v

PR #537, opened 2025-09-03
Explicitly pass `expert_tensor_parallel_size` to `initialize_model_parallel`

nathan-az:
Due to the [default handling in MCore](https://github.com/NVIDIA/Megatron-LM/blob/e0b9fbb254663b3441375deb4a734d936be2f67f/megatron/core/parallel_state.py#L697), if this isn't passed, `expert_tensor_parallel_size` is forced to `tensor_model_parallel_size`. With this PR, we first check whether the parameter is passed, and if not, fall back to the default behaviour with None.

Not extensively tested, but came across the issue and the fix seems straightforward.

(Hopefully) fixes https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/546

### DIFF
--- a/src/megatron/bridge/models/model_provider.py
+++ b/src/megatron/bridge/models/model_provider.py
@@ -215,7 +215,7 @@ class ModelProviderMixin(abc.ABC, Generic[ModelT]):
             virtual_pipeline_model_parallel_size=getattr(self, "virtual_pipeline_model_parallel_size", None),
             context_parallel_size=getattr(self, "context_parallel_size", 1) or 1,
             expert_model_parallel_size=getattr(self, "expert_model_parallel_size", 1) or 1,
-            expert_tensor_parallel_size=getattr(self, "tensor_model_parallel_size", None),
+            expert_tensor_parallel_size=getattr(self, "expert_tensor_parallel_size", None),
             **model_parallel_kwargs,
         )
         if seed is not None:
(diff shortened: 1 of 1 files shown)


ananthsub:
Thanks for your contribution @nathan-az - could you share the full stack trace of the error you ran into?

ananthsub (approved):

nathan-az:
@ananthsub created [an issue](https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/546) with a repro and stacktrace. Let me know if I've overlooked anything. I'm working out the kinks of using `NeMO-RL` and reporting/fixing issues along the way, but I'm not at all experienced in the megatron/nemo stack.

nathan-az:
Also CC @yaoyu-33 since (AFAICT) some EP cases are currently broken, and this fixes them - I'm not sure how often the NeMO-RL submodule branch `yuya/nemo-rl-use-chunkpatch` is sync'd with `main`.

yfw:
Thank you @nathan-az for your contribution! I've verified this fixes an issue I was seeing with expert_tensor_parallel_size not being set correctly when testing NeMo RL with deepseek models. I've included this change in this PR for NeMo RL: https://github.com/NVIDIA-NeMo/RL/pull/1059 to make sure it gets included into NeMo RL.

PR #544, opened 2025-09-03
Fix the assertion conditions of the delay-wgard-compute

youngeunkwon0405:
Copying the conditions in 
https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/merge_requests/3941/diffs#dcdab519f3c4ff103bdb64e1d296a4cda3fdf722_1106_1106

There was an over guarding before.


youngeunkwon0405:
@ananthsub I was asserting the condition too conservatively. Now it is matching with the [mcore](https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/merge_requests/3941/diffs#dcdab519f3c4ff103bdb64e1d296a4cda3fdf722_1106_1106).
For the coverage test, looks like only 2 assertion lines are uncovered by the test. 
image

ananthsub (approved):

PR #560, opened 2025-09-04
Rename qwen 25 to qwen 2P5 to be more specific

yaoyu-33:


ananthsub (approved):

yaoyu-33:
seems many needs to change including llama4 as well, close for now, may discuss later.

PR #565, opened 2025-09-04
Cleanup logs

gautham-kollu:
Print only for rank0

### DIFF
--- a/scripts/performance/run_script.py
+++ b/scripts/performance/run_script.py
@@ -31,6 +31,7 @@ from megatron.bridge.training.utils.omegaconf_utils import (
     create_omegaconf_dict_config,
     parse_hydra_overrides,
 )
+from megatron.bridge.utils.common_utils import get_rank_safe
 
 
 logger: logging.Logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


scripts/performance/run_script.py
```diff
@@ -72,7 +73,9 @@ def main():
     if args.compute_dtype == "bf16":
         recipe.optimizer.use_precision_aware_optimizer = True
 
-    recipe.to_yaml()
+    if get_rank_safe() == 0:
+        recipe.to_yaml()
+
```
ananthsub, line 78: @gautham-kollu is this needed since the final config is printed here? https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/4b0b56061a8e0c66549b62a6077171f1a5f3fb11/src/megatron/bridge/training/setup.py#L266-L271

also, why is this being printed here before any of the overrides below are applied?

gautham-kollu: The original examples has [this](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/examples/recipes/llama/pretrain_llama3_8b.py#L135 
) and [this](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/examples/recipes/llama/pretrain_llama3_8b.py#L164) I guessed they were forked 


I personally prefer the full user-config that is being before calling init/setup() and one after setup() (We could potentially do a diff  in setup() )

I will let @malay-nagda  and @erhoo82 chime in.

malay-nagda: the one after setup is enough as we might want to see just final configs actually used after all overrides...

I think we should just remove it in `run_script.py`

I missed this in my earlier PR

ananthsub:
resolved in https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/584

PR #579, opened 2025-09-05
Add `print_yaml` method for config visualization

ananthsub:
Calling `to_yaml` without an output path results in printing the configuration as a yaml. this isn't obvious from the method name. 

this PR adds a dedicated `print_yaml` method to resolve any confusion. this also adds a deprecation warning to `to_yaml` encouraging users who want to log the config to the console to rely on `print_yaml` instead, and to use `to_yaml` only for saving the yaml as a file.

### DIFF
--- a/examples/recipes/llama/pretrain_llama3_8b.py
+++ b/examples/recipes/llama/pretrain_llama3_8b.py
@@ -21,16 +21,16 @@ both YAML configuration files and command-line overrides using Hydra-style synta
 
 Examples:
     Basic usage with default configuration:
-        $ python pretrain_llama3_8b.py
+        $ torchrun --nproc_per_node=8 pretrain_llama3_8b.py
 
     Using a custom YAML config file:
-        $ python pretrain_llama3_8b.py --config-file my_custom_config.yaml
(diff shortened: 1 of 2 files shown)


src/megatron/bridge/training/utils/config_utils.py
```diff
@@ -202,6 +213,14 @@ def to_yaml(self, yaml_path: Optional[str] = None) -> None:
                     with open(yaml_path, "w") as f:
                         yaml.safe_dump(config_dict, f, default_flow_style=False)
 
+    def print_yaml(self) -> None:
+        """
+        Print the config container to the console in YAML format.
+        """
+        config_dict = self.to_dict()
+        with safe_yaml_representers():
+            print(yaml.safe_dump(config_dict, default_flow_style=False))
```
yaoyu-33, line 222: does it need print rank 0?

ananthsub: i wanted to leave it to the caller to determine on which ranks they want to print to not hardcode it here

maanug-nv: can we make it rank 0 in the example? i was using the example to test some other recipes and i noticed all ranks were dumping the config which wasn't useful.

PR #591, opened 2025-09-07
Add global barrier after checkpoint save to ensure synchronization across ranks

izikgo:
Lack of synchronisation results in a race condition which in-turn results in a crash while reporting wandb artifact.
When saving the config in rank 0 using ```cfg.to_yaml(config_filename)```, a tmp file is created.
During this process, in the last rank (which runs the wandb reporter), walks the checkpoint directory, containing the above tmp file.
Then, when reporting of each file occurs, the tmp file is no longer present which causes a FileNotFound exception.


ananthsub (approved):
great catch!

PR #606, opened 2025-09-08
Add Thread-Safe Configuration Loading to Prevent Race Conditions

yaoyu-33:
**Problem**

When multiple threads or processes simultaneously attempt to load HuggingFace model configurations (via `AutoConfig.from_pretrained()`), race conditions can occur during the download and caching process. This leads to intermittent failures with errors such as:

- File corruption during concurrent downloads
- Cache inconsistencies between processes
- Network timeouts due to resource contention
- Incomplete configuration files being read

This issue was particularly problematic in multi-threaded environments where several workers try to load the same model configuration simultaneously.

## 🔧 **Solution**

Implemented a comprehensive thread-safe and process-safe configuration loading system with the following key features:

### **🔒 File-Based Locking**
- Uses `filelock` library for cross-process synchronization
- Creates unique lock files per model path (hash-based) to avoid conflicts
- 60-second timeout to prevent indefinite blocking
- Graceful fallback when `filelock` is not available

### **🔄 Smart Retry Logic**
- Exponential backoff with jitter to prevent thundering herd effects
- Configurable retry attempts (default: 3)
- Intelligent error classification:
  - **Permanent errors** (404, repo not found, access denied) → No retries
  - **Transient errors** (network timeouts, connection issues) → Retry with backoff

### DIFF
--- a/tests/unit_tests/models/llama/test_llama_bridge.py
+++ b/tests/unit_tests/models/llama/test_llama_bridge.py
@@ -251,7 +251,7 @@ class TestMegatronLlamaBridge:
         result = bridge.provider_bridge(mock_pretrained_llama)
 
         # These should always be set to these values for Llama
-        assert result.share_embeddings_and_output_weights == False
+        assert result.bias_activation_fusion == True
 
 
 class TestAutoBridgeIntegration:


src/megatron/bridge/utils/safe_config_loader.py
```diff
@@ -0,0 +1,133 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
ananthsub, line 1: nit: add this util here instead? https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/src/megatron/bridge/models/hf_pretrained

src/megatron/bridge/utils/safe_config_loader.py
```diff
@@ -0,0 +1,133 @@
...
+import hashlib
+import time
+from pathlib import Path
+from typing import Union
+
+from transformers import AutoConfig
+from transformers.configuration_utils import PretrainedConfig
+
+try:
+    import filelock
```
yfw: should we add this to pyproject.toml?

src/megatron/bridge/utils/safe_config_loader.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    last_exception = None
+    
+    for attempt in range(max_retries + 1):
+        try:
+            if HAS_FILELOCK:
+                # Use file locking for process-safe access
+                # Create a lock file based on the path hash to avoid conflicts
+                path_hash = hashlib.md5(str(path).encode()).hexdigest()
+                lock_file = Path.home() / ".cache" / "huggingface" / f".megatron_config_lock_{path_hash}"
```
yfw: Do we want to give the user a way to override this path (e.g. via environment variable) in case they want to specify a different directory? I can see it being useful if they want to create a lock in a shared directory for running across multiple nodes.

yaoyu-33:
@yfw @ananthsub  comments addressed.

ananthsub (approved):

yfw (approved):
lgtm, thanks for adding!

src/megatron/bridge/models/hf_pretrained/safe_config_loader.py
```diff
@@ -0,0 +1,147 @@
...
+from transformers import AutoConfig
+from transformers.configuration_utils import PretrainedConfig
+
+
+try:
+    import filelock
+
+    HAS_FILELOCK = True
+except ImportError:
+    HAS_FILELOCK = False
```
ko3n1g: What's the motivation for safeguarding filelock?

yaoyu-33: no it's fine, lemme remove the guard.

PR #611, opened 2025-09-09
[megatron FSDP] Support `fsdp_dtensor` checkpoint format 

### LINKED ISSUE #54: Support Megatron FSDP
Megatron FSDP is a custom FSDP implementation available in megatron-core.

Megatron FSDP needs handling in a [few areas](https://github.com/search?q=repo%3ANVIDIA%2FMegatron-LM+use_megatron_fsdp+language%3APython+path%3A%2F%5Emegatron%5C%2F%2F+path%3A%2F%5Emegatron%5C%2Ftraining%5C%2F%2F&type=code)

1. Initialization: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/9ab694d75307f9de2281e45b148ecb5e88594216/src/megatron/bridge/models/model_provider.py#L608-L650
2. Support a new checkpoint format `fsdp_dtensor` doesn't rely on MCore dist checkpointing. This needs to be enforced in the config container validation if use_megatron_fsdp is used
3. handle Megatron FSDP for the parameter synchronization here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/9ab694d75307f9de2281e45b148ecb5e88594216/src/megatron/bridge/training/train.py#L188
4. Calculate L2 Norms: https://github.com/NVIDIA/Megatron-LM/blob/d6301fbaa0ff922dc7f03903fe1906b1e7ed5942/megatron/training/utils.py#L53-L66

ananthsub:
fixes #54 

Megatron FSDP relies on DTensor support, which needs special handling in megatron bridge checkpointing. This PR introduces a few changes to support this:

1. Use `unwrap_model` from megatron core

As this is already in core, there's no need to redefine it in the common utils. All callsites are updated to use the core utility and the copy in megatron bridge is removed.

2. Checkpointing:

A new checkpoint format `fsdp_dtensor` was introduced to distinguish handling for Megatron FSDP vs. megatron core dist checkpointing paths. This directly calls PyTorch DCP instead of going through MCore Dist Ckpt APIs. There are differences in:
- RNG state generation
- model state dict generation
- sharded state dict metadata
- the actual save/load paths

MLM reference: https://github.com/NVIDIA/Megatron-LM/commit/af28b5a559ace0782ad53957267f36e2b36ae011

3. This PR includes a few other fixes synchronizing changes from MLM:
- model config function handling
- adding a dedicated function to force parameter syncing before checkpoint saving / switching to eval
- precision aware optimizer + megatron fsdp should update the ddp config's `preserve_fp32_weights`

4. Testing
- Unit testing for configs/train utils
- Integration testing for checkpointing functions
- Functional testing for end to end training with megatron FSDP enabled + checkpoint saving & resuming

### DIFF
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -431,7 +431,7 @@ class CheckpointConfig:
     exit_on_missing_checkpoint: bool = False
     """If 'load' is set, but checkpoint is not found (e.g., path typo), then exit instead of random initialization."""
 
-    ckpt_format: Literal["torch_dist", "zarr"] = "torch_dist"
+    ckpt_format: Literal["torch_dist", "zarr", "fsdp_dtensor"] = "torch_dist"
     """Checkpoint format to use."""
 
     ckpt_convert_format: Optional[Literal["torch", "torch_dist", "zarr"]] = None
@@ -457,8 +457,10 @@ class CheckpointConfig:
     """Apply full load parallelization across DP for distributed checkpoints."""
 
     ckpt_assume_constant_structure: bool = False
-    """If the model and optimizer state dict structure is constant throughout a *single training job,
-    it allows for different checkpointing performance optimizations."""
+    """Assume the checkpoint structure is constant across saves to enable optimizations."""
+
+    strict_fsdp_dtensor_load: bool = False
+    """Whether to enforce strict loading for FSDP DTensor checkpoints. When False, allows partial loading."""
 
     dist_ckpt_strictness: Literal[
         "assume_ok_unexpected",
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -805,13 +805,22 @@ class ConfigContainer(Container):
                 print_rank_0("average_in_collective is not supported with Megatron FSDP, setting to True")
                 self.ddp.average_in_collective = False
 
-                # Checkpoint
+            if self.optimizer.use_precision_aware_optimizer:
+                self.ddp.preserve_fp32_weights = False
+
+        # Checkpoint
         if self.checkpoint.save is not None or self.checkpoint.load is not None:
             # only check if saving or loading
             if self.checkpoint.ckpt_format == "fsdp_dtensor":
                 assert self.ddp.use_megatron_fsdp and not self.dist.use_torch_fsdp2, (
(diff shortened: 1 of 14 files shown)


src/megatron/bridge/training/config.py
```diff
@@ -803,14 +805,23 @@ def validate(self) -> None:
                 print_rank_0("average_in_collective is not supported with Megatron FSDP, setting to True")
                 self.ddp.average_in_collective = False
 
-                # Checkpoint
+            if self.optimizer.use_precision_aware_optimizer:
+                self.ddp.preserve_fp32_weights = False
+
```
ananthsub, line 810: https://github.com/NVIDIA/Megatron-LM/blob/de48755da00c56b76d7e4f6e395b9ab7601d1e4e/megatron/training/training.py#L970-L971

src/megatron/bridge/training/setup.py
```diff
@@ -283,18 +284,18 @@ def _update_model_config_funcs(
...
+        model_config.no_sync_func = [model_chunk.no_sync for model_chunk in model]
         if len(model) == 1:
-            model_config.grad_sync_func = model_config.grad_sync_func[0]
+            model_config.no_sync_func = model_config.no_sync_func[0]
+        if align_grad_reduce:
+            model_config.grad_sync_func = [model_chunk.start_grad_sync for model_chunk in model]
+            if len(model) == 1:
+                model_config.grad_sync_func = model_config.grad_sync_func[0]
     if ddp_config.overlap_param_gather and ddp_config.align_param_gather:
         model_config.param_sync_func = [model_chunk.start_param_sync for model_chunk in model]
```
ananthsub, line 300: adapted from https://github.com/NVIDIA/Megatron-LM/blob/de48755da00c56b76d7e4f6e395b9ab7601d1e4e/megatron/training/training.py#L2003-L2019

src/megatron/bridge/training/train.py
```diff
@@ -646,6 +646,29 @@ def post_training_step_callbacks(
             gc.collect()
 
 
+def should_disable_forward_pre_hook(
```
ananthsub, line 649: syncing from https://github.com/NVIDIA/Megatron-LM/blob/de48755da00c56b76d7e4f6e395b9ab7601d1e4e/megatron/training/training.py#L2824-L2826

Issue #563, opened 2025-09-04
Config value print issue in the log file


erhoo82:
1. Config values before overriding are printed
2. All ranks repeatedly print config values in the log file

malay-nagda:
@erhoo82 I think this already fixed with https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/534

ananthsub:
closed in https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/584

PR #624, opened 2025-09-09
add zero_rank_only func

liding-nv:
Adding a decorator `rank_zero_only` to enable a func being called only on rank 0.

It fixes the issue that `cfg.to_yaml(yaml_path=xxx)` not compatible with MultiStorageClientFeature when multi-gpu is used. Calling this func from all ranks will crash if msc is enabled.
```
if MultiStorageClientFeature.is_enabled():
    msc = MultiStorageClientFeature.import_package()
    with msc.open(yaml_path, "w") as f:
        yaml.safe_dump(config_dict, f, default_flow_style=False)
```

`rank_zero_only` should also be useful for other functions intended to be called only on rank 0.

### DIFF
--- a/src/megatron/bridge/training/utils/config_utils.py
+++ b/src/megatron/bridge/training/utils/config_utils.py
@@ -22,6 +22,7 @@ import yaml
 from megatron.core.msc_utils import MultiStorageClientFeature
 from omegaconf import OmegaConf
 
+from megatron.bridge.utils.common_utils import rank_zero_only
 from megatron.bridge.utils.instantiate_utils import InstantiationMode, instantiate
 from megatron.bridge.utils.yaml_utils import safe_yaml_representers
 
@@ -181,6 +182,7 @@ class _ConfigContainerBase:
         else:
             return value
 
+    @rank_zero_only
     def to_yaml(self, yaml_path: Optional[str] = None) -> None:
         """
         Save the config container to a YAML file.
--- a/src/megatron/bridge/utils/common_utils.py
+++ b/src/megatron/bridge/utils/common_utils.py
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 import os
-from typing import Type, Union
+from functools import wraps
+from typing import Type, Union, Callable, Any, Optional
 
 import torch
 import torch.distributed
@@ -143,3 +144,15 @@ def use_dist_ckpt(ckpt_format: str) -> bool:
         True if the format is not "torch", False otherwise.
     """
     return ckpt_format != "torch"
+
+
+def rank_zero_only(fn: Callable) -> Callable:
+    """Decorator to enable a function being called only on global rank 0."""
+
+    @wraps(fn)
+    def wrapped_fn(*args: Any, **kwargs: Any) -> Optional[Any]:
+        if get_rank_safe() == 0:
+            return fn(*args, **kwargs)
+        return None
+
+    return wrapped_fn


ananthsub:
@liding-nv could you describe the issue you're facing in more detail? in checkpoint saving, the configuration serialized to yaml occurs explicitly runs on rank 0 only here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/248a962d89e9ab3d4790958f85ab38c48b22b9a1/src/megatron/bridge/training/checkpointing.py#L560-L611

are you seeing issues outside that?

liding-nv:
hi @ananthsub i was running the example code in readme https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main?tab=readme-ov-file#-quickstart (copied here)
```

import megatron.bridge.recipes.llama.llama32_1b as llama32_1b
from megatron.bridge.training.pretrain import pretrain

if __name__ == "__main__":
    # Load Llama from Hugging Face Hub and convert to Megatron
    bridge = AutoBridge.from_hf_pretrained("meta-llama/Llama-3.2-1B")
    model_provider = bridge.to_megatron_provider()

    # Get defaults for other configuration from an existing Llama 3.2 recipe
    cfg = llama32_1b.pretrain_config()
    cfg.model = model_provider
    cfg.train.train_iters = 10

    cfg.dataset.sequence_length = cfg.model.seq_length
    cfg.tokenizer.vocab_size = cfg.model.vocab_size

    pretrain(cfg, forward_step)
```

added one line just above `pretrain` to save yaml config
```
cfg.to_yaml(yaml_path=/path/to/file)

pretrain(cfg, forward_step)
```

run the job with
```
torchrun --nproc_per_node=8 quick_test.py
```
and saw error msg if the output file already exist (e.g., by runing torchrun cmd twice)
```
FileNotFoundError: [Errno 2] No such file or directory: '/path/to/file/.sample_mbridge_llama_config.yaml.tmp' -> '/path/to/file/sample_mbridge_llama_config.yaml'
```
so yea it is outside checkpoint saving. I'm just running some toy example to understand how m-bridge works in general, not launch a real training job

another thing i observed, without `rank_zero_only`, when we call `cfg.to_yaml()` it will print the entire cfg from all ranks, which is a lot of lines on my terminal...

ananthsub:
thanks for sharing @liding-nv - instead of adding the decorator to `to_yaml`, in my opinion we should update the example in the quickstart in the README to make this explicit. what do you think?

liding-nv:
> thanks for sharing @liding-nv - instead of adding the decorator to `to_yaml`, in my opinion we should update the example in the quickstart in the README to make this explicit. what do you think?

do you mean in quick start example
```
if get_rank_safe() == 0:
    cfg.to_yaml(yaml_path=xxx)
```
i guess in this case users need to know which func should be called only from rank 0. Hiding it from user-facing code might be easier for them.

if the question is whether we use `if get_rank_safe() == 0` vs `@rank_zero_only`, then i think both should just work.

liding-nv:
also came across this issue https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/563

> All ranks repeatedly print config values in the log file

which I also find it makes the log hard to check. adding decorator should prevent it and work with existing code, so you don’t need to update/remove it manually. lmk if that makes sense or if I missed anything…

yaoyu-33:
Closing this PR as it has been inactive for 6+ months. Feel free to reopen if still relevant.

PR #627, opened 2025-09-10
fix: Delay megatron core config post init checks to override settings

ananthsub:
In recipes, the configurations are first initialized in the config container, after which overrides can be applied. An issue is that for mcore configs, the post init checks run immediately when the config container is first created. 

Example: 
The issue seen with microbatch_group_size_per_vp_stage is related to the failure of recipe override. For any "config.attr is None" style condition like [this](https://github.com/NVIDIA/Megatron-LM/blob/5e8c9c48b8b12994e53616480d75b42aacdfbac3/megatron/core/model_parallel_config.py#L395-L396), once a wrong initial value is set during the first call. Any following override + post_init won't change that wrong value.

To resolve this:
For megatron bridge, we need to ensure that all config overrides are applied first before any post init initialization logic is run. we do this by inheriting from the mcore configs, overriding the `__post_init__` to be a no-op, and provide a `finalize()` method which calls the superclass's `__post_init__`. For users relying on the mbridge training loop, this is called once during the `ConfigContainer.validate` at the beginning of the setup. This ensures that the post-init initialization for these configs take effect only after all overrides are captured. 

outside of training, the bridge calls finalize where needed when going to_megatron_model. outside of these workflows, it's the caller's responsibility to explicitly call `finalize()` before proceeding


tests/functional_tests/training/test_pretrain.py
```diff
@@ -16,15 +16,16 @@
...
 import torch
-from megatron.core.distributed import DistributedDataParallelConfig
-from megatron.core.optimizer import OptimizerConfig
 
 from megatron.bridge.models.llama import Llama32ModelProvider1B
 from megatron.bridge.training.config import (
     CheckpointConfig,
     ConfigContainer,
+    DistributedDataParallelConfig,
+    DistributedInitConfig,
```
ananthsub:
```suggestion
```

examples/models/multi_gpu_hf.py
```diff
@@ -82,8 +82,12 @@ def main(
         model_provider = bridge.to_megatron_provider(load_weights=False)
         model_provider.tensor_model_parallel_size = tp
         model_provider.pipeline_model_parallel_size = pp
+        model_provider.pipeline_dtype = torch.bfloat16
```
ananthsub, line 85: `pipeline_dtype` was missing when the script set pp > 1. this was found now since `finalize()` below will run the transformer config post init checks for validation and default settings

tests/functional_tests/models/test_nemotron_h_conversion.py
```diff
@@ -195,12 +195,12 @@ def test_nemotronh_conversion_parallelism(self, nemotronh_toy_model_path, tmp_pa
...
-            "run",
-            "--data-file=/workspace/.coverage",
-            "--source=/workspace/",
-            "--parallel-mode",
+            # "-m",
+            # "coverage",
+            # "run",
+            # "--data-file=/workspace/.coverage",
+            # "--source=/workspace/",
+            # "--parallel-mode",
```
ananthsub:
```suggestion
            "-m",
            "coverage",
            "run",
            "--data-file=/workspace/.coverage",
            "--source=/workspace/",
            "--parallel-mode",
```

malay-nagda (approved):
tested successfully for Llama3 8B BF16 pre-training on H100.

gautham-kollu (approved):

maanug-nv (approved):

pruprakash:
## QA RCCA Analysis

### 1. Fix Reference
- **PR**: #627 - fix: Delay megatron core config post init checks to override settings

### 2. Root Cause
MCore configs run post_init checks immediately when config container is created, before recipe overrides can be applied. This caused issues like `microbatch_group_size_per_vp_stage` being set to wrong initial value that can't be overridden.

### 3. Trigger Configuration
- Using recipes with config overrides
- MCore configs with post_init validation

### 4. Nature of the Bug
**Classification**: CODE BUG - Post_init checks running before overrides applied

### 5. Existing Test Coverage
**In Fix PR**: YES - 22 test files across functional and unit tests

**In NMFW Tests**: Not specifically covering config override timing

### 6. Coverage Assessment
| Fix PR unit tests | YES | YES |
| Fix PR functional tests | YES | YES |

### 7. New Regression Test
**NOT NEEDED** - Fix PR includes comprehensive test coverage.

### 8. Conclusion
**Verdict**: ADEQUATE COVERAGE - Fix PR includes 22 test files covering config initialization.

PR #635, opened 2025-09-11
Update build_train_valid_test_data_loaders to handle eval_iters=0

chtruong814:
Update build_train_valid_test_data_loaders to handle eval_iters=0

At least in the case of the short convergence tests, eval_iters is 0 and it fails 
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/tests/end_to_end_tests/train_from_recipe.py#L265

### DIFF
--- a/src/megatron/bridge/data/loaders.py
+++ b/src/megatron/bridge/data/loaders.py
@@ -200,7 +200,7 @@ def build_train_valid_test_data_loaders(
         data_parallel_rank=mpu.get_data_parallel_rank(),
         data_parallel_size=mpu.get_data_parallel_world_size(),
     )
-    if cfg.train.skip_train:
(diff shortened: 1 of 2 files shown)


.github/actions/test-template/action.yml
```diff
@@ -239,6 +239,7 @@ runs:
           --env TRANSFORMERS_OFFLINE=0 \
           --env HYDRA_FULL_ERROR=1 \
           --env HF_HOME=/home/TestData/HF_HOME \
+          --env HF_HUB_DISABLE_XET=1 \
```
chtruong814: I'm going to try to remove this in this PR.  Xet is a storage solution from huggingface and it wasn't working earlier.  This allowed tests to pass, otherwise, it'd get some s3 error.

ananthsub (approved):

PR #647, opened 2025-09-15
Fix Deepseek Recipe

cuichenx:
1. Add 32 node configuration and clean up recompute arguments
2. Add option in test script to use custom config name 
3. Set ETP to 1
4. Sync rope fusion with experimental flag
5. Remove use_mtp in favor of specifying mtp_num_layers

### DIFF
--- a/src/megatron/bridge/recipes/deepseek/deepseek_v3.py
+++ b/src/megatron/bridge/recipes/deepseek/deepseek_v3.py
@@ -310,7 +310,7 @@ def pretrain_config(
     return cfg
 
 
-def pretrain_config_32nodes(**kwargs):    
+def pretrain_config_32nodes(**kwargs):
     """
     Create a pre-training configuration for DeepSeek-V3 (671B) model with minimal number of nodes (32).
 


src/megatron/bridge/recipes/deepseek/deepseek_v3.py
```diff
@@ -302,3 +304,20 @@ def pretrain_config(
         )
 
     return cfg
+
+
+def pretrain_config_32nodes(**kwargs):
```
ananthsub: is this intended to be used anywhere outside of internal convergence testing? if not, we can expose this config just through `train_from_recipe` to get these overrides instead

cuichenx: A few users/customers have asked for a "minimal" recipe of deepseek, so I think it's useful to keep this in the recipe file to share with them

src/megatron/bridge/recipes/deepseek/deepseek_v3.py
```diff
@@ -118,14 +109,17 @@ def model_config(
     cfg.apply_rope_fusion = True
 
     # Pipeline parallelism configs. We infer PP layout from the provided PP and VP size
+    if mtp_num_layers is None:
```
ananthsub, line 113: should this read from `use_mtp` like L92? is the `use_mtp` flag still useful, or would it make more sense to rely entirely on `mtp_num_layers` being not None?

cuichenx: good catch!
I think use_mtp is redundant with mtp_num_layers = None so i'm going to remove it

src/megatron/bridge/recipes/deepseek/deepseek_v3.py
```diff
@@ -184,6 +178,12 @@ def pretrain_config(
     precision_config: Optional[Union[MixedPrecisionConfig, str]] = None,
     comm_overlap_config: Optional[CommOverlapConfig] = None,
     enable_deepep: bool = False,
+    # Recomputation
+    recompute_granularity: str = "selective",
+    recompute_modules: Optional[List[str]] = None,
+    recompute_method: Optional[str] = None,
+    recompute_num_layers: Optional[int] = None,
+    apply_rope_fusion: bool = False,
```
cuichenx, line 186: "false" currently has better perf because of output spamming of experimental feature. Once that mcore MR is merged, i will reevaluate

src/megatron/bridge/recipes/deepseek/deepseek_v3.py
```diff
@@ -295,10 +299,29 @@ def pretrain_config(
...
+        ConfigContainer: Configuration for pre-training.
+    """    
+    return pretrain_config(
+        pipeline_parallelism=8,
+        expert_parallelism=32,
+        recompute_granularity="full",
+        recompute_method="uniform",
+        recompute_num_layers=1,
+        **kwargs,
+    )
```
ananthsub:
```suggestion
    )

```

for the lint error

cuichenx:
/ok to test [9d0bcb1](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/647/commits/9d0bcb185b73062bb1cf14ccbefa1d0a33836cfb)

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/647/commits/83ad569f520f7e50a1676a4c660753a2204f1488

PR #652, opened 2025-09-16
Add use_te_op_fuser knob in config and turn it on in llama3 perf benchmarks

guyueh1:
* Add use_te_op_fuser knob in config, set default as False
* Turn on the knob in llama3 perf benchmarks


malay-nagda (approved):

chtruong814:
Moving forward with merging this despite codecov being slightly below threshold.

PR #669, opened 2025-09-17
Malay/dsv3 2509

malay-nagda:


scripts/performance/run_script.py
```diff
@@ -49,16 +49,37 @@ def main():
...
+            virtual_pipeline_parallelism=vp,
+            enable_deepep=enable_deepep,
         )
         from megatron.bridge.training.utils.moe_token_drop import apply_moe_token_drop
 
-        recipe.model = apply_moe_token_drop(recipe.model)
+        if enable_deepep:
+            recipe.model.moe_router_force_load_balancing = True
+        else:
+            recipe.model = apply_moe_token_drop(recipe.model)
```
erhoo82, line 74: We should add a knob for token drop/dropless, where it defaults to dropless.
When we measure the perf of B200 and GB200, we should be using `tokendrop + enable_deepep:False` and `tokendropless + enable_deepep:True` for H100.

scripts/performance/configs/deepseek/deepseek_v3_llm_pretrain.yaml
```diff
@@ -40,6 +45,9 @@ ConfigContainer:
 
   mixed_precision:
     grad_reduce_in_fp32: false
+  
+  comm_overlap:
+    overlap_grad_reduce: false
```
gdengk: `overlap_grad_reduce: false` is a workaround to fix 1F1B_A2A overlap issue. We could put this to True when 1F1B_A2A is on or the bug fix branch is merged into core0.14.0

scripts/performance/setup_experiment.py
```diff
@@ -64,6 +64,7 @@
                 enable_vboost=args.enable_vboost,
                 nccl_pp_comm_chunksize=2097152 if args.model_size in ["70b", "405b"] else None,
                 gpu_sm100_or_newer=args.gpu.lower() in ["b200", "gb200"],
+                layernorm_sm_margin=20 if (args.gpu.lower() in ["h100"] and args.model_name in ["deepseek"]) else 16,
```
gdengk: DeepEP needs layernorm_sm_margin=20. If DeepEP is not enabled, we don't need this.

scripts/performance/configs/deepseek/deepseek_v3_llm_pretrain.yaml
malay-nagda: I'll add a note for this

erhoo82 (approved):

PR #607, opened 2025-09-08
Remove model providers for different model sizes (Qwen, Llama)

yaoyu-33:
https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/583

PR #676, opened 2025-09-18
decouple lr decay iters and train iters in recipe

ananthsub:
Adds a separate argument for `lr_decay_iters` in the recipe configurations, which is by default None. users who want to explicitly set this can.

Otherwise, training will fallback to setting this based on train_iters: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/138fc543eb2dad42f9a2bdde250424dce84afc9f/src/megatron/bridge/training/config.py#L938-L940

this avoids issues where the recipe is instantiated, train_iters in the train config is subsequently overridden, but the lr_decay_iters is not also kept in sync


ananthsub:
@liding-nv that is tested here:

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/138fc543eb2dad42f9a2bdde250424dce84afc9f/tests/unit_tests/training/test_config.py#L492-L523

liding-nv:
oh i see. I think this test doesn't capture the issue. It directly pass `train_iters=2000` to config, and in this way, both `train_iters` and `lr_decay_iters` are the same when cfg is initialized (so tests passed). Here, the problem is from overriding the parameter after cfg is initialized and two variables are no longer the same..

yaoyu-33 (approved):

tests/functional_tests/recipes/utils.py
```diff
@@ -87,3 +91,24 @@ def run_pretrain_recipe_test(
...
+    # the default seq_length does *not* match the model seq_length.
+    config.model.seq_length = 512
+    config.dataset.sequence_length = 512
+
+    assert config.scheduler.lr_decay_iters is None
+
+    runtime_config_update(config)
+
+    assert config.train.train_iters == 50000
+    assert config.scheduler.lr_decay_iters == config.train.train_iters
```
ananthsub, line 114: this shouldn't be part of the functional_tests, which run on multiple GPUs and are setup now more for mini end-to-end testing. this should be part of a unit-test if it's not already covered

liding-nv: yea i was debating on this as well when i added this test... i did't go with unit tests  because the same test needs to be added to all files under `tests/unit_tests/recipes`, and seemed to be a lot of duplications. i guess https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/607 would address this and we can revisit this once 607 is merged

PR #677, opened 2025-09-18
feat: Add move_model_to_cuda_device option in model configuration

izikgo:
* Introduced a new parameter `move_model_to_cuda_device` in GetModelKwargs to control GPU allocation.
* Updated the `get_model` function to utilize this parameter for conditional model movement to GPU.
* Modified AutoBridge to support the new parameter during model conversion.

### DIFF
--- a/src/megatron/bridge/models/model_provider.py
+++ b/src/megatron/bridge/models/model_provider.py
@@ -403,6 +403,7 @@ class GetModelKwargs(TypedDict, total=False):
         wrap_with_ddp: Whether to wrap model with DDP.
         data_parallel_random_init: Initialize parameters randomly across data parallel ranks.
         use_cpu_initialization: Initialize model on CPU.
+        move_model_to_cuda_device: Move model to GPU.
         init_model_with_meta_device: Initialize model on meta device.
         pre_wrap_hook: A single callable or list of callables that overrides all registered pre-wrap hooks.
         post_wrap_hook: A single callable that overrides all registered post-wrap hooks.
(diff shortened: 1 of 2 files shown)


src/megatron/bridge/models/model_provider.py
```diff
@@ -524,7 +528,11 @@ def get_model(
     # GPU allocation.
     # For FSDP2, we don't allocate GPU memory here. We allocate GPU memory
     # in the fully_shard function of FSDP2 instead.
-    if not (use_torch_fsdp2 and model_config.use_cpu_initialization) and not model_config.init_model_with_meta_device:
+    if (
+        not (use_torch_fsdp2 and model_config.use_cpu_initialization)
```
yaoyu-33: hi, I dont think we need to add `move_model_to_cuda_device` here, it shouldn't move if we use `model_config.use_cpu_initialization` at least not move here.
can we just update the condition to be 
`not use_torch_fsdp2 and not model_config.use_cpu_initialization and not model_config.init_model_with_meta_device:`

yaoyu-33:
sorry @izikgo I forgot to click submit for my comment. Do you think you my comments make sense?

izikgo:
Sure.
I'll change as per your suggestion

yaoyu-33 (approved):

yaoyu-33:
thanks @izikgo for the contribution . we also notice few other issues upon exporting, check https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/738 if you are interested.

PR #678, opened 2025-09-18
llama nemotron provider and bridge

ananthsub:


src/megatron/bridge/models/conversion/utils.py
```diff
@@ -277,7 +277,10 @@ def get_causal_lm_class_via_auto_map(
                 repo_id=repo_id,
             )
         except Exception:
-            return None
+            # If dynamic loading fails (e.g., version incompatibilities),
+            # return the class name as string for string-based bridge registration
+            class_name = auto_map_class.split(".")[-1]  # Extract class name from module.ClassName
+            return class_name
```
ananthsub, line 284: @yaoyu-33 @yfw I needed this fallback to register "DeciLMForCausalLM", which is at odds with https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/937

suiyoubi (approved):
LGTM. 

so the `llama_nemotron_bridge` is actually only doing the conversion for the heterogeneous llama nemotron model, should we name it `llama_nemotron_hete_bridge` for clarification ? Also I think we should mention it somewhere that we directly use LlamaBridge for part of the LlamaNemotron Models.

ananthsub:
> so the llama_nemotron_bridge is actually only doing the conversion for the heterogeneous llama nemotron model, should we name it llama_nemotron_hete_bridge for clarification ? Also I think we should mention it somewhere that we directly use LlamaBridge for part of the LlamaNemotron Models.

I was having trouble with this because the names are so similar and they're bundled in the Llama Nemotron collection: https://huggingface.co/collections/nvidia/llama-nemotron-67d92346030a2691293f200b

The LlamaBridge would cover:
- nvidia/Llama-3.1-Nemotron-Nano-8B-v1
- nvidia/Llama-3.1-Nemotron-Nano-4B-v1.1
- nvidia/Llama-3.1-Nemotron-70B-Instruct-HF

The LlamaNemotronHeterogeneousBridge would cover:
- nvidia/Llama-3_3-Nemotron-Super-49B-v1_5
- nvidia/Llama-3_1-Nemotron-Ultra-253B-v1
- nvidia/Llama-3_3-Nemotron-Super-49B-v1

This is more for developers, since all the routing is handled for users with AutoBridge

src/megatron/bridge/models/llama_nemotron/llama_nemotron_bridge.py
```diff
@@ -0,0 +1,160 @@
...
+    2. They have heterogeneous block configurations (different layers have different specs)
+    3. They require trust_remote_code=True to load from HuggingFace
+
+    Supported models (examples):
+    - nvidia/Llama-3_3-Nemotron-Super-49B-v1 (80 layers, 8192 hidden)
+    - nvidia/Llama-3_3-Nemotron-Super-49B-v1_5 (updated v1.5 release)
+    - nvidia/Llama-3_1-Nemotron-Ultra-253B-v1 (162 layers, 16384 hidden)
+
+    Homogeneous Llama-Nemotron models (Nano/70B) use standard LlamaForCausalLM
+    architecture and are handled by the regular LlamaBridge.
```
ananthsub, line 49: @suiyoubi the distinction for heterogeneous is mentioned here

PR #685, opened 2025-09-19
feat(evaluate): support multiple validation datasets + decouple val/train dataloader config

hawkoli1987:
## Summary

This PR adds **per-dataset validation** support and cleanly **decouples validation dataloader config from training dataloader config** in Megatron-Bridge. When enabled, evaluation runs over multiple validation datasets **sequentially**, logging **per-dataset** metrics and a **total evaluation duration** for the whole multi-set pass. When not enabled, behavior is unchanged.

---

## Motivation

Many training setups need to report validation metrics **per corpus** (e.g., domain A vs. domain B) rather than only a blended/weighted single number. This PR introduces a light-touch mechanism to construct and iterate multiple validation dataloaders, reusing the existing evaluation loop, without changing default behavior for users who do not opt in.

---

### Core
- **Multiple validation sets**: add support to run eval sequentially over a set of validation datasets, logging per-dataset loss/perplexity (e.g., `val_loss/<dataset_name>`).
- **Decoupled val/train dataloader config**: allow validation to use its own dataloader knobs independently of training (e.g., `num_workers`, `pin_memory`, etc.), so eval can stay lightweight even when training uses different settings.
- **Evaluation timing**: log a single **total** duration for the multi-set evaluation pass (sum over all validation datasets) to keep dashboards uncluttered.

> Commits:
> - `[training, data] feat: support multiple validation datasets with decoupled val/train config`
> - `[test] test: add unit tests for multiple validation sets`
---

#### Validation config resolution
**Test file:** `tests/unit_tests/training/test_config.py::TestValidationConfigResolution`
**Source under test:** `megatron.bridge.training.config.ConfigContainer.validate()`

- **Fallbacks:** When validation values are not set, they fall back to training config.
  *Setup:* `TrainingConfig` only defines training batch sizes, `GPTDatasetConfig` only defines training workers.
  *Expected:* Validation inherits training values (e.g., workers, batch sizes adjusted for world size).

- **Explicit values:** When validation values are provided, they override training config.
  *Expected:* Validation-specific settings are preserved (e.g., smaller val batch size, fewer workers).

- **Mixed case:** Combination of explicit and missing validation fields.
  *Expected:* Explicit fields used directly, others fall back or are recomputed.

---

#### Validation dataloader config
**Test file:** `tests/unit_tests/training/test_eval.py::TestValidationDataloaderCreation`
**Source under test:** `megatron.bridge.training.config.ValidationConfig`, `GPTDatasetConfig`, and `LoggerConfig`

- **Attributes exist:** `ValidationConfig` fields (`val_micro_batch_size`, `val_global_batch_size`) are present and hold expected values.
- **Dataloader attributes:** `GPTDatasetConfig` fields (`val_num_workers`, `val_pin_memory`, `val_persistent_workers`) are present and preserve explicitly set values.
- **Multiple validation sets flag:** `GPTDatasetConfig.multiple_validation_sets` attribute exists and is `True` when set.
- **Logger naming:** `LoggerConfig.multiple_validation_sets_use_dataset_name` returns the correct boolean for both `True` and `False` settings.

---

#### Evaluation entrypoint & logging
**Test file:** `tests/unit_tests/training/test_eval.py::TestEvaluateAndPrintResults`
**Source under test:** `megatron.bridge.training.eval.evaluate_and_print_results()`

- **Single dataset:** Calls `evaluate()` once and logs metrics to TensorBoard and WandB.
- **Multiple datasets:** With `blend_per_split` specifying multiple val datasets, calls `evaluate()` sequentially per dataset.
  *Expected:* Logs both per-dataset and aggregated validation metrics.
- **Timelimit:** If `evaluate()` signals timelimit, no metrics are logged.
- **Naming:** Dataset names from `blend_per_split` (e.g., `val_dataset_1`, `val_dataset_2`) appear in logged metric keys.

### Configuration

To evaluate **multiple validation datasets sequentially** and log **per-dataset** metrics,
- pack a list of validation data-prefixes into `valid_data_path`
```python
valid_data_path = [0.5, '/data/path_train_a', 0.5, '/data/path_train_b'] # demo
valid_data_path = ['/data/path_train_a', '/data/path_train_b']
```
- megatron bridge dataset_utils will convert it blend_per_split accordingly, after merging with `train_data_path` and `val_data_path`
```python
from megatron.bridge.recipes.utils.dataset_utils import get_blend_fields_from_data_paths

blend, blend_per_split, split = get_blend_fields_from_data_paths(
    data_paths, data_args_path, train_data_path, valid_data_path, test_data_path, per_split_data_args_path, mock
)
```
- **enable** the flag `dataset.multiple_validation_sets: true`.

**Resulting config (example):**
```yaml
dataset:
  _target_: megatron.bridge.training.config.GPTDatasetConfig
  ...
  blend: null
  blend_per_split:
    # train split: ([train_prefixes], [weights])
    - - - /data/path_train_a
        - /data/path_train_b
      - - 0.5
        - 0.5
    # validation split: ([val_prefixes]), no weights
    - - - /data/path_val_a
        - /data/path_val_b
  multiple_validation_sets: true
```

- **Train split** (first item) is a pair: `([prefixes], [weights])`.
- **Validation split** (second item) is a list of prefixes; **no weights** are used—each dataset is evaluated **individually** in sequence.
- When `multiple_validation_sets: true`, the runner logs **per-dataset** metrics (e.g., `val_loss/<dataset_name>`) and an aggregated summary.
- If you **omit** `multiple_validation_sets` (or set it to `false`), only the **aggregated** validation metric is logged (the individual per-dataset curves are not emitted).

## Backward Compatibility

- **No breaking changes.** Default paths remain identical when `multiple_validation_sets` is not provided.
- Logging keys for per-dataset metrics are **namespaced** (e.g., `val_loss/<name>`), so existing dashboards won’t break.

---

## Performance Considerations

- Multiple validation datasets are evaluated **sequentially** using the existing forward/eval loop.
- Recommended defaults for many eval datasets:
  - `cfg.dataset.val_num_workers: 0` # may test with 1 to reduce latency between eval batches
  - `cfg.dataset.val_persistent_workers: false` # if true, will likely result in RAM OOM
  - `cfg.dataset.val_pin_memory: true` # may test with false to reduce RAM usage, or with `non_blocking=True` in H2D transfer
  - `cfg.validation.val_micro_batch_size: 1` # max it up to vRAM limit
  - `cfg.validation.val_global_batch_size` # defaults to training `global_batch_size` if unset
- Total eval wall-time increases with the number of datasets; this PR logs a **single total evaluation duration** (`evaluate-all-datasets` timer) to track that cost clearly.
 
---

## Testing Notes

- **Ruff check and format passed** (`ruff check`, `ruff format`).
- **11 unit tests passed** (3 in `test_config.py::TestValidationConfigResolution`, 4 in `test_eval.py::TestValidationDataloaderCreation`, 4 in `test_eval.py::TestEvaluateAndPrintResults`).
- **Smoke test** on Qwen3-600M with 2 GPUs using `nemo:25.11.nemotron_3_nano` container. Loss decreased from 9.67 to 7.37 over 97 iterations. "aggregated across 2 datasets" logged at every eval interval. Per-dataset and aggregated metrics visible in WandB.

---

## Documentation

- Added/updated recipe/test materials describing how to enable `dataset.multiple_validation_sets` and how per-dataset metrics appear in logs.

---

## Checklist

- [x] Feature is opt-in; default behavior unchanged
- [x] Lint & format pass locally (`ruff check`, `ruff format`)
- [x] 11 unit tests pass
- [x] 2-GPU smoke test with 2 validation datasets passes
- [x] Logging keys namespaced per dataset (`val_loss/<name>`, `val_ppl/<name>`)

---

## Request for Review

- Functional review of the evaluation changes
- Guidance on naming/placement of the new config field (happy to align with project conventions)
- Feedback on logging keys and whether maintainers prefer an additional blended/weighted summary

## Summary by CodeRabbit

* **New Features**
  * Added support for multiple validation datasets during evaluation with independent configuration options.
  * Introduced validation-specific parameters for batch sizing, worker threads, memory pinning, and persistent workers.
  * Enhanced evaluation logging with per-dataset and aggregated metrics.

* **Tests**
  * Added comprehensive unit tests for validation configuration resolution and multi-dataset evaluation flows.


src/megatron/bridge/data/loaders.py
```diff
@@ -200,52 +201,98 @@ def worker_init_fn(_):
...
-            cfg.dataset.data_sharding,
-            worker_init_fn=maybe_worker_init_fn,
-            collate_fn=valid_ds.collate_fn if hasattr(valid_ds, "collate_fn") else None,
-            pin_memory=cfg.dataset.pin_memory,
-            persistent_workers=cfg.dataset.persistent_workers,
-            data_parallel_rank=mpu.get_data_parallel_rank(),
-            data_parallel_size=mpu.get_data_parallel_world_size(),
-        )
 
-    if cfg.train.eval_iters > 0:
```
ananthsub: with multiple datasets:
1. is `eval_iters` intended to control the total number of validation iters, or the number per dataset?
2. we still need a convention of skipping evaluation. currently this was based on eval_iters = 0, which is no longer checked here. this change was merged relatively recently, so maybe it's a result of not rebasing?

src/megatron/bridge/training/eval.py
```diff
@@ -39,6 +40,7 @@ def evaluate(
     config: ConfigContainer,
     verbose: bool = False,
     non_loss_data_func: Optional[Callable] = None,
+    log_timing: bool = True,  # Add this parameter
```
ananthsub: this seems like it's better served in the logger config, alongside the other logger options

src/megatron/bridge/training/config.py
```diff
@@ -1034,6 +1059,40 @@ def validate(self) -> None:
...
+        self.dataset.val_persistent_workers = (
+            self.dataset.val_persistent_workers
+            if self.dataset.val_persistent_workers is not None
+            else self.dataset.persistent_workers
+        )
+        self.train.val_micro_batch_size = (
+            self.train.val_micro_batch_size
+            if self.train.val_micro_batch_size is not None
+            else self.train.micro_batch_size
+        )
```
ananthsub: nit: since these fields only check against other within the same config, this fallback can be implemented in the individual subconfig's `finalize()` methods. the overall config container validation is intended to validate cross-cutting dependencies across sub config classes

ananthsub (commented):
thanks for the detailed contribution @hawkoli1987 ! 
we'd like to reconcile the implementation here with https://github.com/NVIDIA/Megatron-LM/commit/b6a7f40374410fc93bade2abe33e2a4979b41f75 + https://github.com/NVIDIA/Megatron-LM/commit/237080ba487796940e705985c866cf237430a352 as we aim to merge the MLM and mbridge training loops/configs together. have you already seen the corresponding commits in MLM?

src/megatron/bridge/data/loaders.py
hawkoli1987: Thanks @ananthsub for the review!

1. if we enable `multiple_validation_sets = True`, the `eval_iters` will be number per dataset
2. I think that logic of skipping evaluation remains in the [train.py](https://github.com/hawkoli1987/Megatron-Bridge/blob/6435d94f18c8c2e591954bbe1532097eb4d4f9b5/src/megatron/bridge/training/train.py#L367)

src/megatron/bridge/training/eval.py
hawkoli1987: let me clarify,

this is intended as a 'local' variable to control the logging of timing for individual dataset, which I took a strong assumption that users generally:
- want the logging of evaluation time at each eval interval (hence this flag was not exposed in config before)
- prefer to have a logging of the timing for the total evaluation duration instead of individual val dataset duration when in `multiple_validation_sets` mode, to de-clutter the logs

in the original use case with single evaluation dataset, this `log_timing` should default to 'True', so it doesn't change the default behavior. (I made a mistake [here](https://github.com/hawkoli1987/Megatron-Bridge/blob/6435d94f18c8c2e591954bbe1532097eb4d4f9b5/src/megatron/bridge/training/eval.py#L367) by setting it to False, will correct it)

However, if the intention is to give users the handle if the timing of evaluation should be logged at all, I will add it to the logger config

src/megatron/bridge/training/config.py
hawkoli1987: Good point! Will migrate to `TrainingConfig.finalize()` if that's ok?

hawkoli1987:
> thanks for the detailed contribution @hawkoli1987 ! we'd like to reconcile the implementation here with [NVIDIA/Megatron-LM@b6a7f40](https://github.com/NVIDIA/Megatron-LM/commit/b6a7f40374410fc93bade2abe33e2a4979b41f75) + [NVIDIA/Megatron-LM@237080b](https://github.com/NVIDIA/Megatron-LM/commit/237080ba487796940e705985c866cf237430a352) as we aim to merge the MLM and mbridge training loops/configs together. have you already seen the corresponding commits in MLM?

Sorry @ananthsub , when i check the implementation on mcore side, I only found that when the `multiple_validation_sets' was configured, it returns a list of val_datasets instead of a single val_dataset. That's how I build the rest of evaluation logic in this PR. I wasn't paying attention to the `megatron/training/training.py` as I understood Megatron-Bridge only utilize the `megatron/core`, and has its own set of training logic in `src/megatron/bridge/training/eval.py`?

ezKoya:
Are there any plans or news on adding this feature?

yaoyu-33:
Hi I am following up to get this merged! can you help to resolve the conflicts please? @hawkoli1987

hawkoli1987:
sure @yaoyu-33 , will fix it soon!

hawkoli1987:
hello  @yaoyu-33 
 
to make it clean, instead of rebasing from main, I branched from the latest 'main', and reapplied all changes and tests, incorporating the suggestions as discussed with @ananthsub previously, and force pushed the changes.
 
Thanks!

src/megatron/bridge/training/config.py
```diff
@@ -185,12 +183,6 @@ class DistributedInitConfig:
     enable_megatron_core_experimental: bool = False
     """Enable experimental features for Megatron Core."""
 
-    distributed_timeout_seconds_after_init: int | None = None
```
yaoyu-33, line 198: are these removal in the current file expected? seems non related.

src/megatron/bridge/data/loaders.py
```diff
@@ -204,58 +205,99 @@ def worker_init_fn(_):
         persistent_workers=cfg.dataset.persistent_workers,
         data_parallel_rank=mpu.get_data_parallel_rank(),
         data_parallel_size=mpu.get_data_parallel_world_size(),
-        global_batch_size=cfg.train.global_batch_size,
```
yaoyu-33, line 213: is this change intentional?

yaoyu-33:
@hawkoli1987 : fixed one conflict, plz double check that. 

Mostly looks good now, but some removal I am not sure they are intentional, can you double check the changes please. Thanks.

src/megatron/bridge/training/config.py
hawkoli1987: sorry this wasn't intentional, please revert

src/megatron/bridge/data/loaders.py
hawkoli1987: this wasn't intentional either, please revert

yaoyu-33:
@hawkoli1987  wonder if you have time to revert the un-intentional changes to get this merged? It's on your fork might be easier for you to change. if you need help lemme known

hawkoli1987:
sure, @yaoyu-33 , will do!

maanug-nv:
Hi @hawkoli1987 , don't see any new commits since your last response. any update?

hawkoli1987:
hello @maanug-nv 

Sorry for the delay. I have again reworked  on this PR to make it compatible with upstream changes. Please review, thanks!

yaoyu-33:
@hawkoli1987  : can you fix lint issues? check contribute.md plz

PR #713, opened 2025-09-22
fix: log config container as dict for wandb config

liding-nv:
cfg is not properly uploaded to wandb. In wandb config panel, it only shows as
```
_call_: true
_target_: megatron.bridge.training.config.ConfigContainer
```
need to call `cfg.to_dict()` to convert config container object to dict.

### DIFF
--- a/tests/unit_tests/training/test_state.py
+++ b/tests/unit_tests/training/test_state.py
@@ -376,6 +376,7 @@ class TestGlobalState:
         mock_config.logger.wandb_save_dir = "/tmp/wandb"
         mock_config.logger.wandb_entity = "test_entity"
         mock_config.checkpoint.save = "/tmp/checkpoints"
+        mock_config.to_dict.return_value = {"config": "data"}
         state._cfg = mock_config
 
         mock_wandb = MagicMock()


src/megatron/bridge/training/state.py
```diff
@@ -197,7 +197,7 @@ def wandb_logger(self) -> Optional[Any]:
                     "dir": save_dir,
                     "name": self.cfg.logger.wandb_exp_name,
                     "project": self.cfg.logger.wandb_project,
-                    "config": yaml.safe_load(dump_dataclass_to_yaml(self.cfg)),
+                    "config": self.cfg.to_dict(),
```
ananthsub, line 198: thanks for the fix @liding-nv . could you attach a screenshot to the PR demonstrating the full config is now shown?

liding-nv: This is the old wandb config
image

This is the new wandb config
image

ananthsub: awesome, thanks for fixing this!

please address the DCO check for CI. then you can trigger tests for the PR following the instructions here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci

PR #714, opened 2025-09-22
Fix lr warmup

### LINKED ISSUE #693: Issue with Learning Rate Schedule
When using `lr_warmup_fraction` in `SchedulerConfig`, the learning rate warmup does not appear as expected. In the example below, with `lr_warmup_fraction: 0.1` I expected to see around 35 iterations of warmup, but instead I only saw one.

Image

I can fix the issue if others have it too.

sanjana-inflection:
Fixes #693


ananthsub (approved):
thanks for the fix and updating the tests!

PR #725, opened 2025-09-23
[docs] add parallelism guide

ananthsub:
Fixes https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/740

- port assets and content from https://docs.nvidia.com/nemo-framework/user-guide/latest/nemotoolkit/features/parallelisms.html
- merge context parallelism from https://docs.nvidia.com/nemo-framework/user-guide/latest/longcontext/contextparallel.html into this page too to have one place where different parallelisms are covered


docs/index.md
```diff
@@ -7,6 +7,7 @@
 :hidden:
 
 bridge-guide.md
+parallelisms.md
```
maanug-nv, line 9: I think this can go under training and customization as well. or a technical details section

docs/parallelisms.md
```diff
@@ -0,0 +1,380 @@
...
+### Distributed Data Parallelism
+
+Distributed Data Parallelism (DDP) keeps the model copies consistent by synchronizing parameter gradients across data-parallel GPUs before each parameter update. More specifically, it sums the gradients of all model copies using all-reduce communication collectives.
+
+![Distributed Data Parallelism](images/ddp.gif)
+*Figure: Distributed Data Parallelism synchronizes gradients across multiple GPUs using all-reduce operations.*
+
+### Distributed Optimizer
+
+Distributed optimizer is a memory-optimized data-parallel deployment method. It shards the optimizer states and the high-precision master parameters across data-parallel GPUs instead of replicating them. At the parameter optimizer step, each data-parallel GPU updates its shard of parameters. Since each GPU needs its own gradient shard, the distributed optimizer conducts reduce-scatter of the parameter gradients instead of all-reduce of them. Then, the updated parameter shards are all-gathered across data-parallel GPUs. This approach significantly reduces the memory need of large-scale LLM training.
```
maanug-nv: We could also link to MCore docs here, they are more thorough.

docs/parallelisms.md
```diff
@@ -0,0 +1,380 @@
...
+
+### Enable Data Parallelism
+
+In Megatron Bridge, DDP is the default parallel deployment method. The total number of GPUs corresponds to the size of the DP group, and training an LLM with model parallelism decreases the size of the DP group.
+
+To enable the distributed optimizer, configure the {py:class}`bridge.training.config.OptimizerConfig`:
```
maanug-nv: and the ddp config

docs/parallelisms.md
```diff
@@ -0,0 +1,380 @@
...
+)
+```
+
+For more optimizer options, refer to the {py:class}`bridge.training.config.OptimizerConfig` API documentation.
+
+## Model Parallelism
+
+Model Parallelism (MP) is a distributed model deployment method that partitions the model parameters across GPUs to reduce the need for per-GPU memory. Megatron Bridge supports various model-parallel methods through Megatron Core, which can be mixed to maximize LLM training performance.
+
+### Tensor Parallelism
```
maanug-nv, line 53: would it be useful to go into depth on row parallel vs column parallel?

docs/index.md
ananthsub: i'd kept it at the top level since these settings would also be used for the HF bridge users outside of training

maanug-nv: fair

PR #729, opened 2025-09-23
Malay/qwen3 2509

malay-nagda:


scripts/performance/utils/helpers.py
```diff
@@ -62,6 +62,31 @@
...
 
+def set_megatron_fsdp_overrides(recipe: Any, perf_overrides: Any) -> None:
+    """Set the mcore fsdp overrides from the performance matrix."""
+    use_megatron_fsdp = perf_overrides.get("use_megatron_fsdp", False)
+    if use_megatron_fsdp:
+        recipe.ddp.use_megatron_fsdp = True
+        recipe.ddp.data_parallel_sharding_strategy = "optim_grads_params"
+        recipe.ddp.keep_fp8_transpose_cache = False
+        # average_in_collective is not supported with Megatron FSDP
+        recipe.ddp.average_in_collective = False
```
sanandaraj5597, line 75: We should set this to True Malay?

malay-nagda: - this is still unset here for fsdp- https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/training/config.py#L998-L1004.
     - So, even if I set to True, it will be unset above
- If I comment the average in collective lines in config.py, I get the error-
```
[rank40]:   File "/opt/megatron-lm/megatron/core/distributed/fsdp/mcore_fsdp_adapter.py", line 100, in __init__
[rank40]:     module=MegatronFSDP(
[rank40]:            ^^^^^^^^^^^^^
[rank40]:   File "/opt/megatron-lm/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py", line 248, in __init__
[rank40]:     self._init_fsdp_param_and_grad_buffer()
[rank40]:   File "/opt/megatron-lm/megatron/core/distributed/fsdp/src/megatron_fsdp/megatron_fsdp.py", line 278, in _init_fsdp_param_and_grad_buffer
[rank40]:     self.dist_index.get_dp_group(is_expert_parallel=True).size()
[rank40]:     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank40]: AttributeError: 'NoneType' object has no attribute 'size'
```

malay-nagda: Also, I get relative perf even with this unset

sanandaraj5597: Yeah this adds less than a percentage of perf.

ananthsub: edit: if it errors out then we'll need the fallback

those validation checks in the config were added since we didn't have https://github.com/NVIDIA/Megatron-LM/commit/d91daa0cd69fbd885f7867d05a0ec229ad4f4438 and https://github.com/NVIDIA/Megatron-LM/commit/dd7e1315008a1616e203f22b56bd64ebb7de2ec8 merged

sanandaraj5597:
LGTM.

sanandaraj5597 (approved):

scripts/performance/utils/helpers.py
sanandaraj5597: I agree, lets have average in collective disabling. Lets get the fusion disabling removed in this PR.

PR #738, opened 2025-09-23
Fix few issues for bridge export

yaoyu-33:
1. cpu only export
2. fix checkpoint loading by reset model cfg's parallel
3. delete extra state from state_dict for loading


src/megatron/bridge/training/checkpointing.py
```diff
@@ -137,6 +137,35 @@ def get_checkpoint_version() -> Optional[float]:
     return _CHECKPOINT_VERSION
 
 
+def delete_extra_state(state_dict):
```
ananthsub, line 140: why do we need this?

yaoyu-33: i find the solution on slack: https://nvidia.slack.com/archives/C08ACB3NFGQ/p1757237344633979

yaoyu-33: TE is raising the same error.

ananthsub: is there a stacktrace for the issue you saw? i'm apprehensive about doing this unconditionally across all model loads? is it restricted to specific TE versions?

ananthsub: this would now be called for general checkpoint saving/loading during training since `_generate_model_state_dict` is called by `generate_state_dict`

src/megatron/bridge/training/checkpointing.py
```diff
@@ -137,6 +137,45 @@ def get_checkpoint_version() -> Optional[float]:
...
+                   - A full checkpoint dict with a "model" key, or
+                   - A model state dict directly
+
+    Returns:
+        The modified state dictionary with extra state keys removed.
+    """
+    # Handle both cases: full checkpoint dict with "model" key or direct model state dict
+    if isinstance(state_dict, dict) and "model" in state_dict:
+        # Full checkpoint dict case
+        target_dict = state_dict["model"]
```
ananthsub, line 158: this will miss keys formatted for pipeline parallelism: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/6941a9319c2eaf297db7aa1252874c2f58ccfea5/src/megatron/bridge/training/checkpointing.py#L825-L836

src/megatron/bridge/training/model_load_save.py
```diff
@@ -307,6 +307,7 @@ def load_megatron_model(
     return_state_dict: bool = False,
     use_cpu_init: bool = False,
     skip_temp_dist_context: Optional[bool] = None,
+    override_provider: TransformerConfig = None,
```
```suggestion
    override_provider: TransformerConfig | None = None,
```

src/megatron/bridge/training/model_load_save.py
```diff
@@ -323,13 +324,31 @@ def load_megatron_model(
...
 
     Returns:
         The model instance with loaded weights if return_state_dict is False,
         otherwise returns a dictionary containing the full, unsharded model state_dict.
     """
 
-    model_cfg, mlm_args = load_model_config(checkpoint_path)
+    if override_provider is not None:
+        model_cfg = override_provider
+        mlm_args = None
```
ananthsub: this will break loading MLM models if override_provider is passed in, since mlm_args needs to be available. why not always load the model config and then override model_cfg with override_provider? 

@maanug-nv if you have more thoughts here.

previously, this was a 2-step process for users who needed more customization:

1. call `load_model_config` to get back the transformer config. apply any overrides on this
2. call `build_and_load_model` with the updated config to initialize the megatron model

example: https://github.com/NVIDIA-NeMo/Export-Deploy/blob/df3415779b4bcb1c829368fbb8f8a00a10fac560/nemo_deploy/nlp/inference/inference_base.py#L212-L268

my concern with passing the override is that users could pass a plain TransformerConfig in and breaking the expectation that it's also a model provider for mbridge usage in L248.

could you share an example for where/how `load_megatron_model` is being called now?

docs/parallelisms.md
```diff
@@ -15,7 +15,7 @@ Distributed Data Parallelism (DDP) keeps the model copies consistent by synchron
...
+[Distributed optimizer](https://docs.nvidia.com/megatron-core/developer-guide/latest/distrib_optimizer.html) is a memory-optimized data-parallel deployment method. It shards the optimizer states and the high-precision master parameters across data-parallel GPUs instead of replicating them. At the parameter optimizer step, each data-parallel GPU updates its shard of parameters. Since each GPU needs its own gradient shard, the distributed optimizer conducts reduce-scatter of the parameter gradients instead of all-reduce of them. Then, the updated parameter shards are all-gathered across data-parallel GPUs. This approach significantly reduces the memory need of large-scale LLM training.
```
ananthsub: bad rebase?

src/megatron/bridge/training/checkpointing.py
```diff
@@ -137,6 +137,45 @@ def get_checkpoint_version() -> Optional[float]:
...
+        target_dict = state_dict
+
+    # If target is not a mapping-like object, nothing to clean
+    if not hasattr(target_dict, "keys"):
+        return state_dict
+
+    # Some objects may implement keys() but not be directly iterable into a list (e.g., mocks)
+    try:
+        keys = list(target_dict.keys())
+    except Exception:
```
maanug-nv, line 170: some exception message here is probably useful

Issue #770, opened 2025-09-25
Llama3 8b bf16 cleanup error


salberdi-nvidia:
**Describe the bug**

After finishing training, llama3 8b errors out with the following:
`[2025-09-24 16:08:38] iteration       25/      25 | consumed samples:         3200 | elapsed time per iteration (ms): 13029.8 | learning rate: 3.750000E-06 | global batch size:   128 | lm loss: 3.664289E+00 | loss scale: 1.0 | grad norm: 11.408 | number of skipped iterations:   0 | number of nan iterations:   0 |
[after training is done] datetime: 2025-09-24 16:08:38
[rank0]:[W924 16:08:40.642247764 ProcessGroupNCCL.cpp:1505] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources. For more info, please see https://pytorch.org/docs/stable/distributed.html#shutdown (function operator())
[rank1]:[W924 16:08:40.708499746 ProcessGroupNCCL.cpp:1505] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources. For more info, please see https://pytorch.org/docs/stable/distributed.html#shutdown (function operator())
[rank6]:[W924 16:08:40.728819510 ProcessGroupNCCL.cpp:1505] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources. For more info, please see https://pytorch.org/docs/stable/distributed.html#shutdown (function operator())
[rank7]:[W924 16:08:40.728828322 ProcessGroupNCCL.cpp:1505] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources. For more info, please see https://pytorch.org/docs/stable/distributed.html#shutdown (function operator())
[rank4]:[W924 16:08:40.736044290 ProcessGroupNCCL.cpp:1505] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources. For more info, please see https://pytorch.org/docs/stable/distributed.html#shutdown (function operator())
[rank2]:[W924 16:08:41.150081871 ProcessGroupNCCL.cpp:1505] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources. For more info, please see https://pytorch.org/docs/stable/distributed.html#shutdown (function operator())
[...]

PR #772, opened 2025-09-25
destroy process group at end of performance script

ananthsub:
fixes https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/770

### DIFF
--- a/scripts/performance/run_script.py
+++ b/scripts/performance/run_script.py
@@ -167,6 +167,7 @@ def main():
     pretrain(config=recipe, forward_step_func=forward_step)
 
     if torch.distributed.is_initialized():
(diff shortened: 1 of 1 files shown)


scripts/performance/run_script.py
```diff
@@ -138,6 +139,9 @@ def main():
 
     pretrain(config=recipe, forward_step_func=forward_step)
 
+    if torch.distributed.is_initialized():
+        torch.distributed.destroy_process_group()
```
ananthsub, line 171:
```suggestion
    if torch.distributed.is_initialized():
        torch.distributed.barrier()
        torch.distributed.destroy_process_group()
```

PR #773, opened 2025-09-25
update get mixed precision config helper

ananthsub:
1. add a small convenience to accept either a str or MixedPrecisionConfig so users / recipes don't have to keep adding checks like this:

```
if isinstance(mixed_precision_config, str):
    mixed_precision_config = get_mixed_precision_config(mixed_precision_config)

# apply overrides to config
```

2. register hyphenated names along with underscores to support nemo2 conventions like "bf16-mixed" alongside "bf16_mixed"


maanug-nv (approved):
discussed 1 nitpick offline, lgtm otherwise

PR #781, opened 2025-09-26
add dclm tutorial

dimapihtar:


tutorials/data/dclm/shuffle.py
```diff
@@ -0,0 +1,100 @@
...
+    return parser
+
+
+def shuffle_data(
+    path_to_save: str,
+    source_file: str,
+    num_workers: int = 1,
+    lines_per_split: int = 1000000,
+) -> None:
+    """Merges hundreds of small .jsonl files into single .json file.
```
maanug-nv: This docstring seems incorrect

maanug-nv (changes requested):
I think these scripts are really short. Most of it seems like boilerplate (argparsing, making dirs, moving files), and the core stuff is in a few bash commands and `huggingface_hub.snapshot_download()`. 
To me, that makes this content more suitable for a notebook, especially since it's meant to be a tutorial. Most of the boilerplate could be removed in a notebook format. Any required arguments like save paths would just need a comment or instructions on setting them. 

I guess the one benefit I can see of scripts in this format is that they are convenient for customers to launch a few jobs to download the whole dataset. If we think that's useful to provide, we can have both the scripts and the notebook.

dimapihtar:
> I think these scripts are really short. Most of it seems like boilerplate (argparsing, making dirs, moving files), and the core stuff is in a few bash commands and `huggingface_hub.snapshot_download()`. To me, that makes this content more suitable for a notebook, especially since it's meant to be a tutorial. Most of the boilerplate could be removed in a notebook format. Any required arguments like save paths would just need a comment or instructions on setting them.
> 
> I guess the one benefit I can see of scripts in this format is that they are convenient for customers to launch a few jobs to download the whole dataset. If we think that's useful to provide, we can have both the scripts and the notebook.

good point. I think I can add notebook version also and we can have both approaches.

PR #782, opened 2025-09-26
rename seq_length to sequence_length

dimapihtar:


ananthsub (changes requested):
I don't think we should rename everything to sequence_length:
- `seq-length` is the arg that MLM users are exposed to currently, we should keep it this way for backwards compatibility
- `seq_length` is what NeMo FW already uses

since the confusion stems from `sequence_length` on the megatron core's ``BlendedMegatronDatasetConfig`, what do you think about adding `seq_length` as another attribute to the config here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/938e8a2320caa78c8609d1812f96fd45fcae5c31/src/megatron/bridge/training/config.py#L297-L329

and synchronizing `seq_length` with `sequence_length` here the way it's done for fp8 in the mixed precision config:  https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/938e8a2320caa78c8609d1812f96fd45fcae5c31/src/megatron/bridge/training/mixed_precision.py#L74-L80

this way, we could refer everything to use seq_length while keeping compatibility with mcore. the con is that we are duplicating fields, but this is something we could consider upstreaming

PR #788, opened 2025-09-26
Add workflow to cache a HF model 

maanug-nv:


ko3n1g (approved):

ko3n1g (commented):
Do we need to set HF Home? Might be safer?

maanug-nv:
@ko3n1g mount point is `/home` right? Is 165b00a5920085ded5e3f7af11b79f7c19c82a80 correct?

ko3n1g:
No, please read my comment again. This workload doesn’t run in a container but directly on the node, so the path is inside /mnt

ko3n1g:
You will also need to set up python and install transformers

ananthsub (commented):
@maanug-nv could you also update this? https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md

with an adding new model section, for when this is needed, who can, and how to trigger this workflow

maanug-nv:
Sorry my mistake on the cache dir, lack of understanding. 
Added 'setup-python' step and `pip install transformers` step.

Also wasn't sure what is the best choice for runner and environment here, let me know if that should be changed.

maanug-nv:
Tested with Gemma 3 270m: https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/18051546710/job/51374252053

Re-requesting review, in case there's a better way to handle the HF access token secret or pulling licensed models.

Currently my access token is used, but any maintainer of this repo should be able to replace the secret with their own, eg when next set of Llama models is released and needs to be cached.

ko3n1g (approved):

PR #801, opened 2025-09-27
Support sample-based training

### LINKED ISSUE #720: Support sample-based training
**Is your feature request related to a problem? Please describe.**
Support sample-based training as an alternative to specifying iteration-based training

ananthsub:
Part of #720 

Support sample based training based Megatron-LM support. Changes are made for the following:

1. Include `train_samples` in the TrainingConfig
2. Add sample-based fields to the LR Scheduler Config
3. Validate mutual exclusivity between samples vs iters: users should fully specify either iteration-based training or sample-based training
4. Update data loader logic to account for calculating samples
5. Add a sampler-based optimizer utility for distributed adam w/ cosine annealing to match the utility offered for iteration-based training

TODO: update docs

### DIFF
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -420,6 +420,28 @@ class SchedulerConfig:
         if self.override_opt_param_scheduler:
             assert not self.use_checkpoint_opt_param_scheduler, "both override and use-checkpoint are set."
 
+        # Validate mutual exclusivity between iteration-based and sample-based scheduler fields
+        has_iter_fields = (
+            self.lr_decay_iters is not None or self.lr_warmup_iters != 0 or self.lr_wsd_decay_iters is not None
+        )
+        has_sample_fields = (
+            self.lr_decay_samples is not None or self.lr_warmup_samples != 0 or self.lr_wsd_decay_samples is not None
+        )
+
+        assert not (has_iter_fields and has_sample_fields), (
+            f"Cannot mix iteration-based and sample-based scheduler fields. "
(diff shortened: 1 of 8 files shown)


src/megatron/bridge/training/config.py
```diff
@@ -333,7 +342,7 @@ class SchedulerConfig:
     wd_incr_steps: Optional[int] = field(init=False, default=None)
     wsd_decay_steps: Optional[int] = field(init=False, default=None)
 
-    def finalize(self):
+    def finalize(self) -> None:
```
maanug-nv, line 414: Should we add a similar check of not setting both samples and iters settings here? In the case that a user sets train samples, but both warmup iters and warmup samples, should we error or let training config define priority?

in this config though, both can be None.

src/megatron/bridge/training/config.py
```diff
@@ -1100,6 +1126,60 @@ def validate(self) -> None:
...
         self._sync_and_validate_external_cuda_graph()
 
+    def _validate_training_scheduler_compatibility(self) -> None:
+        """Cross-validation between training and scheduler configs."""
+        has_train_samples = self.train.train_samples is not None
+
+        if has_train_samples:
+            # Sample-based training validation
+            if self.scheduler.lr_decay_iters is not None:
+                raise ValueError("Use lr_decay_samples for sample-based training, not lr_decay_iters")
```
maanug-nv: nit: Why `ValueError` instead of just `assert`? Contributors might be confused why we use both.

tests/unit_tests/training/test_config.py
maanug-nv: nit: naming and comments no longer make sense.

PR #804, opened 2025-09-27
support async saving for CI end to end testing

ananthsub:
async saving is by default disabled in the checkpoint config and recipes. this adds the flag to enable this for testing


dimapihtar (approved):
LGTM. Thank you!

Issue #688, opened 2025-09-19
Make Mbrdige extendable


abhinavg4:
We want to use Mbrdige in our own code base, but there are some registries that don't allow us to simple extend. 

This is a live issue and I will update it with all the issues I find.

1. Data module [registry](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/347624593b89e444930e411a88ca3a0f5dc99219/src/megatron/bridge/data/utils.py#L150)

ananthsub:
for the dataset generation, https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/494 is one proposal to extend this as the current registry is limited

abhinavg4:
Things we need to do in on_train_start and end:

We have a few heavy models like VAEs that are only used for validation and we would like to load them and discard them.

PR #806, opened 2025-09-27
Improve typehints for `forward_step_func` and add tests for functor inputs

ananthsub:
- make checks for whether to inject the global state more robust based on the signature of `forward_step_func`
- improve typehints for `forward_step_func` to characterize the arguments we can support
- add tests for functors to validate that stateful forward step funcs can be passed in

part of #688


src/megatron/bridge/training/forward_step_func_types.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def __call__(
+        self,
+        data_iterator: Iterable,
+        model: GPTModel,
+        return_schedule_plan: bool = False,
+    ) -> tuple[torch.Tensor, LossFunction]: ...
+
+
+class FourArgForwardStep(Protocol):
```
AAnoosheh, line 164: Instead of all these why not just accept the same number of call args, but then use "injection" to update state attributes if they exist.

I.e. update `forward_step.state` and `forward_step.model` before calling `forward_step(minimum_args)`

ananthsub: I didn't want to assume that the user has these set as attributes on their class or try to enforce it on the protocol this way

AAnoosheh: Or what about just create an interface for a `def update_state(self, **kwargs)` and have subclasses implement it. Then call this function if it exists right before calling `forward_step()`

ananthsub: @AAnoosheh for this PR at least I want to add tests + documentation for what forward_step_func accepts. for now, this also keeps the training loop agnostic to whether a class or a function is passed in.

since `update_state` and `forward_step` would be called right after each other, the class could persist the args passed from the loop before going into the `forward_step`. 

for classes, especially with ones that might retain references to the state objects, we'd need to be careful about lifetimes of those objects since they would now live beyond the training loop lifetime

Issue #719, opened 2025-09-23
Cannot fine-tune a base model


thvasilo:
**Describe the bug**

I have a local download of a base model, e.g. Llama3.2-1B. I would like to try fine-tuning from the base model. My model path contains exactly the files listed in https://huggingface.co/meta-llama/Llama-3.2-1B-Instruct/tree/main

The process is able to load my model weights, however in `src/megatron/bridge/training/setup.py` there's a call to 

```
# Load pretrained checkpoint if available
if cfg.checkpoint.pretrained_checkpoint is None or not checkpoint_exists(cfg.checkpoint.pretrained_checkpoint):
    raise ValueError(f"Invalid pretrained checkpoint directory found: {cfg.checkpoint.pretrained_checkpoint}")
```

which call inside `src/megatron/bridge/training/utils/checkpoint_utils.py`

```python
def checkpoint_exists(checkpoints_path: Optional[str]) -> bool:
    """Check if a checkpoint directory exists.

    Args:
        checkpoints_path: Path to the potential checkpoint directory.

    Returns:
        True if the path exists, False otherwise.
    """
    if checkpoints_path is None:
        return False

    train_state_filename = os.path.join(checkpoints_path, f"{TRACKER_PREFIX}_{TRAIN_STATE_FILE}")

    if file_exists(train_state_filename):
        return True

    print(f"Didn't find {train_state_filename=}")

    # Fallback to the Megatron-LM tracker file
    path = get_checkpoint_tracker_filename(checkpoints_path)
    if MultiStorageClientFeature.is_enabled():
        msc = MultiStorageClientFeature.import_package()
        return msc.os.path.isfile(path)
    else:
        return os.path.isfile(path)
```

which tries to find additional files like train_state_filename='/models/meta-llama/Llama-3.2-1B-Instruct/latest_train_state.pt'

As a result I'm getting the error

```
[rank0]: Traceback (most recent call last):
[rank0]:   File "/workdir/llama3_peft_recipe.py", line 67, in <module>
[rank0]:     main()
[rank0]:   File "/workdir/llama3_peft_recipe.py", line 64, in main
[rank0]:     finetune(llama32_1b_config, forward_step_func=forward_step)
[rank0]:   File "/workdir/src/megatron/bridge/utils/decorators.py", line 36, in wrapper
[rank0]:     return func(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #809, opened 2025-09-27
[docs] add notes to checkpointing that save/loading path are expected to be in megatron format

ananthsub:
followup to question in https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/719

### DIFF
--- a/docs/training/checkpointing.md
+++ b/docs/training/checkpointing.md
@@ -2,7 +2,9 @@
 
 The {py:class}`bridge.training.config.CheckpointConfig` controls model checkpointing behavior, including saving and loading checkpoints, checkpoint formats, and various optimization features.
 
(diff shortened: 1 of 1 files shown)


maanug-nv:
Should include this in docstrings as well

docs/training/checkpointing.md
```diff
@@ -2,6 +2,8 @@
 
 The {py:class}`bridge.training.config.CheckpointConfig` controls model checkpointing behavior, including saving and loading checkpoints, checkpoint formats, and various optimization features.
 
+> **Note**: This documentation covers **Megatron-format checkpoints** used during training. For converting between 🤗 Hugging Face and Megatron formats, see the {doc}`../bridge-guide`.
```
maanug-nv: Built docs and this looks good as is. Just fyi, we have a code fence for notes that looks a bit different: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/docs/performance-guide.md?plain=1#L9-L11

maanug-nv (approved):

PR #819, opened 2025-09-29
Fix DP size calculation in config.py

yaoyu-33:


ananthsub (approved):

yaoyu-33:
seems incorrect just to change this value.

PR #830, opened 2025-10-01
Qwen 25vl training pipeline

yaoyu-33:


src/megatron/bridge/training/utils/omegaconf_utils.py
```diff
@@ -168,13 +169,31 @@ def _dataclass_to_omegaconf_dict(val_to_convert: Any, path: str = "") -> Any:
...
+            target = f"{inspect.getmodule(cfg_class).__name__}.{cfg_class.__qualname__}.from_dict"
+            logger.debug(f"Converting {cfg_class.__qualname__} at {current_path} to callable dict")
+            return {
+                "_target_": target,
+                "_call_": True,
+                "config_dict": val_to_convert.to_dict(),
+            }
+    except ModuleNotFoundError:
+        # transformers is optional; if unavailable, fall through to other handlers
+        pass
```
ananthsub, line 188: this should be handled here: 
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/f9aad1f3118813e81a079223db4467e012e9df98/src/megatron/bridge/utils/yaml_utils.py#L62-L77

were you seeing serialization issues outside of this?

src/megatron/bridge/training/vlm_step.py
```diff
@@ -0,0 +1,233 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
ananthsub, line 1: should we offer higher-level functions like `pretrain_gpt` / `finetune_gpt` / `pretrain_qwen25_vl`, `finetune_qwen25_vl` that automatically select the right step for users? 
or how else should we associate the custom forward step func with the recipe? otherwise it might not be obvious to users that an implementation exists

previously this was bound to the model config, but we want to keep the model providers independent of the training side

src/megatron/bridge/recipes/qwen_vl/qwen25_vl_dataset.py
```diff
@@ -0,0 +1,231 @@
...
+    random_seed: int = 0
+    image_size: Tuple[int, int] = (256, 256)
+    pad_to_max_length: bool = True
+    create_attention_mask: bool = True
+
+    # Keep parity with GPTDatasetConfig usage in batching utilities
+    skip_getting_attention_mask_from_dataset: bool = True
+
+    # HF AutoProcessor instance will be set during build
+    processor: Optional[Any] = None
```
ananthsub: this is not documented right now, but private fields are skipped when saving a checkpoint and serializing the config to yaml. so any fields you don't want serialized in the `run_config.yaml` can be prefixed with `_` 

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/c7f081bd841bd689cb5a5f46b72f4e98df9ceccd/src/megatron/hub/training/utils/config_utils.py#L121-L122

src/megatron/bridge/recipes/qwen_vl/qwen25_vl_step.py
```diff
@@ -0,0 +1,233 @@
...
+import torch
+from megatron.core import parallel_state
+from megatron.core.models.gpt import GPTModel
+from megatron.core.utils import get_batch_on_this_cp_rank, get_model_config
+
+from megatron.bridge.training.config import ConfigContainer
+from megatron.bridge.training.gpt_step import (
+    _create_loss_function,
+    get_packed_seq_params,
+)
```
ananthsub: we can spilt these functions out of `gpt_step` to indicate they are more generic

src/megatron/bridge/recipes/qwen_vl/qwen25_vl_step.py
```diff
@@ -0,0 +1,233 @@
...
+            _batch_required_keys[key] = val.cpu() if val is not None else None
+        else:
+            _batch_required_keys[key] = None
+
+    return _batch_required_keys
+
+
+def get_batch_on_this_tp_rank(*args, **kwargs):
+    # Re-export the shared implementation for backward compatibility
+    from megatron.bridge.training.gpt_step import get_batch_on_this_tp_rank as _shared_get_batch_on_this_tp_rank
```
ananthsub: same here

src/megatron/bridge/training/utils/omegaconf_utils.py
yaoyu-33: it's because we are using transformers' model config in the config, that's causing some issues here.

src/megatron/bridge/recipes/qwen_vl/qwen25_vl_step.py
yaoyu-33: i feel gpt_step is the general one lol

src/megatron/bridge/training/utils/omegaconf_utils.py
yaoyu-33: i dont think the yaml would work, here it's not for saving just for omega conf creation

src/megatron/bridge/training/vlm_step.py
yaoyu-33: good point, what do you think let recipe returns cfg and the step function?

ananthsub: returning the config and the step function will be a breaking change for existing users

src/megatron/bridge/training/vlm_step.py
```diff
@@ -0,0 +1,224 @@
...
+        batch.get("cu_seqlens"),
+        batch.get("cu_seqlens_argmin"),
+        batch.get("max_seqlen"),
+        batch.get("pixel_values"),
+        batch.get("image_grid_thw"),
+    )
+
+
+def forward_step(
+    state: GlobalState, data_iterator: Iterable, model: GPTModel, return_schedule_plan: bool = False
```
ananthsub, line 202: is it still a GPTModel?

src/megatron/bridge/recipes/qwen_vl/qwen25_vl_step.py
yaoyu-33: moved them out to utils

src/megatron/bridge/training/vlm_step.py
yaoyu-33: yes, now vlm is extension to llm

src/megatron/bridge/training/utils/loss_utils.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
ananthsub, line 1: you can include add this here instead: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/training/losses.py

we can also rename that that to be masked token loss specific

examples/recipes/qwen_vl/finetune_qwen25_vl.py
```diff
@@ -0,0 +1,184 @@
+#!/usr/bin/env python3
```
ananthsub, line 1: the initial example added here was to demonstrate how to override configs for the recipe: https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/examples/recipes/llama

i dont think we should propagate that style for more examples if it's not needed. going with the more dataclass approach will be easier for users to follow, especially since the config overrides don't add functionality and they make it harder to read the code. wdyt?

src/megatron/bridge/data/vlm_datasets/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+from .conversation_dataset import VLMConversationDataset
+from .hf_dataset_makers import (
+    make_cord_v2_dataset,
+    make_cv17_dataset,
+    make_medpix_dataset,
+    make_rdr_dataset,
+)
+from .hf_provider import HFDatasetConversationProvider
+from .mock_provider import MockVLMConversationProvider
+from .preloaded_provider import PreloadedVLMConversationProvider
```
ananthsub: nit: use absolute imports instead of relative ones

pyproject.toml
```diff
@@ -78,6 +78,7 @@ dependencies = [
     "tqdm>=4.67.1",
     "hydra-core>1.3,<=1.3.2",
     "megatron-core[dev,mlm]>=0.14.0a0,<0.16.0",
+    "qwen-vl-utils",
```
ananthsub, line 81: should we make it an optional dependency? its handled that way in the code but added here directly

src/megatron/bridge/data/vlm_datasets/collate.py
```diff
@@ -0,0 +1,318 @@
...
+    - For each assistant text, tokenize without special tokens and search sequentially
+    - On success, unmask that span; otherwise leave masked
+    """
+    tokenizer = getattr(processor, "tokenizer", processor)
+    ids = input_ids.tolist()
+    mask = [0] * len(ids)
+
+    def try_mark(span_text: str, start_from: int) -> int:
+        """Tokenize a span and mark its occurrence if found. Returns new search start index."""
+        variants = [span_text, span_text + "\n"]
```
cuichenx, line 81: this feels a bit brittle.. would the following work? 
1. exclude the first and last tokens
2. create mask based on the remaining tokens
3. mask out one token before and after the masked token 

the reason is that first and last tokens might have whitespace characters included


also, should this function throw an error if nothing is found? (i.e. mask is all zeros)

yaoyu-33: will update in next pr add "error if not found"

src/megatron/bridge/data/vlm_datasets/hf_provider.py
```diff
@@ -0,0 +1,116 @@
...
+from .hf_dataset_makers import (
+    make_cord_v2_dataset,
+    make_cv17_dataset,
+    make_medpix_dataset,
+    make_rdr_dataset,
+)
+
+
+@dataclass(kw_only=True)
+class HFDatasetConversationProvider(DatasetProvider):
```
ananthsub, line 36: did you have any issues when not specifying the dataloader type, which controls the samplers attached to the dataloader?

src/megatron/bridge/data/vlm_datasets/hf_provider.py
```diff
@@ -0,0 +1,116 @@
...
+    """DatasetProvider that builds VLM conversation datasets from HF datasets.
+
+    This provider leverages simple maker functions that return lists of examples
+    with a "conversation" schema understood by model processors. It binds a
+    HuggingFace `AutoProcessor` for the specified model and selects an
+    appropriate collate function for batching.
+    """
+
+    # Required to match model.seq_length (enforced by ConfigContainer.validate)
+    sequence_length: int
```
ananthsub, line 46: is it strictly needed? or should we update the validation logic to handle the dataset config not having sequence length set?

src/megatron/bridge/training/vlm_step.py
```diff
@@ -0,0 +1,271 @@
...
+) -> tuple[
+    torch.Tensor,
+    torch.Tensor,
+    torch.Tensor,
+    torch.Tensor,
+    torch.Tensor,
+    torch.Tensor,
+    torch.Tensor,
+    torch.Tensor,
+    Qwen2_5_VLVisualInputs,
```
ananthsub: it looks out of place for a generic vlm step to have model-specific inputs. not blocking for now, but if/when more are added, maybe it makes sense to specialize this to a qwenvl step

pyproject.toml
yaoyu-33: i hope we can add it as a hard dependency, it's essential for a few vl variants we are working on, also it's not a big pip package.

src/megatron/bridge/data/vlm_datasets/hf_provider.py
yaoyu-33: it's needed for some cases we need to do truncation or padding. maybe we can somehow consolidate those? but feels like a lower priority

src/megatron/bridge/training/vlm_step.py
yaoyu-33: updated to Any for now

src/megatron/bridge/data/vlm_datasets/hf_provider.py
yaoyu-33: though we dont set values here, we updated the values in recipes to be "single". dont see issues with this setting, maybe should use cyclic instead.

examples/recipes/qwen_vl/finetune_qwen25_vl.py
maanug-nv: +1 to this . unless there's anything unique to demonstrate for VLMs. but this example looks basically identical to llm. besides forward step. if we think users might need to customize that more often for VLMs, then we should write docs for that.

src/megatron/bridge/training/utils/omegaconf_utils.py
```diff
@@ -168,13 +169,31 @@ def _dataclass_to_omegaconf_dict(val_to_convert: Any, path: str = "") -> Any:
     """
     current_path = path
 
+    # Handle Hugging Face GenerationConfig / PretrainedConfig by converting to a callable dict
+    # compatible with our YAML representer logic
+    try:
```
maanug-nv, line 174: I guess this will resolve the issues we are seeing. But can you help me understand why can't we just store as a dict in the provider in the first place? And then when it needs to be consumed, check the type and build GenerationConfig from the dict?

maanug-nv: Also wasn't clear why this needs to be stored in the provider, isn't it already in the hf_pretrained classes (PretrainedCausalLM and PretrainedVLM), so it will be accessible from the AutoBridge? What is it used for?

src/megatron/bridge/recipes/qwen_vl/qwen25_vl_dataset.py
```diff
@@ -0,0 +1,231 @@
...
+from typing import Any, Dict, Optional, Tuple
+
+import numpy
+import torch
+from PIL import Image
+
+from megatron.bridge.training.config import DatasetBuildContext, DatasetProvider
+
+
+class MockQwen25VLDataset(torch.utils.data.Dataset):
```
suiyoubi, line 25: Do we still need this ? Or the common one in the `mock_provider` is sufficient ?

src/megatron/bridge/data/vlm_datasets/mock_provider.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Initialize and store processor
+        self._processor = AutoProcessor.from_pretrained(self.hf_processor_path, trust_remote_code=True)
+
+        base_examples = self._make_base_examples()
+
+        def _maybe_make(size: int) -> Optional[VLMConversationDataset]:
+            if not size or size <= 0:
+                return None
+            return VLMConversationDataset(
```
suiyoubi, line 102: is this intentional ? shouldn't it invoke `MockVLMConversationDataset ` in stead ?

src/megatron/bridge/training/utils/omegaconf_utils.py
yaoyu-33: will remove GenerationConfig.

ananthsub (approved):

thomasdhc (approved):

PR #848, opened 2025-10-01
[nemo run] Allow plugins to customize script args

### LINKED ISSUE #570: Augment run plugins to support custom script argument override styles
The NeMo-Run plugins here are targeted for recipe users: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/recipes/run_plugins.py

For users relying on the `run.Script` launch mechanism, the plugins today pass args according to the hydra-style overrides. however, users may not always configure their scripts to accept args in this way. 

To allow for more customizability, we can augment the plugins to take in a `script_args_converter_fn` which should return a list of args to append to the task's arguments. this would allow users to specify their own conversion function if their training script doens't use hydra-style overrides

ananthsub:
fixes #570 

By default, the plugins assume that scripts accept hydra-style overrides, but not all script users may use this convention. 

for these users to still benefit from nemo run plugins offered by the repo, plugins can now accept a converter function, which transforms a dataclass of options into a list of args the run script accepts. this allows users to customize how args are specified without needing to subclass/extend/write their own plugins

### DIFF
--- a/docs/recipe-usage.md
+++ b/docs/recipe-usage.md
@@ -198,6 +198,42 @@ if __name__ == "__main__":
     run.run(train_script, plugins=plugins, executor=executor)
 ```
 
+##### Custom Argument Converters
+
+By default, plugins convert their configuration to Hydra-style CLI arguments when used with `run.Script` tasks. If your training script uses a different argument format (e.g., argparse), you can provide a custom converter function via the `script_args_converter_fn` parameter.
+
+```python
+import nemo_run as run
+from typing import List
+from megatron.bridge.recipes.run_plugins import (
+    PreemptionPlugin,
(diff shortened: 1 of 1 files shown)


src/megatron/bridge/recipes/run_plugins.py
```diff
@@ -12,10 +12,45 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""
```
maanug-nv, line 15: I think this content should be added to the recipes guide [here](https://docs.nvidia.com/nemo/megatron-bridge/latest/recipe-usage.html#plugins).
Do we expect people to read this file? If so, we can leave it here as well.

maanug-nv (approved):

PR #865, opened 2025-10-03
docs: Revert 0.2.0 push

ko3n1g:


docs/versions1.json
aschilling-nv, line 1: The versions1.json should match this:

[
    {
        "preferred": true,
        "version": "0.1.0",
        "url": "../0.1.0"
    }
]

aschilling-nv (approved):

PR #873, opened 2025-10-03
ModelOpt QAT supportf

yueshen2016:
Support training process with loading and saving quantized checkpoint.


src/megatron/bridge/training/checkpointing.py
```diff
@@ -1387,15 +1386,6 @@ def _load_checkpoint_from_path(
...
         model_sd_kwargs = dict(metadata=sharded_sd_metadata)
 
-        # ModelOpt restoration
-        if has_nvidia_modelopt:
-            if ckpt_type == CheckpointType.LOCAL:
-                print_rank_0("WARNING: Local checkpointing does not support nvidia_modelopt.")
-            elif ckpt_type == CheckpointType.GLOBAL:
-                restore_modelopt_state(model, state_dict)
-            else:
-                restore_sharded_modelopt_state(model, checkpoint_name)
```
ananthsub: could you share why this is being removed?

src/megatron/bridge/training/setup.py
```diff
@@ -170,13 +170,24 @@ def setup(
         cfg.model.register_pre_wrap_hook(peft_hook)
         print_rank_0("Registered PEFT pre-wrap hook")
 
+    if cfg.model.restore_modelopt_state:
+        from megatron.bridge.training.post_training.checkpointing import load_modelopt_state
+
+        def modelopt_pre_wrap_hook(model):
+            checkpoint_path = cfg.checkpoint.pretrained_checkpoint or cfg.checkpoint.load
```
ananthsub: if model_opt state is in both `pretrained_checkpoint` and `load` directory paths, this will prioritize the pretrained checkpoint. is that expected? do you need to check explicitly for which path has the model opt state?

src/megatron/bridge/training/post_training/checkpointing.py
```diff
@@ -19,23 +19,55 @@
...
+    # Check for iter_* folders (inspired by AutoBridge implementation)
+    iter_folders = [
+        f
+        for f in os.listdir(checkpoint_path)
+        if os.path.isdir(os.path.join(checkpoint_path, f)) and f.startswith("iter_")
+    ]
+
+    if iter_folders:
+        # Find the folder with the largest iteration number
+        def get_iter_number(folder_name: str) -> int:
```
yaoyu-33: iter number better get from state dict not file name?
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/training/checkpointing.py#L2128

examples/recipes/llama/pretrain_llama3_8b.py
```diff
@@ -61,6 +61,7 @@
 from megatron.bridge.recipes.llama import llama3_8b_pretrain_config as pretrain_config
 from megatron.bridge.training.config import ConfigContainer
 from megatron.bridge.training.gpt_step import forward_step
+from megatron.bridge.training.post_training.checkpointing import has_modelopt_state
```
yaoyu-33: I would duplicate this script in quantization folder, not use this one.
If i was an external user, I would go to examples/quantization for this.

src/megatron/bridge/models/gpt_provider.py
```diff
@@ -89,12 +89,13 @@ def local_layer_spec(config: "GPTModelProvider") -> ModuleSpec:
 
 def quantization_layer_spec(config: "GPTModelProvider") -> ModuleSpec:
     """Layer specification for quantization with ModelOpt."""
+    use_arbitrary_attention_mask = parallel_state.get_context_parallel_world_size() == 1
```
yaoyu-33, line 98: add comments about this condition, explain why use_arbitrary_attention_mask depends on cp

src/megatron/bridge/training/config.py
```diff
@@ -1168,6 +1168,13 @@ def validate(self) -> None:
             if self.optimizer.use_precision_aware_optimizer:
                 self.ddp.preserve_fp32_weights = False
 
+        # ModelOpt/Quantization checks
+        if self.model.restore_modelopt_state and self.model.gradient_accumulation_fusion:
+            print_rank_0(
+                "Gradient accumulation fusion is not supported with ModelOpt/Quantized models, setting to False"
+            )
+            self.model.gradient_accumulation_fusion = False
```
yaoyu-33: I would rather fail here, and let user manually set `model.gradient_accumulation_fusion = False`.

src/megatron/bridge/models/gpt_provider.py
ChenhanYu: @yueshen2016 , we use arbitrary mask for speculative decoding training. For CP>1, only causal_mask_type can be used (since the attention backend requires causal_mask_type).

src/megatron/bridge/training/checkpointing.py
yueshen2016: First, I check with @yaoyu-33 and learnt that CheckpointType.LOCAL and CheckpointType.GLOBAL are not adopted anymore in Megatron-Bridge, so only restore_sharded_modelopt_state is needed for Megatron-Bridge. Second, load_modelopt_state function is called [here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/29933ed6c38dbcbe7fafd6c04a6dbabc684e9fb6/src/megatron/bridge/training/setup.py#L178) in setup.py, before this part, and restore_sharded_modelopt_state function is wrapped inside load_modelopt_state. That's why we could dump this code block.

src/megatron/bridge/models/gpt_provider.py
yueshen2016: Sure, the comments have been added.

src/megatron/bridge/training/post_training/checkpointing.py
yueshen2016: Thanks for the suggestion! Now iter number is fetched from state dict.

src/megatron/bridge/training/setup.py
yueshen2016: Thanks for the catch, now it will check which one has modelopt state dict first.

yueshen2016:
okay to test f3cf4372adc1bce26c80b1cf2e86bba926d8f338

yaoyu-33 (approved):

examples/recipes/llama/pretrain_llama3_8b.py
yueshen2016: Okay, could you explain why we need this duplication in examples/quantization?

src/megatron/bridge/training/config.py
yueshen2016: Sure, have fixed this part.

.github/workflows/cicd-main.yml
```diff
@@ -221,6 +221,8 @@ jobs:
           - script: L2_Launch_recipes_mamba
           - script: L2_Launch_recipes_qwen
           - script: L2_Launch_data
+          - script: L2_Launch_quantization
+            timeout: 50
```
chtruong814: It looks like this test runs in 15 minutes?  The default timeout is 30 minutes anyway so we probably don't need this override.

chtruong814 (commented):
I think we can remove the extended timout for the new test?

.github/workflows/cicd-main.yml
yueshen2016: Fixed.

PR #875, opened 2025-10-06
removed dependencies

salberdi-nvidia:
When run from a login node, using megatron-core[dev,mlm] would pull in many unnecessary packages that needed GPU libraries to be installed which would be unnaccessible from a login node

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,7 +77,7 @@ dependencies = [
     "pyyaml>=6.0.2",
     "tqdm>=4.67.1",
     "hydra-core>1.3,<=1.3.2",
-    "megatron-core[dev,mlm]>=0.14.0a0,<0.16.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -77,7 +77,7 @@ dependencies = [
     "pyyaml>=6.0.2",
     "tqdm>=4.67.1",
     "hydra-core>1.3,<=1.3.2",
-    "megatron-core[dev,mlm]>=0.14.0a0,<0.16.0",
+    "megatron-core>=0.14.0a0,<0.16.0",
```
nv-mollys, line 80: What is this fixing? Can we add a comment in the PR description please

salberdi-nvidia: Added comment in the description

nv-mollys (approved):

PR #885, opened 2025-10-07
[docs] Add docs for inproc restart

ananthsub:
Follow up for docs: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/499#issuecomment-3294449644

### DIFF
--- a/docs/training/resiliency.md
+++ b/docs/training/resiliency.md
@@ -539,3 +539,230 @@ The system uses `RerunDataIterator` to handle data replay:
 - **State Saving**: Captures data iterator state for reproducible re-runs.
 - **Replay Capability**: Can rewind and replay the same data batches.
 - **Checkpoint Support**: Saves/restores iterator state across job restarts.
+
+## In-Process Restart
+
+> **Warning**: This is a highly experimental feature and is subject to change in backwards incompatible ways without notice.
+
+The in-process restart mechanism provides automatic fault recovery by restarting the training function within the same operating system process when failures occur. Unlike traditional scheduler-level restarts, in-process restart eliminates the overhead of launching new jobs, starting containers, initializing Python interpreters, and creating new CUDA contexts.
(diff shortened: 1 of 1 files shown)


docs/training/resiliency.md
```diff
@@ -539,3 +539,230 @@ The system uses `RerunDataIterator` to handle data replay:
...
+Where:
+- `SLURM_JOB_ID`: The Slurm job ID from the `SLURM_JOB_ID` environment variable
+- `hostname`: The hostname of the node where rank 0 is running
+- `SLURM_PROCID`: The global rank from the `SLURM_PROCID` environment variable
+- `SLURM_LOCALID`: The local rank on the node from the `SLURM_LOCALID` environment variable
+
+**Python Configuration:**
+
+```python
+config.inprocess_restart = InProcessRestartConfig(
```
maanug-nv: nit: this is mentioned before Configuration section. Maybe move Slurm section after Config, Config options sections.

docs/training/resiliency.md
```diff
@@ -539,3 +539,230 @@ The system uses `RerunDataIterator` to handle data replay:
 - **State Saving**: Captures data iterator state for reproducible re-runs.
 - **Replay Capability**: Can rewind and replay the same data batches.
 - **Checkpoint Support**: Saves/restores iterator state across job restarts.
+
+## In-Process Restart
```
maanug-nv, line 543: iirc, in-process restart won't work for all types of failures, and it's usually best to use it in combination with in-job restarts (i think Szymon recommended this before, assuming it's still true). eg switch failure and several nodes that were part of the job become inaccessible. 

So I'm just suggesting we include that recommendation somewhere, probably at the end of this header, before key features.

maanug-nv (approved):
couple comments, but lgtm otherwise. fftm after addressing

PR #910, opened 2025-10-09
Interface for NVFP4 recipe

WanZzzzzz:


src/megatron/bridge/training/mixed_precision.py
```diff
@@ -53,6 +53,9 @@ class MixedPrecisionConfig:
     fp8_multi_head_attention: bool = False
     fp8_param: Optional[bool] = None
     fp8_param_gather: bool = False
+    # fp4 related
+    fp4: Optional[str] = None
+    fp4_recipe: str = "nvfp4"
```
ananthsub, line 59: thanks for sending this out @WanZzzzzz !

could we also validate in the `finalize` method below that neither to match this: 

- https://github.com/NVIDIA/Megatron-LM/blob/93aa335b952e8e53078c9dbd36686e72e0c2b1f9/megatron/training/arguments.py#L635-L636

- https://github.com/NVIDIA/Megatron-LM/blob/93aa335b952e8e53078c9dbd36686e72e0c2b1f9/megatron/training/arguments.py#L662-L672


should we also expose `fp4_param` here?

WanZzzzzz: Made the change. Actually "FP4 native weights" is not supported for now, so users can not enable fp4_param for now. I will also add the check in MCore.

ananthsub: if it's not supported yet, we shouldn't expose it here then

src/megatron/bridge/training/mixed_precision.py
```diff
@@ -86,6 +99,13 @@ def finalize(self):
                 "When fp8_param_gather=True and fp8_recipe='mxfp8', "
                 "reuse_grad_buf_for_mxfp8_param_ag must be set to True"
             )
+        # FP4 and FP8 are mutually exclusive
+        if self.fp4 and self.fp8:
+            raise ValueError("fp4 and fp8 cannot be used simultaneously. Please choose one.")
+
```
ananthsub, line 96: we also should have the TE version check for fp4: https://github.com/NVIDIA/Megatron-LM/blob/93aa335b952e8e53078c9dbd36686e72e0c2b1f9/megatron/training/arguments.py#L671

PR #655, opened 2025-09-16
Mamba (Nemotron-H incl Nano-v2) Model Support Enhancements

JRD971000:
# Changelog for Megatron Bridge MR

### **Mamba Model Support Enhancements**
- **Added specialized parameter mappings for Mamba architecture**:
  - `MambaInProjMapping`: Handles input projection weights with z, x, B, C, dt components and tensor-parallel distribution
  - `MambaConv1dMapping`: Manages 1D convolution weights for x, B, C components with proper TP distribution
- **Enhanced NemotronH bridge implementation**:
  - Replaced generic AutoMapping with specialized mappings for better Mamba SSM layer handling
  - Added proper tensor-parallel support for Mamba mixer components (A_log, D, dt_bias, norm.weight)
  - Implemented custom activation function (`nemotron_h_activation_func`) with `torch.pow(F.relu(x), 2)`

### **Custom Modeling File Preservation**
- **Added support for preserving custom modeling files** when using `trust_remote_code=True`:
  - Automatically copies custom modeling files (modeling_*.py, configuration_*.py, tokenization_*.py, etc.) during model saving
  - Supports both local directory and HuggingFace Hub model sources
  - Graceful fallback handling for file copy failures
- **Enhanced AutoBridge with custom modeling support**:
  - Added `set_custom_modeling_source()` method for specifying custom file sources
  - Preserved `trust_remote_code` setting during model loading operations

### **Configuration Serialization**
- **Enhanced config serialization** to handle complex Python objects:
  - Added support for callable functions (converted to `_target_` references)
  - Improved handling of `torch.dtype` objects (converted to string representation)
  - Better support for Enum types and other torch objects
  - Graceful handling of non-JSON serializable objects

### **Model Bridge Robustness**
- **Fixed module config handling** for non-MegatronModule components (e.g., `torch.nn.Conv1d`)
- Added automatic config inheritance from parent modules when needed

### **NemotronH Provider Enhancements**
- **Custom mamba stack specification** with proper activation function handling
- **Improved activation function management** with validation and fallback mechanisms
- Better integration with Megatron Core's MLP layer specifications

### **Added comprehensive unit tests** for custom modeling file preservation:
- Test coverage for basic file copying functionality
- Error handling tests for permission issues and missing files
- Integration tests for `save_artifacts()` method with different `trust_remote_code` settings

## 📋 **Files Modified**

- `src/megatron/bridge/models/config.py` - Enhanced config serialization
[...]

PR #917, opened 2025-10-10
Copy custom modeling logic

maanug-nv:
Splitting these changes from #655 into a separate PR.

Adds a helper to copy custom modeling files like `modeling_nemotron_h.py`, which is usually found in the HF repo when the model isn't in `transformers` library. This copying is done when exporting back to HF format.

This is necessary for exporting. Given the workflow, 1) import from HF, 2) finetune, 3) export back to HF, the exported checkpoint should be valid, even if `trust_remote_code=True` was set when importing. With these changes, the exported HF checkpoint can be used right away (eg to instantiate a model using `transformers` and generate text), without the user needing to copy files.

### DIFF
--- a/src/megatron/bridge/models/hf_pretrained/base.py
+++ b/src/megatron/bridge/models/hf_pretrained/base.py
@@ -55,15 +55,14 @@ class PreTrainedBase(ABC):
         self._state_dict_accessor: Optional[StateDict] = None
         self.init_kwargs = kwargs
 
-        # Common custom modeling file patterns
-        self.custom_file_patterns = [
-            "modeling_*.py",
-            "configuration_*.py",
-            "tokenization_*.py",
-            "processing_*.py",
-            "feature_extraction_*.py",
-            "*.py",  # fallback for anything not captured above
-        ]
+        # File patterns used to copy files used for custom modeling, e.g.
+        # modeling_*.py, configuration_*.py, tokenization_*.py,
+        # processing_*.py, feature_extraction_*.py. HF repos
+        # may have additional python files that are imported in the
+        # aforementioned, so we just copy all python files
+        # instead of above specific prefixes.
+        # Currently, we can capture all json files via ARTIFACTS.
+        self.custom_file_patterns = ["*.py"]
 
     def get_artifacts(self) -> Dict[str, str]:
         """Get the artifacts dictionary mapping artifact names to their attribute names."""


src/megatron/bridge/models/hf_pretrained/base.py
```diff
@@ -53,15 +55,90 @@ def _load_model(self) -> "PreTrainedModel":
...
+        target_path = Path(target_path)
+        
+        # Common custom modeling file patterns
+        custom_file_patterns = [
+            "modeling_*.py",
+            "configuration_*.py", 
+            "tokenization_*.py",
+            "processing_*.py",
+            "feature_extraction_*.py"
+        ]
```
cuichenx: Should we expose this list to the individual model bridges? Each models will have different files to copy over

cuichenx: or, is there any harm to just copy over all .py files?

maanug-nv: i made it an attribute for now, so it can overriden if needed

cuichenx: this is the `PreTrainedBase` class, do you have an example of how to override this in the model bridge? (maybe for nemotron h?)

maanug-nv: not sure how to do that, i don't think MegatronModelBridge instances have access to the auto bridge. afaiu, auto bridge just calls/attaches appropriate methods of model bridge. So I guess user would have to set this list manually after creating an auto bridge.

i'll update this list of patterns to include all python files, if there's functional/security concerns that come up, we can address it at that time.

src/megatron/bridge/models/hf_pretrained/base.py
```diff
@@ -54,14 +55,92 @@ def __init__(self, **kwargs):
...
         self.init_kwargs = kwargs
 
+        # Common custom modeling file patterns
+        self.custom_file_patterns = [
+            "modeling_*.py",
+            "configuration_*.py",
+            "tokenization_*.py",
+            "processing_*.py",
+            "feature_extraction_*.py",
+            "*.py",  # fallback for anything not captured above
```
cuichenx: is there any reason we still need to keep the top five and not just "*.py"?

maanug-nv: ill move those to a comment, just to document intent

PR #918, opened 2025-10-10
Add Param Mappings for NemotronH 

maanug-nv:
Splitting these changes from #655 into a separate PR.

Adds special param mapping types for MambaMixer in projection and conv1d parameters.

### DIFF
--- a/src/megatron/bridge/models/conversion/param_mapping.py
+++ b/src/megatron/bridge/models/conversion/param_mapping.py
@@ -1395,8 +1395,14 @@ class MambaInProjMapping(MegatronParamMapping[Dict[str, torch.Tensor]]):
         megatron_module: Optional[nn.Module],
     ) -> Dict[str, torch.Tensor]:
         """Gather Mamba in_proj shards and merge into single HF tensor."""
-        if megatron_weights is not None:
-            megatron_weights = self.maybe_dequantize(megatron_weights)
+        # Handle cross-PP broadcast
+        megatron_weights = self.broadcast_from_pp_rank(megatron_weights, cache_key=str(self.hf_param))
+
+        if megatron_weights is None:
+            return {}
+
+        # Dequantize if needed
+        megatron_weights = self.maybe_dequantize(megatron_weights)
 
         # Broadcast config to all PP ranks for collective communication
         if megatron_module is None:
@@ -1429,7 +1435,6 @@ class MambaInProjMapping(MegatronParamMapping[Dict[str, torch.Tensor]]):
         # Gather each component across TP ranks
         full_weights = []
         for component in local_components:
-            component = self.broadcast_from_pp_rank(component)
             if self.tp_size == 1:
                 full_weight = component
             else:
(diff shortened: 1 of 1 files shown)


src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -830,6 +830,10 @@ def build_conversion_tasks(
                         continue
 
                 local_module, local_weights = get_module_and_param_from_name(megatron_model, local_name, vp_stage)
+                if local_module is not None and not hasattr(local_module, "config"):
+                    # If module is not a MegatronModule (e.g. torch.nn.Conv1d or a module list) we need
+                    # to get the config from the model
+                    setattr(local_module, "config", model_config)
```
maanug-nv, line 836: This technically works, though I don't like arbitrarily using `setattr()`. Can we make the model config an attribute of the conversion task or the param mapping base class?
@yaoyu-33 wdyt?

yaoyu-33: @maanug-nv , we can make config into conversion task. But maybe do in another pr and leave this as it is in this PR.

src/megatron/bridge/models/conversion/param_mapping.py
```diff
@@ -1339,6 +1339,206 @@ def resolve(self, captures: Tuple[str, ...]) -> "MegatronParamMapping":
...
+        n_heads_local = config.mamba_num_heads // self.tp_size
+
+        # Extract local components
+        z_shard_idx = torch.arange(d_inner_local)
+        x_shard_idx = torch.arange(d_inner_local) + d_inner_local
+        B_shard_idx = torch.arange(d_tot_ssm_local) + 2 * d_inner_local
+        C_shard_idx = torch.arange(d_tot_ssm_local) + 2 * d_inner_local + d_tot_ssm_local
+        dt_shard_idx = torch.arange(n_heads_local) + 2 * (d_inner_local + d_tot_ssm_local)
+
+        local_components = [
```
yfw, line 1427: For the PP case, `megatron_weights` may be None here. I think we need to call `self.broadcast_from_pp_rank` on the `megatron_weights` first and remove it from the for loop below.

src/megatron/bridge/models/conversion/param_mapping.py
```diff
@@ -1339,6 +1339,206 @@ def resolve(self, captures: Tuple[str, ...]) -> "MegatronParamMapping":
...
+
+        d_inner_local = (config.mamba_num_heads * config.mamba_head_dim) // self.tp_size
+        d_tot_ssm_local = (config.mamba_state_dim * config.mamba_num_groups) // self.tp_size
+
+        # Extract local components
+        x_shard_idx = torch.arange(d_inner_local)
+        B_shard_idx = torch.arange(d_tot_ssm_local) + d_inner_local
+        C_shard_idx = torch.arange(d_tot_ssm_local) + d_inner_local + d_tot_ssm_local
+
+        local_components = [
```
yfw, line 1532: Similar comment as above, `megatron_weights` may be None here in the PP case

src/megatron/bridge/models/conversion/param_mapping.py
maanug-nv: Yeah makes sense, thanks for catching that. I'm thinking start of this method should look the same as this: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/maanug/nemotronh-mappings/src/megatron/bridge/models/conversion/param_mapping.py#L822-L828

maanug-nv: Updated, lmk if there's still any issues

yfw: looks good, thanks for fixing!

PR #938, opened 2025-10-13
Clean up fusion checks

### LINKED ISSUE #437: Clean up fusion checks
TE is a hard requirement for this repo. As such, some of the fusion checks [here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/utils/fusions.py) are redundant. Some of these checks, however, are for fusions that come from apex. Based on my investigation in MCore:

`can_enable_apply_rope_fusion()` will always be true, given our TE and MCore version requirements. can be removed
for `can_enable_gradient_accumulation_fusion()`, I think MCore still uses an apex module for this, so it should probably be kept.
`can_enable_bias_dropout_fusion()` can be removed, b/c MCore version requirement
`can_enable_masked_softmax_fusion()` , this is an apex fusion that is only used with MCore-local specs (MCore local attention impl). probably can be removed since MBridge will always default to TE specs
`validate_rope_fusion_compatibility()` is useful, but could be called in post init instead of provide().

ananthsub:
fix #437

### DIFF
--- a/src/megatron/bridge/models/gemma/gemma3_provider.py
+++ b/src/megatron/bridge/models/gemma/gemma3_provider.py
@@ -75,6 +75,7 @@ class Gemma3ModelProvider(GPTModelProvider):
     rope_scaling_factor: float = 1.0
     # Disable cuDNN attention since TE 1.8 does not support head dim > 128
     attention_backend: AttnBackend = AttnBackend.flash
+    softmax_scale: float = 1.0 / math.sqrt(256)
 
     # mlp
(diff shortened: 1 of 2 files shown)


maanug-nv:
can we also cleanup `validate_rope_fusion_compatibility()`, has some similar checks about TE version. We could instead maybe add [this](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/arguments.py#L1004) check there, not sure if we already have it.

PR #939, opened 2025-10-14
Support Qwen3next

yuzhongw-nvidia:


src/megatron/bridge/models/conversion/param_mapping.py
```diff
@@ -16,6 +16,7 @@
 import re
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Generic, List, Optional, Tuple, TypeVar, Union
+import zlib
```
yaoyu-33: seems not used?

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -547,6 +537,7 @@ def import_ckpt(
         cls,
         hf_model_id: str | Path,
         megatron_path: str | Path,
+        use_cpu_initialization: bool = False,
```
yaoyu-33: seems not used?

src/megatron/bridge/models/model_provider.py
```diff
@@ -109,6 +109,7 @@ def provide_distributed_model(
         wrap_with_ddp: bool = True,
         data_parallel_random_init: bool = True,
         use_cpu_initialization: None | bool = False,
+        return_cpu_model: bool = False,
```
yaoyu-33: can you explain why need return cpu model other than cpu init?

src/megatron/bridge/training/checkpointing.py
```diff
@@ -2048,7 +1876,7 @@ def _build_sharded_state_dict_metadata(
         else:
             metadata["distrib_optim_sharding_type"] = "dp_zero_gather_scatter"
     metadata["chained_optim_avoid_prefix"] = True
-    metadata["singleton_local_shards"] = False
+    metadata["singleton_local_shards"] = True
```
yaoyu-33, line 2051: why is this change needed?

src/megatron/bridge/models/qwen/qwen3_next_bridge.py
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        # Main decoder params
+        param_mappings = {
+            # Embedding and output
+            "embedding.word_embeddings.weight": "model.embed_tokens.weight",
+            "output_layer.weight": "lm_head.weight",
+            "decoder.final_layernorm.weight": "model.norm.weight",
+        }
+        for megatron_param, hf_param in layer_mappings.items():
```
yaoyu-33: can we just write down the mapping directly, not use layer_mappings. It's a bit inconsistent with other model.
We feel it is clearer just write the whole name together, though we need to duplicate the prefix.

yaoyu-33:
Hi, thanks for contribution. unit tests/ functional tests need to be added. also it will be nice if you can attach your validation results.
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/docs/adding-new-models.md#6-validate-with-examples

src/megatron/bridge/models/qwen/qwen3_next_bridge.py
yuzhongw-nvidia: Yes. I will do the change later.

src/megatron/bridge/models/model_provider.py
yuzhongw-nvidia: Well, I checked `model_provider.py` again, and found it was fixed. The condition of moving the model to the device was wrong before ([code](https://github.com/yuzhongw-nvidia/Megatron-Bridge/blob/a28e16104e84820bb7388aa670211f5eed5fe8e0/src/megatron/bridge/models/model_provider.py#L534)). Anyway, let me remove this argument now.

src/megatron/bridge/training/checkpointing.py
yuzhongw-nvidia: According to this line of [code](https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/blob/dev/megatron/training/checkpointing.py#L391) in megatron-lm, `singleton_local_shards` is set to `False` if and only if `args.dist_ckpt_save_pre_mcore_014 or force_pre_mcore_014`.

Honestly, I don't fully understand this flag, but in my actual runs, I've found that setting `metadata["singleton_local_shards"] = False` causes the tensor parallel sharding to be incorrect in MCore.

src/megatron/bridge/models/qwen/qwen3_next_bridge.py
yuzhongw-nvidia: Thanks. Changes done. Does it look good now?

src/megatron/bridge/models/conversion/auto_bridge.py
yuzhongw-nvidia: Thanks. Removed.

src/megatron/bridge/models/conversion/param_mapping.py
yuzhongw-nvidia: Thanks, removed.

yaoyu-33:
moved the PR to https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1068

PR #950, opened 2025-10-14
Check pretrained checkpoint actually exists

cuichenx:
When PEFT is disabled and checkpoint.pretrained_checkpoint is provided, add a check to confirm the directory actually contains a checkpoint. If it does not, a clear ValueError is raised immediately instead of silently starting from random weights.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for pretrained checkpoint paths. Invalid paths are now detected early with a clear error message, improving error handling and user experience.

### DIFF
--- a/tests/unit_tests/training/test_checkpointing.py
+++ b/tests/unit_tests/training/test_checkpointing.py
@@ -2160,7 +2160,7 @@ class TestCheckpointIterationResolution:
         assert "Please set checkpoint.load to the base checkpoint directory" in str(exc_info.value)
 
     def test_ckpt_step_with_only_pretrained_raises(self):
-        """ckpt_step with only pretrained_checkpoint should raise (ckpt_step applies to load, not pretrained)."""
+        """ckpt_step with only pretrained_checkpoint should raise because pretrained path does not exist."""
         from megatron.bridge.training.config import CheckpointConfig
 
         config = CheckpointConfig(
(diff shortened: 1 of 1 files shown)


src/megatron/bridge/training/setup.py
```diff
@@ -196,6 +196,11 @@ def setup(
             # This is switched off here in order to load these states from the checkpoint
             cfg.checkpoint.finetune = False
     else:
+        # Validate the supplied pretrained checkpoint path when PEFT is not used.
```
ananthsub: we have this option in the checkpoint config: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/edfb63d23b03f2a1c5404d903185a63f18f852ff/src/megatron/bridge/training/config.py#L624

by default we don't exit, matching MLM's defaults. setting this would cover if either `pretrained` or `load` is not specified correctly

cuichenx: I just tested setting `exit_on_missing_checkpoint` to true and training still started from scratch..

Besides, shouldn't we default this to True? `load` can be a different story, but if a user specifies the `pretrained_checkpoint` arg, they clearly want to finetune, not pretrain, so we should error out on a misspelled checkpoint dir by default

ananthsub: that makes sense to me. how about adding this validation here so it happens sooner? https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/3b9db5ffaec1e622702d00fa6aa5af1b7b974605/src/megatron/bridge/training/config.py#L698-L705

then for PEFT, we can simply assert that pretrained checkpoint is set and avoid checking if the checkpoint exists again

src/megatron/bridge/training/config.py
```diff
@@ -897,6 +897,13 @@ class CheckpointConfig:
 
     def finalize(self) -> None:
         """Post-initialization checks for checkpoint config."""
+        if self.pretrained_checkpoint is not None:
+            from megatron.bridge.training.utils.checkpoint_utils import file_exists
+
+            assert file_exists(self.pretrained_checkpoint), (
```
ananthsub, line 903: isn't pretrained_checkpoint a directory? how would this check work?

cuichenx: os.path.exists() returns true for both file and directory

yaoyu-33 (approved):

Issue #934, opened 2025-10-13
Megatron "cuda_graph_impl" api change


yaoyu-33:
Some configs need to change after 
https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/merge_requests/3823/diffs (internal)

Issue #503, opened 2025-08-27
Full Cuda-graph support for Mbridge


gautham-kollu:

erhoo82:
@gautham-kollu What is the current status of this?

erhoo82:
Verify/Fix full-iteration cuda graph with LM3 8B and 70B with and w/o PP.
Verify/Fix full-iteration cuda graph with DSv3 tokendrop

PR #951, opened 2025-10-14
Add support for full-iteration cuda graph 

gautham-kollu:
Fixes https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/503
Fixes https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/934

Description: 
 Change to support full_iteration cudagraph.

Set the following increase the scope after the enabling cuda graphs.
```
 model.cuda_graph_scope ="full_iteration"
```

Testing:
   - Tested Llama3-8b on single node Hopper using GBS=8

### DIFF
--- a/tests/functional_tests/recipes/test_llama_recipes_pretrain_1b.py
+++ b/tests/functional_tests/recipes/test_llama_recipes_pretrain_1b.py
@@ -29,7 +29,6 @@ LLAMA_PRETRAIN_RECIPES = [
     # (config_func, name, parallelism_overrides, model_overrides)
     (llama32_1b_config, "llama32_1b", {}, {"num_layers": 2}),
     (llama32_3b_config, "llama32_3b", {}, {"num_layers": 2}),
-    (llama32_3b_config, "llama32_8b", {}, {"num_layers": 2, "cuda_graph_impl": "local", "cuda_graph_scope": "full_iteration", "check_for_nan_in_grad" : False, "use_te_rng_tracker": True}),
 ]
 
 
--- a/tests/functional_tests/recipes/utils.py
+++ b/tests/functional_tests/recipes/utils.py
@@ -110,9 +110,6 @@ def run_pretrain_recipe_test(
             for attribute_name, attribute_value in model_overrides.items():
                 setattr(config.model, attribute_name, attribute_value)
 
-        config.rng.te_rng_tracker = True
-        config.rerun_state_machine.check_for_nan_in_loss = False
-        config.ddp.check_for_nan_in_grad = False
         pretrain(config, forward_step)
 
         # Basic verification that training completed successfully


src/megatron/bridge/training/train.py
```diff
@@ -523,7 +524,13 @@ def train_step(
         )
 
         # Forward pass.
-        forward_backward_func = get_forward_backward_func()
+        if cfg.model.enable_cuda_graph and cfg.model.cuda_graph_scope == "full_iteration":
+            print_rank_0("Using full cuda graph")
+            forward_backward_func = FullCudaGraphWrapper(get_forward_backward_func())
```
ananthsub: do we need to expose the warmup steps through the config? https://github.com/NVIDIA/Megatron-LM/blob/c9746e49784c47573bccf17cd98ba714ab58b9f3/megatron/core/full_cuda_graph.py#L101C47-L101C70

ananthsub: we also need this for evaluation https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/2eea3d1d50cbd4f6f14d6ea72eac3297d71a7e76/src/megatron/bridge/training/eval.py#L93

src/megatron/bridge/training/eval.py
```diff
@@ -90,7 +91,13 @@ def evaluate(
             if verbose:
                 print_rank_0(f"Evaluating iter {iteration}/{state.cfg.train.eval_iters}")
 
+        if state.cfg.model.enable_cuda_graph and state.cfg.model.cuda_graph_scope == "full_iteration":
+            forward_backward_func = FullCudaGraphWrapper(
+                get_forward_backward_func(), cuda_graph_warmup_steps=state.cfg.model.cuda_graph_warmup_steps
+            )
+        else:
             forward_backward_func = get_forward_backward_func()
+
```
ananthsub: nit: can turn this into a helper function since its copied across train and eval

src/megatron/bridge/models/gpt_full_te_layer_autocast_spec.py
```diff
@@ -225,7 +225,7 @@ def __init__(self, config, layer_number=1, hidden_dropout=None, **kwargs):
             transformer_layer_args["ub_atomic_gemm_rs"] = config.tp_comm_atomic_rs
         self.transformer_layer = AutocastTransformerLayer(**transformer_layer_args)
 
-        if self.config.enable_cuda_graph and self.training:
+        if self.config.enable_cuda_graph and self.training and self.config.cuda_graph_scope != "full_iteration":
```
ananthsub: we also need to assert that `check_for_nan_in_loss_and_grad` isnt set for full_iteration CUDA graph support. this can go into configcontainer.validate

src/megatron/bridge/training/config.py
```diff
@@ -1253,7 +1226,9 @@ def validate(self) -> None:
         # Validate DeepEP is supported for the current GPU architecture
         validate_deepep(self.model)
 
-        self._sync_and_validate_external_cuda_graph()
+        # Sync config. If TE RNG tracker is set in either ways, set them in both places.
+        if self.rng.te_rng_tracker or self.model.use_te_rng_tracker:
+            self.model.use_te_rng_tracker = self.rng.te_rng_tracker = True
```
ananthsub: as a sanity check, in case there are any dependent attributes relying on `use_te_rng_tracker` in the transformer config's `__post_init__`, we should set this *before* L1120 above

src/megatron/bridge/training/train.py
```diff
@@ -319,7 +320,7 @@ def train(
                     model_config.param_sync_func = param_sync_func
                     pre_hook_enabled = True
                     # Set the manual hooks when CUDA Graphs are used.
-                    if model_config.external_cuda_graph:
+                    if model_config.cuda_graph_impl == "transformer_engine":
```
ananthsub, line 333: there also needs to be handling for the skipped iterations in L280 in case there's overlap with the graph capture steps

ananthsub (dismissed):

src/megatron/bridge/models/model_provider.py
```diff
@@ -567,14 +567,15 @@ def get_model(
...
     if wrap_with_ddp:
-        model = _ddp_wrap(
-            model,
-            data_parallel_random_init,
-            ddp_config,
-            overlap_param_gather_with_optimizer_step,
-            use_megatron_fsdp=use_megatron_fsdp,
-            use_torch_fsdp2=use_torch_fsdp2,
-        )
+        with torch.cuda.stream(torch.cuda.Stream()):
```
ananthsub: we have the side stream allocation here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8e440bb767f16a2b4c52628d673921ceee4b8fd9/src/megatron/bridge/models/model_provider.py#L680-L691

so we shouldn't have it again applied over all of `_ddp_wrap`

malay-nagda (dismissed):

tests/functional_tests/recipes/test_llama_recipes_pretrain_1b.py
```diff
@@ -29,6 +29,7 @@
     # (config_func, name, parallelism_overrides, model_overrides)
     (llama32_1b_config, "llama32_1b", {}, {"num_layers": 2}),
     (llama32_3b_config, "llama32_3b", {}, {"num_layers": 2}),
+    (llama32_3b_config, "llama32_8b", {}, {"num_layers": 2, "cuda_graph_impl": "local", "cuda_graph_scope": "full_iteration", "check_for_nan_in_grad" : False, "use_te_rng_tracker": True}),
```
ananthsub: IMO it's better to have a separate functional test for training for full cuda graphs than piggybacking on the recipe tests here

gautham-kollu: reverted them. will handle that in another pr

Issue #721, opened 2025-09-23
Stop using temporary nemo-rl-use branch in Megatron-Bridge; standardize on main for nemo-rl


yaoyu-33:
Converge nemo-rl development on the Megatron-Bridge main branch and deprecate the temporary nemo-rl-use workflow/branch. This avoids constant submodule branch switching, reduces divergence, and makes ownership/CI clearer.

terrykong:
Shall we use this issue to also track the changes necessary from MLM?

yaoyu-33:
RL PR at : https://github.com/NVIDIA-NeMo/RL/pull/1358/files

PR #955, opened 2025-10-14
Sync RL custom branch into main branch 

yaoyu-33:
As mentioned in https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/721, NeMo RL can start to use tot branch after this PR.
This PR opens few APIs for RL being able to customize few hooks.

### DIFF
--- a/src/megatron/bridge/models/gpt_oss/gpt_oss_provider.py
+++ b/src/megatron/bridge/models/gpt_oss/gpt_oss_provider.py
@@ -27,6 +27,7 @@ from megatron.bridge.models.gpt_provider import GPTModelProvider
 try:
     from megatron.core.fusions.fused_bias_geglu import quick_gelu
 except ImportError:
+    # TODO(yuya): remove backup path once versioning issue solved
     # Fallback if fused_bias_geglu is not available
     quick_gelu = torch.nn.functional.gelu
 


src/megatron/bridge/models/model_provider.py
```diff
@@ -551,8 +560,8 @@ def get_model(
         for model_module in model:
             model_module.cuda(torch.cuda.current_device())
 
-    if model_config.fp16 or model_config.bf16:
-        model = [Float16Module(model_config, model_module) for model_module in model]
+    if (model_config.fp16 or model_config.bf16) and mixed_precision_wrapper is not None:
```
ananthsub, line 563: is there a risk that fp16/bf16 is enabled and mixed_precision_wrapper is not specified causing this to silently fall through? should this assert instead that if the mixed precision flags are specified that the wrapper module must be set?

src/megatron/bridge/models/gpt_oss/gpt_oss_provider.py
```diff
@@ -17,14 +17,20 @@
...
 from megatron.core.utils import is_te_min_version
 
 from megatron.bridge.models.gpt_provider import GPTModelProvider
 
 
+try:
+    from megatron.core.fusions.fused_bias_geglu import quick_gelu
+except ImportError:
+    # Fallback if fused_bias_geglu is not available
+    quick_gelu = torch.nn.functional.gelu
```
ananthsub, line 32: the import fallbacks are hard to follow since they don't line up with the MLM submodule commit pinned in the repo now. i understand its done to manage RL's own MLM dependency but could you leave a comment/TODO to remove this once RL's separate MLM pinned commit is resolved?

ananthsub (approved):

src/megatron/bridge/models/model_provider.py
yaoyu-33: RL has a case they have bf16 but they defer the wrapping later. That's actually one their use case. hopefully no one will abuse this...

src/megatron/bridge/models/gpt_oss/gpt_oss_provider.py
yaoyu-33: cool

PR #963, opened 2025-10-15
added config overrides

salberdi-nvidia:
added parallelism config overrides

### DIFF
--- a/scripts/performance/argument_parser.py
+++ b/scripts/performance/argument_parser.py
@@ -146,6 +146,71 @@ def parse_cli_args():
         action="store_true",
     )
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


scripts/performance/argument_parser.py
```diff
@@ -145,6 +145,71 @@ def parse_cli_args():
...
+        required=False,
+        default=None,
+    )
+    parser.add_argument(
+        "-gb",
+        "--global_batch_size",
+        type=int,
+        required=False,
+        default=None,
+    )
```
nv-mollys, line 212: Where are these arguments being applied? Do they need to be added to common.py as well?

PR #970, opened 2025-10-16
ci: Enforce test time SLA

ko3n1g:
* Enforce 500ms for each single unit test (and comments out a few that exceed this threshold)
* Moves the data tests to functional
* Max timeout of 15mins for all unit tests
* Max timeout of 30mins for L2 tests
* Splits out llama recipe tests to keep the 30mins

In another PR, yesterday, we enforced job concurrency to 5 (for both MBridge and Automodel). This allows interleaving jobs across repos into the shared infra, thereby harmonizing waiting times and leading to a fair allocation.


tests/unit_tests/peft/test_canonical_lora.py
```diff
@@ -779,6 +779,7 @@ def test_canonical_lora_with_gpt_model(self):
         efficiency_ratio = trainable_params / total_params
         assert efficiency_ratio < 0.3, f"CanonicalLoRA should be parameter efficient, got ratio: {efficiency_ratio}"
 
+    @pytest.mark.pleasefixme  # This test is too slow for unit tests (>Xs)
```
ananthsub, line 782: can we restrict this to run on gpus only?

tests/unit_tests/peft/test_dora.py
```diff
@@ -693,6 +693,7 @@ def test_dora_transform_idempotent_megatron_model(self):
             assert first_name == second_name
             assert second_module is first_module, f"DoRA module {first_name} should be identical object"
 
+    @pytest.mark.pleasefixme  # This test is too slow for unit tests (>0.5s)
```
ananthsub, line 696: same here

tests/unit_tests/peft/test_lora.py
```diff
@@ -679,6 +679,7 @@ def test_lora_with_gpt_model(self):
         efficiency_ratio = trainable_params / total_params
         assert efficiency_ratio < 0.3, f"LoRA should be parameter efficient, got ratio: {efficiency_ratio}"
 
+    @pytest.mark.pleasefixme  # This test is too slow for unit tests (>Xs)
```
ananthsub: same here

tests/unit_tests/training/utils/test_checkpoint_utils.py
```diff
@@ -435,6 +435,7 @@ def read_config_worker():
         assert len(results) == 10
         assert all(result == config_data for result in results)
 
+    @pytest.mark.pleasefixme  # This test is too slow for unit tests (>Xs)
```
ananthsub: this test can probably be deleted. it's not a useful unit test since it's intentionally slow

tests/unit_tests/peft/test_canonical_lora.py
ko3n1g: I don't think we're using a GPU marker at this point. All tests run on a GPU node

tests/unit_tests/training/utils/test_checkpoint_utils.py
ko3n1g: Do you see value in moving this into a functional test bucket?

tests/functional_tests/training/test_checkpoint_utils.py
```diff
@@ -0,0 +1,36 @@
+import time
+from unittest.mock import patch
+
+import yaml
+
+from megatron.bridge.training.utils.checkpoint_utils import read_run_config
+
+
+def test_performance_large_config_file(self, tmp_path):
```
ananthsub: imo this test is not helpful - the run config will be loaded during the functional tests when resuming the checkpoint. we can remove this entirely

ananthsub (approved):

PR #971, opened 2025-10-16
dataclass cfgs - perf

malay-nagda:
This adds dataclass configs for LLMs per GPU per precision for pre-training optimized for performance.

Why move to dataclass?
        - This makes it easier to use "optimal" config directly as a Python object
        - We don't have to deal with file reading and post processing from csv,yaml,etc

User still has the option to override parallelism configs, batch size, cuda graphs, etc.

### DIFF
--- a/scripts/performance/configs/deepseek/workload_base_configs.py
+++ b/scripts/performance/configs/deepseek/workload_base_configs.py
@@ -26,7 +26,9 @@ DEEPSEEK_V3_GB300_256GPUS_BF16_PARALLEL_CONFIG = WorkloadBaseConfig(
     expert_tensor_parallel_size=1,
     global_batch_size=2048,
     micro_batch_size=1,
-    recompute_modules=["mla_up_proj"],
+    cuda_graph_impl="transformer_engine",
+    cuda_graph_scope="attn",
+    recompute_modules=["mlp", "moe_act"],
 )
 
 
@@ -39,7 +41,9 @@ DEEPSEEK_V3_GB300_256GPUS_FP8_CS_PARALLEL_CONFIG = WorkloadBaseConfig(
     expert_tensor_parallel_size=1,
     global_batch_size=2048,
     micro_batch_size=1,
-    recompute_modules=["mla_up_proj"],
+    cuda_graph_impl="transformer_engine",
+    cuda_graph_scope="attn",
+    recompute_modules=["mlp", "moe_act"],
 )
 
 


scripts/performance/run_script.py
```diff
@@ -53,6 +61,9 @@ def main():
 
     if args.model_name == "llama3" and args.model_size == "8b":
         recipe = llama3_8b_pretrain_config(mock=True, precision_config=precision_config)
+        if args.gpu.lower() == "h100":
+            cfg_str = f"llama3_8b_{args.gpu.lower()}_{args.compute_dtype.lower()}_config"
+            recipe = globals()[cfg_str](args.fp8_recipe)
```
ko3n1g: How about making this more explicit, like

```suggestion
        if args.gpu.lower() == "h100":
            if args.compute_dtype.lower() == "bf16":
              recipe = llama3_8b_h100_bf16
            elif ...:
            else:
               raise ValueError("Dtype %s not supported", args.compute_dtype.lower())
        elif ...        
```

malay-nagda: Over time, `run_script` will explode with if-else block considering the number of current models we track and will add more. 
the current way I think keeps it a bit more concise.

ko3n1g: I see your point. No worries, let’s keep it as it is

ananthsub: if the branching is an issue, you could do what the mixed precision configs do and maintain a registry:

the perf scripts can have a way of generating the entry key based on the model/pretrain or finetune/gpu type/precision type. it would remove the need to look up globals()

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/3b9db5ffaec1e622702d00fa6aa5af1b7b974605/src/megatron/bridge/training/mixed_precision.py#L147-L168


https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/3b9db5ffaec1e622702d00fa6aa5af1b7b974605/src/megatron/bridge/training/mixed_precision.py#L406-L422

scripts/performance/configs/qwen3/qwen3_235b_a22b_llm_pretrain.py
```diff
@@ -0,0 +1,313 @@
...
+    cfg.train.micro_batch_size = 1 if mbs is None else mbs
+    cfg.model.seq_length = 4096
+    cfg.dataset.sequence_length = 4096
+
+    cfg.mixed_precision.grad_reduce_in_fp32 = False
+    cfg.ddp.grad_reduce_in_fp32 = False
+
+    set_qwen3_235b_a22b_specific_overrides(cfg)
+
+    cuda_graph_impl = "local" if kwargs.get("cuda_graph_impl") is None else kwargs.get("cuda_graph_impl")
```
ko3n1g:
```suggestion
    cuda_graph_impl =  kwargs.get("cuda_graph_impl", "local")
```

malay-nagda: This does not work.
it will set to "local" only if `cuda_graph_impl ` has not set attribute value. But, we do set the default to `None`, in which case get will return `None` and not "local"

scripts/performance/run_script.py
malay-nagda: This looks good. I just use importlib instead of globals

scripts/performance/argument_parser.py
```diff
@@ -229,6 +265,115 @@ def list_of_strings(arg):
...
+    )
+    parser.add_argument(
+        "-gb",
+        "--global_batch_size",
+        type=int,
+        required=False,
+        default=None,
+    )
+    parser.add_argument(
+        "--moe_a2a",
```
erhoo82: change to `moe_a2a_overlap`?

scripts/performance/argument_parser.py
erhoo82, line 1: Can we also add wandb interface?
`project name`, `job name`?

scripts/performance/configs/llama3/llama3_70b_llm_pretrain.py
```diff
@@ -0,0 +1,205 @@
...
+    cfg.dataset.sequence_length = 8192
+
+    cfg.comm_overlap.tp_comm_overlap_cfg = userbuffers_bf16_h100_h8192_tp4_mbs1_seqlen8192
+
+    return cfg
+
+
+def llama3_70b_h100_64gpus_fp8_config(fp8_recipe: str = "cs") -> ConfigContainer:
+    """H100, 64xGPU, FP8 preset with selectable recipe (ds/cs/mx/ss)."""
+    cfg = llama3_70b_pretrain_config(mock=True, precision_config=get_precision_config("fp8", fp8_recipe))
```
gautham-kollu: In the .yaml configs we have _common_ . Looks like we don't have something similar.  I understand it is not necessary but isn't it helpful ?

scripts/performance/configs/deepseek/deepseek_llm_pretrain.py
```diff
@@ -0,0 +1,205 @@
...
+    set_deepseek_v3_common_configs(cfg)
+
+    apply_parallelism_and_batch_config(cfg, parallelism_and_batch_cfg)
+
+    cfg.model.recompute_modules = ["mla_up_proj"]
+    cfg.comm_overlap.overlap_grad_reduce = True
+
+    cfg.model = apply_moe_token_drop(cfg.model)
+
+    cfg.dataset.num_workers = 0
```
erhoo82, line 105: Add a comment on top `Currently, num_workers=0 and pin_memory=False gives greater performance on GB200. We are under debugging`

scripts/performance/configs/deepseek/deepseek_llm_pretrain.py
```diff
@@ -0,0 +1,205 @@
...
+
+    apply_parallelism_and_batch_config(cfg, parallelism_and_batch_cfg)
+
+    cfg.model.recompute_modules = ["mla_up_proj", "mlp"]
+
+    cfg.model.moe_router_force_load_balancing = True
+
+    set_moe_a2a_1f1b_overrides(cfg)
+
+    cfg.comm_overlap.overlap_grad_reduce = False
```
erhoo82, line 163: Why disable this?

scripts/performance/configs/deepseek/deepseek_llm_pretrain.py
```diff
@@ -0,0 +1,205 @@
...
+        enable_deepep=True,
+        layout="Et|(tt|)*30mL",
+    )
+    set_deepseek_v3_common_configs(cfg)
+
+    apply_parallelism_and_batch_config(cfg, parallelism_cfg.DEEPSEEK_V3_H100_1024GPUS_BF16_PARALLEL_CONFIG)
+
+    cfg.model.recompute_modules = ["mla_up_proj", "mlp"]
+
+    cfg.model.moe_router_force_load_balancing = True
```
erhoo82: Why this is set only for h100 configs?

malay-nagda: Because we added this for dropless which is default for only H100. 

But, if user overrides from cmdline to use dropless, in this case it will be set for blackwell too.

malay-nagda: do we want to default to dropless for blackwell too from now? @erhoo82

scripts/performance/configs/deepseek/deepseek_llm_pretrain.py
malay-nagda: shared over Slack @erhoo82

scripts/performance/configs/llama3/llama3_70b_llm_pretrain.py
malay-nagda: @gautham-kollu added a common method that applies to all

- `llama3` models [here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/971/files#diff-81bc3a21ae6001eb67f2bcf98fbd9fc8f2852d65fc24462f3985f7b555ec93a8R41)
- models  [here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/971/files#diff-b45195eb6a822345e139507969c63b332d1ea7714ad2471aa045a01c77cb07a9R71)

this is what you mean, right?

sanandaraj5597 (dismissed):
LGTM, has all the changes we discussed internally.

scripts/performance/configs/deepseek/deepseek_llm_pretrain.py
```diff
@@ -0,0 +1,207 @@
...
+    set_deepseek_v3_common_configs(cfg)
+
+    set_parallelism_and_batch_configs(cfg, parallelism_and_batch_cfg)
+
+    cfg.model.recompute_modules = ["mla_up_proj"]
+    cfg.comm_overlap.overlap_grad_reduce = True
+
+    cfg.model = apply_moe_token_drop(cfg.model)
+
+    cfg.dataset.num_workers = 0
```
erhoo82, line 105: Please add the same comment above
`    # Setting num_workers and pin_memory to 0 and False respectively gives better performance.
    # we are debugging this and might change this in the future.`

scripts/performance/configs/deepseek/deepseek_llm_pretrain.py
```diff
@@ -0,0 +1,207 @@
...
+
+    cfg.comm_overlap.overlap_grad_reduce = False
+
+    return cfg
+
+
+def deepseek_v3_h100_1024gpus_fp8_config(fp8_recipe: str = "cs") -> ConfigContainer:
+    """H100, 1024xGPU, FP8 baseline config."""
+    parallelism_and_batch_cfg = parallelism_cfg.DEEPSEEK_V3_H100_1024GPUS_FP8_CS_PARALLEL_CONFIG
+    if fp8_recipe == "ss":
```
erhoo82: let's change `ss` to `sc`, which is short for sub-channel.
Also, `FP8_SS` to `FP8_SC`

scripts/performance/configs/deepseek/deepseek_llm_pretrain.py
```diff
@@ -0,0 +1,207 @@
...
+
+    cfg.model.recompute_modules = ["mla_up_proj", "mlp"]
+
+    cfg.model.moe_router_force_load_balancing = True
+
+    set_moe_a2a_overlap_overrides(cfg)
+
+    cfg.comm_overlap.overlap_grad_reduce = False
+
+    return cfg
```
erhoo82, line 165: Can we combine fp8 and bf16 like?
Otherwise, we would have 11 configs (=[H100, B200, GB200, GB300] x [bf16, fp8, (nvfp4)]) per model? 

```
def deepseek_v3_h100_1024gpus_config(precision: str = "bf16", fp8_recipe: str = "cs") -> ConfigContainer:
    """H100, 1024xGPU config."""
    if precision == "bf16":
        parallelism_and_batch_cfg = parallelism_cfg.DEEPSEEK_V3_H100_1024GPUS_BF16_PARALLEL_CONFIG
        precision_config = get_precision_config("bf16")
    else:
        parallelism_and_batch_cfg = parallelism_cfg.DEEPSEEK_V3_H100_1024GPUS_FP8_CS_PARALLEL_CONFIG
        if fp8_recipe == "ss":
            parallelism_and_batch_cfg = parallelism_cfg.DEEPSEEK_V3_H100_1024GPUS_FP8_SS_PARALLEL_CONFIG
        precision_config = get_precision_config("fp8", fp8_recipe)

    cfg = pretrain_config(
        mock=True,
        precision_config=precision_config,
        pipeline_parallelism=parallelism_and_batch_cfg.pipeline_model_parallel_size,
        virtual_pipeline_parallelism=parallelism_and_batch_cfg.virtual_pipeline_model_parallel_size,
        enable_deepep=True,
        layout="Et|(tt|)*30mL",
    )
    set_deepseek_v3_common_configs(cfg)

    set_parallelism_and_batch_configs(cfg, parallelism_cfg.DEEPSEEK_V3_H100_1024GPUS_BF16_PARALLEL_CONFIG)

    cfg.model.recompute_modules = ["mla_up_proj", "mlp"]

    cfg.model.moe_router_force_load_balancing = True

    set_moe_a2a_overlap_overrides(cfg)

    cfg.comm_overlap.overlap_grad_reduce = False

    return cfg
```

erhoo82:
BTW, wouldn't `PARALLEL_CONFIG` contain other knobs not shared among configs (e.g., `recompute_modules`)make it more concise?

So, basically, things shared among model and device are co-defined and set in the model file and everything else specfiic to precision and recipe are set in the `PARALLEL_CONFIG`?

scripts/performance/configs/deepseek/deepseek_llm_pretrain.py
malay-nagda: updated default to dropless for Blackwell

scripts/performance/configs/deepseek/deepseek_llm_pretrain.py
malay-nagda: added comment

scripts/performance/utils/utils.py
```diff
@@ -0,0 +1,105 @@
...
+    model_size: str,
+    gpu: str,
+    num_gpus: int,
+    compute_dtype: str,
+    fp8_recipe: Optional[str] = None,
+) -> Dict[str, int]:
+    """Get the parallelism defaults for a given model, size, GPU, number of GPUs, compute dtype, and FP8 recipe."""
+    parallelism_name = f"{model_name}_{model_size}_{gpu}_{num_gpus}gpus_{compute_dtype}"
+    if compute_dtype == "fp8":
+        fp8_recipe = "cs" if fp8_recipe == "ds" else fp8_recipe  # we don't maintain FP8-DS parallelism configs
```
erhoo82: Remove `ds` from fp8_recipe instead of internally changing here.

malay-nagda:
@ko3n1g @chtruong814 is this CI fine?
 
I get the following error in one test-
`Error response from daemon: No such container: nemo_container_19111796246`

erhoo82 (approved):

erhoo82 (approved):

PR #975, opened 2025-10-16
Qwen3 finetune recipes

ananthsub:
1. Import the pretrained checkpoints using AutoBridge: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/examples/conversion/convert_checkpoints.py

2. In the user train script, import qwen3 finetuning configs + specify paths to pretrained checkpoints from step 1 to begin finetuning

train script: https://gist.github.com/ananthsub/69139c3a8e30f57b4aa09eef422559cb

### DIFF
--- a/src/megatron/bridge/recipes/qwen/qwen3.py
+++ b/src/megatron/bridge/recipes/qwen/qwen3.py
@@ -555,7 +555,7 @@ def _qwen3_finetune_common(
 
     opt_cfg, scheduler_cfg = distributed_fused_adam_with_cosine_annealing(
         lr_warmup_iters=lr_warmup_iters,
-        lr_decay_iters=lr_decay_iters,  # None is fine
+        lr_decay_iters=lr_decay_iters,
         max_lr=finetune_lr,
         min_lr=min_lr,
         adam_beta2=0.98,
@@ -598,6 +598,7 @@ def _qwen3_finetune_common(
         checkpoint=CheckpointConfig(
             save_interval=save_interval,
             save=checkpoint_dir,
+            load=checkpoint_dir,
             pretrained_checkpoint=pretrained_checkpoint,
             ckpt_format="torch_dist",
             fully_parallel_save=True,


src/megatron/bridge/recipes/qwen/qwen3.py
```diff
@@ -41,26 +42,25 @@ class Qwen3CommonKwargs(TypedDict, total=False):
 
     # Core identifiers
     hf_path: str
-    dir: Optional[str]
+    dir: str | None
```
yaoyu-33, line 45: open an issue to change update it for good in all places?

src/megatron/bridge/recipes/qwen/qwen3.py
```diff
@@ -338,3 +355,252 @@ def _qwen3_common(
...
+        "tensor_parallelism": 8 if is_full_sft else 1,  # Match NeMo2: TP=8 for SFT, TP=1 for LoRA
+        "pipeline_parallelism": 1,
+        "peft": peft_value,
+        "finetune_lr": 5e-6 if is_full_sft else 1e-4,  # Match NeMo2: lower LR for SFT
+    }
+    combined_kwargs: Qwen3FinetuneKwargs = {**recommended_kwargs, **user_kwargs}
+    return _qwen3_finetune_common(**combined_kwargs)
+
+
+def _qwen3_finetune_common(
```
yaoyu-33, line 497: how is finetune common different from common? do we need 2 of them? @cuichenx  as well

ananthsub: i find it easier to read when they're split up. otherwise the user has to keep track of branching between pretraining vs sft vs peft across parallelisms, lr, dataset, and tokenizers. in addition, the finetuning recipe configs will have a lot of code in common, at least based on https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/collections/llm/recipes/finetune_default.py

if anything, having `_qwen3_sft_common` and `_qwen3_peft_common` helpers will make the split even more obvious to users/developers. wdyt?

cuichenx: i think it makes sense to separate finetune and pretrain. the data config, checkpoint config and tokenizer are all different. i think it even makes sense to have a common finetune helper for all models

src/megatron/bridge/recipes/qwen/qwen3.py
```diff
@@ -338,3 +355,252 @@ def _qwen3_common(
...
+
+    Default configuration: 1 node, 8 GPUs (TP=1, PP=1)
+    """
+    recommended_kwargs: Qwen3FinetuneKwargs = {
+        "hf_path": "Qwen/Qwen3-0.6B",
+        "tensor_parallelism": 1,
+        "pipeline_parallelism": 1,
+        "peft": "lora",
+        "finetune_lr": 1e-4,
+    }
```
cuichenx, line 376: add `is_full_sft` to this recipe as well?

src/megatron/bridge/recipes/qwen/qwen3.py
```diff
@@ -338,3 +355,252 @@ def _qwen3_common(
...
+
+    Default configuration: 1 node, 8 GPUs (TP=1, PP=1)
+    """
+    recommended_kwargs: Qwen3FinetuneKwargs = {
+        "hf_path": "Qwen/Qwen3-1.7B",
+        "tensor_parallelism": 1,
+        "pipeline_parallelism": 1,
+        "peft": "lora",
+        "finetune_lr": 1e-4,
+    }
```
cuichenx: add is_full_sft to this recipe as well?

src/megatron/bridge/recipes/qwen/qwen3.py
```diff
@@ -338,3 +355,252 @@ def _qwen3_common(
...
+    # Core model configuration
+    tensor_parallelism: int = 1,
+    pipeline_parallelism: int = 1,
+    pipeline_parallelism_dtype: torch.dtype | None = None,
+    virtual_pipeline_parallelism: int | None = None,
+    context_parallelism: int = 1,
+    sequence_parallelism: bool = False,
+    use_megatron_fsdp: bool = False,
+    # Finetuning-specific params
+    pretrained_checkpoint: str | None = None,
```
cuichenx, line 510: nit, pretrained checkpoint should not be optional right?

ananthsub: i kept it as optional here because users may do the following:

```
cfg = qwen3_32b_finetune_config()
apply_overrides(cfg, path_to_config)  # config file contains all overrides, including pretrained_checkpoint
finetune(cfg, forward_step)
```

if it were required, those users would need to either specify just the `pretrained_checkpoint` separately, or add a dummy value

```
cfg = qwen3_32b_finetune_config(pretrained_checkpoint="dummy")
apply_overrides(cfg, path_to_yaml_config)  # actually sets pretrained_checkpoint
finetune(cfg, forward_step)
```

since it's open-ended for how the final config is ultimately built, i wanted to be lenient here. we still assert that if calling `finetune`, the `pretrained_checkpoint` or `load` is set
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/2eea3d1d50cbd4f6f14d6ea72eac3297d71a7e76/src/megatron/bridge/training/finetune.py#L47-L49

cuichenx: I see, makes sense!

PR #976, opened 2025-10-16
Ability to run perf scripts without installing megatron-bridge

gautham-kollu:
Copy the run plugins required for perf scripts to individual file to allow running perf scripts without installing megatron-bridge.

Tests:
```
python scripts/performance/setup_experiment.py -m llama3 -s 8b --task pretrain --num_gpus 8  --config_file scripts/performance/configs/llama3/llama3_8b_llm_pretrain.yaml  -a <a> -p <p> -i <container> -g h100 --gpus_per_node 8 -hf <HF_TOKEN>
```

### DIFF
--- a/scripts/performance/perf_plugins.py
+++ b/scripts/performance/perf_plugins.py
@@ -32,6 +32,7 @@ import os
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Callable, List, Optional, Union
 
+
 MISSING_NEMO_RUN_MSG = "nemo-run is not available. Please install it with `pip install nemo-run`."
 
 


malay-nagda:
With this, don't we have to maintain the same plugins in 2 places- `scripts/performance/perf_plugins.py` and `megatron/bridge/recipes/run_plugins.py`?

i.e. we have to make sure changes in one location are copied to the other location for every change in either of the files.

malay-nagda:
@gautham-kollu this adds confusion where are we supposed to make new updates since there are 2 copies for the same plugins.

https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/976#issuecomment-3413997285

cc: @erhoo82 @ananthsub

gautham-kollu:
> @gautham-kollu this adds confusion where are we supposed to make new updates since there are 2 copies for the same plugins.
> 
> [#976 (comment)](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/976#issuecomment-3413997285)
> 
> cc: @erhoo82 @ananthsub

This is a valid concern. We chose to do for backward compatibility and in the future we will deprecate the original run_plugins.py in favor of a common places like  megatron-bridge/tools/. Doing this because it is an improvement in user-experience for customers

ananthsub:
@malay-nagda to add to that, https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/972 will remove the dependency that these plugins have on the megatron bridge src. after that, we can consolidate the plugins under `tools/` and make use of them in the perf scripts without duplication

PR #983, opened 2025-10-17
feat: changelog config

pablo-garay:


chtruong814:
How did you get to the labels in the config?  I'm not sure the PRs are really being labeled in such a way. Would everything just be in the default category then?

pablo-garay:
We'll use the default - all in one case - at least until we find a way to label properly (discussed with Ananth)

Issue #749, opened 2025-09-24
[docs] Automatic versioning upgrade for each release


ananthsub:
Follow up to https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/689

As much as possible we should automate the version bumps for docs during each release: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/692

yaoyu-33:
Closing — this issue has been open for 7+ months with no activity. If this is still relevant, please reopen or file a new issue with updated context. Thank you!

PR #986, opened 2025-10-17
Adding explainers to PyPi version switch

aschilling-nv:
- Adds explainers to versions1.json, conf.py, and project.json for PyPi version switcher.
- Sets the latest in the main branch to support the "Latest" drop-down in the version switcher.
- Sets absolute URL in version switcher.

Part of #749


ko3n1g (changes requested):
This is great. Thanks a lot for doing this. Let's not merge this now since I'm syncing the timing with @ananthsub.

I will lift the block when they confirmed

PR #989, opened 2025-10-17
github CI workflow improvement

yaoyu-33:
2. reduce num layers to 2 for most of tests
3. reduce train steps to 10 for most of tests
4. reduce gbs to 8 and mbs to 1 for most of tests
5. switch to offline HF

### DIFF
--- a/tests/functional_tests/recipes/test_mamba_recipes_pretrain.py
+++ b/tests/functional_tests/recipes/test_mamba_recipes_pretrain.py
@@ -16,7 +16,7 @@
 
 import pytest
 
-from megatron.bridge.recipes.mamba.mamba2_130m import pretrain_config as mamba2_130m_config
+from megatron.bridge.recipes.mamba.mamba2 import mamba2_130m_pretrain_config as mamba2_130m_config
 from tests.functional_tests.recipes.utils import run_pretrain_recipe_test
 
 
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -83,6 +83,9 @@ jobs:
     runs-on: self-hosted-nemo
     name: Launch_Unit_Tests
     environment: nemo-ci
+    env:
+      TRANSFORMERS_OFFLINE: "1"
+      HF_HUB_OFFLINE: "1"
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -127,6 +130,8 @@ jobs:
     environment: nemo-ci
     env:
       HF_HOME: /home/TestData/HF_HOME
+      TRANSFORMERS_OFFLINE: "1"
+      HF_HUB_OFFLINE: "1"
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/src/megatron/bridge/data/builders/finetuning_dataset.py
+++ b/src/megatron/bridge/data/builders/finetuning_dataset.py
@@ -94,12 +94,6 @@ class FinetuningDatasetBuilder:
         if self.packed_sequence_size > 0:
             from megatron.bridge.data.datasets.packed_sequence import prepare_packed_sequence_data
 
-            # Guard: require source jsonl files to exist. Unit tests expect KeyError here.
-            if not (self.train_path.exists() and self.validation_path.exists()):
-                raise KeyError(
-                    f"Required dataset files missing for packing: train={self.train_path}, val={self.validation_path}"
-                )
-
             if not self.train_path_packed.is_file():
                 print_rank_0(f"Preparing packed training data at {self.train_path_packed}")
                 prepare_packed_sequence_data(


scripts/run_ci_tests.sh
```diff
@@ -0,0 +1,262 @@
...
...-m coverage run --data-file=/opt/Megatron-Bridge/.coverage --source=/opt/Megatron-Bridge/ --parallel-mode -m pytest -o log_cli=true -o log_cli_level=INFO -v -s -x -m 'not pleasefixme' --tb=short -rA tests/functional_tests/training -k 'not test_inprocess_restart' && coverage run --data-file=/opt/Megatron-Bridge/.coverage --source=/opt/Megatron-Bridge/ --parallel-mode -m pytest -o log_cli=true -o log_cli_level=INFO -v -s -x -m 'not pleasefixme' --tb=short -rA tests/functional_tests/converter -k 'not gemma and not gemma2 and not gemma3 and not glm45' && coverage run --data-file=/opt/Megatron-Bridge/.coverage --source=/opt/Megatron-Bridge/ --parallel-mode -m pytest -o log_cli=true -o log_cli_level=INFO -v -s -x -m 'not pleasefixme' --tb=short -rA tests/functional_tests/models -k 'not gemma and not gemma2 and not gemma3 and not glm45' && python -m torch.distributed.run --nproc_per_node=2 --nnodes=1 -m coverage run --data-file=/opt/Megatron-Bridge/.coverage --source=/opt/Megatron-Bridge/ --parallel-mode -m pytest -o log_cli=true -o log_cli_level=INFO -v -s -x -m 'not pleasefixme' --tb=short -rA tests/functional_tests/recipes -k 'not gemma and not gemma2 and not gemma3 and not glm45'"; fi
```
ko3n1g: How about we call the `L2_XXX.sh` scripts here? That might be easier to maintain over time?

yaoyu-33: I just notice we have such L2 runs. Looking at recipes L2 scripts, seems not covering all recipe tests in recipe folder?

ananthsub: > seems not covering all recipe tests in recipe folder?

we can only run the small models on github CI

tests/functional_tests/models/test_gemma2_conversion.py
```diff
@@ -86,18 +86,8 @@ def gemma2_toy_model_path(self, tmp_path_factory):
...
-        if Path(pre_downloaded_path).exists():
-            print(f"Loading tokenizer from pre-downloaded path: {pre_downloaded_path}")
-            tokenizer = GemmaTokenizer.from_pretrained(pre_downloaded_path)
-        else:
-            # Fall back to downloading from HuggingFace
-            print("Pre-downloaded tokenizer not found, attempting to download from HuggingFace")
-            tokenizer = GemmaTokenizer.from_pretrained("google/gemma-2b")
+        # Download and save tokenizer from a reference Gemma model directly from Hugging Face
+        tokenizer = GemmaTokenizer.from_pretrained("google/gemma-2b")
         tokenizer.save_pretrained(model_dir)
```
ananthsub, line 91: why is this being removed? these are on the azure mount for CI. is it that not everyone running these tests will have access? in that case, what's the issue with the fallback logic?

tests/functional_tests/recipes/test_mamba_recipes_pretrain.py
```diff
@@ -19,27 +19,28 @@
...
 
 MAMBA_PRETRAIN_RECIPES = [
-    # (config_func, name, parallelism_overrides)
-    (mamba2_130m_config, "mamba2_130m", {}),  # Small model, use recipe defaults
-    (mamba2_370m_config, "mamba2_370m", {}),  # Small model, use recipe defaults
-    (mamba2_780m_config, "mamba2_780m", {}),  # Small model, use recipe defaults
+    # (config_func, name, parallelism_overrides, model_overrides)
+    (mamba2_130m_config, "mamba2_130m", {}, {"num_layers": 2, "hybrid_override_pattern": "MM"}),
+    (mamba2_370m_config, "mamba2_370m", {}, {"num_layers": 2, "hybrid_override_pattern": "MM"}),
+    (mamba2_780m_config, "mamba2_780m", {}, {"num_layers": 2, "hybrid_override_pattern": "MM"}),
```
ananthsub: what do you think about only having 1 such test to ensure its functional? for capacity concerns, i dont think we can offer comprehensive testing for each recipe at PR time. considering the similarities in the models, we run just mamba2_780m and rely on nightly testing for the other variants

tests/functional_tests/recipes/test_qwen_recipes_pretrain.py
```diff
@@ -26,17 +26,23 @@
 
 
 QWEN_PRETRAIN_RECIPES = [
-    # (config_func, name, parallelism_overrides)
-    (qwen2_500m_config, "qwen2_500m", {}),
-    (qwen25_500m_config, "qwen25_500m", {}),
+    # (config_func, name, parallelism_overrides, model_overrides)
+    (qwen2_500m_config, "qwen2_500m", {}, {"num_layers": 2}),
+    (qwen25_500m_config, "qwen25_500m", {}, {"num_layers": 2}),
 ]
```
ananthsub, line 32: we should add qwen3 600m here too

tests/functional_tests/training/test_inprocess_restart.py
```diff
@@ -46,7 +46,7 @@
 
 def build_test_config(
     save_dir: str,
-    train_iters: int = 20,
```
ananthsub, line 49: if timing is a concern, we can also have a custom model smaller than llama 32 1B below. other tests use 145M. the model size is not relevant for this functional test

tests/functional_tests/training/test_megatron_fsdp.py
```diff
@@ -325,8 +325,8 @@ def test_fsdp_pretrain_save_resume(self, tmp_path):
 
         try:
             seq_length = 512
-            total_iters = 20
-            checkpoint_iters = 10
+            total_iters = 10
```
ananthsub, line 328: update the comments below too. they're outdated now

tests/functional_tests/training/test_pretrain.py
```diff
@@ -66,7 +66,7 @@ def test_pretrain_with_checkpoint(self, tmp_path):
             global_batch_size = 8
             micro_batch_size = 1
             seq_length = 512
-            total_iters = 100
+            total_iters = 10
 
             model_cfg = Llama32ModelProvider1B(
```
ananthsub, line 71: @maanug-nv had mentioned this before. we should split out a small model provider config that's used for functional testing as a util in https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/tests/functional_tests/utils.py

1B is overkill for these types of tests

scripts/run_ci_tests.sh
```diff
@@ -0,0 +1,262 @@
+#!/usr/bin/env bash
```
ananthsub, line 1: IMO it's best to split this and the github CI workflow changes into a separate PR, and keep this as improvements to make tests run faster. what do you think?

tests/functional_tests/models/test_gemma2_conversion.py
yaoyu-33: it should get from HF_HOME folder with Transformer offline, not a custom folder, too much trouble for future.

tests/functional_tests/recipes/test_mamba_recipes_pretrain.py
yaoyu-33: ok with this, even removing mamba

tests/functional_tests/training/test_inprocess_restart.py
yaoyu-33: yeah, i dont want to deviate too much from original config, just reduce layers and experts for now

tests/functional_tests/training/test_pretrain.py
yaoyu-33: the layer is update to 1

ko3n1g (approved):
This is so cool!

PR #992, opened 2025-10-17
Temporary workaround for TP initialization: keep initialization enabled

yaoyu-33:


ananthsub (approved):
the real fix should go here: https://github.com/NVIDIA/Megatron-LM/blob/572f0198aa6a5d6b3f46364049ed5a0d4ff72cf2/megatron/core/tensor_parallel/layers.py#L131

https://github.com/NVIDIA/Megatron-LM/blob/572f0198aa6a5d6b3f46364049ed5a0d4ff72cf2/megatron/core/tensor_parallel/layers.py#L259-L260

where `set_tensor_model_parallel_attributes` should be called even if the weight initialization isn't performed: https://github.com/NVIDIA/Megatron-LM/blob/572f0198aa6a5d6b3f46364049ed5a0d4ff72cf2/megatron/core/tensor_parallel/layers.py#L131-L133

accepting in the meantime to fix for bridge users

PR #994, opened 2025-10-18
Handle variable sequence lengths in sampler + train step explicitly

ananthsub:
This is more efficient than what was implemented in https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/974
and more closely follows NeMo's assumptions for SFT datasets/ handling in Megatron Parallel


dimapihtar (approved):
LGTM. Thank you!

PR #999, opened 2025-10-20
Add fallback logic for slurm for initializing process group

ananthsub:
Not all clusters may set `RANK`, `WORLD_SIZE`, `MASTER_ADDR` or `MASTER_PORT` when launching. these environment variables are relied on for `torch.distributed.init_process_group` (and other places for logging before the global pg is set up). This PR adds fallback support for SLURM environment variables to `common_utils`. the PR is intended to introduce the smallest change for SLURM support without exposing any new interfaces to users. if this needs to be pluggable for users to customize for other clusters, we can expose an interface for this then.

### DIFF
--- a/src/megatron/bridge/utils/slurm_utils.py
+++ b/src/megatron/bridge/utils/slurm_utils.py
@@ -19,6 +19,7 @@ distributed training configuration from SLURM environment variables.
 """
 
 import os
+import warnings
 
 
 def is_slurm_job() -> bool:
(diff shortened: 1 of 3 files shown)


src/megatron/bridge/utils/common_utils.py
```diff
@@ -69,10 +94,75 @@ def get_last_rank() -> int:
...
+
+    Returns:
+        The master port.
+    """
+    # 1. Check standard environment variable
+    if "MASTER_PORT" in os.environ:
+        return int(os.environ["MASTER_PORT"])
+
+    # 2. Check if in SLURM and use default
+    slurm_port = _resolve_slurm_master_port()
```
malay-nagda: @ananthsub this call always returns `29500` for Slurm. Is this expected?

ananthsub: @malay-nagda I've updated the port to be determined based on the job id if available, matching the support provided previously through pytorch lightning's cluster environment plugin

src/megatron/bridge/utils/slurm_utils.py
```diff
@@ -0,0 +1,129 @@
...
+    Returns:
+        The master node address, or None if not in SLURM environment.
+    """
+    if not is_slurm_job():
+        return None
+
+    # Try both SLURM_NODELIST and SLURM_JOB_NODELIST
+    nodelist = os.environ.get("SLURM_NODELIST") or os.environ.get("SLURM_JOB_NODELIST")
+    if not nodelist:
+        return None
```
gautham-kollu: np: Line 83 should be unknown state as we are already checking. Consider adding a warning here

src/megatron/bridge/utils/slurm_utils.py
```diff
@@ -0,0 +1,129 @@
...
+        # Split into base and range part
+        # e.g., "node[001-004]" -> base="node", range_part="001-004"
+        base = nodelist.split("[")[0]
+        range_part = nodelist.split("[")[1].split("]")[0]
+
+        # Handle both ranges (001-004) and lists (001,003,005)
+        # Extract first element
+        first_element = range_part.split(",")[0].split("-")[0]
+
+        return f"{base}{first_element}"
```
gautham-kollu: np: Consider moving this into a util method

src/megatron/bridge/utils/common_utils.py
```diff
@@ -69,10 +104,67 @@ def get_last_rank() -> int:
...
+    """
+    if "MASTER_PORT" in os.environ:
+        return int(os.environ["MASTER_PORT"])
+
+    slurm_port = resolve_slurm_master_port()
+    if slurm_port is not None:
+        return slurm_port
+
+    warnings.warn("Could not determine master port from MASTER_PORT or SLURM environment. Defaulting to 29500.")
+    return 29500
```
gautham-kollu, line 167: np: Consider making global variable as it is getting used in 2 places.

gautham-kollu (approved):
LGTM!

PR #1001, opened 2025-10-20
log experimental function warning on rank 0 only

ananthsub:
to reduce warnings in logs


maanug-nv:
I think you'd need to add `patch("torch.distributed.get_rank", return_value=0)` to `test_pretrain_raises_error_when_process_group_initialized_with_inprocess_restart` now.

PR #895, opened 2025-10-08
Update deepseek recipes to use bridge, add recipe tests

yaoyu-33:

PR #1006, opened 2025-10-20
Make nemotronh recipe imports consistent with codebase

ananthsub:
follow example of #607 and #895 for a consistent import structure. swapping out the model provider to use the bridge can come separately after this refactor


maanug-nv (approved):
Looks great much simpler, thanks!

PR #1007, opened 2025-10-20
Make Nemotron nano recipe imports consistent with codebase

ananthsub:
depends on #1006

follow example of https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/607 and https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/895 for a consistent import structure. swapping out the model provider to use the bridge can come separately after this refactor


maanug-nv (commented):
👍

PR #1014, opened 2025-10-21
[Draft] Re-write HF LLM Dataset

yaoyu-33:


Phlip79:
Can you add some tests, similar to what is found at `tests/unit_tests/data/vlm_datasets`?

PR #1023, opened 2025-10-21
sequence_length -> seq_length

dimapihtar:


src/megatron/bridge/training/config.py
```diff
@@ -320,6 +323,10 @@ def finalize(self) -> None:
         This method calls the original Megatron Core GPTDatasetConfig.__post_init__()
         and then performs Bridge-specific validation.
         """
+        # Bridge seq_length → sequence_length
+        if self.seq_length is not None:
+            self.sequence_length = self.seq_length
```
ananthsub: this doesn't necessarily synchronize that any changes to `sequence_length` are mirrored to `seq_length` and vice versa

we still rely on `sequence_length` elsewhere too:
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/063498590a2923a63c04df1fd3e7a5467a519880/src/megatron/bridge/training/utils/train_utils.py#L447

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/063498590a2923a63c04df1fd3e7a5467a519880/src/megatron/bridge/training/config.py#L1240-L1245

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/063498590a2923a63c04df1fd3e7a5467a519880/scripts/performance/utils/helpers.py#L156

tests could be updated too: 
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/063498590a2923a63c04df1fd3e7a5467a519880/tests/unit_tests/recipes/test_llama_recipes.py#L101

docs also need to be updated:
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/063498590a2923a63c04df1fd3e7a5467a519880/docs/training/config-container-overview.md?plain=1#L169

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/063498590a2923a63c04df1fd3e7a5467a519880/docs/nemo2-migration-guide.md?plain=1#L159

src/megatron/bridge/training/config.py
```diff
@@ -299,7 +299,6 @@ def build_datasets(self, context: DatasetBuildContext) -> Tuple[Optional[Any], O
         pass
 
 
-@dataclass
```
ananthsub, line 305: could you explain why this is no longer a dataclass? what does the serialized config container yaml now look like with this change? 

why not do something like this to keep `seq_length` and `sequence_length` in sync? https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/b245756e35943083a77aff6217fe60dd1704f6ca/src/megatron/bridge/training/mixed_precision.py#L74-L80

ananthsub (approved):

PR #1033, opened 2025-10-22
followup to add tests for tokenizer changes in #969

ananthsub:
improve test coverage for eos property + null multimodal tokenizer introduced in sync #969


dimapihtar (approved):
LGTM. Thank you!

PR #1039, opened 2025-10-22
Cleanup ci test import fallback

ananthsub:
Depends on https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1038 and #1007 

now that the recipe imports are all updated to use the same style, we can remove the fallback support in `train_from_recipe` for testing


maanug-nv (approved):
LGTM. Doubtful this will cause breakages, but since all our CI tests depend on this, can we just run a nightly pipeline with latest commit on this branch before merging?

PR #1040, opened 2025-10-22
support loading checkpoints from a given step

ananthsub:
Add support for `ckpt_step` in the checkpoint config. 

This allows users to specify a specific checkpoint directory to resume from based on the iteration. This is an opt-in feature. The default is still to resume from the latest checkpoint saved.

Special handling is needed for PEFT where there are 2 checkpoint loads:
1. an initial load from `pretrained_checkpoint` 
2. potentially a load from  the `load` directory in case of resuming the adapter states within a training run. 

 
`ckpt_step` is intended to only take effect for `load` directories as `pretrained_checkpoint` should be unchanged over the course of training.


dimapihtar (approved):
LGTM. Thank you!

PR #1043, opened 2025-10-22
Add script to allow people run ci tests locally before use GH action.

yaoyu-33:


scripts/run_ci_tests.sh
```diff
@@ -0,0 +1,262 @@
...
...-m coverage run --data-file=/opt/Megatron-Bridge/.coverage --source=/opt/Megatron-Bridge/ --parallel-mode -m pytest -o log_cli=true -o log_cli_level=INFO -v -s -x -m 'not pleasefixme' --tb=short -rA tests/functional_tests/training -k 'not test_inprocess_restart' && coverage run --data-file=/opt/Megatron-Bridge/.coverage --source=/opt/Megatron-Bridge/ --parallel-mode -m pytest -o log_cli=true -o log_cli_level=INFO -v -s -x -m 'not pleasefixme' --tb=short -rA tests/functional_tests/converter -k 'not gemma and not gemma2 and not gemma3 and not glm45' && coverage run --data-file=/opt/Megatron-Bridge/.coverage --source=/opt/Megatron-Bridge/ --parallel-mode -m pytest -o log_cli=true -o log_cli_level=INFO -v -s -x -m 'not pleasefixme' --tb=short -rA tests/functional_tests/models -k 'not gemma and not gemma2 and not gemma3 and not glm45' && python -m torch.distributed.run --nproc_per_node=2 --nnodes=1 -m coverage run --data-file=/opt/Megatron-Bridge/.coverage --source=/opt/Megatron-Bridge/ --parallel-mode -m pytest -o log_cli=true -o log_cli_level=INFO -v -s -x -m 'not pleasefixme' --tb=short -rA tests/functional_tests/recipes -k 'not gemma and not gemma2 and not gemma3 and not glm45'"; fi
```
ananthsub: this seems like its going to be difficult to maintain. now there are 2 different places to update:
- this script for local testing
- the `L2_*` scripts that CI uses

why not call the L2 scripts here?

scripts/run_ci_tests.sh
```diff
@@ -0,0 +1,262 @@
...
+    export TORCH_CPP_LOG_LEVEL="error"
+    ft_launcher \
+      --rdzv_backend=c10d --rdzv_endpoint=127.0.0.1:29500 \
+      --nnodes=1 --nproc-per-node=2 \
+      --ft-param-rank_section_timeouts=setup:600,step:180,checkpointing:420 \
+      --ft-param-rank_out_of_section_timeout=300 \
+      --monitor-interval=5 --max-restarts=3 \
+      --ft-restart-policy=min-healthy \
+      -m pytest -o log_cli=true -o log_cli_level=INFO -v -s -x -m "not pleasefixme" --tb=short -rA \
+      tests/functional_tests/training/test_inprocess_restart.py
```
ananthsub, line 181: same here - as we need to sync changes from L2_training to this. maybe we can add an option to the script for wehther to use uv run python or not to reduce the duplication?

as a concrete example, we will need to add nsys profiling functional tests, which will require changes to L2_Launch_training that we'd then need to mirror here

PR #1051, opened 2025-10-22
feat: Dependabot automerge if successful

pablo-garay:
Dependabot : automerge PR if checks all complete successful 

Workflow tested worked successfully here:
https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/18758427697/job/54498962342

it did not merge PR only because of PR merge conflicts at the very end of the workflow

### DIFF
--- a/.github/workflows/_update_dependencies.yml
+++ b/.github/workflows/_update_dependencies.yml
@@ -107,7 +107,7 @@ jobs:
           git_user_signingkey: true
           git_commit_gpgsign: true
 
-      - name: Create Bump PR: Commit and push changes
+      - name: Create Bump PR Commit and push changes
         env:
           GH_TOKEN: ${{ secrets.PAT }}
         run: |


.github/workflows/_update_dependencies.yml
```diff
@@ -107,23 +107,207 @@ jobs:
...
+        if: steps.find-pr.outputs.pr-number != ''
+        env:
+          PR_NUMBER: ${{ steps.find-pr.outputs.pr-number }}
+        run: |
+          echo "Monitoring PR #${PR_NUMBER} for CI check completion..."
+          
+          MAX_ATTEMPTS=144  # Wait up to 12 hours (144 attempts * 5 minutes)
+          ATTEMPT=0
+          
+          while [ $ATTEMPT -lt $MAX_ATTEMPTS ]; do
```
chtruong814, line 258: Is there not a way to just set to auto-merge upfront rather than polling for results?

.github/workflows/_update_dependencies.yml
```diff
@@ -107,23 +107,207 @@ jobs:
...
+        run: |
+          echo "========================================="
+          echo "Preparing to commit changes"
+          echo "========================================="
+          
+          # Configure git
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git config --global user.signingkey ""
+          git config --global commit.gpgsign false
```
ko3n1g, line 122: Why are not using the provided PGP key?

ko3n1g (commented):
I'm not sure if this works if main branch has protection rules. Can we clarify which rules were applied in the private fork's main branch?

.github/workflows/_update_dependencies.yml
pablo-garay: That could be another approach
I came with this current approach
i guess we could do that if wanted (i personally have no preference)
though i would suggest we do that in a follow up PR since this was already tested & works

.github/workflows/_update_dependencies.yml
pablo-garay: let's please update that if wanted in follow up PR

ko3n1g (approved):

PR #1053, opened 2025-10-22
fix Nemotron h provider and bridge

ananthsub:
fix issues with nemotron nano v2
- set kv_channels and map to head_dim from the HF config
- don't use lambdas in model provider so that we can reload the provider class from the yaml saved in the checkpoint

the following now works as expected:

```
python examples/conversion/convert_checkpoints.py import \
    --hf-model nvidia/NVIDIA-Nemotron-Nano-9B-v2-Base\
    --megatron-path <output_path>
```

```
torchrun --standalone --nnodes=1 --nproc-per-node=2   examples/conversion/hf_to_megatron_generate_text.py   --hf_model_path "nvidia/NVIDIA-Nemotron-Nano-9B-v2-Base"   --megatron_model_path "<output_path>   --prompt "The capital of France is"   --max_new_tokens 20   --tp 2   --pp 1
```

```
======== GENERATED TEXT OUTPUT ========
Prompt: The capital of France is
Generated: <s>The capital of France is Paris." This is a statement of fact, not a question. It's not a question, so
=======================================
```

### DIFF
--- a/src/megatron/bridge/models/nemotronh/nemotron_h_bridge.py
+++ b/src/megatron/bridge/models/nemotronh/nemotron_h_bridge.py
@@ -59,7 +59,7 @@ class NemotronHBridge(MegatronModelBridge):
             add_bias_linear=hf_config.use_bias,
             num_attention_heads=hf_config.num_attention_heads,
             num_query_groups=hf_config.num_key_value_heads,
-            kv_channels=getattr(hf_config, "head_dim"),
+            kv_channels=getattr(hf_config, "head_dim", None),
             init_method_std=hf_config.initializer_range,
             layernorm_epsilon=hf_config.layer_norm_epsilon,
             make_vocab_size_divisible_by=self.make_vocab_size_divisible_by(hf_config.vocab_size),


src/megatron/bridge/models/mamba/mamba_provider.py
```diff
@@ -72,7 +85,7 @@ class MambaModelProvider(TransformerConfig, ModelProviderMixin[MCoreMambaModel])
     deallocate_pipeline_outputs: bool = True
     bias_dropout_fusion: bool = True
     cross_entropy_loss_fusion: bool = True
-    mamba_stack_spec: Union[ModuleSpec, Callable[[], ModuleSpec]] = lambda: default_mamba_stack_spec
+    mamba_stack_spec: Union[ModuleSpec, Callable[[], ModuleSpec]] = get_default_mamba_stack_spec
```
ananthsub, line 88: todo: check for lambdas in the finalize

src/megatron/bridge/models/nemotronh/nemotron_h_provider.py
```diff
@@ -26,6 +26,20 @@
 logger = logging.getLogger(__name__)
 
 
+def nemotron_h_squared_relu(x: torch.Tensor) -> torch.Tensor:
```
yaoyu-33: seems already exists in src/megatron/bridge/training/mlm_compat/activations.py

ananthsub: good catch, i'll flip this such that the models/ defines the activation func and mlm compat imports from here

PR #1058, opened 2025-10-23
llama3 finetune recipes

ananthsub:
Screenshot 2025-10-29 at 7 40 33 AM

### DIFF
--- a/src/megatron/bridge/training/train.py
+++ b/src/megatron/bridge/training/train.py
@@ -589,7 +589,6 @@ def train_step(
         seq_length = model_config.seq_length  # Default for pretraining
         forward_backward_data_iterator = data_iterator  # Default for pretraining
 
-
         if cfg.dataset.dataloader_type == "batch":
             # Finetuning path to support variable-length sequences
             from megatron.bridge.data.finetuning import prepare_finetuning_batch


tests/unit_tests/recipes/test_llama_recipes.py
cuichenx, line 1: can we add a test for DoRA since the recipes enable it?

src/megatron/bridge/training/eval.py
```diff
@@ -99,15 +101,36 @@ def evaluate(
             if verbose:
                 print_rank_0(f"Evaluating iter {iteration}/{state.cfg.train.eval_iters}")
 
+            # Handle finetuning vs pretraining data consumption
+            seq_length = state.cfg.model.seq_length  # Default for pretraining
+            eval_data_iterator = data_iterator  # Default for pretraining
+
+            if state.cfg.dataset.dataloader_type == "batch":
```
yaoyu-33, line 108: is this check to decide finetune a bit implicit

ananthsub: it's tied to the sampler used, specifically for parity with this from nemo since we use the same global batch sampler: https://github.com/NVIDIA-NeMo/NeMo/blob/10327b8e58f300d0af40c90b22c0641412e951c5/nemo/lightning/megatron_parallel.py#L1481-L1497

Issue #786, opened 2025-09-26
Model Support Tracker


yaoyu-33:
| Model Name   | Status | Ticket | Bridge Conversion | Pretrain Recipes | SFT / LoRA Recipes |
|--------------|--------|--------|-------------------|------------------|--------------------|
| Gemma3       |        | #780   |                   |                  |                    |
| GLM4.5v      |        | #779   |                   |                  |                    |
| GLM4.5       |        | #778   |                   |                  |                    |
| Falcon H1    |        | #657   |                   |                  |                    |
| GLM-4.5V     |        | #637   |                   |                  |                    |
| GPT-OSS      |        | #556   |                   |                  |                    |
| OLMoE        |        | #515   |                   |                  |                    |
| GPT-OSS  |        | #380   |                   |                  |                    |
| Bert         |    | #154   |                   |                  |                    |
| Nemotron     |        | #150   |                   |                  |                    |
| Gemma        |        | #147   |                   |                  |                    |
| Qwen2        |        | #145   |                   |                  |                    |

PR #1068, opened 2025-10-23
[Model Support] Qwen3-Next

yaoyu-33:
part of #786

### DIFF
--- a/tests/unit_tests/models/qwen/test_qwen3_next_bridge.py
+++ b/tests/unit_tests/models/qwen/test_qwen3_next_bridge.py
@@ -214,7 +214,7 @@ class TestQwen3NextBridge:
         result = bridge.provider_bridge(mock_pretrained_qwen3_next)
 
         # Check MTP configuration
-        assert result.mtp_num_layers == 1
+        assert result.mtp_num_layers == 0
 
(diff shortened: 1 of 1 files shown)


yaoyu-33:
torchrun --nproc_per_node=8 examples/conversion/hf_to_megatron_generate_text.py --hf_model_path="Qwen/Qwen3-Next-80B-A3B-Instruct" --prompt="Hello, how are you?" --ep 8
were you able to do any verifications?
Prompt: Hello, how are you?
Generated: Hello, how are you? liberty rig rig rig定arellaijn翻 van-INF定itecture翻定itecture翻翻 rig定itecture

Output non-sense right now

ananthsub (approved):

pyproject.toml
```diff
@@ -66,6 +66,7 @@ classifiers = [
     "Topic :: Utilities",
 ]
 dependencies = [
+    "transformers==4.57.1",
```
ko3n1g, line 69: Could we just a looser pin than this? If we know that the lower boundary is 4.57.1, I would pledge for `>=`. Even better would be to have no pin, and to gracefully raise exception on the affected codepaths when transformers is `<4.57.1`. That way we would be more flexible

ko3n1g (approved):
Discussed offline that a follow up PR will follow

PR #1070, opened 2025-10-24
fix: resuming failed because wct was reinitialized

hawkoli1987:


ananthsub:
merged this fix in https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1110

PR #1105, opened 2025-10-27
set default trust remote code to True and use torch_dtype

liding-nv:
fix some minor issues in conversion scripts
- add default `trust_remote_code=True` to `AutoBridge. from_hf_pretrained`
- change `dtype` to `torch_dtype` to `from_pretrained`

### DIFF
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -20,9 +20,11 @@ from dataclasses import dataclass, field, fields
 from pathlib import Path
 from typing import Any, Literal, Optional, Tuple, Union
 
+import torch
 from megatron.core.datasets.gpt_dataset import GPTDatasetConfig as MCoreGPTDatasetConfig
 from megatron.core.distributed import DistributedDataParallelConfig as MCoreDistributedDataParallelConfig
 from megatron.core.optimizer import OptimizerConfig as MCoreOptimizerConfig
+from megatron.core.transformer.enums import AttnBackend
 
 from megatron.bridge.data.datasets.packed_sequence import PackedSequenceSpecs
 from megatron.bridge.models import GPTModelProvider, T5ModelProvider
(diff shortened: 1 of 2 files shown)


examples/conversion/compare_hf_and_megatron/compare.py
```diff
@@ -503,7 +503,7 @@ def _load_megatron_model(args):
         )
     else:
         # Convert from HF to Megatron
-        bridge = AutoBridge.from_hf_pretrained(args.hf_model_path)
+        bridge = AutoBridge.from_hf_pretrained(args.hf_model_path, trust_remote_code=True)
```
ananthsub, line 506: why make `trust_remote_code=True` as the default?

liding-nv: i was testing a local nemotron model and need to set this to `True`. This script does not expose this option as one of arguments. 

I looked at other scripts and they seems to default this to `True`, e.g.,
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/838c61600951719a4dad7ee1ee9b8bc7cef68f24/examples/conversion/hf_to_megatron_generate_text.py#L114

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/838c61600951719a4dad7ee1ee9b8bc7cef68f24/examples/conversion/hf_megatron_roundtrip_multi_gpu.py#L79

Is this the correct way to do it?

ananthsub: ah okay, yes that is the right way. https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/838c61600951719a4dad7ee1ee9b8bc7cef68f24/src/megatron/bridge/models/conversion/auto_bridge.py#L222-L224

since it's a script, `trust_remote_code=True` should be okay

ananthsub (approved):

PR #1110, opened 2025-10-27
fix throughput logging when resuming

ananthsub:
carry forward https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1070


dimapihtar (approved):
LGTM. Thank you!

PR #1113, opened 2025-10-28
Support Slurm Entrance for ModelZoo

Wohox:
This merge request adds slurm entrance for modelZoo(https://github.com/yanring/Megatron-MoE-ModelZoo).

Example: 
```bash
cd <Megatron-Bridge>/scripts/moe_model_zoo
bash examples/qwen3/launch_qwen235b.sh
```


src/megatron/bridge/training/config.py
```diff
@@ -432,10 +432,10 @@ def finalize(self) -> None:
 
         # Validate mutual exclusivity between iteration-based and sample-based scheduler fields
         has_iter_fields = (
-            self.lr_decay_iters is not None or self.lr_warmup_iters != 0 or self.lr_wsd_decay_iters is not None
+            self.lr_decay_iters is not None or self.lr_warmup_iters is not None or self.lr_wsd_decay_iters is not None
```
yaoyu-33, line 500: why is this change?
do we need to double check not none and not 0?

src/megatron/bridge/models/transformer_config.py
```diff
@@ -66,6 +66,14 @@ def finalize(self) -> None:
         """
         if self.pipeline_model_parallel_size > 1 and self.pipeline_dtype is None:
             self.pipeline_dtype = self.params_dtype
+        if isinstance(self.attention_backend, str):
```
yaoyu-33: need some comments on this code block. what attention backend value it supports now? why the change is required?

src/megatron/bridge/training/config.py
Wohox: Mainly 2 reasons:
* Default value of iters should be consistent, mixed usage of 0 and None is not a good idea.
* Setting to 0 will trigger mixed iter/sample usage assertion.

src/megatron/bridge/models/transformer_config.py
Wohox: Reason: User can only pass string or numeric values in yaml, therefore it's necessary to map these values to its corresponding dtype. I know that torch.dtype conversion is handled in `_apply_overrides`(https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/training/utils/omegaconf_utils.py#L369), but this method contains no business logic, therefore I add the attn backend conversion to transformer_config class.

All the listed types are supported: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/enums.py#L60-L67

PR #1120, opened 2025-10-28
ci: Refactor perf setup experiment

ko3n1g:
* Breaks down the script into functions that allow injecting custom executors
* Adds a detach argument for cases where we want to listen to the execution


malay-nagda (approved):
LGTM

PR #1133, opened 2025-10-29
Add moonlight 16B recipe

suiyoubi:
serialization requires for pp layout

### DIFF
--- a/.github/actions/test-template/action.yml
+++ b/.github/actions/test-template/action.yml
@@ -239,6 +239,7 @@ runs:
           --runtime=nvidia --gpus all \
           --shm-size=64g \
           --env TRANSFORMERS_OFFLINE=0 \
+          --cpus=40 \
           --env HYDRA_FULL_ERROR=1 \
           --env HF_HOME=/home/TestData/HF_HOME \
           --env NEMO_HOME=/home/TestData/nemo_home \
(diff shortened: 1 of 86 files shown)


suiyoubi:
/ok to test [2426242](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1133/commits/2426242e5d321e11cdf92b61de1d2b705d1bb842)

src/megatron/bridge/training/state.py
```diff
@@ -191,11 +192,35 @@ def wandb_logger(self) -> Optional[Any]:
...
+                        if not isinstance(result, str):
+                            # __str__ returned non-string type, use type name instead
+                            return f"<{type(obj).__name__}>"
+                        return result
+                    except Exception:
+                        # __str__ raised an exception, use type name as fallback
+                        return f"<{type(obj).__name__}>"
+
+                config_dict = self.cfg.to_dict()
+                sanitized_config = json.loads(json.dumps(config_dict, default=safe_serialize))
```
ananthsub, line 217: did you see any errors with checkpoint saving as well? or was this only an issue with wandb?

suiyoubi: checkpoint is fine I only observed this serialization issue with wandb.

suiyoubi: the root cause of this is the pp layout which is a list

ananthsub: https://github.com/NVIDIA/Megatron-LM/pull/2055 for the update to MLM

PR #1150, opened 2025-10-30
fix: hf attention mask in qwen vl training is different from megatron

yaoyu-33:


src/megatron/bridge/models/qwen_vl/modeling_qwen25_vl.py
```diff
@@ -190,12 +190,16 @@ def forward(
             if self.config.sequence_parallel:
                 inputs_embeds = scatter_to_sequence_parallel_region(inputs_embeds)
 
+        # Megatron attention mask (B,1,S,S) or (1,1,S,S) boolean mask (True = masked)
+        # HuggingFase attention mask (B,S) mask (1 = keep).
```
ananthsub:
```suggestion
        # HuggingFace attention mask (B,S) mask (1 = keep).
```

ananthsub (approved):

chtruong814:
Moving forward with merging since it's a minor change.  Testing this will be covered elsewhere.

PR #1155, opened 2025-10-31
Add gpt-oss finetune recipes + fixes

cuichenx:
Works with https://github.com/NVIDIA/Megatron-LM/pull/2076

### DIFF
--- a/scripts/performance/configs/deepseek/deepseek_llm_pretrain.py
+++ b/scripts/performance/configs/deepseek/deepseek_llm_pretrain.py
@@ -47,12 +47,12 @@ def set_deepseek_v3_common_configs(cfg: ConfigContainer) -> None:
 def deepseek_v3_gb300_256gpus_config(precision: str = "bf16", fp8_recipe: str = "cs") -> ConfigContainer:
     """GB300, 256xGPU, baseline config."""
     if precision == "bf16":
-        base_cfg = base_cfgs.DEEPSEEK_V3_GB300_256GPUS_BF16_PARALLEL_CONFIG
+        base_cfg = base_cfgs.DEEPSEEK_V3_GB300_256GPUS_BF16_BASE_CONFIG
         precision_config = get_precision_config(precision)
     else:
-        base_cfg = base_cfgs.DEEPSEEK_V3_GB300_256GPUS_FP8_CS_PARALLEL_CONFIG
+        base_cfg = base_cfgs.DEEPSEEK_V3_GB300_256GPUS_FP8_CS_BASE_CONFIG
         if fp8_recipe == "mx":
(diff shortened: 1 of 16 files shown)


src/megatron/bridge/models/conversion/param_mapping.py
```diff
@@ -790,7 +790,8 @@ def hf_to_megatron(
         # not the actual value, so it is safe to do this.
         normalized_param = self._normalize_expert_param_name(self.megatron_param)
         _, target_param = get_module_and_param_from_name(megatron_module, normalized_param)
-
+        if target_param.device.type == "cpu":
+            target_param = target_param.to(torch.cuda.current_device())
```
cuichenx: this was added for TP import of attention sink, where it must be initialized on the gpu instead of cpu. I added a fix in megatron core to initialize this on GPU as well. but I think it's best in general to ensure parameter is on gpu before any tp communication.

yaoyu-33: do you need to move it back?

src/megatron/bridge/models/gpt_oss/gpt_oss_provider.py
```diff
@@ -63,8 +63,8 @@ class GPTOSSProvider(GPTModelProvider):
     yarn_beta_fast: float = 32.0
     yarn_beta_slow: float = 1.0
     yarn_correction_range_round_to_int: bool = False
-    yarn_mscale: float = 1.0
-    yarn_mscale_all_dim: float = 1.0
+    yarn_mscale: Optional[float] = None
```
yaoyu-33, line 66: this is changing the default value, will it break anything?

cuichenx: it's handled in mcore: https://github.com/NVIDIA/Megatron-LM/pull/2038/files#diff-2590b64ed17e41f658591a5907a1c723dfa85831007cd0be1bd5615432161d3dR248-R250
turns out mscale should be None to match HF's implementation

cuichenx:
/ok to test [7367429](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1155/commits/7367429371012ca6aa4f78f56091b9dbb325c12d)

src/megatron/bridge/models/conversion/param_mapping.py
cuichenx: reverted this change

cuichenx:
/ok to test [195467f](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1155/commits/195467f384bf41241908a48f15f2e46be8424777)

cuichenx:
ci error can be resolved after 2076 is merged

ananthsub (dismissed):

Issue #787, opened 2025-09-26
Add unit tests for flop utils


gautham-kollu:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

yaoyu-33:
Closing — this issue has been open for 7+ months with no activity. If this is still relevant, please reopen or file a new issue with updated context. Thank you!

PR #1162, opened 2025-11-01
Add unit-test for flop-calculator

gautham-kollu:
Add unit-test to test flop-calculator a few known model types. 
 - Llama for dense
 - Qwen for MoE

Addresses #787  and #850


tests/unit_tests/utils/test_flop_utils.py
```diff
@@ -0,0 +1,57 @@
...
+            ("llama", "llama3_70b_pretrain_config", 8192, 151936, 3.68e15),
+            ("llama", "llama31_405b_pretrain_config", 8192, 151936, 2.07e16),
+            ("qwen", "qwen3_30b_a3b_pretrain_config", 4096, 151643, 9.42e13),
+            ("qwen", "qwen3_235b_a22b_pretrain_config", 4096, 151643, 6.06e14),
+        ],
+    )
+    def test_flops(self, model_family, model_config_func_name, seq_length, vocab_size, expected_flops):
+        model_family_module = importlib.import_module(f"megatron.bridge.recipes.{model_family}")
+        cfg = getattr(model_family_module, model_config_func_name)()
+        cfg.model.finalize()
```
maanug-nv, line 40: do we need to run validate/finalize on the whole config for the flops calc? im not sure what other configs besides the model are used by the flops calc, so just trying to understand first.

tests/unit_tests/utils/test_flop_utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+from megatron.bridge.training.utils.flop_utils import num_floating_point_operations
+from megatron.bridge.utils.vocab_utils import calculate_padded_vocab_size
+
+
+class TestFlops:
+    @pytest.mark.parametrize(
+        "model_family,model_config_func_name, seq_length, vocab_size, expected_flops",
+        [
+            ("llama", "llama3_8b_pretrain_config", 8192, 151936, 4.22e14),
```
ananthsub, line 30: this doesn't work in the unit test suite because we don't supply the HF token. the recipes pull the model configs from HF, and since this is a gated repo, the test is failing.

maanug-nv: ~~do we mount the fileshare? can't we just cache the llama models?~~
llama 3 8b is already cached , so it seems we don't mount the fileshare

tests/unit_tests/utils/test_flop_utils.py
gautham-kollu: We are using the model and the tokenizer for now.  This is more simulated just for tests.

PR #1178, opened 2025-11-04
Low Precision Recipes for LLama3-8B

adityavavreNVDA:
This PR adds recommended args for LLama3-8B low precision recipes. Following recipes are verified (long convergence 1T tokens):
- FP8 current scaling
- MXFP8
- NVFP4

Attaching a short convergence loss curve for reference:
image

### DIFF
--- a/src/megatron/bridge/recipes/llama/llama3.py
+++ b/src/megatron/bridge/recipes/llama/llama3.py
@@ -258,7 +258,7 @@ def llama3_8b_low_precision_pretrain_config(
         "bf16_with_mxfp8_mixed",
         "bf16_with_fp8_current_scaling_mixed",
         "bf16_with_nvfp4_mixed",
-    ], f"Invalid low precision recipe: {mixed_precision_recipe}"
+    ], f"Invalid low precision recipe: {mixed_precision_recipe}. This recipe has not been tested yet."
     precision_config = get_mixed_precision_config(mixed_precision_recipe)
     recommended_kwargs: Llama3CommonKwargs = {
         "hf_path": "meta-llama/Meta-Llama-3-8B",
--- a/tests/unit_tests/recipes/test_llama_recipes.py
+++ b/tests/unit_tests/recipes/test_llama_recipes.py
@@ -70,12 +70,6 @@ def _safe_overrides_for(name: str) -> dict:
                 # Note: Finetuning recipes set parallelism internally based on PEFT vs full SFT
             }
         )
-    elif "low_precision" in lname:  # low precision recipes need an additional precision argument
-        overrides.update(
-            {
-                "mixed_precision_recipe": "bf16_with_fp8_current_scaling_mixed",
-            }
-        )
     else:
         # Pretrain-specific overrides
         overrides.update(
@@ -88,6 +82,13 @@ def _safe_overrides_for(name: str) -> dict:
                 "context_parallel_size": 1,
             }
         )
+        # Low precision recipes require an additional mixed_precision_recipe argument
+        if "low_precision" in lname:
+            overrides.update(
+                {
+                    "mixed_precision_recipe": "bf16_with_fp8_current_scaling_mixed",
+                }
+            )
 
         # Large models/variants may set additional flags in pretrain recipes
         if "70b" in lname or "405b" in lname:


adityavavreNVDA:
/ok to test [0c67d7b](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1178/commits/0c67d7bd77c119a727644dfc1f618473fc6f7a36)

src/megatron/bridge/recipes/llama/llama3.py
```diff
@@ -192,7 +192,9 @@ def llama3_8b_128k_pretrain_config(**user_kwargs: Unpack[Llama3CommonKwargs]) ->
     combined_kwargs: Llama3CommonKwargs = {**recommended_kwargs, **user_kwargs}
     return _llama3_common(**combined_kwargs)
 
-def llama3_8b_low_precision_pretrain_config(mixed_precision_recipe: str, **user_kwargs: Unpack[Llama3CommonKwargs]) -> ConfigContainer:
+def llama3_8b_low_precision_pretrain_config(
+    mixed_precision_recipe: str, **user_kwargs: Unpack[Llama3CommonKwargs]
```
ananthsub, line 241: n00b question: what is the advantage of having a dedicated function for this vs. users passing in or overriding the default precision setting to use the ones listed below?

adityavavreNVDA: Wanted to make it clear to users that these recipes are well tested for convergence and specify the hyperparams used in long convergence testing. The default params specified for bf16 precision don't seem to work well with low precision. For example, we found that AdamW epsilon has a significant effect on convergence.

adityavavreNVDA:
/ok to test [8ab27c8](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1178/commits/8ab27c8a0eb149ba0075b709931505fe7f923102)

adityavavreNVDA:
/ok to test [402563b](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1178/commits/402563b5a0f4da0b9d5a44fe8b1944156032703a)

src/megatron/bridge/recipes/llama/llama3.py
ananthsub: what do you think about exposing these as individual recipe configs? e.g. `llama3_8b_bf16_mxfp8_mixed_pretrain_config`, `llama3_8b_bf16_fp8_cs_mixed_pretrain_config`, `llama3_8b_bf16_nvfp4_mixed_pretrain_config`

this might be clearer for which low-precision recipes have been tested for long convergence, and can be easier to follow when hyperparams vary across recipes

adityavavreNVDA: Apart from the precision configs, there are no other differences in the recipes. Hence I decided to just define one function for all. To be very clear with with which recipes have been tested I have an assert statement inside the function which checks if its one of FP8CS, MXFP8, and NVFP4 or not. @cuichenx  / @yaoyu-33 any thoughts?

adityavavreNVDA: Maybe in the assert statement instead of saying "Invalid Recipe" I can just say the recipe has not been verified for long convergence?

cuichenx: I would support individual recipe configs like llama3_8b_bf16_mxfp8_mixed_pretrain_config, llama3_8b_bf16_fp8_cs_mixed_pretrain_config, llama3_8b_bf16_nvfp4_mixed_pretrain_config

Or maybe at least write in the doc string which recipes are explicitly tested. We can't guarantee a newly added low precision config would work with this, for example.

adityavavreNVDA: Exactly why I check inside the function which type of recipe is used and return an error if its not fp8cs, mxfp8, or nvfp4.

adityavavreNVDA: Because other than the mixed precision configs there is no other difference between these recipes.

ananthsub (approved):

adityavavreNVDA:
/ok to test [8bdd0cd](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1178/commits/8bdd0cd0fdd803148e7629fcc7e8b8fdec5334c9)

src/megatron/bridge/recipes/llama/llama3.py
```diff
@@ -192,6 +193,46 @@ def llama3_8b_128k_pretrain_config(**user_kwargs: Unpack[Llama3CommonKwargs]) ->
...
+        "bf16_with_fp8_current_scaling_mixed",
+        "bf16_with_nvfp4_mixed",
+    ], f"Invalid low precision recipe: {mixed_precision_recipe}"
+    precision_config = get_mixed_precision_config(mixed_precision_recipe)
+    recommended_kwargs: Llama3CommonKwargs = {
+        "hf_path": "meta-llama/Meta-Llama-3-8B",
+        "tensor_parallelism": 1,
+        "pipeline_parallelism": 1,
+        "context_parallelism": 2,
+        "sequence_parallelism": False,
```
ananthsub: these were updated in https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1199 so we use the megatron core terminology everywhere

```suggestion
        "tensor_model_parallel_size": 1,
        "pipeline_model_parallel_size": 1,
        "context_parallel_size": 2,
        "sequence_parallel": False,
```

adityavavreNVDA:
/ok to test [8541129](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1178/commits/854112909f0ab31c1f86449e61d165f45a731f19)

ananthsub (approved):

adityavavreNVDA:
/ok to test [d899be7](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1178/commits/d899be79ac2e3937b6126f0ef90b3258946ab7b6)

adityavavreNVDA:
/ok to test [75ad7af](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1178/commits/75ad7af0f306bc499237677c5b352e28a7db632e)

adityavavreNVDA:
/ok to test [213aca1](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1178/commits/213aca159b22bb7489dc1953d46a0230f746a383)

ananthsub (approved):

PR #1203, opened 2025-11-06
Bridge instantiate_utils: drop unexpected config keys with warning

yaoyu-33:
Risks of Using *Lenient* Mode for Config Handling**

Using *lenient* mode when loading configs (instead of *strict*) can avoid errors when config fields are missing or changed, but it introduces several concerns:

#### ✅ **Pros**

* Prevents errors when config keys are missing or removed across versions.
* Makes backward compatibility easier initially.

#### ⚠️ **Risks / Potential Issues**

1. **Hidden Config Mismatches**

   * Missing or outdated config fields won’t throw errors, so silent differences may go unnoticed.
   * This can lead to unexpected behaviors and makes debugging harder.

2. **Reproducibility Concerns**

   * With lenient mode, there are *no guarantees* that runs are reproducible across PyTorch/TE/MCore versions.
   * Config drift over time becomes hard to track.

3. **Version Guarding Becomes MCore’s Problem**

   * If configs are dropped or renamed, a lenient approach pushes the burden to MCore side to guard changes properly.

4. **Technical Debt Over Time**

   * Without strict checking, config schema changes accumulate “silent failures”.
   * Eventually, migrations or retroactive fixes become harder.

  - Introduces defensive handling in `megatron.bridge.utils.instantiate_utils.instantiate_node` to gracefully ignore unexpected config keys when instantiating targets.
  - Emits a `logging.WARNING` identifying dropped keys and the config `full_key` to aid debugging.
  - Adds unit tests verifying both the drop-and-warn behavior and that targets accepting `**kwargs` are unaffected.

- Why
  - Backward compatibility with older checkpoints/config YAMLs that may contain stale fields (e.g., `use_megatron_fsdp`) caused failures like:
    - TypeError("OptimizerConfig.__init__() got an unexpected keyword argument 'use_megatron_fsdp'")
  - We prefer a non-fatal path with clear warnings rather than hard failures during resume/migration.

- What changed
  - `instantiate_utils.py`
    - New `_filter_kwargs_for_target(target, kwargs, full_key)`:
      - Uses `inspect.signature` to determine accepted keyword parameters.
      - If the target has `**kwargs`, forwards all keys unchanged.
      - Otherwise, drops keys not present in keyword-capable params and logs a warning including the dropped key list and `full_key`.
    - Integrated the filter just before calling the resolved `_target_`.
  - Tests
    - `tests/unit_tests/bridge/test_instantiate_utils.py`:
      - `test_drops_unexpected_kwargs_and_warns`: ensures unexpected keys are removed and a warning mentioning them is emitted.
      - `test_allows_kwargs_when_target_accepts_var_kwargs`: ensures no warning is produced and keys are preserved when the target accepts `**kwargs`.

### DIFF
--- a/src/megatron/bridge/training/utils/checkpoint_utils.py
+++ b/src/megatron/bridge/training/utils/checkpoint_utils.py
@@ -33,6 +33,7 @@ TRACKER_PREFIX = "latest"
 CONFIG_FILE = "run_config.yaml"
 
 logger = logging.getLogger(__name__)
+_RUNTIME_ONLY_TARGETS = frozenset({"megatron.core.timers.Timers"})
 
 
 def file_exists(path: str) -> bool:
@@ -276,7 +277,7 @@ def read_run_config(run_config_filename: str) -> dict[str, Any]:
                 else:
                     with open(run_config_filename, "r") as f:
                         config_dict = yaml.safe_load(f)
-                config_obj[0] = config_dict
+                config_obj[0] = _sanitize_run_config_object(config_dict)
             except Exception as e:
                 error_msg = f"ERROR: Unable to load config file {run_config_filename}: {e}"
                 sys.stderr.write(error_msg + "\n")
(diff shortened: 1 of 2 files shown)


src/megatron/bridge/utils/instantiate_utils.py
```diff
@@ -356,6 +347,53 @@ def _convert_target_to_string(t: Any) -> Any:
...
+        return kwargs
+
+    target_str = _convert_target_to_string(target)
+    if mode == InstantiationMode.LENIENT:
+        # Warn and drop the unexpected keys
+        warning_msg = f"Dropping unexpected config keys for target '{target_str}': {sorted(unexpected)}"
+        if full_key:
+            warning_msg += f"\nfull_key: {full_key}"
+        logging.warning(warning_msg)
+        return {k: v for k, v in kwargs.items() if k in allowed_keys}
```
ananthsub: this avoids erroring out when fields are removed from the config. but in the case of changes like this, the semantics are lost: https://github.com/NVIDIA/Megatron-LM/pull/1917

for example:
- a user previously used set `external_cuda_graph=True` and the config did not yet have `cuda_graph_scope` and has checkpoints saved with this config
- in a newer version where `external_cuda_graph` is removed, this PR will drop the arg in lenient mode.

it might not always be possible to infer the new setting from an old one so this might be inevitable, but we should make no claims about full reproducibility across versions in this case

tests/unit_tests/bridge/test_instantiate_utils.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python3
```
ananthsub: move this into the existing test file: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/tests/unit_tests/utils/test_instantiate_utils.py

ananthsub: this file can be removed now

ananthsub (approved):

PR #1289, opened 2025-11-11
fix: set default value for `cuda_graph_impl`

HollowMan6:
# What does this PR do ?

Set default value for `cuda_graph_impl` according to https://github.com/NVIDIA/Megatron-LM/blob/1f6cde85d23ff0c307a47bbdd8bfd778b95a161f/megatron/core/transformer/transformer_config.py#L646, as this is needed in:

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/b94ef9c73a95de8655da8347b999413183d423eb/src/megatron/bridge/models/gpt_provider.py#L199

So that users don't need to pass it manually to the provider everytime if they don't want to set it

```log
  File "megatron/bridge/models/model_provider.py", line 179, in provide_distributed_model
    model = get_model(
            ^^^^^^^^^^
  File "megatron/bridge/models/model_provider.py", line 524, in get_model
    model = _create_model(model_provider, model_type)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "megatron/bridge/models/model_provider.py", line 630, in _create_model
    model = model_provider.provide(
            ^^^^^^^^^^^^^^^^^^^^^^^
  File "megatron/bridge/models/gpt_provider.py", line 199, in provide
    if self.cuda_graph_impl != "none":
       ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Qwen2ModelProvider' object has no attribute 'cuda_graph_impl'. Did you mean: 'cuda_graph_scope'?. Did you mean: '_return_value'?
```

# Changelog

- Set default value for `cuda_graph_impl` to be `none`

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/src/megatron/bridge/models/gpt_provider.py
+++ b/src/megatron/bridge/models/gpt_provider.py
@@ -167,6 +167,7 @@ class GPTModelProvider(TransformerConfig, ModelProviderMixin[MCoreGPTModel]):
 
     # Additional parameters that might be needed
     init_model_with_meta_device: bool = False
+    cuda_graph_impl: Literal["none", "local", "te", "transformer_engine"] = "none"
     use_te_rng_tracker: bool = False
     virtual_pipeline_model_parallel_size: Optional[int] = None
     account_for_embedding_in_pipeline_split: bool = False


src/megatron/bridge/models/gpt_provider.py
```diff
@@ -167,6 +167,7 @@ class GPTModelProvider(TransformerConfig, ModelProviderMixin[MCoreGPTModel]):
 
     # Additional parameters that might be needed
     init_model_with_meta_device: bool = False
+    cuda_graph_impl: Literal["none", "local", "te", "transformer_engine"] = "none"
```
ananthsub, line 170: could you explain what error you're seeing? this is the same as the default value here: https://github.com/NVIDIA/Megatron-LM/blob/56019e6376934aac077f593607a9517697e09682/megatron/core/transformer/transformer_config.py#L685

and since GPTModelProvider inherits from TransformerConfig, we shouldn't need to redefine this field here

HollowMan6: I just checked again and found that actually my local mcore source code doesn't have this `cuda_graph_impl` defined, and it looks like this commit hasn't been tagged by any version: https://github.com/NVIDIA/Megatron-LM/commit/5769804591c2c72b4173541631ec4baf75f0dfdd

ananthsub: this will be available in the next megatron core v0.15 release

HollowMan6:
> this will be available in the next megatron core v0.15 release

Ok, sounds reasonable, I'll close this for now.

PR #1291, opened 2025-11-11
guard trust_remote_code

dimapihtar:
# What does this PR do ?

1. Avoids direct usage of `trust_remote_code=True` for security reasons.
2. Introduces "safe list" of HF directories we can trust using `trust_remote_code=True`.
3. Introduces `is_safe_repo` function which handles `trust_remote_code` given by user & safe list of HF repos.

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/examples/conversion/hf_megatron_roundtrip_multi_gpu.py
+++ b/examples/conversion/hf_megatron_roundtrip_multi_gpu.py
@@ -84,8 +84,8 @@ def main(
         trust_remote_code=is_safe_repo(
             trust_remote_code=trust_remote_code,
             hf_path=hf_model_id,
-            torch_dtype=torch.bfloat16,
         ),
+        torch_dtype=torch.bfloat16,
     )
 
     if megatron_load_path:


README.md
```diff
@@ -51,7 +51,8 @@ Conversion-only quickstart (✅ Core):
 from megatron.bridge import AutoBridge
 
 # 1) Create a bridge from a Hugging Face model (hub or local path)
-bridge = AutoBridge.from_hf_pretrained("meta-llama/Llama-3.2-1B", trust_remote_code=True)
+# Set trust_remote_code to True only if you trust the repository
+bridge = AutoBridge.from_hf_pretrained("meta-llama/Llama-3.2-1B", trust_remote_code=trust_remote_code)
```
ananthsub: for examples, isn't it fine if the user explicitly opts into this? why do we need this here? trust_remote_code isn't defined anywhere

docs/adding-new-models.md
```diff
@@ -183,7 +183,8 @@ A minimal bidirectional end-to-end check:
 from megatron.bridge import AutoBridge
 
 # HF → Megatron
-bridge = AutoBridge.from_hf_pretrained("<org>/<model-id>", trust_remote_code=True)
+# Set trust_remote_code to True only if you trust the repository
+bridge = AutoBridge.from_hf_pretrained("<org>/<model-id>", trust_remote_code=trust_remote_code)
```
ananthsub: same here

docs/adding-new-models.md
```diff
@@ -183,7 +183,8 @@ A minimal bidirectional end-to-end check:
 from megatron.bridge import AutoBridge
 
 # HF → Megatron
-bridge = AutoBridge.from_hf_pretrained("<org>/<model-id>", trust_remote_code=True)
+# Set trust_remote_code to True only if you trust the repository
```
ananthsub: this doc needs to be added for updating the allow list when adding a new model to the bridge

examples/conversion/compare_hf_and_megatron/compare.py
```diff
@@ -203,7 +203,7 @@ def get_model_class(model_class_name: str = None, is_vl_model: bool = False):
         return AutoModelForCausalLM
 
 
-def is_vision_language_model(model_path: str) -> bool:
+def is_vision_language_model(model_path: str, trust_remote_code: bool = False) -> bool:
```
ananthsub: in these and other cases, does it make sense to have bool | None, such that we always opt into trust_remote_code for supported model repos without extra user input?

src/megatron/bridge/utils/common_utils.py
```diff
@@ -39,6 +39,20 @@
...
+    "Qwen",
+    "deepseek-ai",
+    "meta-llama",
+    "google",
+    "openai",
+    "mistralai",
+    "moonshotai",
+    "llava-hf",
+    "gpt2",
+]
```
ananthsub: common_utils is too broad for this. i'd recommend creating a utils file here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/src/megatron/bridge/models/hf_pretrained

tests/functional_tests/models/deepseek/test_deepseek_conversion.py
```diff
@@ -56,7 +58,10 @@ def deepseek_toy_model_path(self, tmp_path_factory):
         model_dir = temp_dir / "deepseek_toy"
 
         # Create a minimal config and model using Auto classes to avoid direct imports
-        config = AutoConfig.from_pretrained("deepseek-ai/DeepSeek-V3", trust_remote_code=True)
+        config = AutoConfig.from_pretrained(
+            "deepseek-ai/DeepSeek-V3",
+            trust_remote_code=is_safe_repo(hf_path="deepseek-ai/DeepSeek-V3"),
```
ananthsub: i dont get how this is an improvement for security..could you explain why its needed here?

README.md
dimapihtar: > for examples, isn't it fine if the user explicitly opts into this? why do we need this here? trust_remote_code isn't defined anywhere

I remember @ericharper mentioned that it's not good we have `trust_remote_code=True` in readme files as well. Or do you want to remove this param from readme?

tests/functional_tests/models/deepseek/test_deepseek_conversion.py
dimapihtar: I think it's fine have this function usage in the tests as well since we introduced it in MBridge.

maanug-nv: I'd argue that this is not necessary in tests since we already use HF_HUB_OFFLINE=1 for all tests.

tests/functional_tests/models/glm/test_glm45_conversion.py
```diff
@@ -176,7 +178,10 @@ def test_toy_model_creation(self, glm45_toy_model_path):
                 glm45_toy_model_path,
                 torch_dtype=torch.bfloat16,
                 low_cpu_mem_usage=False,  # Ensure full loading
-                trust_remote_code=True,
+                trust_remote_code=is_safe_repo(
+                    trust_remote_code=True,
```
maanug-nv: I don't see the point of calling the function with `trust_remote_code=True` . It's just going to return True anyway.

Also it doesn't make sense to check toy model paths. These are local dirs created by the test itself. They won't fit the expected '<org>/<model_id>' structure, since `glm45_toy_model_path` here is a local directory.

src/megatron/bridge/utils/common_utils.py
```diff
@@ -264,3 +278,25 @@ def extract_expert_number_from_param(param_name: str) -> int:
...
+        True if remote code execution is allowed; False otherwise.
+    """
+    if trust_remote_code is not None:
+        return trust_remote_code
+
+    hf_repo = hf_path.split("/")[0]
+    if hf_repo in SAFE_REPOS:
+        return True
+
+    return False
```
maanug-nv: Probably should log a warning if returning False.

README.md
maanug-nv: I don't think `trust_remote_code=True` is necessary for Llama-3.2-1B . It should be in `transformers`, at least for the version we require.

src/megatron/bridge/training/config.py
```diff
@@ -231,6 +231,9 @@ class DataloaderConfig:
     persistent_workers: bool = False
     """Whether to keep data loading workers persistent across epochs."""
 
+    trust_remote_code: bool = None
```
maanug-nv: `Optional[bool]`

src/megatron/bridge/models/hf_pretrained/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    "google",
+    "openai",
+    "mistralai",
+    "moonshotai",
+    "llava-hf",
+    "gpt2",
+]
+
+
+def is_safe_repo(trust_remote_code: bool | None = None, hf_path: str = None) -> bool:
```
ananthsub:
```suggestion
def is_safe_repo(hf_path: str, trust_remote_code: bool | None) -> bool:
```

why is hf_path optional?

imo it makes more sense for the default for `trust_remote_code` in the other callsites in the repo to be `None`. then if the `trust_remote_code` is None and the model path passes the allow list in `is_safe_repo`, `is_safe_repo` returns True. otherwise, we return back whatever `trust_remote_code` is.

src/megatron/bridge/models/hf_pretrained/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    Determine whether remote code execution should be trusted for a given
+    Hugging Face repository path.
+
+    Args:
+        trust_remote_code (bool): whther to define repo as safe w/o checking SAFE_REPOS.
```
ananthsub: this comment reads more like an implementation detail for how this check is implemented. it won't be helpful to users on the API docs

src/megatron/bridge/models/hf_pretrained/utils.py
```diff
@@ -0,0 +1,59 @@
...
+        hf_path (str): path to HF's model or dataset.
+
+    Returns:
+        True if remote code execution is allowed; False otherwise.
+    """
+    if trust_remote_code is not None:
+        if trust_remote_code is False:
+            logger.warning(
+                "`trust_remote_code=False`. Remote code may not be executed. "
+                "Set `trust_remote_code=True` only if you fully trust the Hugging Face repository."
```
ananthsub, line 62: doesn't it make sense to log this warning if `trust_remote_code=True` instead ?

src/megatron/bridge/models/hf_pretrained/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    "Qwen",
+    "deepseek-ai",
+    "meta-llama",
+    "google",
+    "openai",
+    "mistralai",
+    "moonshotai",
+    "llava-hf",
+    "gpt2",
+]
```
ananthsub: super nit: keep alphabetized

```suggestion
SAFE_REPOS: list[str] = [
    "deepseek-ai",
    "gpt2",
    "google",
    "llava-hf",
    "meta-llama",
    "mistralai",
    "moonshotai",
    "nvidia",
    "openai",
    "Qwen",
]
```

examples/conversion/hf_megatron_roundtrip_multi_gpu.py
```diff
@@ -63,6 +64,7 @@ def main(
     etp: int = 1,
     megatron_save_path: str | None = None,
     megatron_load_path: str | None = None,
+    trust_remote_code: bool | None = False,
```
ananthsub:
```suggestion
    trust_remote_code: bool | None = None,
```
if we set this to None, then users don't need to set anything else as long as the path is from an allowed organization

examples/conversion/hf_megatron_roundtrip.py
```diff
@@ -36,21 +36,28 @@
...
 from megatron.bridge.models.conversion import weights_verification_table
+from megatron.bridge.models.hf_pretrained.utils import is_safe_repo
 
 
 console = Console()
 HF_MODEL_ID = "meta-llama/Llama-3.2-1B"
 
 
-def main(hf_model_id: str = HF_MODEL_ID, output_dir: str = None) -> None:
+def main(hf_model_id: str = HF_MODEL_ID, output_dir: str = None, trust_remote_code: bool | None = False) -> None:
```
ananthsub:
```suggestion
def main(hf_model_id: str = HF_MODEL_ID, output_dir: str = None, trust_remote_code: bool | None = None) -> None:
```

examples/conversion/compare_hf_and_megatron/compare.py
```diff
@@ -203,7 +204,7 @@ def get_model_class(model_class_name: str = None, is_vl_model: bool = False):
         return AutoModelForCausalLM
 
 
-def is_vision_language_model(model_path: str) -> bool:
+def is_vision_language_model(model_path: str, trust_remote_code: bool | None = False) -> bool:
```
ananthsub:
```suggestion
def is_vision_language_model(model_path: str, trust_remote_code: bool | None = None) -> bool:
```

src/megatron/bridge/models/hf_pretrained/utils.py
dimapihtar: I'd warn only if `trust_remote_code=False` because it will may user know why code was not executed (HF error is not so clear).

maanug-nv: i agree with dmytro. trust_remote_code should be always opt-in at the user's level now with this PR. if a user is already deciding to trust remote code, they are aware already.

ananthsub (approved):

PR #1416, opened 2025-11-19
Update performance scripts to align with llmb requirements

scsudhakaran:
This PR updates performance scripts to add the following features

* Support for additional Slurm parameters.
* `nsys_extra_args` attribute to the `NsysPlugin`
* Nemotron-H 56B B200 recipe

### DIFF
--- a/scripts/performance/argument_parser.py
+++ b/scripts/performance/argument_parser.py
@@ -450,6 +450,7 @@ def parse_cli_args():
     args, cli_dotlist_overrides = parser.parse_known_args()
     return args, cli_dotlist_overrides
 
+
 def parse_additional_slurm_params(params_str):
     """
     Parse additional SLURM parameters from a string of key=value pairs.
@@ -475,10 +476,10 @@ def parse_additional_slurm_params(params_str):
     params = {}
 
     # Try semicolon separation first (most reliable for complex values)
-    if ';' in params_str:
-        parts = params_str.split(';')
+    if ";" in params_str:
+        parts = params_str.split(";")
     # Try space separation next
-    elif ' ' in params_str:
+    elif " " in params_str:
         parts = params_str.split()
     # No separators found - treat as single parameter
     else:
@@ -489,8 +490,8 @@ def parse_additional_slurm_params(params_str):
         if not part:
             continue
 
-        if '=' in part:
-            key, value = part.split('=', 1)
+        if "=" in part:
+            key, value = part.split("=", 1)
             params[key.strip()] = value.strip()
         else:
             # Boolean flag (no value)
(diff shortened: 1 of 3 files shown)


scripts/performance/setup_experiment.py
malay-nagda, line 1: we should be using `profiling_start_step` and end step from args where we instantiate `NsysPlugin` here.

scripts/performance/argument_parser.py
```diff
@@ -418,6 +418,76 @@ def parse_cli_args():
...
+    parser.add_argument(
+        "--profiling_stop_step", type=int, help="Defines stop step for profiling", required=False, default=11
+    )
 
+    parser.add_argument(
+        "-pgm",
+        "--profiling_gpu_metrics",
+        help="Enable nsys gpu metrics. Disabled by default.",
+        action="store_true",
+    )
```
malay-nagda, line 440: this arg is not used anywhere. do we need it?

scripts/performance/perf_plugins.py
```diff
@@ -116,6 +117,13 @@ def setup(self, task: Union["run.Partial", "run.Script"], executor: "run.Executo
         launcher = executor.get_launcher()
         launcher.nsys_profile = True
         launcher.nsys_trace = self.nsys_trace or ["nvtx", "cuda"]
+        launcher.nsys_extra_args = self.nsys_extra_args or [
+            "--force-overwrite=true",
+            "--capture-range=cudaProfilerApi",
+            "--capture-range-end=stop",
+            "--cuda-graph-trace=node",
+            "--cuda-event-trace=false",
+        ]
```
malay-nagda: I think this can just be 
```
launcher.nsys_extra_args = self.nsys_extra_args or launcher.nsys_extra_args
```
given `launcher.nsys_extra_args` is exactly same as https://github.com/NVIDIA-NeMo/Run/blob/main/nemo_run/core/execution/launcher.py#L34-L39

scripts/performance/setup_experiment.py
scsudhakaran: done

scripts/performance/argument_parser.py
scsudhakaran: Yes, it is used during profiling.

scripts/performance/perf_plugins.py
scsudhakaran: done

scripts/performance/argument_parser.py
malay-nagda: gotcha! I see you added it and use it in new commits.

malay-nagda (approved):

malay-nagda (dismissed):

PR #1441, opened 2025-11-20
feat: mcore trigger mbridge

pablo-garay:
Adapts Megatron-Bridge so that tests can be run from : megatron-lm (mcore)

image

image

image

Tests have passed:
image

image

### DIFF
--- a/tests/unit_tests/Launch_Unit_Tests.sh
+++ b/tests/unit_tests/Launch_Unit_Tests.sh
@@ -16,17 +16,13 @@
 set -xeuo pipefail # Exit immediately if a command exits with a non-zero status
 
 echo "=================================================="
-echo "🧪 UNIT TESTS - MCore Commit Information"
+echo "🧪 UNIT TESTS"
 echo "=================================================="
 
-# Display MCore commit SHA if available
-MCORE_SHA_FILE="/opt/Megatron-Bridge/.mcore_commit_sha"
-if [ -f "${MCORE_SHA_FILE}" ]; then
-    MCORE_COMMIT=$(cat "${MCORE_SHA_FILE}")
-    echo "📦 MCore commit SHA: ${MCORE_COMMIT}"
+# Display MCore commit SHA if triggered from MCore CI
+if [ -f "/opt/Megatron-Bridge/.mcore_commit_sha" ]; then
+    echo "📦 MCore commit: $(cat /opt/Megatron-Bridge/.mcore_commit_sha)"
 fi
-
-echo "=================================================="
 echo ""
 
 CUDA_VISIBLE_DEVICES="0,1" uv run coverage run -a --data-file=/opt/Megatron-Bridge/.coverage --source=/opt/Megatron-Bridge/ -m pytest \


.github/workflows/cicd-main.yml
```diff
@@ -20,6 +20,32 @@ on:
...
+  workflow_dispatch:
+    inputs:
+      mcore_commit:
+        description: 'MCore commit SHA to test against'
+        required: false
+        type: string
+      mcore_branch:
+        description: 'MCore branch name (for reference)'
+        required: false
+        type: string
```
ko3n1g, line 34: What is this input for?

pablo-garay: 1. Set in Mcore:
- name: Get MCore commit info
  id: mcore_info
  run: |
    MCORE_BRANCH=${GITHUB_REF#refs/heads/}
    echo "branch=$MCORE_BRANCH" >> $GITHUB_OUTPUT
    
2. Passed to MBridge
gh workflow run install-test.yml \
  --field mcore_commit=${{ steps.mcore_info.outputs.sha }} \
  --field mcore_branch=${{ steps.mcore_info.outputs.branch }}  # ← Sent here

3. Received by MBridge (as workflow input)
mcore_branch:
  description: 'MCore branch name (for reference)'
  required: false
  type: string

4. Used in MBridge Logs (for informational purposes useful to the user)
echo "📌 MCore branch: ${{ github.event.inputs.mcore_branch || 'unknown' }}"

pablo-garay: Current Usage: Logging Only (optional, its very helpful to the user to know)
That's it . It's only used for displaying in logs. It doesn't affect:
 Which code gets tested / Which tests run / Any conditional logic . It's pre-computed & pre-populated by mcore - as its value is passed from mcore

pablo-garay:
> I'm not convinced the git submodule bump would be properly reflected. Can you think of a way to verify that during testing the right MCore commit is being used?

For megatron-bridge i added checks . To compare:
ACTUAL_COMMIT against EXPECTED_COMMIT

image

image

image

they are all matching as seen in screenshots attached. If they werent matching/dont match , the job would fail instantly 

so its working as expected :)

pablo-garay:
> I'm not convinced the git submodule bump would be properly reflected. Can you think of a way to verify that during testing the right MCore commit is being used?

Yes I set comparisons in place in each place: unit tests, functional tests, install tests and also after container is built. Verification compares the input megatron-lm version (to be used) with the submodule's used version

Tested & all checks have verified (matched, exact commit hash matched)

.github/actions/test-template/action.yml
```diff
@@ -137,6 +152,117 @@ runs:
...
+            HAS_CHANGES=0
+            if ! git diff-index --quiet HEAD --; then
+              echo "   ⚠️  WARNING: Working directory has uncommitted changes!"
+              git status --short
+              echo "   Modified files:"
+              git diff HEAD --name-only
+              HAS_CHANGES=1
+            else
+              echo "   ✅ Working directory is clean (no uncommitted changes)"
+            fi
```
ko3n1g: Why would a freshly cloned repo have uncommitted changes?

.github/actions/test-template/action.yml
```diff
@@ -137,6 +152,117 @@ runs:
...
+            
+            # Check file hashes at source
+            echo "   Source file hashes (from working directory):"
+            for file in megatron/core/__init__.py megatron/core/package_info.py pyproject.toml; do
+              if [ -f "${file}" ]; then
+                HASH=$(sha256sum "${file}" | awk '\''{print substr($1,1,16)}'\'')
+                FULL_HASH=$(sha256sum "${file}" | awk '\''{print $1}'\'')
+                echo "     ${file}: ${HASH}... (full: ${FULL_HASH})"
+              fi
+            done
```
ko3n1g: I think we can trust `git rev-parse HEAD`. Let's add additional checks beyond that once we see that its necessary

.github/actions/test-template/action.yml
pablo-garay: This was for testing purposes  just to check since was having issues before

.github/actions/test-template/action.yml
pablo-garay: I added several checks & spent a really long time making sure checks are fine . Sounds good, we remove them for now & add beyond that if seen necessary later in the future

docker/Dockerfile.ci
```diff
@@ -30,13 +30,24 @@ COPY src/megatron/bridge/__init__.py src/megatron/bridge/package_info.py /opt/Me
...
+        uv sync --link-mode copy --all-extras --all-groups; \
+    else \
+        uv sync --locked --only-group build && \
+        uv sync --link-mode copy --locked --all-extras --all-groups; \
+    fi && \
     uv cache prune
+
+# Store MCore commit SHA for verification
+RUN echo "${MCORE_COMMIT_SHA}" > /opt/Megatron-Bridge/.mcore_commit_sha && \
+    echo "📝 Stored MCore commit SHA: ${MCORE_COMMIT_SHA}"
```
ko3n1g, line 53: Please remove this in a follow-up PR

ko3n1g (commented):
Thanks, @pablo-garay. There's one additional cleanup ask, but feel free to submit that with a follow-up PR

PR #1447, opened 2025-11-21
feat: Add whitelist support for mismatched params in load_hf_weights

yaoyu-33:
This PR introduces a whitelist mechanism for `load_hf_weights` to allow skipping specific parameters that have shape mismatches during HuggingFace to Megatron conversion. This is particularly useful when dealing with models that have known structural differences or when intentionally ignoring certain weights (e.g., specific biases or auxiliary heads) that would otherwise cause the loading process to fail with a `ValueError`.

### Key Changes
*   **Core Logic**: Modified `load_weights_hf_to_megatron` in `MegatronModelBridge` (both in `models/model_bridge.py` and `models/conversion/model_bridge.py`) to accept a `mismatch_params_whitelist` argument.
*   **Pattern Matching**: Implemented `fnmatch`-based matching, allowing users to specify exact parameter names or glob patterns (e.g., `"*.bias"`, `"layers.0.*"`) in the whitelist.
*   **AutoBridge Integration**: Exposed `mismatch_params_whitelist` argument in `AutoBridge.load_hf_weights` to pass it down to the underlying model bridge.
*   **Behavior**: Mismatched parameters found in the whitelist are skipped with a warning log (on rank 0) instead of raising a `ValueError`.

### Testing
*   Added a new unit test `test_load_hf_weights_with_whitelist` in `tests/unit_tests/models/test_auto_bridge.py` to verify that the whitelist is correctly passed and handled.
*   Updated existing tests in `test_auto_bridge.py` to align with the updated method signature.

### Usage Example
```python
# Load weights, skipping specific mismatched parameters
bridge.load_hf_weights(
    megatron_model,
    mismatch_params_whitelist=[
        "decoder.layers.0.self_attention.linear_qkv.weight", 
        "*.bias"

### DIFF
--- a/src/megatron/bridge/models/conversion/auto_bridge.py
+++ b/src/megatron/bridge/models/conversion/auto_bridge.py
@@ -274,7 +274,7 @@ class AutoBridge(Generic[MegatronModelT]):
         self,
         model: list[MegatronModelT],
         hf_path: str | Path | None = None,
-        mismatch_params_whitelist: list[str] | None = None,
+        allowed_mismatched_params: list[str] | None = None,
     ) -> None:
         """
         Load HuggingFace weights into a Megatron model.
@@ -287,7 +287,7 @@ class AutoBridge(Generic[MegatronModelT]):
             model: List of Megatron model instances (one per virtual pipeline stage)
             hf_path: Optional path to load weights from. If None, uses weights
                 from the bridge's hf_pretrained instance
-            mismatch_params_whitelist: Optional list of parameter names or patterns
+            allowed_mismatched_params: Optional list of parameter names or patterns
                 to allow mismatch (skip instead of raise error).
 
         Returns:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/megatron/bridge/models/conversion/auto_bridge.py [hunk 1]
```diff
@@ -282,6 +287,8 @@ def load_hf_weights(self, model: list[MegatronModelT], hf_path: str | Path | Non
             model: List of Megatron model instances (one per virtual pipeline stage)
             hf_path: Optional path to load weights from. If None, uses weights
                 from the bridge's hf_pretrained instance
+            mismatch_params_whitelist: Optional list of parameter names or patterns
```
ananthsub: nit: it could be helpful to add an example for the supported pattern here

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -270,7 +270,12 @@ def can_handle(cls, path: Union[str, Path], trust_remote_code: bool = False) ->
         except Exception:
             return False
 
-    def load_hf_weights(self, model: list[MegatronModelT], hf_path: str | Path | None = None) -> None:
+    def load_hf_weights(
+        self,
+        model: list[MegatronModelT],
+        hf_path: str | Path | None = None,
+        mismatch_params_whitelist: list[str] | None = None,
```
ananthsub:
```suggestion
        allowed_mismatched_params: list[str] | None = None,
```
nit: naming might be clearer like this?

ananthsub (approved):

chtruong814:
Overall test coverage healthy.  Moving forward to merge this.

PR #1451, opened 2025-11-21
Added Qwen 3 next perf scripts

sanandaraj5597:
Add Qwen3 perf scripts to main.

### DIFF
--- a/scripts/performance/configs/qwen3/qwen3_llm_pretrain.py
+++ b/scripts/performance/configs/qwen3/qwen3_llm_pretrain.py
@@ -20,6 +20,7 @@ from utils.helpers import (
 )
 
 from megatron.bridge.recipes.qwen.qwen3_moe import qwen3_30b_a3b_pretrain_config, qwen3_235b_a22b_pretrain_config
+from megatron.bridge.recipes.qwen.qwen3_next import qwen3_next_80b_a3b_pretrain_config
 from megatron.bridge.training.comm_overlap import CommOverlapConfig
 from megatron.bridge.training.config import ConfigContainer
 from megatron.bridge.training.utils.moe_token_drop import apply_moe_token_drop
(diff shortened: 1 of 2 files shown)


scripts/performance/configs/qwen3/qwen3_llm_pretrain.py
```diff
@@ -217,3 +233,61 @@ def qwen3_30b_a3b_h100_config(precision: str = "bf16") -> ConfigContainer:
...
+
+    cfg = qwen3_next_80b_a3b_pretrain_config(
+        mock=True,
+        precision_config=precision_config,
+        comm_overlap_config=CommOverlapConfig(tp_comm_overlap=True),
+    )
+    set_qwen3_next_common_configs(cfg)
+    set_workload_base_configs(cfg, base_cfg)
+
+    return cfg
```
malay-nagda, line 292: Are you going to add B200 config as well?

scripts/performance/configs/qwen3/qwen3_llm_pretrain.py
```diff
@@ -217,3 +233,61 @@ def qwen3_30b_a3b_h100_config(precision: str = "bf16") -> ConfigContainer:
     set_workload_base_configs(cfg, base_cfg)
 
     return cfg
+
+def qwen3_next_80b_a3b_gb200_config(precision: str = "bf16") -> ConfigContainer:
```
malay-nagda, line 237: we need to remove `_next_` from these methods as [get_model_recipe](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/scripts/performance/utils/utils.py#L70) would fail otherwise

scripts/performance/configs/qwen3/workload_base_configs.py
```diff
@@ -212,6 +217,51 @@
 )
 
 
+# Qwen3 Next Presets --------------------------------------
+
+QWEN3_NEXT_80B_A3B_GB200_FP8_MX_BASE_CONFIG = replace(
```
malay-nagda, line 222: same as the methods-
we need to remove `_NEXT_` from these names as [get_workload_base_config](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/scripts/performance/utils/utils.py#L97) would fail otherwise

scripts/performance/configs/qwen3/qwen3_llm_pretrain.py
sanandaraj5597: The tracker has just GB300, GB200 and H100. So I skipped B200.

PR #1462, opened 2025-11-21
Add falcon h1

dhiaEddineRhaiem:
This PR adds support for FalconH1 in MegatronBridge.
FalconH1 is introduced as a new ParallelHybrid layer in the hybrid design of MegatronBridge.
cc @sbhavani

## Release Notes

* **New Features**
  * Added support for FalconH1 model family with multiple sizes (500M, 1.5B, 7B, 34B parameters)
  * Enabled configurable hybrid Mamba/attention/MLP layer architectures for flexible model composition
  * Added HuggingFace pretrained model integration for seamless loading and conversion


src/megatron/bridge/models/falcon_h1/modeling_falconh1/falconh1_block.py
```diff
@@ -0,0 +1,460 @@
+# Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
```
yaoyu-33, line 1: can we create a "modeling_faconh1" folder and move all modeling files into "modeling_faconh1" folder, block, layer, model, mamba_hybrid.

Top level will have "modeling_faconh1" folder + "faconh1_provider" + "faconh1_bridge"

dhiaEddineRhaiem: done! @yaoyu-33

yaoyu-33:
@dhiaEddineRhaiem  thanks a lot for the contribution. 
Overall it looks very good!
We will need to verify the conversion between hf / megatron. Can you check https://docs.nvidia.com/nemo/megatron-bridge/latest/adding-new-models.html#validate-with-examples and maybe attach you validation results here?

dhiaEddineRhaiem:
hello @yaoyu-33 ,

 this the [output](https://gist.github.com/dhiaEddineRhaiem/f798edd7624cd49f450f4db8cf71fc3e) of the hf conversion of falcon h1 after the previous fixes but before adding [Falcon h1 Mup Fwd multipliers](https://huggingface.co/tiiuae/Falcon-H1-7B-Instruct/blob/main/config.json#L51). 
 
 For reference , here is the complete list of FalconH1 MuP multipliers (from [config](https://huggingface.co/tiiuae/Falcon-H1-0.5B-Instruct/blob/main/config.json)):

embedding_multiplier:  scales embedding outputs
lm_head_multiplier:scales final logits
attention_in_multiplier: scales attention input
attention_out_multiplier: scales attention output
key_multiplier: scales K projection ([ref](https://github.com/huggingface/transformers/blob/b59386dc0a44eef0b6c671ed6ed09a76b75235e8/src/transformers/models/falcon_h1/modeling_falcon_h1.py#L373))
mlp_multipliers:  gate/down projections ([ref](https://github.com/huggingface/transformers/blob/b59386dc0a44eef0b6c671ed6ed09a76b75235e8/src/transformers/models/falcon_h1/modeling_falcon_h1.py#L1037))
ssm_in_multiplier: scales SSM input
ssm_out_multiplier: cales SSM output
ssm_multipliers: z/x/B/C/dt components ([ref](https://github.com/huggingface/transformers/blob/b59386dc0a44eef0b6c671ed6ed09a76b75235e8/src/transformers/models/falcon_h1/modeling_falcon_h1.py#L615))

So the gibberish output was expected because we were not applying [Falcon h1 Mup Fwd multipliers](https://huggingface.co/tiiuae/Falcon-H1-7B-Instruct/blob/main/config.json#L51)(MLP , attention and ssm multipliers) yet. 

I managed to add some of them in bridge except [MLP gate and down proj multipliers](https://huggingface.co/tiiuae/Falcon-H1-34B-Instruct/blob/main/config.json#L36) , [ssm_multipliers](https://huggingface.co/tiiuae/Falcon-H1-34B-Instruct/blob/main/config.json#L51), [key_multiplier](https://huggingface.co/tiiuae/Falcon-H1-34B-Instruct/blob/main/config.json#L18) which i think should be added to core in the fwd of [MambaMixer](https://github.com/NVIDIA/Megatron-LM/blob/8daf046f0dacbd80d968d483dd00d2df160a78e7/megatron/core/ssm/mamba_mixer.py#L413) , [MLP](https://github.com/NVIDIA/Megatron-LM/blob/8daf046f0dacbd80d968d483dd00d2df160a78e7/megatron/core/transformer/mlp.py#L140) and [selfAttention](https://github.com/NVIDIA/Megatron-LM/blob/8daf046f0dacbd80d968d483dd00d2df160a78e7/megatron/core/transformer/attention.py#L779). 
More info on how these multipliers are being applied could be found here 

1. [MLP gate and down proj](https://github.com/huggingface/transformers/blob/b59386dc0a44eef0b6c671ed6ed09a76b75235e8/src/transformers/models/falcon_h1/modeling_falcon_h1.py#L1037)

2.[ssm_multipliers](https://github.com/huggingface/transformers/blob/b59386dc0a44eef0b6c671ed6ed09a76b75235e8/src/transformers/models/falcon_h1/modeling_falcon_h1.py#L615)

3.[key_multiplier](https://github.com/huggingface/transformers/blob/b59386dc0a44eef0b6c671ed6ed09a76b75235e8/src/transformers/models/falcon_h1/modeling_falcon_h1.py#L373)

so IMO , they should be add to core in the fwd of MambaMixer , MLP and self attention. 

Any help on this please ?

Separius:
Hi @dhiaEddineRhaiem, how bad would it be to fuse the multipliers into the weights (during conversion)? I'm guessing changing the mcore is a no go, will the training break or will it be just worse? or can it be accounted for in lr? can we multiply the learning rate of those fused weights with 1/multiplier and get the same updates? (I think wd needs to be changed too)

sbhavani:
@dhiaEddineRhaiem thanks for adding the muP multipliers. I think this is a good approach while we figure out general muP support in Megatron Core. 

One potential bug in `falconh1_layer.py`: The attention out multiplier overwrites the tensor:

`attn_output = self.config.attention_out_multiplier`

yaoyu-33:
@dhiaEddineRhaiem : sry for a bit late in reply.
MCore also accept PRs. These configs doesn't seem very intrusive so we can add it in mcore.
Another way we can do is, if we want to merge faster, we can override some the mixer class / mlp class directly in mcore. Example is https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl where we have to do deepstack etc. for qwen3vl

yaoyu-33 (dismissed):

gautham-kollu:
@dhiaEddineRhaiem  Can you please follow the instructions here https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-linting-and-formatting  to fix linting ?

yaoyu-33:
Falcon H1 takeover update pushed to this PR branch.

What changed:
- Rebased the Falcon H1 support onto latest `origin/main` and pushed via a non-force merge commit to preserve the existing PR branch ancestry.
- Kept Falcon H1-specific fixes in the Bridge side rather than modifying `3rdparty/Megatron-LM`.
- Applied Falcon H1 MuP/scaling multipliers in the active hybrid layer path:
  - `key_multiplier` in Falcon H1 attention
  - `mlp_multipliers` in the Falcon H1 MLP path
  - `ssm_multipliers` in the Falcon H1 Mamba/SSM path
  - fixed `attention_out_multiplier` so it scales the attention tensor instead of overwriting it
  - fixed `lm_head_multiplier` so it scales unpacked logits instead of the output-layer tuple
- Updated Falcon H1 RoPE config handling for current Transformers compatibility.
- Ran formatting/lint via `uv run pre-commit run --all-files` after the latest-main integration.

Recommended next step:
- Run CI on the pushed head and complete reviewer sign-off for the Falcon H1 support.

src/megatron/bridge/models/__init__.py
```diff
@@ -32,6 +32,14 @@
     DeepSeekV2Bridge,
     DeepSeekV3Bridge,
 )
+from megatron.bridge.models.falcon_h1 import (
```
yaoyu-33, line 35: we will need example folder with conversion / inference / readme to cover the correctness check of the model.
Also missing necessary tests, unit test at least required.

src/megatron/bridge/models/__init__.py
```diff
@@ -32,6 +32,14 @@
...
+    FalconH1Bridge,
+    FalconH1ModelProvider,
+    FalconH1ModelProvider1P5BDeep,
```
yaoyu-33: we no longer provide size/version specific configs. should remove/ clean up .

src/megatron/bridge/models/falcon_h1/falconh1_bridge.py
```diff
@@ -0,0 +1,183 @@
...
+    Example:
+        >>> from megatron.bridge import AutoBridge
+        >>> bridge = AutoBridge.from_hf_pretrained("tiiuae/Falcon-H1-7B-Instruct", trust_remote_code=True)
+        >>> provider = bridge.to_megatron_provider()
+    """
+
+    def provider_bridge(self, hf_pretrained: PreTrainedCausalLM) -> FalconH1ModelProvider:
+        hf_config = hf_pretrained.config
+
+        return FalconH1ModelProvider(
```
yaoyu-33: should use super config mapping and additionally override if necessary for config part, check other models implementation.

yaoyu-33:
Addressed the latest Falcon H1 review feedback in 6908cb6e91ed5ff9d06c3abc8353ec95fc9cc451.

Changes:
- Added `examples/models/falcon_h1` with README, conversion wrapper, and inference wrapper for HF -> Megatron conversion, roundtrip export, and generation workflows.
- Added focused Falcon H1 unit tests covering bridge/provider config behavior, public exports, pipeline pre/post preservation, vocab padding, layer allocation, and mapping assumptions.
- Removed size/version-specific Falcon H1 provider exports and kept the generic `FalconH1ModelProvider` API.
- Refactored `FalconH1Bridge` to use the superclass config/provider mapping path, with only Falcon-specific overrides applied after the shared mapping.

Note: the earlier inherited unsigned-history DCO blocker is unchanged by this patch.

examples/models/falcon_h1/conversion.sh
```diff
@@ -0,0 +1,58 @@
...
+    --torch-dtype bfloat16 \
+    --trust-remote-code
+
+uv run python examples/conversion/convert_checkpoints.py export \
+    --hf-model "$HF_MODEL_ID" \
+    --megatron-path "${MEGATRON_MODEL_PATH}/iter_0000000" \
+    --hf-path "$HF_EXPORT_PATH" \
+    --trust-remote-code
+
+uv run python - "$HF_MODEL_ID" "$HF_EXPORT_PATH" <<'PY'
```
yaoyu-33, line 40: why do we need these, clean up if not needed

examples/models/falcon_h1/README.md
```diff
@@ -0,0 +1,62 @@
...
+`TP=1, PP=1` and `NPROC_PER_NODE=1`, which is sufficient for the 0.5B model.
+
+```bash
+bash examples/models/falcon_h1/conversion.sh
+```
+
+To run the same check with a different parallelism layout:
+
+```bash
+TP=2 PP=1 NPROC_PER_NODE=2 bash examples/models/falcon_h1/conversion.sh
```
yaoyu-33: does default value TP=1 work? if not , update default value please

yaoyu-33: Addressed in f454d169c1f3bd6acfb09f2d6b19a63b3d248c3b0. The default `TP=1, PP=1, NPROC_PER_NODE=1` path was validated for `tiiuae/Falcon-H1-0.5B-Instruct`, and the README now states that this default layout works for the 0.5B model.

src/megatron/bridge/models/falcon_h1/modeling_falconh1/falconh1_block.py
```diff
@@ -0,0 +1,443 @@
...
+                        )
+                    if isinstance(layer, FalconH1Layer):
+                        hidden_states = layer(
+                            hidden_states=hidden_states,
+                            attention_mask=attention_mask,
+                            inference_context=inference_context,
+                            rotary_pos_emb=rotary_pos_emb,
+                            sequence_len_offset=sequence_len_offset,
+                        )
+                    else:  # MambaLayer
```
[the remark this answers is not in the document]
yaoyu-33: Fixed in f454d169c1f3bd6acfb09f2d6b19a63b3d248c3b0. The second dispatch branch is now `elif isinstance(layer, FalconH1Layer)`, so `TransformerLayer` instances no longer fall through into the Mamba path.

examples/models/falcon_h1/conversion.sh
yaoyu-33: Addressed in f454d169c1f3bd6acfb09f2d6b19a63b3d248c3b0. I kept the tokenizer copy and documented why it is needed: this export flow reconstructs the HF config/weights from a Megatron checkpoint, but tokenizer artifacts are not written by that config-only export path. The exported HF directory is later used as a standalone `AutoTokenizer.from_pretrained(...)` path by the inference example, so the tokenizer files need to be copied from the source HF checkpoint.

src/megatron/bridge/models/falcon_h1/__init__.py
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
yaoyu-33: Fixed in f454d169c1f3bd6acfb09f2d6b19a63b3d248c3b0.

src/megatron/bridge/models/falcon_h1/falconh1_provider.py
```diff
@@ -0,0 +1,161 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
yaoyu-33: Fixed in f454d169c1f3bd6acfb09f2d6b19a63b3d248c3b0.

yaoyu-33:
Addressed the latest Falcon H1 review summary in f454d169c1f3bd6acfb09f2d6b19a63b3d248c3b0.

Changes:
- Fixed `FalconH1Stack.forward()` dispatch so standalone `TransformerLayer` instances do not also run through the Mamba branch.
- Updated Falcon H1 copyright headers requested in review.
- Documented why the Falcon H1 conversion example copies tokenizer artifacts into the exported HF directory.
- Clarified that the default `TP=1, PP=1, NPROC_PER_NODE=1` example layout works for the 0.5B Falcon H1 checkpoint.
- Added focused Falcon H1 unit coverage for config validation, MLP multipliers, attention key scaling, and Mamba SSM multipliers.
- Removed the no-op size-specific-provider public API test.

Validation:
- `uv run pre-commit run --all-files` passed.
- Targeted Falcon H1 validation passed: unit tests, default TP=1 conversion, and default TP=1 inference flow.

Residual blocker:
- DCO is still failing from inherited unsigned history and is unchanged by this patch.

yaoyu-33 (approved):

PR #1478, opened 2025-11-24
Update performance scripts

scsudhakaran:


scripts/performance/utils/helpers.py
```diff
@@ -269,6 +269,23 @@ def set_user_overrides(recipe: ConfigContainer, kwargs: Dict[str, Any]) -> None:
...
+
+    layout_map = {
+        (1, 1): None,
+        (4, 1): [["embedding"] + ["decoder"] * 16, ["decoder"] * 16, ["decoder"] * 16, ["decoder"] * 13 + last_layer],
+        (8, 1): [["embedding"] + ["decoder"] * 8] + [["decoder"] * 8] * 6 + [["decoder"] * 5 + last_layer],
+        (4, 2): [["embedding"] + ["decoder"] * 8] + [["decoder"] * 8] * 6 + [["decoder"] * 5 + last_layer],
+        (16, 1): [["embedding"] + ["decoder"] * 4] + [["decoder"] * 4] * 14 + [["decoder"] + last_layer],
+        (8, 2): [["embedding"] + ["decoder"] * 4] + [["decoder"] * 4] * 14 + [["decoder"] + last_layer],
+        (4, 4): [["embedding"] + ["decoder"] * 4] + [["decoder"] * 4] * 14 + [["decoder"] + last_layer],
+    }
```
bdubauski, line 288: What is the source material for this layout map? It appears to be unique to ds v3 recipe.

scsudhakaran: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/r0.2.0/src/megatron/bridge/recipes/deepseek/deepseek_v3.py#L233

We had to set this layout again because of the issue mentioned here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/1502

* Megatron-Bridge creates the recipe with default values.
* If user passes parallelism configs it overrides the recipe attributes with user given ones.
* The layout needs to be updated with the final PP-VP setting. This code block handles that.

PR #1584, opened 2025-12-04
ci: Add DGXC executor

ko3n1g:
# What does this PR do ?

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/scripts/performance/setup_experiment.py
+++ b/scripts/performance/setup_experiment.py
@@ -231,11 +231,19 @@ def main(
 
     if use_recipes:
         script_name = ENTRYPOINT_RECIPE
-        exp_name = f"{model_recipe_name.replace('_pretrain_config', '')}_{task}_{num_gpus}gpu_{gpu}"
+        exp_name = (
+            wandb_experiment_name
+            if wandb_experiment_name is not None
+            else f"{model_recipe_name.replace('_pretrain_config', '')}_{task}_{num_gpus}gpu_{gpu}"
+        )
 
     else:
         script_name = ENTRYPOINT_PEFORMANCE
-        exp_name = f"{model_recipe_name.replace('_pretrain_config', '')}_{task}_{num_gpus}gpu_{gpu}_{compute_dtype}"
+        exp_name = (
+            wandb_experiment_name
+            if wandb_experiment_name is not None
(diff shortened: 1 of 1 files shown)


scripts/performance/argument_parser.py
```diff
@@ -308,13 +367,71 @@ def parse_cli_args():
...
+        type=parse_additional_slurm_params,
         help="Additional SLURM parameters as key=value pairs. "
         "Use semicolons (;) to separate parameters when values contain commas. "
         "Examples: 'nodelist=node001,node002;constraint=gpu' or 'reservation=my_res;exclusive'",
         required=False,
     )
 
+    # DGXCloud
+    dgxc_args = parser.add_argument_group("DGXCloud arguments")
+    dgxc_args.add_argument(
```
malay-nagda: nit: 
- can we skip `--use_dgxc` arg and just use `--dgxc_cluster`/`--dgxc_base_url`- if either of these two are not `None`, we know we're using DGXC?
- or maybe even further just have `--dgxc_base_url`?

this is to as tight list of CLI args as possible- also users don't need to specify more args than necessary

scripts/performance/utils/executors.py
```diff
@@ -151,3 +151,67 @@ def slurm_executor(
...
+    custom_env_vars: List[str] = None,
+    dgxc_pvc_mount_path: str = "/nemo-workspace",
+    container_image: str = "nvcr.io/nvidia/nemo:dev",
+):
+    """
+    DGXCloud cluster definition with appropriate cluster params and NeMo container params needed for pre-training
+    and fine-tuning experiments
+    """
+
+    env_vars = {
```
malay-nagda, line 177: - better to have script level `DGXC_ENV_VARS`?
- why not add `PERF_ENV_VARS` too?

ko3n1g: I didnt think about this so far.. I think it takes a second round of review to check which perf env vars we want for DGXC too

scripts/performance/argument_parser.py
ko3n1g: `use_dgxc` we can drop.. but `dgxc_cluster` I think we need, as we add a custom spec for GCP only..

malay-nagda (approved):

PR #1596, opened 2025-12-04
Added 25.11 starter pack

sanandaraj5597:
Add 25.11 perf details.


docs/performance-summary.md
```diff
@@ -41,41 +41,56 @@ The performance data includes:
 
 ---
 
-## 25.09 NeMo Container
```
ananthsub, line 44: why was the 25.09 table removed? we should keep the history across container versions. 25.11 could have been added as a new section

ananthsub: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1667

PR #1640, opened 2025-12-08
Remove deprecated NCCL_NET_GDR_LEVEL and NCCL_NET_GDR_C2C environment variables

sudostock:
These variables were a workaround for NCCL versions older than 2.27.x and are no longer needed, with NeMo containers having 2.27x since 25.07.
- Removed setting of these variables in .
- Updated  to remove references to these variables from the 'Activation offloading to host memory' section and the 'Index - List of Tuning Knobs'.

Additionally since we aren't checking the NCCL version and blanket applying, should be removed.

### DIFF
--- a/docs/performance-guide.md
+++ b/docs/performance-guide.md
@@ -330,12 +330,7 @@ Additionally, because CP shards activations, it also partitions optimizer states
    >    > 2. `TransformerConfig.cpu_offloading_weights=False`
    >    > 3. `TransformerConfig.cpu_offloading_num_layers= <int:≤activation_offload_layers>`
    >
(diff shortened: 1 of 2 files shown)


bdubauski (approved):
Looks good to me.

malay-nagda:
> Can we add conditional checks instead of removing? We can do a torch NCCL version check.

from @sanandaraj5597 comment on corresponding NeMo PR- https://github.com/NVIDIA-NeMo/NeMo/pull/15161

sanandaraj5597:
Can we add conditional checks instead of removing? We can do a torch NCCL version check.

PR #1642, opened 2025-12-08
Support a user provided no weight decay condition

jstjohn:
# What does this PR do ?

Allow the user to supply a custom lambda as the no weight decay condition.

NOTE: this will be deprecated in megatron soon, however we are wroking on a new API to support this style of feature in the future.


jstjohn:
This PR is no longer necessary due to changes in mcore that support change weight decay condition etc in better ways.

PR #1657, opened 2025-12-09
fixed experiment folder naming convention

rsalagame-nvidia:
Fixed experiment folder naming convention where we provide values if available, otherwise we fall back to workload defaults

Previously, folder naming convention was:
**/lustre/fsw/infra_rd_gsw/rsalagame/llama3.1-405B-70B-8B-GB300-latest-test/test-env/workloads/pretrain_llama3.1/experiments/pretrain_llama3_8b_nvfp4_gpus8_tpNone_ppNone_cpNone_vpNone_epNone_mbsNone_gbsNone/pretrain_llama3_8b_nvfp4_gpus8_tpNone_ppNone_cpNone_vpNone_epNone_mbsNone_gbsNone_1765307130/pretrain_llama3_8b_nvfp4_gpus8_tpNone_ppNone_cpNone_vpNone_epNone_mbsNone_gbsNone**

After fix, folder naming convention contains the workload defaults:
**/lustre/fsw/infra_rd_gsw/rsalagame/llama3.1-405B-70B-8B-GB300-latest-test/test-env/workloads/pretrain_llama3.1/experiments/pretrain_llama3_8b_nvfp4_gpus8_tp1_pp1_cp1_vpNone_ep1_mbs4_gbs128/pretrain_llama3_8b_nvfp4_gpus8_tp1_pp1_cp1_vpNone_ep1_mbs4_gbs128_1765309590/pretrain_llama3_8b_nvfp4_gpus8_tp1_pp1_cp1_vpNone_ep1_mbs4_gbs128**

### DIFF
--- a/scripts/performance/utils/helpers.py
+++ b/scripts/performance/utils/helpers.py
@@ -240,7 +240,7 @@ def set_user_overrides(recipe: ConfigContainer, kwargs: Dict[str, Any]) -> None:
         recipe.model.pipeline_model_parallel_size = kwargs.get("pipeline_model_parallel_size")
     if kwargs.get("context_parallel_size") is not None:
         recipe.model.context_parallel_size = kwargs.get("context_parallel_size")
-    if kwargs.get("virtual_pipeline_model_parallel_size") != -1:
+    if kwargs.get("virtual_pipeline_model_parallel_size") is not None:
         recipe.model.virtual_pipeline_model_parallel_size = kwargs.get("virtual_pipeline_model_parallel_size")
     if kwargs.get("expert_model_parallel_size") is not None:
         recipe.model.expert_model_parallel_size = kwargs.get("expert_model_parallel_size")


scripts/performance/setup_experiment.py
```diff
@@ -138,11 +140,33 @@ def main(
...
+    
+    # Handle global batch size with scaling
+    if gbs is not None:
+        gbs_final = gbs
+    else:
+        # Scale GBS based on number of GPUs if different from default
+        if num_gpus != workload_base_config.num_gpus:
+            gbs_final = int(workload_base_config.gbs_scaling_factor * num_gpus)
+        else:
+            gbs_final = workload_base_config.global_batch_size
```
bdubauski: - When would be the case that gbs is not provided for the recipe? How common is that?
- This logic needs to be tested on different recipe to ensure it doesn't introduce regressions
- Couldn't this code be simplified?

```suggestion
    else:
        # Scale GBS based on number of GPUs 
        gbs_final = int(workload_base_config.gbs_scaling_factor * num_gpus)
```

bdubauski:
adding @scsudhakaran to take a second look as well.

scripts/performance/setup_experiment.py
```diff
@@ -138,11 +140,33 @@ def main(
...
+    # Load workload base config to get default values if not explicitly provided
+    workload_base_config = get_workload_base_config(
+        model_name, model_size, gpu, compute_dtype, domain, task
+    )
+    
+    # Use provided values if available, otherwise fall back to workload defaults
+    tp_size_final = tp_size if tp_size is not None else workload_base_config.tensor_model_parallel_size
+    pp_size_final = pp_size if pp_size is not None else workload_base_config.pipeline_model_parallel_size
+    cp_size_final = cp_size if cp_size is not None else workload_base_config.context_parallel_size
+    vp_size_final = vp_size if vp_size not in (None, -1) else workload_base_config.virtual_pipeline_model_parallel_size
```
scsudhakaran: This condition results an incorrect value if user explicitly provides vp_size as `None`

scripts/performance/setup_experiment.py
rsalagame-nvidia: yup, no situation where it is not provided. Should always just be the product of the scaled factor and the number of gpus.

scripts/performance/setup_experiment.py
```diff
@@ -138,11 +140,30 @@ def main(
...
+    # Load workload base config to get default values if not explicitly provided
+    workload_base_config = get_workload_base_config(
+        model_name, model_size, gpu, compute_dtype, domain, task
+    )
+    
+    # Use provided values if available, otherwise fall back to workload defaults
+    tp_size_final = tp_size if tp_size is not None else workload_base_config.tensor_model_parallel_size
+    pp_size_final = pp_size if pp_size is not None else workload_base_config.pipeline_model_parallel_size
+    cp_size_final = cp_size if cp_size is not None else workload_base_config.context_parallel_size
+    vp_size_final = vp_size if vp_size !=  -1 else workload_base_config.virtual_pipeline_model_parallel_size
```
bdubauski:
```suggestion
    vp_size_final = vp_size if vp_size is not None else workload_base_config.virtual_pipeline_model_parallel_size
```

bdubauski: @rsalagame-nvidia please also change argument_parser.py to set default to None for vp_size

bdubauski: @scsudhakaran FYI ^

bdubauski (approved):

PR #1663, opened 2025-12-09
Propagate PERF_ENV_VARS and overrides to SlurmExecutor container_env

sudostock:
Expose the container_env parameter of SlurmExecutor, explicitly set container_env for any environment variable being set. If you're setting performance variables for your run you expect them to be set in the container.

The pyxis/enroot defaults are to set all variables in the user environment in the container environment, except when that variable already exists in the container. Then it won't overwrite it unless '--container-env=VAR' is set.

### DIFF
--- a/scripts/performance/setup_experiment.py
+++ b/scripts/performance/setup_experiment.py
@@ -145,6 +145,10 @@ def main(
         f"_vp{vp_size}_ep{ep_size}_mbs{mbs}_gbs{gbs}"
     )
 
+    # Ensure all environment variables (including those added above) are propagated to the container
+    current_container_env = set(executor.container_env) if executor.container_env else set()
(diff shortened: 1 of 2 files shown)


scripts/performance/setup_experiment.py
```diff
@@ -145,6 +145,10 @@ def main(
         f"_vp{vp_size}_ep{ep_size}_mbs{mbs}_gbs{gbs}"
     )
 
+    # Ensure all environment variables (including those added above) are propagated to the container
+    current_container_env = set(executor.container_env) if executor.container_env else set()
+    executor.container_env = list(current_container_env | set(executor.env_vars))
+
```
sudostock, line 151: This section was carried over from the llmb-r0.1.0 branch which did set other variables in the setup_executor script. Not sure if we want to keep in this branch.

PR #1745, opened 2025-12-16
Always set segment for gb systems.

sudostock:
We should be explicit and always set segment where applicable. The existing version used assumptions about slurm defaults, however on at least one internal cluster segment unset is changed to 'segment=2' which can have negative performance implications.

Copy of https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1315 for llmb-r0.2.0


bdubauski (approved):
LGTM

PR #1747, opened 2025-12-16
feat: Validate PEFT target modules

yaoyu-33:
## Summary
- add alias tracking in ModuleMatcher so mis-typed target_modules raise clearer errors
- propagate alias registration to CanonicalLoRA/DoRA and validate in PEFT base call
- add regression test ensuring LoRA raises when a target module is missing

## Summary by CodeRabbit

* **Bug Fixes**
  * PEFT configurations now validate that all specified target modules exist in the model, raising an error with clear feedback when requested targets are missing.

* **Tests**
  * Added validation test for missing target module handling in LoRA configurations.


HollowMan6:
Maybe this can be coordinated with #1799 so that Canonical LoRA also supports targeting at MLA linear layers.

cuichenx (approved):

PR #1761, opened 2025-12-17
Remove HF token requirement for Qwen3 to enable offline mode.

sudostock:
We are pre-downloading the tokenizers for Qwen, no need for the hard hf_token assert (which will force online mode again)


bdubauski (approved):
Looks good

Issue #1750, opened 2025-12-16
PEFT Training fails for dense models when PP = 1 when activation checkpointing is on


erictang000:
**Describe the bug**

<img width="358" height="231" alt="Image" src="https://github.com/user-attachments/assets/07fa7d00-1e42-40f7-9179-1b9ec67816e0" />

When integrating training with Megatron-Bridge PEFT with SkyRL, we're noticing issues with convergence for dense models that seem to be only happening with `pp = 1`. Upon digging further, I'm able to see that when `pp=1` for these dense models, the LoRA weights are not being updated after training, leading me to believe there's an issue in Megatron-Bridge or Megatron-Core code leading to this being the case. This also seems to be appearing on the verl integration: https://github.com/volcengine/verl/pull/4063#issuecomment-3661475306.

**Steps/Code to reproduce bug**

In this PR: https://github.com/NovaSky-AI/SkyRL/pull/743, running `skyrl-train/examples/megatron/run_megatron_lora_qwen3-0.6b.sh` with pp=1 will not converge, while pp > 1 will converge.

This might be too heavyweight for megatron-bridge to reproduce, but I'm wondering if there are PEFT training scripts (maybe for SFT) on the Megatron-Bridge end that might be able to reproduce this behavior by using pp=1 for dense models (like Qwen/Qwen3-0.6B). I wasn't able to find any recipes with those settings in this repo, so any pointers for repro'ing on Megatron-Bridge would be helpful too!

**Expected behavior**

I expect applying LoRA via Megatron-Bridge on a dense model with no model parallelism to correctly update weights.

**Additional context**

HollowMan6:
I just found that for Verl's case, it will work (with non-zero grad norm) for PP=1 if we disable the activation checkpointing: https://github.com/volcengine/verl/pull/4063#issuecomment-3661475306 I'm not sure if this is also the case for you, but it looks like Megatron's CheckpointFunction.backward() is only invoked by PyTorch autograd when at least one input tensor requires grad.

- With PP>1, received tensors from other stages have requires_grad=True, so checkpoint backward is always called.
- With PP=1 and frozen base model, embedding outputs have requires_grad=False. This means CheckpointFunction.backward() is never called, and LoRA gradients inside the checkpoint are never computed.

erictang000:
@HollowMan6 Thanks for the pointer! That worked for me as well. Seeing convergence with pp=1, tp=1, dp=4 now.

[...]

PR #1762, opened 2025-12-17
fix: Fix PEFT recompute hook

yaoyu-33:
## Summary
- add a reusable recompute helper (with attribution) to ensure checkpointed transformer blocks see grad-requiring inputs during adapter-only finetuning
- invoke the helper when PEFT transforms are applied so adapter-only runs no longer drop gradients
- cover the helper with a focused unit test exercising the grad-enable path and duplicate patch guard

### DIFF
--- a/tests/unit_tests/peft/test_recompute.py
+++ b/tests/unit_tests/peft/test_recompute.py
@@ -49,7 +49,10 @@ class DummyModel(torch.nn.Module):
         self.base.weight.requires_grad = False
 
         # Trainable adapter parameter whose name contains ".adapter."
-        self.adapter = DummyAdapter()
+        # Use a ModuleDict with key "adapter" so that the full parameter
+        # name includes the expected substring (".adapter.") used by
+        # maybe_enable_recompute_inputs_grad.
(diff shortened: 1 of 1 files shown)


HollowMan6:
@yaoyu-33 Is it possible to apply the patching logic inside `__call__(self, model: ModelType, training: bool = True)` in https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/174298a35c87f2d0e14ef0d3cf91f1c030de70b6/src/megatron/bridge/peft/base.py directly? I just found that ~the current patching place seems to only apply to SFT (~ Verl doesn't call the logics in `src/megatron/bridge/training/setup.py`~)~, that way this will benefit for all downstream projects. https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/1750#issuecomment-3667522218 

Reference: https://github.com/volcengine/verl/blob/16a6c4791c5f829c4a0c207ee3a086e90f855157/verl/utils/megatron_utils.py#L218-L222

yaoyu-33:
@HollowMan6  updated.

erictang000:
confirmed that this is working for me!
image

pink is with the fix (blue is without activation checkpointing, purple is without LoRA)

PR #1764, opened 2025-12-18
GBS was not being scaled correctly across larger llama3.1 runs. If it…

rsalagame-nvidia:
GBS was not being scaled correctly across larger llama3.1 runs. If it wasn't being set explicitly, was reading directly from workload base config instead of scaling up. Needed to be the # gpus * gbs scaling factor


bdubauski (approved):
LGTM

Issue #1649, opened 2025-12-09
[Question] Is it expected for `pretrained_checkpoint` loading to fail silently


kannankumar:
**Describe the bug**
Even when `pretrained_checkpoint` is specified for finetuning /continued pretraining recipes, the training starts from random initialized weights instead of failing, when the checkpoint is not loadable (incorrect path or incompatible checkpoint type).
I notice that `exit_on_missing_checkpoint = False` by default (which might not be obvious to new users of Mbridge). 

**Steps/Code to reproduce bug**

Any finetuning recipe with the `pretrained_checkpoint` pointing to an incorrect path (to simulate a typo)

**Expected behavior**

The training recipe fails early when the pretrained_checkpoint is not found. This would allow to surface the error early instead of wasting compute with pretraining from random initialized weights.
This could be possible if either `exit_on_missing_checkpoint = True` by default or possibly with a stricter check when pretrained_checkpoint is specified in config.

**Additional suggestion**
It would also help if [docstrings](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/b3571fbc96d6e2149215734a1e9e5e9ef74d9c1b/src/megatron/bridge/training/config.py#L694) mention that expected pretrained checkpoint is in Megatron format.

PR #1822, opened 2025-12-29
fix: correct throughput calculation with sequence packing and warn on missing checkpoint

### LINKED ISSUE #1636: Potential bug in `report_throughput()` with sequence packing
Hi,

Suppose that I have:
1. a dataset with `dataset.seq_length=2048` -- since single samples can be up to 2048 tokens (many other are shorter)
2. a model with `model.seq_length=2048` -- forced to be equal to `dataset.seq_length`
3. packing enabled, with `dataset.packed_sequence_specs.packed_sequence_size=8192` -- to pack many sequences together ("simulating" `mbs=4`)

My understanding after https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/1539 (cc @yaoyu-33) is that each micro computes `8192` tokens during training, like a `mbs=4` with `seq_length=2048` (with _many_ more samples given packing). However, if this is right, I suspect Bridge is computing a wrong throughput when packing is enabled since [report_throughput() receives _just_](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/58f04619dc650164278dec749a08c95580ced0d0/src/megatron/bridge/training/utils/train_utils.py#L453) `dataset.seq_length` thus not knowing the *true* packing size.

Indeed I'm seeing the same `TFLOP/s/GPU` even if I double the packing size that made me suspect either 1) the model is receiving 2048 tokens truncating the 8192 or 2) the throughput computation is wrong.

yurekami:
## Summary

- **Fix throughput/runtime metrics when sequence packing is enabled** (#1636): Uses `packed_sequence_size` instead of `seq_length` for token calculations, ensuring accurate TFLOP/s and tokens/sec metrics
- **Add warning for silent checkpoint loading failure** (#1649): Warns when `pretrained_checkpoint` is set but `exit_on_missing_checkpoint=False`, preventing silent training from random weights due to typos or invalid paths
- Enhanced `pretrained_checkpoint` docstring to clarify checkpoint format requirements (Megatron distributed format)

## Test plan

- [ ] Verify throughput metrics are correctly calculated when using sequence packing (should show higher tokens/sec proportional to `packed_sequence_size`)
- [ ] Verify warning is printed when `pretrained_checkpoint` is specified with `exit_on_missing_checkpoint=False`
- [ ] Run existing unit tests to ensure no regressions

Fixes #1636, #1649

### DIFF
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -693,8 +693,8 @@ class CheckpointConfig:
     pretrained_checkpoint: Optional[str] = None
     """Directory containing a pretrained model checkpoint for finetuning.
     The checkpoint must be in Megatron distributed checkpoint format (torch_dist, zarr, or fsdp_dtensor).
-    If this path is set but the checkpoint is not found, training will fail unless
-    exit_on_missing_checkpoint is explicitly set to False."""
+    If this path is set but the checkpoint is not found, training will start from random
+    initialization (scratch) unless exit_on_missing_checkpoint is set to True."""
 
     ckpt_step: Optional[int] = None
     """Checkpoint step to load model from."""
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -691,7 +691,10 @@ class CheckpointConfig:
     Assumed when loading a release checkpoint."""
 
     pretrained_checkpoint: Optional[str] = None
-    """Directory containing a pretrained model checkpoint for finetuning."""
+    """Directory containing a pretrained model checkpoint for finetuning.
+    The checkpoint must be in Megatron distributed checkpoint format (torch_dist, zarr, or fsdp_dtensor).
+    If this path is set but the checkpoint is not found, training will fail unless
+    exit_on_missing_checkpoint is explicitly set to False."""
 
     ckpt_step: Optional[int] = None
     """Checkpoint step to load model from."""
@@ -1394,6 +1397,18 @@ class ConfigContainer(Container):
 
         if self.peft is not None:
             assert self.checkpoint.pretrained_checkpoint is not None, "PEFT requires a pretrained checkpoint path"
+
+        # When pretrained_checkpoint is specified for finetuning, ensure we don't silently
+        # start from random weights if the checkpoint is missing. This catches common mistakes
+        # like typos in the checkpoint path early, rather than wasting compute on training
+        # from random initialization.
+        if self.checkpoint.pretrained_checkpoint is not None and not self.checkpoint.exit_on_missing_checkpoint:
+            print_rank_0(
+                "Warning: pretrained_checkpoint is specified but exit_on_missing_checkpoint=False. "
+                "If the checkpoint at the specified path is not found or invalid, training will "
+                "silently start from random initialization. Consider setting exit_on_missing_checkpoint=True "
+                "to fail early on missing checkpoints."
+            )
 
         if self.dataset is not None:
             # Only validate sequence length for GPTDatasetConfig or FinetuningDatasetConfig
--- a/src/megatron/bridge/data/datasets/packed_sequence.py
+++ b/src/megatron/bridge/data/datasets/packed_sequence.py
@@ -15,6 +15,7 @@ import json
 import logging
 from dataclasses import dataclass
 from pathlib import Path
+from typing import Optional
 
 import numpy as np
 from megatron.core.msc_utils import MultiStorageClientFeature
@@ -158,6 +159,22 @@ class PackedSequenceSpecs:
     If less than or equal to 0, sequence packing is disabled. Defaults to -1.
     Note: This arg is distinct from `seq_length` because `seq_length` specifies the maximum length
     of the original sequence (i.e. the length to truncate long sequences in the input data).
+    """
+
+    avg_sequences_per_bin: Optional[float] = None
+    """
+    Average number of sequences packed into each bin.
+
+    This value is critical for accurate TFLOP/s calculation because attention
+    FLOPs scale quadratically per sequence, not per total tokens. For example:
+    - 4 sequences of 1024 tokens packed into 4096 -> attention FLOPs ~ 4 * 1024^2
+    - Treating as single 4096 sequence -> attention FLOPs ~ 4096^2 (4x overestimate)
+
+    If not specified, this will be estimated as:
+        packed_sequence_size / seq_length
+
+    For accurate metrics, set this based on your dataset's packing statistics.
+    Future versions may track this automatically during packing.
     """
 
     tokenizer_model_name: str = None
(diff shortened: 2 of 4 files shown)


yaoyu-33:
@yurekami thanks for the contribution. This fix is partially correct, but we cannot simply treat the packed sequence size as the sequence length. In the attention, each sequence in the bin is calculated individually. Using the entire pack size will overestimate the calculation.

A rough estimation would be to set a pseudo batch size equal to the average number of sequences in each bin, and set seq length = seq pack size // pseudo batch size.

yurekami:
Thanks for the catch, @yaoyu-33 — you're right that attention FLOPs scale as O(seq_len²) per sequence, so using the full packed sequence size overestimates the cost (e.g., 4×2048² = 16M vs 8192² = 67M, a 4x overestimate).

**Proposed fix:**
- **Tokens/sec**: Continue using `packed_sequence_size` (this is correct)
- **TFLOP/s**: Compute pseudo batch size as average sequences per bin, then use `effective_seq_len = packed_seq_size // pseudo_batch_size` for attention FLOP estimation

Question: Does the pipeline already expose `avg_sequences_per_bin` or similar packing statistics, or should I add that tracking alongside the existing packing metrics?

Happy to implement whichever approach fits best with the codebase!

yaoyu-33:
@yurekami : thanks for the proposal. that's all good to me. we dont yet have avg_sequences_per_bin statistics, we can make it a argument or placeholder for now, we can clean it up in another pr.

Thanks for contribution!

yurekami:
# PR #1822 - Revised Implementation

## Response to @yaoyu-33's Review

Thank you for the insightful feedback. You're absolutely right that attention FLOPs scale quadratically per sequence, and the naive approach of using `packed_sequence_size` directly would overestimate TFLOP/s by a factor equal to the packing ratio.

### The Problem Illustrated

```
Scenario: 4 sequences × 1024 tokens → 4096 packed tokens

┌─────────────────────────────────────────────────────────┐
│                    Packed Bin (4096 tokens)             │
├──────────────┬──────────────┬──────────────┬────────────┤
│   Seq 1      │    Seq 2     │    Seq 3     │   Seq 4    │
│   (1024)     │   (1024)     │   (1024)     │  (1024)    │
│              │              │              │            │
│  Attention:  │  Attention:  │  Attention:  │ Attention: │
│   1024²      │   1024²      │   1024²      │  1024²     │
└──────────────┴──────────────┴──────────────┴────────────┘

NAIVE (wrong):   4096² = 16,777,216 attention units
CORRECT:     4 × 1024² =  4,194,304 attention units
                          ─────────
                          4× overestimate!
```

### Solution Implemented

Per your suggestion, I've added `avg_sequences_per_bin` as an optional parameter with fallback estimation:

```python
@dataclass
class PackedSequenceSpecs:
    packed_sequence_size: int = -1
    avg_sequences_per_bin: Optional[float] = None  # NEW
    ...
```

**Two separate functions for different metrics:**

| Metric | Function | Formula |
|--------|----------|---------|
| tokens/sec | `get_packed_sequence_size()` | `packed_sequence_size` (linear) |
| TFLOP/s | `get_effective_seq_length_for_tflops()` | `packed_size / avg_sequences_per_bin` |

### Fallback Behavior

When `avg_sequences_per_bin` is not explicitly set:

```python
# Estimate from the ratio of packed size to base sequence length
avg_seqs = packed_sequence_size / seq_length

# Example: 4096 / 1024 = 4.0 estimated sequences per bin
```

A warning is logged to encourage users to set the actual value:

```
[Throughput] avg_sequences_per_bin not configured.
Estimating 4.00 sequences/bin from packed_size(4096) / seq_length(1024).
Set avg_sequences_per_bin for more accurate TFLOP/s metrics.
```

### Future Enhancement Path

The fallback estimation assumes uniform sequence lengths (≈ seq_length). For more accurate metrics with variable-length sequences:

1. **Phase 1 (this PR):** Manual configuration via `avg_sequences_per_bin`
2. **Phase 2 (future):** Track packing statistics during dataset creation
3. **Phase 3 (future):** Store/load stats in `packed_metadata_path`

### Files Changed

1. **`packed_sequence.py`** - Add `avg_sequences_per_bin` field
2. **`train_utils.py`** - Split into two throughput functions

### Testing

See `pr-1822-test-cases.py` for comprehensive test coverage including:
- No packing scenarios
- Explicit `avg_sequences_per_bin`
- Fallback estimation
- Edge cases (disabled packing, zero values)
- Mathematical verification of the 4× overestimate problem

---

Let me know if you'd like any adjustments to this approach!

src/megatron/bridge/training/config.py
```diff
@@ -691,7 +691,10 @@ class CheckpointConfig:
     Assumed when loading a release checkpoint."""
 
     pretrained_checkpoint: Optional[str] = None
-    """Directory containing a pretrained model checkpoint for finetuning."""
+    """Directory containing a pretrained model checkpoint for finetuning.
+    The checkpoint must be in Megatron distributed checkpoint format (torch_dist, zarr, or fsdp_dtensor).
+    If this path is set but the checkpoint is not found, training will fail unless
```
yaoyu-33: change "training will fail" to 
If this path is set but the checkpoint is not found, training will start from scratch (might be unexpected) unless exit_on_missing_checkpoint is explicitly set to False.

yaoyu-33 (dismissed):

yaoyu-33:
add one comment otherwise lgtm, thanks again!

yurekami:
@yaoyu-33 Thanks for the thorough review! The docstring clarification makes sense — explicitly warning about the silent fallback to training from scratch is important for users who expect their checkpoint to load.

The split approach (tokens/sec using `packed_sequence_size`, TFLOP/s using `effective_seq_len`) correctly handles the quadratic attention scaling issue. 👍

yaoyu-33:
@yurekami  I wonder if you can update the docstring, so I can start the merging process?

yaoyu-33 (approved):

yaoyu-33:
@yurekami  can you plz check unit test failure? get back to me if you have difficulty

yaoyu-33:
gentle ping to keep this active

yurekami:
Closing this, with thanks for the approval in January and apologies for going quiet on the failing test. Both halves have since been fixed on main in better ways than this patch: `num_floating_point_operations` now takes `seqlen_sum` / `seqlen_squared_sum`, and `train.py` feeds it per-batch stats that `flop_utils` derives from `cu_seqlens` for packed batches, which is the right primitive where this PR had an `avg_sequences_per_bin` knob the user had to fill in by hand. And `CheckpointConfig` now asserts that `pretrained_checkpoint` exists (config.py, `file_exists` check in the validator), which closes the silent-start case in #1649 harder than the warning here. If #1636 still reproduces on main it is a question of how the packed loader populates those stats, not of this diff.

PR #1867, opened 2026-01-06
feat: Enable THD context-parallel packed sequences

yaoyu-33:
# Enable THD context-parallel packed sequences

## Summary
- Add `context_parallel_size` to packed-sequence specs and thread it through finetuning dataset builds so packing can pre-pad for THD/CP alignment.
- Pre-pad packed SFT samples when `context_parallel_size>1` so each packed sample length is divisible by `2 * cp_size`, matching THD slicing requirements.
- Partition packed batches for CP with `thd_get_partitioned_indices` instead of generic CP slicing; propagate padded/unpadded cu_seqlens to the forward pass.
- Extend packed-sequence params to surface both padded and unpadded cu_seqlens for THD kernels.
- Add a functional test that runs the Llama-3.2-1B SFT example with packed sequences and `context_parallel_size=2`.

### DIFF
--- a/tests/functional_tests/training/test_seqpacking_cp_example.py
+++ b/tests/functional_tests/training/test_seqpacking_cp_example.py
@@ -89,6 +89,7 @@ class TestPeftSftExample:
                 tokenizer_model_name="meta-llama/Llama-3.2-1B",
                 pad_seq_to_mult=cfg.model.context_parallel_size * 2,
             ),
+            rewrite=False,
         )
 
         cfg.model.seq_length = 256


yaoyu-33:
image

src/megatron/bridge/data/builders/finetuning_dataset.py
cuichenx, line 99: if a user has previous run packing with CP=1, it would use the same packing file right? so train_path_packed should contain the cp size in the path

src/megatron/bridge/data/datasets/packed_sequence.py
```diff
@@ -186,6 +242,12 @@ class PackedSequenceSpecs:
     If True, pad cu_seqlens to a constant size, which is required for use with cudagraphs.
     """
 
+    context_parallel_size: int = 1
```
cuichenx: is user expected to pass in this manually?

cuichenx (commented):
looks great, couple of comments. I pushed my suggestions

src/megatron/bridge/data/datasets/packed_sequence.py
cuichenx: if so i suggest we update all recipe files.

src/megatron/bridge/training/gpt_step.py
```diff
@@ -33,6 +34,44 @@
...
+    """
+
+    try:
+        import transformer_engine_torch as tex
+    except ModuleNotFoundError as e:
+        logger.error("Please update Transformer Engine to >= 1.10 to use Context Parallel with THD format data")
+        raise e
+
+    cp_rank = parallel_state.get_context_parallel_rank()
+    cu_seqlens = batch["cu_seqlens"].squeeze()
```
liding-nv: worth checking if micro batch size is 1 for seq packing (e.g., `cu_seqlens.dim()==1`)?

src/megatron/bridge/training/gpt_step.py
```diff
@@ -33,6 +34,44 @@
...
+def _partition_packed_batch_for_cp(batch: dict[str, torch.Tensor], cp_size: int) -> dict[str, torch.Tensor]:
+    """Partition THD/packed batches across context-parallel ranks.
+
+    Uses transformer_engine's `thd_get_partitioned_indices` to slice sequence
+    dimension aligned with packed cu_seqlens. This avoids the generic
+    `get_batch_on_this_cp_rank` slicing which assumes contiguous sequence tokens.
+    """
+
+    try:
+        import transformer_engine_torch as tex
```
liding-nv, line 52: need to check tex version after import `is_te_min_version("1.10.0")`

src/megatron/bridge/training/gpt_step.py
```diff
@@ -33,6 +34,44 @@
...
+        logger.error("Please update Transformer Engine to >= 1.10 to use Context Parallel with THD format data")
+        raise e
+
+    cp_rank = parallel_state.get_context_parallel_rank()
+    cu_seqlens = batch["cu_seqlens"].squeeze()
+    cu_seqlens_unpadded = batch.get("cu_seqlens_unpadded")
+    if cu_seqlens_unpadded is not None:
+        batch["cu_seqlens_unpadded"] = cu_seqlens_unpadded.squeeze()
+
+    skip_keys = {
```
liding-nv, line 70: `attention_mask` should also be skipped?

yaoyu-33: why? attention mask will not be skipped because it will also do some shuffling.
Though attn mask will mostly be None.

This part of code was from nemo1.

liding-nv: right... you have `val is None` check below

yaoyu-33:
```
from megatron.bridge.data.builders.hf_dataset import HFDatasetConfig
from megatron.bridge.data.datasets.packed_sequence import PackedSequenceSpecs
from megatron.bridge.data.hf_processors.squad import process_squad_example
from megatron.bridge.recipes.llama.llama3 import llama32_1b_finetune_config as finetune_config
from megatron.bridge.training.finetune import finetune


"""
python examples/conversion/convert_checkpoints.py import \
               --hf-model meta-llama/Llama-3.2-1B \
               --megatron-path /opt/ckpt/meta-llama/Llama-3.2-1B
"""


if __name__ == "__main__":
    cfg = finetune_config(peft=None)

    # Configure tokenizer
    cfg.tokenizer.tokenizer_type = "HuggingFaceTokenizer"
    cfg.tokenizer.tokenizer_model = "meta-llama/Llama-3.2-1B"
    cfg.model.calculate_per_token_loss = True
    cfg.ddp.average_in_collective = False
    cfg.model.context_parallel_size = 2
    cfg.model.tensor_model_parallel_size = 1
    # FinetuningDatasetConfig
    cfg.dataset = HFDatasetConfig(
        dataset_name="squad",
        process_example_fn=process_squad_example,
        seq_length=2048,
        dataloader_type="batch",
        packed_sequence_specs=PackedSequenceSpecs(
            packed_sequence_size=8192,
            tokenizer_model_name="meta-llama/Llama-3.2-1B",
            context_parallel_size=cfg.model.context_parallel_size,
        ),
        num_workers=2,
    )

    # Short test run (10 iterations for demo
    cfg.logger.log_interval = 1
    cfg.train.eval_iters = 0
    cfg.checkpoint.save = "/opt/experiments/runs"
    cfg.checkpoint.pretrained_checkpoint = "/opt/ckpt/meta-llama/Llama-3.2-1B"

    cfg.logger.wandb_project = "mbridge_demo"
    cfg.logger.wandb_exp_name = "llama32_1b_squad_thd_cp1"
    # BF16 is used by default (no mixed_precision setting needed)
    print(cfg)

    finetune(cfg, forward_step)

```

cuichenx (approved):
loss curve matches after padding fix

Image

filippo82:
Woot woot!

Issue #1776, opened 2025-12-19
[Bug] Sequence Packing of Qwen3-VL


lostkevin:
**Describe the bug**

Hi devs! Recently we attempt to train Megatron-Bridge-based Qwen3-VL with slime and find large logprobs diff between train and rollout backends. After deeply investigating the modeling, we find that:

1. `get_rope_index` called in `model.py` cannot correctly process packed input_ids.
2. `_apply_rotary_pos_emb_thd` only support RoPE of traditional mapping without offsets when `apply_rope_fusion` is enabled. `--no-rope-fusion` must be set to get the correct behavior.

Here is a draft patch [fix](https://gist.github.com/lostkevin/fce38435f3b64fe35b47e777ad22c98c) (for dense and MoE, not strictly tested) on 5c7ebe7d8c31e75ddf882e1e6ced69d10e250786 .

nanjiangwill:
hey @lostkevin, thanks for this patch! i am trying to train qwen3-vl-moe with this patch, the log prob diff is smaller but it is still very high(with megatron, the logprob diff between rollout and train is like 0.73 but with fsdp it is 0.03)

lostkevin:
> hey [@lostkevin](https://github.com/lostkevin), thanks for this patch! i am trying to train qwen3-vl-moe with this patch, the log prob diff is smaller but it is still very high(with megatron, the logprob diff between rollout and train is like 0.73 but with fsdp it is 0.03)

We've reached 0.82 on Geo-3K with Qwen3-VL-4B on an internal slime impl (also based on Megatron-Bridge) after ~1K steps with the above patch, the logprob diff is ~0.01 during training process.

nanjiangwill:
> > hey [@lostkevin](https://github.com/lostkevin), thanks for this patch! i am trying to train qwen3-vl-moe with this patch, the log prob diff is smaller but it is still very high(with megatron, the logprob diff between rollout and train is like 0.73 but with fsdp it is 0.03)
> 
> We've reached 0.82 on Geo-3K with Qwen3-VL-4B on an internal slime impl (also based on Megatron-Bridge) after ~1K steps with the above patch, the logprob diff is ~0.01 during training process.

interesting, because previously i ran megatron with dense vl model and without this patch the logprob diff is 0.01 for qwen3-vl-4b.

the issue right now is that the qwen3-vl-moe does not work

lostkevin:
[...]

PR #1868, opened 2026-01-06
Fix Qwen3-VL packed RoPE offsets and MoE export merging

yaoyu-33:
## Summary
- handle packed sequence attention masking when building Qwen3-VL mRoPE indices to fix packed seq logprob divergence (#1776)
- pass packed_seq_params through model to rope utils and use fp32 position ids for qwen3 mrope

Results can be verified via
```
uv run python -m torch.distributed.run --nproc_per_node=8 \
   examples/conversion/hf_to_megatron_generate_vlm.py --hf_model_path="Qwen/Qwen3-VL-30B-A3B-Instruct" \
   --image_path="https://qianwen-res.oss-cn-beijing.aliyuncs.com/Qwen-VL/assets/demo.jpeg" --prompt="Describe this image." \
   --ep 8

uv run python -m torch.distributed.run --nproc_per_node=8 examples/conversion/hf_megatron_roundtrip_multi_gpu.py \
      --hf-model-id Qwen/Qwen3-VL-30B-A3B-Instruct  --ep 2 --tp 2 --pp 2
```

### DIFF
--- a/tests/unit_tests/models/qwen_vl/modelling_qwen3_vl/test_text_model_forward.py
+++ b/tests/unit_tests/models/qwen_vl/modelling_qwen3_vl/test_text_model_forward.py
@@ -1,0 +1,76 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for Qwen3VL text model forward behavior."""
+
+import torch
+
+from megatron.bridge.models.qwen_vl.modelling_qwen3_vl.text_model import Qwen3VLGPTModel
+
+
+class _DummyDecoder:
+    def __init__(self):
+        self.called_with = None
+
+    def __call__(self, **kwargs):
+        self.called_with = kwargs
+        return torch.zeros(1, 1, 1)
+
+
+class _DummyModel:
+    def __init__(self):
+        self.decoder = _DummyDecoder()
+        self.mtp_process = False
+        self.preprocess_output = None
+        self.postprocess_args = None
+
+    def _preprocess(self, **_):
+        self.preprocess_output = (
+            torch.randn(1, 1, 1),
+            torch.randn(1, 1),
+            torch.randn(1, 1),
+            torch.randn(1, 1),
+            torch.tensor([0]),
+            torch.randn(1, 1),
+        )
+        return self.preprocess_output
+
+    def _postprocess(self, **kwargs):
+        self.postprocess_args = kwargs
+        return "ok"
+
+
+def test_forward_accepts_extra_preprocess_output():
+    """Ensure forward ignores extra values returned by _preprocess."""
+    dummy = _DummyModel()
+    input_ids = torch.zeros((1, 4), dtype=torch.long)
+    position_ids = torch.zeros((1, 4), dtype=torch.long)
+    attention_mask = torch.ones((1, 4), dtype=torch.long)
+
+    output = Qwen3VLGPTModel.forward(
+        dummy,
+        input_ids=input_ids,
+        position_ids=position_ids,
+        attention_mask=attention_mask,
+    )
+
+    preproc = dummy.preprocess_output
+    assert output == "ok"
+    assert dummy.decoder.called_with["hidden_states"] is preproc[0]
+    assert dummy.decoder.called_with["rotary_pos_emb"] is preproc[1]
+    assert dummy.decoder.called_with["rotary_pos_cos"] is preproc[2]
+    assert dummy.decoder.called_with["rotary_pos_sin"] is preproc[3]
+    assert dummy.decoder.called_with["sequence_len_offset"] is preproc[4]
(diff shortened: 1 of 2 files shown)


src/megatron/bridge/utils/common_utils.py
```diff
@@ -267,6 +268,59 @@ def extract_expert_number_from_param(param_name: str) -> int:
...
+        return converted_weights_dict
+
+    assert len(converted_weights_dict) == 1, (
+        f"There should be only one key in the converted_weights_dict, got keys: {converted_weights_dict.keys()}"
+    )
+
+    for key, value in converted_weights_dict.items():
+        cache = hf_weights_cache.setdefault(key, {})
+
+        if ep_size == 1:
```
lostkevin: always False?

src/megatron/bridge/models/qwen_vl/qwen3_vl_bridge.py
```diff
@@ -359,38 +360,89 @@ def mapping_registry(self) -> MegatronMappingRegistry:
...
+            return {}
+
+        fused: Dict[str, torch.Tensor] = {}
+
+        for name, tensor in converted.items():
+            if name.endswith(".gate"):
+                base_name = name[: -len(".gate")]
+                gate_tensor = tensor
+                up_tensor = converted.get(f"{base_name}.up")
+                if up_tensor is None:
```
lostkevin: I'm curious that when up_tensor will be None

src/megatron/bridge/models/qwen_vl/qwen3_vl_bridge.py
```diff
@@ -359,38 +360,89 @@ def mapping_registry(self) -> MegatronMappingRegistry:
...
+            hf_weights_cache=self.hf_weights_cache,
+        )
+
 
 class ExpertMLPDownProjMapping(AutoMapping):
     """Mapping for expert MLP down projection weights between HF and Megatron formats."""
 
+    def __init__(self, *args, **kwargs):
+        # Down projection weights need a transpose; rely on AutoMapping permutation support.
+        permute_dims = kwargs.pop("permute_dims", (1, 0))
```
lostkevin: Will bias mapping call torch.permute in AutoMapping?

yaoyu-33: in this case, ExpertMLPDownProjMapping doesn't have bias in the mapping definition. But if it does, it will call permute as well.

src/megatron/bridge/models/qwen_vl/qwen3_vl_bridge.py
yaoyu-33: moved to a assert. it shouldn't happend.

src/megatron/bridge/utils/common_utils.py
yaoyu-33: why always False?

lostkevin: ep_size 1 will return at L286 in advance

src/megatron/bridge/models/qwen_vl/qwen3_vl_bridge.py
lostkevin: I see，but bias is an 1-d tensor. Will permute raise error if this happens？（GPT-OSS has biases，therefore I don’t know if it works on that model）

src/megatron/bridge/utils/common_utils.py
yaoyu-33: lol my bad.

src/megatron/bridge/models/qwen_vl/qwen3_vl_bridge.py
yaoyu-33: yeah it will raise an error. We haven't synchronized this projection part here. We can sync later and yeah maybe this need to changed. (or we update permute logic)

lostkevin (approved):

src/megatron/bridge/models/qwen_vl/qwen3_vl_bridge.py
```diff
@@ -429,20 +476,18 @@ def megatron_to_hf(self, megatron_weights: torch.Tensor, megatron_module: nn.Mod
...
+                # [ep_size, mlp_in, up_out]
+                up_tensor = converted.get(f"{base_name}.up").transpose(1, 2).contiguous()
                 assert up_tensor is not None
-                # Back to HF layout.
-                gate_tensor = gate_tensor.transpose(0, 1).contiguous()
-                up_tensor = up_tensor.transpose(0, 1).contiguous()
-                fused[base_name] = torch.cat([gate_tensor, up_tensor], dim=0)
+                # Back to HF fused layout: stack [gate; up] along dim 0.
+                # [ep_size, 2, mlp_in, gate_out/up_out]
+                fused[base_name] = torch.stack([gate_tensor, up_tensor], dim=1)
```
lostkevin: why not concat here?

yaoyu-33: if concat here needs to split later.
if concat we will have [gate_expert_1, up_expert_1, gate_expert_2, up_expert_2, ...], we kinda need to uninterleave it later

src/megatron/bridge/models/gpt_oss/gpt_oss_bridge.py
```diff
@@ -240,10 +240,6 @@ def megatron_to_hf(self, megatron_weights: torch.Tensor, megatron_module: nn.Mod
             megatron_weights = megatron_weights.transpose(0, 1)
         return super().megatron_to_hf(megatron_weights.contiguous(), megatron_module)
 
-    def _validate_patterns(self, *args, **kwargs):
-        # allow number of wildcards to mismatch in this mapping
-        pass
-
```
cuichenx, line 246: This way gpt-oss would print a warning statement but it should be ignored by the user, right? It might be counter intuitive for new users

PR #1948, opened 2026-01-14
Pass EP size CLI override to PerfEnvPlugin

dingqingy-nv:
## Summary

This PR fixes a bug where `PerfEnvPlugin` does not receive the expert parallel size (`-ep`) from command-line overrides.

## Problem

When using `-ep <size>` to override expert parallelism, the value was correctly applied to the model config via `set_user_overrides()`, but was **not** passed to `PerfEnvPlugin`. As a result, HybridEP environment variables like `NUM_OF_HYBRID_EP_RANKS_PER_NVLINK_DOMAIN` were set based on the workload base config default instead of the user-specified value.

This mismatch causes runtime errors.

## Fix

Pass `args.expert_model_parallel_size` through `main()` to `PerfEnvPlugin`


ko3n1g:
Is this linked to open issues we've documented?

dingqingy-nv:
Not that I am aware of. Just encountered this today myself. Quickly searched issues and didn't find any related

erhoo82 (approved):

PR #1955, opened 2026-01-15
Allow custom checkpoint impl via checkpoint manager interface

ananthsub:
# What does this PR do ?

Currently, checkpointing save/load is implemented purely through functional calls, making customization by users impossible without direct modification/forking.

This PR introduces a CheckpointManager protocol to allow users to inject custom save/load implementations without needing to fork the code. 

Protocol overview:
- `CheckpointManager`: main protocol to implement save/load through. finalizing async calls is also routed through this protocl now too for cleanup
- Auxiliary dataclasses to encapsulate save/load argument inputs

`CheckpointConfig` is augmented with a field to specify the path to the custom manager class. If specified, the class is imported and used during training.  A factory function is added to instantiate the manager to use for training. A default implementation is provided, which wraps around the existing functions. The default implementation also stores the checkpoint context map, used for local checkpointing. There are no changes made to these functions in this PR, so the core checkpoint saving/loading logic is untouched.

Integration points:
- `setup.py` is updated to create the checkpoint manager and use it for the initial checkpoint load
- in `train.py` in the main training loop, the checkpoint manager is used for all saves + async finalization
- in `pretrain.py`, the checkpoint manager is used for the end of train checkpoint save

# Changelog

- Enable custom checkpoint implementations via CheckpointManager interface

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

## Summary by CodeRabbit

* **New Features**
  * Custom checkpoint manager configuration - users can now specify a custom checkpoint manager class through configuration to customize checkpoint save/load behavior.

* **Documentation**
  * Comprehensive documentation added for creating and using custom checkpoint managers, including configuration guidance, usage examples, and implementation patterns.

### DIFF
--- a/src/megatron/bridge/training/setup.py
+++ b/src/megatron/bridge/training/setup.py
@@ -244,9 +244,10 @@ def setup(
     # checkpoints are independent of global ones — they don't write
     # latest_train_state.pt to load_dir, so checkpoint_exists() won't
     # find them.
+    _ckpt_ctx = getattr(checkpoint_manager, "checkpointing_context", {})
     has_local_checkpoint = (
-        "local_checkpoint_manager" in checkpointing_context
-        and checkpointing_context["local_checkpoint_manager"].find_latest() != -1
+        "local_checkpoint_manager" in _ckpt_ctx
+        and _ckpt_ctx["local_checkpoint_manager"].find_latest() != -1
     )
 
     # For PEFT, the pretrained checkpoint is loaded in the pre-wrap hook


src/megatron/bridge/training/checkpointing.py
```diff
@@ -437,6 +438,247 @@ class CheckpointType(Enum):
...
+
+        Args:
+            state: The GlobalState object (needed for async_calls_queue access).
+            blocking: If True, wait for all pending saves to complete.
+            terminate: If True, close the async queue after finalization.
+        """
+        ...
+
+
+class DefaultCheckpointManager:
```
maanug-nv, line 563: maybe im missing something, but shouldn't this extend the `CheckpointManager` protocol?

src/megatron/bridge/training/checkpointing.py
```diff
@@ -437,6 +438,247 @@ class CheckpointType(Enum):
...
+                f"Expected fully qualified class name like 'mypackage.module.ClassName'."
+            )
+
+        try:
+            module = importlib.import_module(module_path)
+        except ImportError as e:
+            raise ImportError(f"Could not import module '{module_path}' for custom checkpoint manager: {e}") from e
+
+        try:
+            custom_manager_class = getattr(module, class_name)
```
maanug-nv, line 701: Should we add some checks like `isinstance(CheckpointManager)` , or check if it has required functions?

src/megatron/bridge/training/setup.py
maanug-nv: What if the user's custom logic does not save the same `train_state.pt` or changes it's name, location, etc.? then we never load right?

src/megatron/bridge/training/checkpointing.py
ananthsub: for a protocol, users don't have to extend `CheckpointManager` since it relies on structural subtyping.

ananthsub:
> I think several functions in model_load_save.py would fail if using a custom checkpoint. probably we don't need to support HF <-> custom format. but what about loading for inference? i think we would either have to make all the functions that model_load_save.py imports from checkpointing part of the protocol (load model weights, get run config), or just document/warn that it is undefined behavior.

> if a user wants to do PEFT on a custom checkpoint, seems like loading would fail because we call _load_checkpoint_from_path() in the peft_pre_wrap_hook(). How would I replace that hook after setup? With a callback? Generally less familiar with PEFT, so maybe there's other concerns.

I think supporting a custom checkpoint format is beyond the initial scope at the moment. This customization is intended for users to be able to handle the save/load operations, but we should expect to have the same checkpoint format on disk. Otherwise there are too many dependencies both within bridge and outside that would then need to access the checkpoint protocol to figure out how to load the checkpoint, which will be very complicated for users to deal with. 

I've updated the docs with a limitations section to make this clear

docs/training/checkpointing.md
```diff
@@ -224,4 +224,145 @@ Local checkpointing leverages the [NVIDIA Resiliency Extension](https://nvidia.g
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
 | `dist_ckpt_optim_fully_reshardable` | `bool` | `False` | Make optimizer distributed checkpoint fully reshardable (TP/PP/EP/DP) as opposed to plain DP reshardability |
-| `distrib_optim_fully_reshardable_mem_efficient` | `bool` | `False` | Use as little memory as possible during save and load by using Gloo. Has affect only with `dist_ckpt_optim_fully_reshardable` flag |
\ No newline at end of file
+| `distrib_optim_fully_reshardable_mem_efficient` | `bool` | `False` | Use as little memory as possible during save and load by using Gloo. Has affect only with `dist_ckpt_optim_fully_reshardable` flag |
+
+## Custom Checkpoint Manager
```
g-husam, line 280: 👋 Hey @ananthsub , I'm one of the users who requested this, thank you so much for putting this together!! Just a heads up: I plan on going through the PR to see if there's anything else we need, will leave comments with any questions I have.

Cheers!

src/megatron/bridge/training/pretrain.py
```diff
@@ -148,22 +148,23 @@ def _pretrain(
...
                 state,
-                ckpt_context,
+                checkpoint_manager,
                 pg_collection,
                 callback_manager=callback_manager,
             )
 
         barrier_and_log("after training is done")
         ckpt_config = config.checkpoint
         if ckpt_config.save and state.train_state.step != 0 and ckpt_config.save_interval != 0:
```
g-husam: Hey @ananthsub , question: will this checkpoint every step so long as these properties are truthy? I'm wondering where the evaluation of something like `if curr_step % save_interval == 0: do save` is happening, to know when to call the save.

Alternatively, the step info can be passed in to `checkpoint_manager.save` (I think it is in `state.training_state`), and then the manager can make the decision. That would be ideal for us actually, but doesn't look like the default impl does that check, so I presume it is happening elsewhere, unless I missed it.

(I'm still reviewing to see if I have any other q's, but wanted to post this q in the meantime to avoid a long delay)

docs/training/checkpointing.md
g-husam, line 79: Btw unrelated, but when reading the rendered doc I noticed this seems to throw off the formatting (I think the backticks are not closed off at the end)

Could also use

```
> [!NOTE]
> The `load` parameter should ...
```

src/megatron/bridge/training/checkpointing.py
```diff
@@ -460,6 +461,256 @@ class CheckpointType(Enum):
...
+    """
+
+    state: GlobalState
+    model: list[MegatronModule]
+    optimizer: MegatronOptimizer | None
+    opt_param_scheduler: Any | None
+    num_floating_point_operations_so_far: int
+
+    train_data_iterator: Any | None = None
+    non_persistent_ckpt: bool = False
```
g-husam, line 489: to confirm, what exactly is considered `non_persistent_ckpt` and when is it set to True? What determines the value set

src/megatron/bridge/training/checkpointing.py
```diff
@@ -460,6 +461,256 @@ class CheckpointType(Enum):
...
+            module = importlib.import_module(module_path)
+        except ImportError as e:
+            raise ImportError(f"Could not import module '{module_path}' for custom checkpoint manager: {e}") from e
+
+        try:
+            custom_manager_class = getattr(module, class_name)
+        except AttributeError as e:
+            raise AttributeError(f"Module '{module_path}' does not have class '{class_name}': {e}") from e
+
+        manager = custom_manager_class(checkpoint_config)
```
g-husam, line 705: we would need to provide some additional config props to our custom checkpoint managers, which we can do by creating a subtype of `CheckpointConfig`, and using that in the recipe rather than `CheckpointConfig`. That looks like it should work fine, but would be great to get some confirmation/guarantee that subclassing will be supported, perhaps with documentation and some test case

docs/training/checkpointing.md
```diff
@@ -224,4 +224,145 @@ Local checkpointing leverages the [NVIDIA Resiliency Extension](https://nvidia.g
...
+
+checkpoint = CheckpointConfig(
+    save="/path/to/checkpoints",
+    custom_manager_class="mypackage.checkpoint.MyCheckpointManager",
+)
+```
+
+### Implementing a Custom Manager
+
+Your custom manager must implement the `CheckpointManager` protocol defined in `megatron.bridge.training.checkpointing`:
```
g-husam, line 313: is CheckpointManager a singleton that is reused throughout the training job? That would be needed for caching/referencing shared data across steps

src/megatron/bridge/training/setup.py
g-husam: similar q's -


1. I suspect cfg.checkpoint.load is expected to be the same as the save prop? i.e. the parent dir, not a specific checkpoint version? Is there any defaulting logic for it to be the same as save, or do users have to explicitly set it?

1. Is it possible to make this condition customizable, so that the checkpoint manager can determine if it can load or not, and then how to load? E.g. add an API to CheckpointManager like `should_load(...) -> bool` or something. That way it can also take into account any custom configuration, plus specific logic it wants to run.
   * There are 2 conditions I'm thinking of for loading a checkpoint - whether the user has configured loading (usually yes to auto handle recovering after an interruption, for example) and whether there is an available checkpoint to load from. Both have to be true to successfully load, otherwise either train from scratch or fail if the user wants to force loading from a checkpoint.

1. Where is the downstream logic that will use this? How does it know whether to train from scratch or from a pre-loaded checkpoint?

src/megatron/bridge/training/checkpointing.py
ananthsub: @g-husam yes, that is the intended way to customize this. I will update the documentation to make that clearer

docs/training/checkpointing.md
ananthsub: the same instance is shared across steps. the manager is created once before training begins

g-husam:
👋 Hey @ananthsub , just checking in on my prior questions and to see when this could be merged. Thanks!

g-husam:
> 👋 Hey @ananthsub , just checking in on my prior questions and to see when this could be merged. Thanks!

Hi @ananthsub, just checking in again on this. Thanks!

PR #1957, opened 2026-01-15
fix: guard config-only export paths

yaoyu-33:
## Summary
- guard export/save paths when bridge is config-only
- add tests for config-only export/save errors


cuichenx (approved):
LGTM

PR #1959, opened 2026-01-15
feat(mimo): phase 1 (config, builders, tests)

aroshanghias-nvd:
Here’s a filled‑in version you can paste into the PR template.

---

# What does this PR do ?

Add Phase‑1 MIMO scaffolding: config/validation, builder utilities, encoder provider interface, and unit tests.

# Changelog

- Add `mimo_config.py` with per‑module parallelism + deployment validation.
- Extend `ConfigContainer` with MIMO helpers, LLM parallelism cross‑validation, and encoder provider checks.
- Add `mimo_builder.py` for HyperCommGrid and MimoModelConfig construction.
- Add encoder provider interface + stub (`GenericVisionEncoderProvider`).
- Add MIMO config unit tests under `tests/unit_tests/training/mimo/`.
- Document `encoder_providers` YAML usage in `models/README.md`.

# GitHub Actions CI

**Pre checks**:

- [x] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [x] Reviewer: Does the PR have correct import guards for all optional libraries?

# Additional Information

- Related to # (issue)
None

### DIFF
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -1233,8 +1233,8 @@ class ConfigContainer(Container):
         return self.mimo is not None and self.mimo.deployment_mode == "homogeneous"
 
     @property
-    def is_separate_mimo(self) -> bool:
-        return self.mimo is not None and self.mimo.deployment_mode == "separate"
+    def is_heterogeneous_mimo(self) -> bool:
+        return self.mimo is not None and self.mimo.deployment_mode == "heterogeneous"
 
     def get_data_parallel_size(self, world_size: int) -> int:
         """Calculate the data parallel size based on the model configuration."""
--- a/src/megatron/bridge/training/mimo_config.py
+++ b/src/megatron/bridge/training/mimo_config.py
@@ -60,7 +60,7 @@ class MIMOConfig:
     llm_module_name: str
     module_parallelisms: dict[str, ModuleParallelismConfig]
     special_token_ids: dict[str, int] = field(default_factory=dict)
-    deployment_mode: Literal["colocated", "separate", "homogeneous"] = "colocated"
+    deployment_mode: Literal["colocated", "heterogeneous", "homogeneous"] = "colocated"
     # TODO: Add optional topology when supporting non-encoder-to-LLM flows.
 
     def get_parallelism(self, module_name: str) -> ModuleParallelismConfig:
@@ -104,13 +104,12 @@ class MIMOConfig:
             elif values != first:
                 raise ValueError("All modules must have identical parallelism in homogeneous deployment.")
 
-    def _validate_separate(self) -> None:
-        # We use "separate" to describe rank placement and avoid overloading
-        # "heterogeneous", which is already used elsewhere for layer-level configs.
+    def _validate_heterogeneous(self) -> None:
+        # "heterogeneous" describes rank placement across distinct modules.
         ranges = []
         for parallelism in self.module_parallelisms.values():
             if parallelism.data_parallel is None:
-                raise ValueError("data_parallel must be set for separate deployment.")
+                raise ValueError("data_parallel must be set for heterogeneous deployment.")
             ranges.append((parallelism.rank_offset, parallelism.rank_offset + parallelism.total_ranks))
 
         ranges.sort()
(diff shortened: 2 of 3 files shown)


src/megatron/bridge/models/encoder_provider.py
```diff
@@ -0,0 +1,50 @@
...
+    hidden_size: int
+    num_attention_heads: int
+    seq_length: int
+
+
+class EncoderProvider(ABC):
+    """Interface for encoder providers used in MIMO setups."""
+
+    @abstractmethod
+    def provide_model(self, pg_collection) -> object:
```
yashaswikarnati, line 24: curious if the return type would be nn.Module

src/megatron/bridge/models/encoder_provider.py
```diff
@@ -0,0 +1,50 @@
...
+        """Create the encoder module (unwrapped)."""
+
+    @abstractmethod
+    def get_transformer_layer_spec(self) -> ModuleSpec:
+        """Return the ModuleSpec for the encoder stack."""
+
+    @abstractmethod
+    def get_projection_spec(self) -> Optional[ModuleSpec]:
```
yashaswikarnati, line 32: nit; curious when the return would be None for Optional

src/megatron/bridge/training/config.py
```diff
@@ -35,6 +35,8 @@
 from megatron.bridge.training.mixed_precision import MixedPrecisionConfig, get_mixed_precision_config
 from megatron.bridge.training.tokenizers.config import TokenizerConfig
 from megatron.bridge.training.tokenizers.tokenizer import MegatronTokenizer
+from megatron.bridge.models.encoder_provider import EncoderProvider
+from megatron.bridge.training.mimo_config import MIMOConfig
```
yashaswikarnati, line 39: nit; MimoConfig - wondering if camelcase would look better for consistency but option to this option as well

src/megatron/bridge/training/mimo_config.py
```diff
@@ -0,0 +1,143 @@
...
+
+        if self.expert_parallel > 1 and self.pipeline_parallel > 1:
+            warnings.warn(
+                "Using expert_parallel > 1 with pipeline_parallel > 1 is complex and may be unsupported.",
+                stacklevel=2,
+            )
+
+
+@dataclass
+class MIMOConfig:
```
yashaswikarnati, line 57: [link](https://github.com/NVIDIA/Megatron-LM/blob/main/examples/mimo/model_providers/llava_vlm.py#L29-L117)

[link](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/models/mimo/config/base_configs.py#L11)

I was thinking we should refer to MimoConfig in MCore and usage in MLM for us to be consistent. also not sure it be a bit fragmented if encoder providers is a separate factory function, ideally MiMo config should encapsulate both the encoders configs/specs and language model configs/specs

src/megatron/bridge/training/mimo_config.py
```diff
@@ -0,0 +1,143 @@
...
+from dataclasses import dataclass, field
+from typing import Literal, Optional
+import warnings
+
+
+@dataclass
+class ModuleParallelismConfig:
+    """Parallelism config for a single module in a MIMO model."""
+
+    tensor_parallel: int = 1
```
yashaswikarnati, line 12: nit; tensor_parallel_size

src/megatron/bridge/models/encoder_provider.py
aroshanghias-nvd: You're right. Updated the return type.

src/megatron/bridge/models/encoder_provider.py
aroshanghias-nvd: When the encoder doesn't need a projector, e.g. it produces the same size as the input to the LLM.

src/megatron/bridge/training/config.py
aroshanghias-nvd: Changed all cases to MimoConfig.

src/megatron/bridge/training/config.py
yaoyu-33, line 1196: this is going to be upstreamed to mcore, and we finalized the fields. also this is shared in all models, not sure it makes sense to save for gpt model as well for example.

This seems to be both model config not an additional config to me.

aroshanghias-nvd: Thanks for the feedback! This is addressed in PR #2040, which refactors MIMO to use MimoModelProvider as a first-class model type (like GPTModelProvider) rather than an add-on config. That PR treats MIMO as a model provider pattern rather than a separate mimo field on ConfigContainer:
> model: GPTModelProvider | T5ModelProvider | MambaModelProvider | MimoModelProvider

PR #1994, opened 2026-01-20
Support direct export of block-wise FP8 weights and scaling factors

eternally-z:
# What does this PR do ?

This PR adds support for directly exporting block-wise **FP8 weights** and their corresponding inverse scales (`scale_inv`) when calling `export_hf_weights`.

### Motivation
In scenarios such as LLM Reinforcement Learning with FP8 training and rollout, the current workflow typically requires de-quantizing FP8 weights back to BF16 during export, only for them to be re-quantized during the rollout phase.

This "de-quantize -> re-quantize" process introduces:
1.  Redundant computational overhead.
2.  Potential precision loss due to repeated conversion.
3.  Increased transfer size: Exporting as BF16 consumes 2x the storage compared to the native FP8 format.

By allowing direct export of FP8 weights and scales, we enable a seamless and efficient pipeline from training to rollout.

### Implementation Details
1.  `export_hf_weights`:
    - Introduced a new function `build_export_fp8_tasks`.
    - this function allows the exporter to dump raw FP8 integer weights and `scale_inv` factors directly, bypassing the conversion to BF16.
    
2.  `load_hf_weights`:
    - Modified the loading process to retain the original `state_dict` when `fp8_param_gather` is enabled. 
    - This ensures that Megatron initializes the model using the **original BF16 weights**, rather than using the de-quantized results from FP8 parameters .

### Scope & Limitations
- Supported: Dense models & MoE models.
- Verification: Verified on Qwen3-8B (Dense) and Qwen3-30B-A3B (MoE).

@AniZpZ
# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

## Summary by CodeRabbit

* **New Features**
  * Introduced a new weight export format option providing enhanced control over model precision during conversion.
  * Added support for adapter weight conversion during model merging operations.

* **Improvements**
  * Enhanced weight conversion with improved handling of quantized parameters and scale data.
  * Strengthened QKV weight tensor handling to support diverse weight configurations during model transformation.

### DIFF
--- a/src/megatron/bridge/models/conversion/model_bridge.py
+++ b/src/megatron/bridge/models/conversion/model_bridge.py
@@ -45,6 +45,7 @@ from megatron.core.utils import (
     unwrap_model,
 )
 from rich.progress import BarColumn, Progress, TextColumn, TimeRemainingColumn
+from transformers.configuration_utils import PretrainedConfig
 from transformers.modeling_utils import PreTrainedModel
 
 from megatron.bridge.models.conversion.mapping_registry import MegatronMappingRegistry
@@ -316,6 +317,7 @@ class MegatronModelBridge(MegatronPeftBridge, Generic[HFPreTrained, ModelProvide
         ("num_experts_per_tok", "moe_router_topk"),
         ("moe_intermediate_size", "moe_ffn_hidden_size"),
         ("aux_loss_alpha", "moe_aux_loss_coeff"),
+        ("router_aux_loss_coef", "moe_aux_loss_coeff"),
         ("scoring_func", "moe_router_score_function"),
         ("n_routed_experts", "num_moe_experts"),
         ("n_group", "moe_router_num_groups"),
@@ -1188,6 +1190,9 @@ class MegatronModelBridge(MegatronPeftBridge, Generic[HFPreTrained, ModelProvide
                         expected_keys = hf_pretrained.state.source.get_all_keys()
                         if "lm_head.weight" in expected_keys:
                             yield HFWeightTuple("lm_head.weight", final_tensor.clone().detach())
+                    elif isinstance(hf_pretrained, PretrainedConfig):
+                        # Always emit lm_head.weight for config-only
+                        yield HFWeightTuple("lm_head.weight", final_tensor.clone().detach())
                 elif embeddings_are_tied and hf_name == "lm_head.weight":
                     # This should not happen when embeddings are tied - assert error
                     raise ValueError(
@@ -1389,12 +1394,11 @@ class MegatronModelBridge(MegatronPeftBridge, Generic[HFPreTrained, ModelProvide
         populated.
         """
 
-        # Ensure hf_pretrained has the required state structure
-        if not (hasattr(hf_pretrained, "state") and hasattr(hf_pretrained.state, "source")):
+        has_hf_state = hasattr(hf_pretrained, "state") and hasattr(hf_pretrained.state, "source")
+        if not isinstance(hf_pretrained, PretrainedConfig) and not has_hf_state:
             raise ValueError("hf_pretrained.state.source is required for weight ordering")
 
-        hf_keys: Iterable[str] = hf_pretrained.state.source.get_all_keys()
-
+        hf_keys: Optional[Iterable[str]] = hf_pretrained.state.source.get_all_keys() if has_hf_state else None
         mapping_registry = self.mapping_registry()
         unwrapped_model = unwrap_model(megatron_model)[0]
         model_config = unwrapped_model.config
@@ -1430,8 +1434,8 @@ class MegatronModelBridge(MegatronPeftBridge, Generic[HFPreTrained, ModelProvide
                     logger.warning(f"WARNING: No mapping found for megatron_param: {global_name}")
                     continue
 
-                # ensure hf weights exist
-                if not mapping.allow_hf_name_mismatch:
+                # Ensure hf weights exist (skip for config-only export where hf_keys is None)
+                if hf_keys is not None and not mapping.allow_hf_name_mismatch:
                     if isinstance(mapping.hf_param, str):
                         if mapping.hf_param not in hf_keys:
                             logger.warning(f"WARNING: Can't find {mapping.hf_param} in hf_keys")
@@ -1874,7 +1878,7 @@ def register_bridge_implementation(
     ) -> Iterable[HFWeightTuple]:
         bridge = bridge_class()
 
-        # allow bridge to access model config (config-only shims or raw configs lack .config)
+        # Allow bridge to access model config (config-only shims or raw configs lack .config)
         bridge.hf_config = hf_pretrained.config if hasattr(hf_pretrained, "config") else hf_pretrained
 
         return bridge.stream_weights_megatron_to_hf(
--- a/docs/adding-new-models.md
+++ b/docs/adding-new-models.md
@@ -159,6 +159,7 @@ Goal: Implement a bridge class that connects an HF model class to a Megatron mod
 Tasks:
 - Add `@MegatronModelBridge.register_bridge(source=<HFClass>, target=GPTModel)`.
 - Implement `provider_bridge(self, hf_pretrained)` to read `hf_pretrained.config` and return `YourModelProvider(...)` with mapped fields (layers, hidden size, FFN, heads, groups, RoPE, dtype via `self.dtype_from_hf(cfg)`).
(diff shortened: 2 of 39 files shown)


src/megatron/bridge/models/conversion/param_mapping.py
```diff
@@ -2292,8 +2292,37 @@ def split_qkv_weights(
         hidden_size = 1
         qkv_reshaped = qkv.view(qkv_total_dim, head_size)
     else:
-        hidden_size = qkv.shape[-1]
-        qkv_reshaped = qkv.view(qkv_total_dim, head_size, hidden_size)
+        # NOTE: For standard (BF16/FP16) weights, `head_size` is the usual kv_channels/head_dim.
```
yaoyu-33, line 2587: will fp8 affect other module layers, other than qkv?

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -1276,6 +1335,242 @@ def build_conversion_tasks(
...
+        for d in fp8_flags_list:
+            if not d:
+                continue
+            for k, v in d.items():
+                if v:
+                    global_fp8_flags[k] = True
+
+        return global_fp8_flags
+
+    def build_export_fp8_tasks(
```
yaoyu-33, line 1565: is this only for export, import will not work

src/megatron/bridge/models/conversion/param_mapping.py
eternally-z: Yes, besides QKV, we found that gdn_linear_weights and kv_weights are also affected.
The root cause is similar: the original code reshapes tensors using standard model dimensions (e.g., `qkvz = qkvz.reshape(-1, hidden_size)`). This logic fails for FP8 scale parameters because their shapes are determined by the block size (scaled down), causing dimension mismatches.
We verified this on Qwen3-8B and Qwen3-30B-A3B. For now, we have temporarily only updated the split logic for the QKV module to accommodate the FP8 scale dimensions, and we can easily apply this to gdn_linear_weights and kv_weights.

src/megatron/bridge/models/conversion/model_bridge.py
eternally-z: Yes, this PR specifically targets the export of block-wise FP8 weights and scales. Importing FP8 weights/scales is not supported at this stage.

yaoyu-33:
@eternally-z: The lint is failing, plz check Contribute.md, need to run pre-commit

eternally-z:
> /ok to test [fd9ba1d](https://github.com/NVIDIA-NeMo/Megatron-Bridge/commit/fd9ba1dd54b443984ef017e1d899d58fa4dddfc8)

@yaoyu-33  hi, we updated the access method for unquantized_state_dict.
Previously, we added it to the return values of load_weights_hf_to_megatron. However, this caused tests to fail. This made us realize that modifying the function's return signature breaks the tests and risks breaking other downstream usages.
To fix this, it is now stored and accessed as an attribute of the model_bridge instance instead.

eternally-z:
@yaoyu-33 hi, I noticed the CI passed but the codecov/patch failed. I'm not sure how to fix the problem. Should I add the tests myself?

gautham-kollu:
@eternally-z There is a codecov failure and merge conflict. Can you please resolve that ?

eternally-z:
> @eternally-z There is a codecov failure and merge conflict. Can you please resolve that ?

@gautham-kollu Yes, I noticed that. I have already resolved the merge conflict.
However, I'm not  sure how to handle the codecov/patch failure. It shows that the test coverage for my newly added code is insufficient. How should I go about fixing this? Should I add the tests myself?

cuichenx:
@eternally-z can you please add a unit test for fp8 export?

eternally-z:
> @eternally-z can you please add a unit test for fp8 export?

I've added a unit test for FP8 export. Please take a look.

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1994/commits/b32e0c0cc6a2f03d0c9c356f368c9950ae0d0deb

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -1398,6 +1465,255 @@ def build_conversion_tasks(
...
 
+    def _detect_fp8_params(
+        self,
+        megatron_model: List[MegatronModel],
+        model_config: TransformerConfig,
+        sorted_global_param_names_all_pp_ranks: List[str],
+        pp_group: Any,
+        fp8_scale_inv_attr: str,
+    ) -> Dict[str, bool]:
+        """Detect which global parameters are blockwise FP8 and gather flags across pipeline parallel ranks.
```
cuichenx, line 1499: does this not work with tensorwise fp8? if so we should guard it

cuichenx:
I'm curious how we should test this. Would the `hf_megatron_roundtrip.py` script for a model like [Qwen/Qwen3-4B-Instruct-2507-FP8](https://huggingface.co/Qwen/Qwen3-4B-Instruct-2507-FP8) work after this PR? Or does it require a checkpoint trained in FP8? If you could comment on what type of FP8 configs you verified (and which configs you think would not work) that would be great

src/megatron/bridge/models/conversion/model_bridge.py
eternally-z: > does this not work with tensorwise fp8? if so we should guard it

Yes, that is right. Currently, this implementation only supports blockwise FP8. I will add a guard check to explicitly handle this and update the PR soon. Thanks for pointing it out!

eternally-z:
> I'm curious how we should test this. Would the `hf_megatron_roundtrip.py` script for a model like [Qwen/Qwen3-4B-Instruct-2507-FP8](https://huggingface.co/Qwen/Qwen3-4B-Instruct-2507-FP8) work after this PR? Or does it require a checkpoint trained in FP8? If you could comment on what type of FP8 configs you verified (and which configs you think would not work) that would be great

Regarding the testing, it does not require a trained checkpoint in FP8. The conversion logic works as long as the Megatron configuration has the corresponding FP8 flags enabled during the process.
Verified/Supported Configuration:
- fp8: 'e4m3' or 'hybrid'
- fp8_recipe: 'blockwise'
- fp8_param(fp8_param_gather): True 

Unsupported Configurations:
Other configurations, such as the delayed scaling (tensorwise) recipe, are currently not supported. I will update the PR with the guarding code soon so that unsupported configs will be caught explicitly.

eternally-z:
Hi, @cuichenx
I have updated the PR with the guarding logic as discussed.
Specifically, I added a helper function `_validate_fp8_export_config` which is called within the export_hf_weight path. It ensures that the current Megatron configuration is compatible with the FP8 export requirements:
- fp8 must be enabled.
- fp8_recipe must be set to 'blockwise'.
- fp8_param must be True 

Any other configurations (such as delayed scaling / tensorwise) will now raise a descriptive ValueError to prevent incorrect exports.

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1994/commits/53c4d1dcfcef1e8ab833e594cf2d7717c08e6eb6

yaoyu-33:
@eternally-z your tests failed.

eternally-z:
> @eternally-z your tests failed.

@yaoyu-33  Sorry about that. I have fixed the tests in the latest commit. Could you please help re-trigger the CI?

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1994/commits/f05fffc3b34c2f4474e02ebe962a462d63741494

eternally-z:
@cuichenx hi,I noticed the Codecov patch coverage was below the 80% target, so I have refactored the test file to improve it.  Could you please trigger the CI again?

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1994/commits/af859d8ce04aa362f75d18b75764d08f3495fb90

eternally-z:
@cuichenx hi, the CI failed due to a timeout. The environment setup (building transformer-engine) consumed most of the time limit, leaving insufficient time for the tests to complete. It seems this failure is unrelated to the logic or changes in this PR. Could you please help re-run the CI?

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1994/commits/53d747578d8e2f7130f43fe67cc449f55a89a1e5

PR #2010, opened 2026-01-21
Refactor the optimizer override function so that users can swap in their own

jstjohn:
# What does this PR do ?

Makes the private `optimizer.py:_build_config_overrides` into a builder object that a user can swap in their config. This allows new models (eg evo2) to have their own custom override function without adding booleans/modifying the core function in mbridge optimizer.py.

# Changelog

- Adds OptimizerConfigOverrideProvider to the ConfigContainer that replaces the `_build_config_overrides` function.
- Adds missing handling of decoupled lr to the config override builder
- Adds note about how we can simplify further once https://github.com/NVIDIA/Megatron-LM/pull/2753 is pulled into `dev`.
- Builds off of https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1935
- Depended on by https://github.com/NVIDIA/bionemo-framework/pull/1429

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -276,7 +276,9 @@ class OptimizerConfigOverrideProviderContext:
 class OptimizerConfigOverrideProvider:
     """Abstract base class for providing config overrides."""
 
-    def build_config_overrides(self, context: OptimizerConfigOverrideProviderContext) -> dict[str, Any]:
+    def build_config_overrides(
+        self, context: OptimizerConfigOverrideProviderContext
+    ) -> dict[ParamKey, ParamGroupOverride] | None:
         """Build config overrides for weight decay based on scheduler configuration.
 
         This function creates parameter-specific overrides for weight decay behavior.


src/megatron/bridge/training/config.py
```diff
@@ -258,6 +263,81 @@ class DatasetBuildContext:
...
+            no_wd_key = ParamKey(name=tuple(no_wd_names))
+            config_overrides[no_wd_key] = ParamGroupOverride(wd_mult=0.0)
+
+        # Now handle decoupled LR:
+        if optimizer_config.decoupled_lr is not None:
+            decoupled_lr_config: ParamGroupOverride = {"max_lr": optimizer_config.decoupled_lr}
+            decoupled_param_key = ParamKey(attr="is_embedding_or_output_parameter")
+            if optimizer_config.decoupled_min_lr is not None:
+                decoupled_lr_config["min_lr"] = optimizer_config.decoupled_min_lr
+            config_overrides[decoupled_param_key] = decoupled_lr_config
```
jstjohn, line 336: @FDecaYed take a look here and make sure that this is ok for you. I pulled this from the default in mcore. If you don't want it, I don't need it, so I can remove it. I figured it wasn't intentionally left out though, just trying to be helpful by adding it in.

src/megatron/bridge/training/config.py
```diff
@@ -1163,6 +1243,9 @@ class ConfigContainer(Container):
     train: TrainingConfig
     model: GPTModelProvider | T5ModelProvider | MambaModelProvider
     optimizer: OptimizerConfig
+    optimizer_config_override_provider: OptimizerConfigOverrideProvider = field(
+        default_factory=OptimizerConfigOverrideProvider
+    )
```
jstjohn, line 1248: This is the key thing for me as a 3rd party user, that I can swap in a custom override class (possibly with configuration options).

src/megatron/bridge/training/optim.py
```diff
@@ -91,8 +48,13 @@ def setup_optimizer(
     Returns:
         tuple containing the optimizer and scheduler
     """
+    if optimizer_config_override_provider is None:
+        optimizer_config_override_provider = OptimizerConfigOverrideProvider()
+
```
jstjohn, line 53: This should handle users of the previous API where they did not provide this function override. In that case it uses the default that we use in training.

src/megatron/bridge/training/config.py
FDecaYed: from my understanding it was simply not used in mbridge before initial opt refactor, except in logging. so I just removed it in the initial PR support it and didn't bother add it back. it looks ok add here, thanks

jstjohn: No problem!

PR #2038, opened 2026-01-22
Set HF_HUB_OFFLINE=1 by default, enabled with '--hf_token' flag.

sudostock:
This makes it consistent with the TRANSFORMER_OFFLINE variable which is offline by default unless '--hf_token' is provided.

As right now recipes are still 'online' even if '--hf_token' is not passed.


sudostock:
@bdubauski I agree in principle but when they hit rate limits they still file bugs with us. I think we should encourage / require a token for the recipes.

PR #2039, opened 2026-01-22
fix: Packed sequence performance regression and missing pad_seq_to_mult parameter

yaoyu-33:
## Summary
When `pad_seq_to_mult == 1` (no padding for CP), skip computing `cu_seqlens_unpadded` to avoid triggering the slower TE kernel path. Only compute and include `cu_seqlens_unpadded` when `pad_seq_to_mult > 1` (actual padding for THD CP).

## Changes
- Add `pad_seq_to_mult` parameter to `GPTSFTPackedDataset`
- Conditionally compute `cu_seqlens_unpadded` based on `pad_seq_to_mult`
- Update `get_packed_seq_params` to only set padded variants when unpadded is present

## Files Modified
- `src/megatron/bridge/data/datasets/sft.py`
- `src/megatron/bridge/training/utils/packed_seq_utils.py`

### DIFF
--- a/tests/unit_tests/data/datasets/test_chat_template.py
+++ b/tests/unit_tests/data/datasets/test_chat_template.py
@@ -1011,6 +1011,7 @@ def _create_minimal_packed_dataset(tokenizer_eos_id: int = 0):
     dataset.pad_to_max_length = False
     dataset.max_seq_length = 32
     dataset.pad_seq_length_to_mult = 1
+    dataset._pad_seq_to_mult = 2  # Used in collate_fn for cu_seqlens_unpadded check (must be > 1 to compute)
     dataset.ceil_to_power_2 = False
     dataset.tokenizer = SimpleNamespace(eos_id=tokenizer_eos_id)
     dataset.answer_only_loss = False
--- a/tests/unit_tests/data/datasets/test_sft.py
+++ b/tests/unit_tests/data/datasets/test_sft.py
@@ -15,6 +15,7 @@
 import datetime
 import json
 import os
+from unittest.mock import MagicMock
 
 import megatron.core.parallel_state as parallel_state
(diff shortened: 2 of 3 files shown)


src/megatron/bridge/training/utils/packed_seq_utils.py
```diff
@@ -56,14 +56,23 @@ def get_packed_seq_params(batch: dict[str, torch.Tensor]) -> PackedSeqParams:
...
+            qkv_format="thd",
+        )
+    else:
+        return PackedSeqParams(
+            cu_seqlens_q=cu_seqlens_padded,
+            cu_seqlens_kv=cu_seqlens_padded,
+            max_seqlen_q=max_seqlen,
+            max_seqlen_kv=max_seqlen,
+            qkv_format="thd",
+        )
```
cuichenx, line 78: nit: is this change necessary? as long as cu_seqlens_unpadded is None, passing None to PackedSeqParams should also be fine

src/megatron/bridge/data/datasets/sft.py
```diff
@@ -98,6 +98,7 @@ def create_sft_dataset(
     get_attention_mask_from_fusion: bool = True,
     pack_metadata_file_path: Path = None,
     pad_cu_seqlens: bool = False,
+    pad_seq_to_mult: int = 1,
```
cuichenx, line 101: should you update the caller of `create_sft_dataset` in `finetuning_datasets.py` as well?

src/megatron/bridge/training/utils/packed_seq_utils.py
yaoyu-33: i want to make it more clear here since we encounter the slow issue. just to make more explicit.

PR #2040, opened 2026-01-22
refactor(mimo): unified MimoModelProvider with ModuleSpec-based API

aroshanghias-nvd:
## Summary

This PR introduces a unified `MimoModelProvider` that accepts `ModuleSpec`s directly, removing the `EncoderProvider` abstraction layer.

### Key Changes
- Add `MimoModelProvider` accepting language_model_spec and modality_submodules_spec
- Remove `EncoderProvider` abstraction
- Rename `MIMOConfig` → `MimoParallelismConfig`
- Fix `ColocatedCommConfig` logic for colocated and homogeneous modes

### Tests
✅ All MIMO tests passing

## Release Notes

* **New Features**
  * Added support for multi-module MIMO (vision-language) models with heterogeneous parallelism configurations enabling different parallel strategies per module.
  * Introduced pre-configured LLaVA vision-language model provider for integrated vision-language training.
  * Extended training configuration system to support MIMO models.

* **Tests**
  * Added comprehensive unit tests for MIMO configuration and provider functionality.

### DIFF
--- a/src/megatron/bridge/models/mimo/llava_provider.py
+++ b/src/megatron/bridge/models/mimo/llava_provider.py
@@ -45,7 +45,7 @@ class LlavaMimoProvider(MimoModelProvider):
 
     # Optional custom configs
     language_config: Optional[TransformerConfig] = None
-    
+
     # Make parent's required field optional (we build it in __post_init__)
     language_model_spec: Optional[ModuleSpec] = None
 
--- a/tests/unit_tests/models/mimo/test_mimo_provider.py
+++ b/tests/unit_tests/models/mimo/test_mimo_provider.py
@@ -807,7 +807,7 @@ class TestMimoModelProviderDistributed:
         mock_device_count.return_value = 8  # Mock sufficient GPUs
         mock_set_device.return_value = None  # Mock set_device to avoid CUDA calls
         mock_is_available.return_value = True  # Mock CUDA availability
-        
+
         # Mock the stream context manager
         mock_stream_ctx.return_value.__enter__ = MagicMock(return_value=None)
         mock_stream_ctx.return_value.__exit__ = MagicMock(return_value=None)
@@ -846,9 +846,7 @@ class TestMimoModelProviderDistributed:
         )
 
         ddp_config = DistributedDataParallelConfig()
-        provider.provide_distributed_model(
-            ddp_config=ddp_config, wrap_with_ddp=True, data_parallel_random_init=True
-        )
+        provider.provide_distributed_model(ddp_config=ddp_config, wrap_with_ddp=True, data_parallel_random_init=True)
 
         # Should wrap with DDP
         assert mock_ddp.called
@@ -989,9 +987,7 @@ class TestMimoModelProviderDistributed:
 
         provider = MimoModelProvider(language_model_spec=language_spec, fp16=True)
 
-        result = provider.provide_distributed_model(
-            wrap_with_ddp=False, mixed_precision_wrapper=mock_custom_wrapper
-        )
+        result = provider.provide_distributed_model(wrap_with_ddp=False, mixed_precision_wrapper=mock_custom_wrapper)
 
         # Should use custom wrapper instead of Float16Module
         mock_custom_wrapper.assert_called_once_with(mock_config, mock_model)
@@ -1122,7 +1118,7 @@ class TestMimoModelProviderDistributed:
         mock_device_count.return_value = 8  # Mock sufficient GPUs
         mock_set_device.return_value = None  # Mock set_device to avoid CUDA calls
         mock_is_available.return_value = True  # Mock CUDA availability
-        
+
         # Mock the stream context manager
         mock_stream_ctx.return_value.__enter__ = MagicMock(return_value=None)
         mock_stream_ctx.return_value.__exit__ = MagicMock(return_value=None)
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,28 @@
+{                                                                                                                                                                                      
+  "permissions": {                                                                                                                                                                     
+    "alwaysAllow": [                                                                                                                                                                         
+      "WebFetch(domain:github.com)",                                                                                                                                                   
+      "WebFetch(domain:api.github.com)",                                                                                                                                               
+      "Bash(gh *)",  
+      "Bash(gh * | jq *)",                                                                                                                                                        
+      "Bash(ls *)",                                                                                                                                                                    
+      "Bash(cat *)",                                                                                                                                                                   
+      "Bash(head *)",                                                                                                                                                                  
+      "Bash(tail *)",                                                                                                                                                                  
+      "Bash(tree *)",                                                                                                                                                                  
+      "Bash(grep *)",                                                                                                                                                                  
+      "Bash(find *)",                                                                                                                                                                  
+      "Bash(which *)",                                                                                                                                                                 
+      "Bash(cd *)",                                                                                                                                                                    
+      "Bash(pwd *)",                                                                                                                                                                   
+      "Bash(sed *)",                                                                                                                                                                   
+      "Bash(du *)",                                                                                                                                                                    
+      "Bash(df *)",                                                                                                                                                                    
+      "Bash(date *)",                                                                                                                                                                  
+      "Bash(git status *)",                                                                                                                                                            
+      "Bash(git log *)",                                                                                                                                                               
+      "Bash(git diff *)",                                                                                                                                                              
+      "Bash(git branch *)"                                                                                                                                                             
+    ]                                                                                                                                                                                  
+  }                                                                                                                                                                                    
+}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,673 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Repository Overview
+
+**NeMo Megatron Bridge** is a PyTorch-native library for training and converting LLM/VLM models. It serves as a bridge between 🤗 Hugging Face and Megatron Core, enabling:
+- Bidirectional checkpoint conversion (HF ↔ Megatron)
+- High-performance distributed training with tensor/pipeline/context/expert parallelism
+- Pre-configured training recipes for popular models
+- PEFT methods (LoRA, DoRA) for efficient fine-tuning
+
+## Build and Development Commands
+
+### Environment Setup
+```bash
+# Using NeMo Framework container (recommended)
+docker run --rm -it -w /workdir -v $(pwd):/workdir \
+  --entrypoint bash --gpus all \
+  nvcr.io/nvidia/nemo:${TAG}
+
+# Or build development container
+docker build -f docker/Dockerfile.ci -t megatron-bridge .
+docker run --rm -it -w /workdir -v $(pwd):/opt/Megatron-Bridge \
+  --entrypoint bash --gpus all megatron-bridge
+
+# For local development, install with uv
+uv sync
+```
+
+### Testing
+```bash
+# Run all tests (linting, unit, functional)
+bash scripts/run_ci_tests.sh
+
+# Run only unit tests
+uv run pytest tests/unit_tests/ -v
+
+# Run specific test file
+uv run pytest tests/unit_tests/training/test_model_load_save.py -v
+
+# Run specific test function
+uv run pytest tests/unit_tests/training/test_model_load_save.py::test_function_name -v
+
+# Run functional tests (requires GPUs)
+uv run pytest tests/functional_tests/ -v
+
+# Run with coverage
+uv run pytest --cov=megatron.bridge --cov-report=html tests/
+
+# Run tests with specific markers
+uv run pytest -m unit  # Only unit tests
+uv run pytest -m integration  # Only integration tests
+```
+
+### Linting and Formatting
+```bash
+# Auto-fix linting and formatting issues
+uv run ruff check --fix .
+uv run ruff format .
+
+# Run pre-commit hooks manually
+uv run pre-commit run --all-files
+
+# Install pre-commit hooks
+uv run pre-commit install
+```
+
+### Dependency Management
+```bash
+# Add new required dependency
+uv add <package-name>
+
+# Add optional dependency to specific group
+uv add --optional --extra <group-name> <package-name>
+
+# Update dependencies (regenerates uv.lock)
+# Commit both pyproject.toml and uv.lock after changes
+```
+
+### Documentation
+```bash
+# Build docs locally
+cd docs
+uv run sphinx-build -b html . _build
+
+# Live-reload docs during development
+uv run sphinx-autobuild . _build
+```
+
+### Running Training
+```bash
+# Single-node training with torchrun
+torchrun --nproc-per-node=8 /path/to/train_script.py
+
+# Multi-node training (example with slurm)
+srun --ntasks-per-node=8 --gpus-per-node=8 python /path/to/train_script.py
+```
+
+## Architecture Overview
+
+### Core Abstractions
+
+#### 1. **AutoBridge** - Conversion Entry Point
+The main user-facing API for HF ↔ Megatron conversion. Auto-detects model architectures.
+
+```python
+from megatron.bridge import AutoBridge
+
+# Load HF model and get bridge
+bridge = AutoBridge.from_hf_pretrained("meta-llama/Llama-3.2-1B")
+
+# Convert to Megatron provider and configure parallelism
+provider = bridge.to_megatron_provider()
+provider.tensor_model_parallel_size = 8
+provider.pipeline_model_parallel_size = 4
+provider.finalize()
+
+# Instantiate model with weights loaded
+model = provider.provide_distributed_model(wrap_with_ddp=False)
+
+# Export back to HF
+bridge.save_hf_pretrained(model, "./exported_model")
+```
+
+**Key files:**
+- `src/megatron/bridge/models/conversion/auto_bridge.py`
+- `src/megatron/bridge/models/conversion/model_bridge.py`
+
+#### 2. **ModelProvider** - Model Instantiation Pattern
+Standardized pattern for creating Megatron models with distributed training support.
+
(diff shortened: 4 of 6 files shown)


src/megatron/bridge/models/mimo/mimo_provider.py
```diff
@@ -0,0 +1,303 @@
...
+    
+    # Parallelism config (Bridge's value-add)
+    mimo_parallelism_config: Optional[MimoParallelismConfig] = None
+    
+    # Freezing options
+    freeze_language_model: bool = False
+    freeze_modality_encoders: Dict[str, bool] = field(default_factory=dict)
+    freeze_modality_projections: Dict[str, bool] = field(default_factory=dict)
+    
+    def _extract_pg_collections_from_grids(
```
yashaswikarnati: minor nit; may be get_* instead of extract_*

src/megatron/bridge/training/config.py
```diff
@@ -1162,6 +1199,7 @@ class ConfigContainer(Container):
     rerun_state_machine: RerunStateMachineConfig = field(default_factory=RerunStateMachineConfig)
     train: TrainingConfig
     model: GPTModelProvider | T5ModelProvider | MambaModelProvider
```
yashaswikarnati: MiMoModelProvider

src/megatron/bridge/training/config.py
```diff
@@ -1207,6 +1264,53 @@ def set_data_parallel_size(self) -> None:
...
 
+    def get_module_data_parallel_size(self, module_name: str) -> int:
+        """Return data parallel size for a specific MIMO module."""
+        if self.mimo is None:
+            raise ValueError("MIMO configuration is not set.")
+        world_size = get_world_size_safe()
+        self.mimo.finalize(world_size if world_size and world_size > 1 else None)
+        return self.mimo.get_parallelism(module_name).data_parallel
+
+    def _validate_mimo(self) -> None:
```
yashaswikarnati: lets may be move to MiMoModelProvider if it makes sense

src/megatron/bridge/models/mimo/mimo_builder.py
```diff
@@ -0,0 +1,68 @@
+from __future__ import annotations
+
+from typing import Dict, List, Optional
+
+from megatron.bridge.training.mimo_config import MimoParallelismConfig
+
+
+def build_hypercomm_grids(
```
yashaswikarnati, line 14: may be move to mimo folder.

src/megatron/bridge/models/mimo/mimo_provider.py
aroshanghias-nvd: Done.

src/megatron/bridge/training/config.py
aroshanghias-nvd: MimoModelProvider is probably not the best location for this. This validation depends on train and model configs (global batch size, model parallel sizes), which are outside the provider’s responsibility. Putting it there introduces inverted dependencies (config → provider), makes config validation happen late (after you’re already building specs/models), and reduces reuse for config‑only flows (CI checks, static validation). A better fit is either 1) leave it here as is, 2) MimoParallelismConfig (with validate_against(model_cfg, train_cfg, world_size) and then call that within config validate)

src/megatron/bridge/models/mimo/mimo_builder.py
aroshanghias-nvd: Moved to mimo folder.

src/megatron/bridge/training/config.py
aroshanghias-nvd: Added.

yashaswikarnati:
lgtm

yashaswikarnati:
@yaoyu-33 / @maanug-nv  could one of you help review

src/megatron/bridge/training/config.py
```diff
@@ -1207,6 +1249,53 @@ def set_data_parallel_size(self) -> None:
         if self.comm_overlap is not None:
             self.comm_overlap.data_parallel_size = self.data_parallel_size
 
+    def get_module_data_parallel_size(self, module_name: str) -> int:
```
yashaswikarnati: I think we can skip this here, this is needed for enabling overlap grad reduce etc, which is part of DDP config, and can be handled when we wrap the model object with DDP. We don't have to change this here for brevity

src/megatron/bridge/models/mimo/mimo_config.py
```diff
@@ -0,0 +1,143 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Literal, Optional
+import warnings
+
+
+@dataclass
+class ModuleParallelismConfig:
```
yashaswikarnati, line 11: I think we should move this inside the mimo folder

aroshanghias-nvd: Makes sense. Moved.

src/megatron/bridge/training/config.py
aroshanghias-nvd: Makes sense. Removed.

src/megatron/bridge/models/mimo/mimo_provider.py
```diff
@@ -0,0 +1,303 @@
...
+    """
+    
+    # Model specs (user provides, like llava_vlm.py example)
+    # Note: Each key in modality_submodules_spec should match a module name
+    # in mimo_parallelism_config.module_parallelisms for per-encoder parallelism
+    language_model_spec: ModuleSpec
+    modality_submodules_spec: Dict[str, ModuleSpec] = field(default_factory=dict)
+    special_token_ids: Dict[str, int] = field(default_factory=dict)
+    
+    # Parallelism config (Bridge's value-add)
```
yashaswikarnati, line 104: nit; we could remove this comment

src/megatron/bridge/training/config.py
aroshanghias-nvd: This is outdated, and addressed in this commit: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2040/changes/acc6ec744509fd7a85e40cf25315462e0f0ab674

src/megatron/bridge/models/mimo/mimo_config.py
```diff
@@ -0,0 +1,143 @@
...
+    """Parallelism config for a single module in a MIMO model."""
+
+    tensor_parallel: int = 1
+    pipeline_parallel: int = 1
+    context_parallel: int = 1
+    expert_parallel: int = 1
```
yaoyu-33: do we support expert tp?

yaoyu-33: tensor_parallel are these sizes? we should keep the naming consistent with tranformer config

src/megatron/bridge/models/mimo/mimo_config.py
```diff
@@ -0,0 +1,143 @@
...
+                "Using expert_parallel > 1 with pipeline_parallel > 1 is complex and may be unsupported.",
+                stacklevel=2,
+            )
+
+
+@dataclass
+class MimoParallelismConfig:
+    """Configuration for multi-module (MIMO) heterogeneous parallelism."""
+
+    llm_module_name: str
```
yaoyu-33: why llm is special?

src/megatron/bridge/models/mimo/mimo_provider.py
```diff
@@ -0,0 +1,339 @@
...
+        
+        mimo_model = MimoModel(mimo_model_config)
+        
+        # Step 7: Apply freezing
+        self._apply_freezing(mimo_model)
+        
+        # Step 8: Move to device/dtype
+        mimo_model.to(device).to(dtype)
+        
+        return MimoModelProviderResult(
```
yaoyu-33: we want to keep provide only yield a model, so we can blend in train loop.
Maybe have another method to yield meta data and keep provider only yield model instance?
Otherwise it's confusing design.

src/megatron/bridge/models/mimo/mimo_config.py
aroshanghias-nvd: Updated naming to match transformer config.

aroshanghias-nvd: Also, Yash's prototype doesn’t really do EP: it’s defined but not used in sizing (total_ranks ignores EP) and tests hard‑code ep=1. So EP is currently placeholder/wiring only in the configs.

src/megatron/bridge/models/mimo/mimo_config.py
aroshanghias-nvd: The canonical MiMo always has an LLM as the middle component, hence the special treatment for topology.

src/megatron/bridge/models/mimo/mimo_provider.py
aroshanghias-nvd: Completely rewrote mimo_provider.py. Now it has mixin inheritance, provide() returns model directly, new build_infra(), provide_distributed_model() override, MimoStubModel for non-participating ranks.

src/megatron/bridge/models/mimo/llava_provider.py
```diff
@@ -0,0 +1,137 @@
...
+            },
+        )
+        
+        # Build vision modality spec
+        self.modality_submodules_spec = {
+            "images": self._build_vision_submodule_spec()
+        }
+        
+        # Set special token IDs
+        self.special_token_ids = {"images": self.image_special_token_id}
```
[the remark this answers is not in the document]
aroshanghias-nvd: MimoModelProvider has no __post_init__ method, so there's nothing to call.

src/megatron/bridge/models/mimo/llava_provider.py
```diff
@@ -0,0 +1,137 @@
...
+            module=self.vision_encoder_module,
+            params=self.vision_encoder_params,
+        )
+        
+        # Vision projector (2-layer MLP)
+        projection_config = TransformerConfig(
+            num_layers=2,
+            hidden_size=self.language_config.hidden_size,
+            num_attention_heads=1,
+        )
```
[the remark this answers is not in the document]
aroshanghias-nvd: TransformerConfig requires num_attention_heads as a mandatory field (no default value). Removing it would cause an instantiation error. The value 1 is a placeholder required by the config schema, even though the MLP projector doesn't use attention.

tests/unit_tests/models/mimo/test_mimo_provider.py
```diff
@@ -0,0 +1,334 @@
...
+        
+        provider = MimoModelProvider(
+            language_model_spec=language_spec,
+            freeze_language_model=True,
+        )
+        
+        result = provider.provide()
+        
+        # Check parameter was frozen
+        assert mock_param.requires_grad is False
```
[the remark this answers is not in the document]
aroshanghias-nvd: This is intentional. The test verifies the side effect of provide() (freezing parameters), not the return value. The call to provide() is necessary to trigger the freezing logic.

src/megatron/bridge/models/mimo/mimo_provider.py
aroshanghias-nvd: Please review, and let me know if it is now compatible with the train loop.

src/megatron/bridge/models/mimo/mimo_builder.py
```diff
@@ -0,0 +1,68 @@
...
+    return {name: [llm] for name in mimo_parallelism_config.module_names if name != llm} | {llm: []}
+
+
+def build_colocated_comm_config(
+    mimo_parallelism_config: MimoParallelismConfig, grids: Dict[str, "HyperCommGrid"]
+) -> "ColocatedCommConfig":
+    """Build ColocatedCommConfig with default encoder-to-LLM topology."""
+    from megatron.core.models.mimo.config.base_configs import ColocatedCommConfig
+
+    module_to_grid_map = {name: grid for name, grid in grids.items()}
```
maanug-nv: nit, this seems the same as `grids` , is it meant to be a copy? does `.copy()` not work?

src/megatron/bridge/models/mimo/mimo_provider.py
```diff
@@ -0,0 +1,598 @@
...
+        >>> # For training loop integration:
+        >>> model = provider.provide_distributed_model(ddp_config=ddp_config)
+        >>> # Or for manual usage:
+        >>> model = provider.provide()
+        >>> infra = provider.build_infra()
+    """
+    
+    # Model specs (user provides, like llava_vlm.py example)
+    language_model_spec: ModuleSpec
+    modality_submodules_spec: Dict[str, ModuleSpec] = field(default_factory=dict)
```
maanug-nv, line 101: Would like to understand the design choice of using a string label for each module. seems this way the module state/components (spec, parallelism config, etc.) are broken up into separate dicts. 
to me, it would make more sense to bundle the contents of a single module into a `MimoModuleWrapper` object that contains spec, parallelism config, commgrid, etc. 

not asking for a refactor, just trying to discuss for my understanding.

src/megatron/bridge/models/mimo/mimo_provider.py
```diff
@@ -0,0 +1,598 @@
...
+    
+    def set_input_tensor(self, input_tensor):
+        """No-op for pipeline parallel compatibility."""
+        pass
+
+
+
+
+@dataclass
+class MimoModelProvider(ModelProviderMixin[MimoModel]):
```
maanug-nv, line 63: im quite ignorant on MIMO design, but is there no need for a dict of `TransformerConfig`s per module?

src/megatron/bridge/models/mimo/mimo_provider.py
```diff
@@ -0,0 +1,598 @@
...
+        )
+        
+        mimo_model = MimoModel(mimo_model_config)
+        
+        # Apply freezing
+        self._apply_freezing(mimo_model)
+        
+        return mimo_model
+
+    def provide_distributed_model(
```
maanug-nv, line 305: @yaoyu-33 just noting that we already have a use case for overriding VP/DDP model wrapping logic, so it needs to be easily customizable in the new design.

maanug-nv (dismissed):
overall don't see any issues. but left several comments to hopefully improve my understanding

src/megatron/bridge/models/mimo/mimo_builder.py
aroshanghias-nvd: Good catch! Changed to grids.copy().

src/megatron/bridge/models/mimo/mimo_provider.py
aroshanghias-nvd: The string-keyed dict follows Megatron-Core's existing MimoModelConfig.modality_submodules_spec: Dict[str, ModuleSpec] API - the provider passes it directly without translation. 
Extending this pattern to MimoParallelismConfig keeps the interface consistent. A MimoModuleWrapper could work as a convenience layer on top if needed, but the current design keeps components decoupled (e.g., specs can be used without parallelism config for testing).

src/megatron/bridge/models/mimo/mimo_provider.py
aroshanghias-nvd: Each ModuleSpec carries its config in params['config'] (e.g., language_model_spec.params['config'] holds the LLM's TransformerConfig). The projector spec has its own config too. Some modules like HF wrappers manage config internally. So there's no need for a separate Dict[str, TransformerConfig] at the provider level - configs live within their respective specs.

aroshanghias-nvd:
Ready for review - branch is up to date with main.

**Latest changes:**
- Removed `ColocatedCommConfig` dependency and colocated/homogeneous modes from Phase 1
- Simplified to heterogeneous-only deployment to make the PR self-contained (no dependency on unmerged Megatron-LM features)

Colocated/homogeneous modes will be added once we have a new design for colocated/homogeneous.

@yaoyu-33 @maanug-nv - could you please review when you have a chance?

Also, the DCO check is failing because an early commit (c7b06c41) is missing the sign-off trailer. Rebasing to fix it causes merge conflicts with the long history. Could this PR be squash-merged instead? That would create a single properly signed commit.

aroshanghias-nvd:
/ok to test [223fb37](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2040/commits/223fb37ff53c588e0199f35eaac17564b5adf6b2)

yashaswikarnati:
/ok to test [223fb37](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2040/commits/223fb37ff53c588e0199f35eaac17564b5adf6b2)

aroshanghias-nvd:
/ok to test [9fb76b3](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2040/commits/9fb76b33ad98858e6c555181afa4f6e9b35e3e1c)

yaoyu-33 (dismissed):

aroshanghias-nvd:
/ok to test [9fb76b3](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2040/commits/9fb76b33ad98858e6c555181afa4f6e9b35e3e1c)

aroshanghias-nvd:
/ok to test [9fb76b3](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2040/commits/9fb76b33ad98858e6c555181afa4f6e9b35e3e1c)

aroshanghias-nvd:
/ok to test [ed75bae](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2040/commits/ed75baeff4d523e915d06405f115adc1cb9b6ce4)

aroshanghias-nvd:
/ok to test [1527a1b](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2040/commits/1527a1bc54c7cc0d159911336d48a6d03a98c761)

aroshanghias-nvd:
/ok to test [a7c46c9](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2040/commits/a7c46c9e3f8eafda6f8e16d7efa3dc02b490e052)

aroshanghias-nvd:
/ok to test [1ed7296](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2040/commits/1ed72960db65d30e23ef575d3d8d3a2cb2236aec)

Issue #1728, opened 2025-12-15
NVFP4 support DSv3 and QWEN3


erhoo82:
Add NVFP4 recipes of DSv3 and QWEN3

erhoo82:
@sanandaraj5597 Can we use NVFP4 for MoE using the staged TE and MCORE codes for 26.02?

sanandaraj5597:
Yes, let me work with @dingqingy-nv for this.

dingqingy-nv:
I will try it in RC1 container and let @sanandaraj5597 know

erhoo82:
@sanandaraj5597 Does 26.02 RC1 container have everything to run fine-grain Moe with NVFP4?

sanandaraj5597:
I think we have everything ready, Dingqing is testing it.

erhoo82:
QWEN3 test pushed to 26.02 patch

PR #2076, opened 2026-01-27
DSV3 NVFP4 recipe on GB300

dingqingy-nv:
# What does this PR do ?

This PR adds DeepSeek V3 NVFP4 recipe configuration for GB300. 

Achieves 1185 TFLOP (V1) and 1220 TFLOPs (V2).

Update: Fast Math improves V2 to 1240 TFLOPs.

To test GB200.

This PR also add `pp_layout` field to WorkloadBaseConfig to support custom pipeline parallelism layouts. This is needed because different workload configs (e.g., V1, V2, large_scale) share the same `deepseek_v3_pretrain_config_gb300` function but may require different PP layouts—for instance, large-scale runs prefer FSDP and work better with `pp_layout=None`, so this must be configurable at the workload base config level.

# Additional Information

- Related to #1728 

## Summary by CodeRabbit

* **New Features**
  * Added GB300 NVFP4 pretraining configurations (V1 and V2 variants) with configurable batch sizes and pipeline topology.
  * Introduced pipeline parallelism layout configuration support.

* **Improvements**
  * Extended compute dtype support to include NVFP4 for optimizer parameter gathering optimization.

### DIFF
--- a/scripts/performance/configs/deepseek/deepseek_workload_base_configs.py
+++ b/scripts/performance/configs/deepseek/deepseek_workload_base_configs.py
@@ -56,15 +56,11 @@ DEEPSEEK_V3_PRETRAIN_CONFIG_GB300_FP8_CS_V1 = DEEPSEEK_V3_PRETRAIN_CONFIG_GB300_
 DEEPSEEK_V3_PRETRAIN_CONFIG_GB300_FP8_MX_V1 = DEEPSEEK_V3_PRETRAIN_CONFIG_GB300_V1
 DEEPSEEK_V3_PRETRAIN_CONFIG_GB300_NVFP4_V1 = replace(
     BASE_DEEPSEEK_V3_CONFIG,
-    num_gpus=256,
-    global_batch_size=2048,
     micro_batch_size=2,
     pipeline_model_parallel_size=2,
     virtual_pipeline_model_parallel_size=8,
     pp_layout="Et*4|(t*4|)*14tmL",
     expert_model_parallel_size=32,
-    moe_flex_dispatcher_backend="hybridep",
-    moe_a2a_overlap=False,
     cuda_graph_scope=[],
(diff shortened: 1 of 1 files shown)


ko3n1g:
/ok to test [1afc195](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2076/commits/1afc195053cd7b41beea9840036381ec1138b14b)

scripts/performance/configs/deepseek/deepseek_workload_base_configs.py
```diff
@@ -54,7 +54,20 @@
...
+    global_batch_size=2048,
+    micro_batch_size=2,
+    pipeline_model_parallel_size=2,
+    virtual_pipeline_model_parallel_size=8,
+    pp_layout="Et*4|(t*4|)*14tmL",
+    expert_model_parallel_size=32,
+    moe_flex_dispatcher_backend="hybridep",
+    moe_a2a_overlap=False,
+    cuda_graph_scope=[],
+    recompute_modules=["mla_up_proj"],
```
erhoo82: We need to override only the changes?
I see that most of the lines remain the same as the baseline.

scripts/performance/utils/utils.py
```diff
@@ -62,6 +62,9 @@ class WorkloadBaseConfig:
     moe_a2a_overlap: Optional[bool] = False
     peft: Optional[str] = None
 
+    # Pipeline parallelism layout
+    pp_layout: Optional[str] = None
```
erhoo82, line 66: This is a nice change to make the PP layout configurable.

scripts/performance/configs/deepseek/deepseek_workload_base_configs.py
dingqingy-nv: Thanks, updated

erhoo82 (approved):

PR #2185, opened 2026-02-03
Add chat template from HF tokenizer

cuichenx:
# What does this PR do ?
This PR adds support for applying HuggingFace tokenizer chat templates during dataset preprocessing for supervised fine-tuning (SFT). When enabled, the tokenizer's apply_chat_template method is used to format input/output pairs according to the model's expected chat format, ensuring proper special tokens and formatting are applied during data preprocessing.

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information
Two runs comparing finetuning with and without chat template. It's expected to see different initial loss but similar converged loss.
image

- Related to # (issue)

### DIFF
--- a/src/megatron/bridge/training/tokenizers/tokenizer.py
+++ b/src/megatron/bridge/training/tokenizers/tokenizer.py
@@ -2,12 +2,14 @@
 
 """Megatron tokenizers."""
 
+from __future__ import annotations
+
 import base64
 import json
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import TYPE_CHECKING, Dict, List, Optional
 
-from megatron.core.tokenizers import MegatronTokenizer, MegatronTokenizerBase
+from megatron.core.tokenizers import MegatronTokenizer
 
 from megatron.bridge.training.tokenizers.bert_tokenization import FullTokenizer as FullBertTokenizer
 from megatron.bridge.training.tokenizers.config import TokenizerConfig
@@ -23,8 +25,11 @@ except ImportError:
     # Fallback to old path
     from megatron.core.datasets.megatron_tokenizer import MegatronTokenizer as MegatronTokenizerCore
 
+if TYPE_CHECKING:
+    from megatron.core.tokenizers import MegatronTokenizerBase
 
-def _compute_space_sensitive(tokenizer_instance: "MegatronTokenizer", default: bool = True) -> bool:
+
+def _compute_space_sensitive(tokenizer_instance: MegatronTokenizer, default: bool = True) -> bool:
     """
     Determine if a tokenizer is space-sensitive.
 


src/megatron/bridge/data/hf_processors/squad.py
```diff
@@ -50,6 +56,8 @@ def process_squad_example(
     """
     # Format input as: "Context: ... Question: ... Answer:"
     _input = f"Context: {example['context']} Question: {example['question']} Answer:"
+    if apply_tokenizer_chat_template and tokenizer is not None and tokenizer.library == "huggingface":
+        _input = tokenizer.apply_chat_template([{"role": "user", "content": _input}], tokenize=False)
```
liding-nv, line 60: Need to check if tokenizer has `chat_template` or not.
Base model typically does not have chat template, and it will raise an error when you run `tokenizer.apply_chat_template`

src/megatron/bridge/data/hf_processors/squad.py
```diff
@@ -50,6 +56,8 @@ def process_squad_example(
...
 
     # Use the first answer as the primary output
     _output = example["answers"]["text"][0]
```
liding-nv, line 63: Q: do we need to process output as well?
If the sample is
```
{"Context": XXXX, "Question": YYYY, "Answer": ZZZZ}
```
Then we construct the chat as 
```
[
{"role": "user", "content": "Context: XXXX Question: YYYY Answer"}, 
{"role": "assistant", "content": "ZZZZ"}
]
```
After chat template, it becomes
```
<|im_start|>system
<|im_end|>
<|im_start|>user
Context: XXXX Question: YYYY Answer<|im_end|>
<|im_start|>assistant
ZZZZ<|im_end|>
```
So output part also has special tokens around it

src/megatron/bridge/training/tokenizers/utils.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from megatron.core.tokenizers import MegatronTokenizer
+from megatron.core.tokenizers import MegatronTokenizer, MegatronTokenizerBase
 
 from megatron.bridge.training.tokenizers.config import TokenizerConfig
```
liding-nv, line 18: use `TYPE_CHECKING` here?

src/megatron/bridge/training/tokenizers/tokenizer.py
```diff
@@ -7,7 +7,7 @@
 from pathlib import Path
 from typing import Dict, List, Optional
 
-from megatron.core.tokenizers import MegatronTokenizer
+from megatron.core.tokenizers import MegatronTokenizer, MegatronTokenizerBase
```
liding-nv: use `TYPE_CHECKING` here?

src/megatron/bridge/data/hf_processors/squad.py
cuichenx: I think this is fine right? output can have special tokens. We'll need to modify the evaluation script to take these tokens into account. What do you suggest as the alternative?

src/megatron/bridge/data/hf_processors/squad.py
cuichenx: good point

src/megatron/bridge/data/hf_processors/squad.py
cuichenx: Ah I understand you now, yes we should process output as well. updated code

PR #2153, opened 2026-01-30
[misc] feat: mlflow improvements

ryxli:
# What does this PR do ?

Ability to use mlflow logging without enabling tensorboard logging.
Add missing log_metrics equivalents for mlflow.
Stylistic choices to organize metrics better in mlflow.

Also recommend env `MLFLOW_ENABLE_ASYNC_LOGGING=True`

# Changelog

- remove conditional check for existence of tensorboard logger, to allow for mlflow only tracking
- (quality of life change) change mlflow metrics dict key sanitizing
  - mlflow treats "/" as directories, meaning that they can be organized in their own section
  - see attached screenshot for example
- add missing metrics logged to mlflow which exist for wandb / tensorboard (such as critical validation loss metrics)
Screenshot 2026-01-30 at 12 24 43 PM

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci)in the Contributing doc for how to trigger 

the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

# Before your PR is "Ready for review"

**Pre checks**:

- [ x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?

If you haven't finished some of the above items you can still open "Draft" PR.

# Additional Information

- Related to # (issue)

## Release Notes

* **New Features**
  * Added MLflow integration to log validation metrics, providing an additional monitoring and tracking option alongside existing systems

* **Improvements**
  * Enhanced metric key formatting with improved special character handling for better compatibility
  * Improved logging reliability by conditionally executing logging operations only when appropriate backends are configured and available

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2199, opened 2026-02-03
Bridging perf from NeMo2 to Mbridge for certain configs

gautham-kollu:
# What does this PR do ?

Code changes
 1. Instead of creating one dataset iterator per model_chunk when VP is enabled, creates one and making the other things shallow point the first one.  Previously this was already implemented for finetuning, extending this to pretraining
 2. Adding a new config to not do numeric_checks post_training. (This is only for advanced users)
 3. Skip computing log metrics if logging is not enabled.
 4. Precompute a few things outside of the training loop to avoid redundant compute

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

## Release Notes

* **New Features**
  * Added `numeric_checks` configuration option for training validation.

* **Improvements**
  * Enhanced multi-model training support with improved iterator handling.
  * Optimized training loop logging behavior.
  * Refined fault tolerance and profiling operation execution.
  * Improved checkpoint decision logic with early-exit handling.

### DIFF
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -968,8 +968,8 @@ class LoggerConfig:
     to progress.txt file in checkpoint directory.
     """
 
-timing_log_level: Literal[-1, 0, 1, 2] = 0
-"""Granularity level to measure and report timing.
+    timing_log_level: Literal[-1, 0, 1, 2] = 0
+    """Granularity level to measure and report timing.
     -1: To disable timing logging as the timer start from 0 and above.
     0: report only iteration time and make sure timing does not introduce extra overhead.
     1: report timing for operations that are executed very limited times (basically once) during each iteration


src/megatron/bridge/training/train.py
```diff
@@ -457,46 +459,49 @@ def train(
...
-        if config.logger.log_params_norm:
-            params_norm = calc_params_l2_norm(model, model_config, use_megatron_fsdp=config.dist.use_megatron_fsdp)
-        learning_rate = None
-        decoupled_learning_rate = None
-        for param_group in optimizer.param_groups:
-            if len(param_group) == 0:
-                continue
-            if param_group["is_decoupled_lr"]:
-                decoupled_learning_rate = param_group["lr"]
+        if config.logger.tensorboard_dir is not None:  # Skip logging as tensorboard logging is disabled.
```
ananthsub: with changes proposed in #2153, this will no longer be true. after 2153, users could log to tensorboard, W&B, and mlflow independently vs the current state where tensorboard is required to log to w&b. we can have some check up front whether logging will be enabled. this will be constant throughout training and can be done outside of the loop

src/megatron/bridge/training/profiling.py
```diff
@@ -58,6 +58,9 @@ def handle_profiling_step(
     Returns:
         NVTX context if nsys profiling was started at this step, None otherwise
     """
+    if config is None:
```
ananthsub, line 61: this is the first thing checked in `should_profile_rank` - why do we need to repeat this here?

src/megatron/bridge/training/config.py
```diff
@@ -681,6 +681,9 @@ class TrainingConfig:
     check_weight_hash_across_dp_replicas_interval: Optional[int] = None
     """Interval to check weight hashes are same across DP replicas. If not specified, weight hashes not checked."""
 
+    numeric_checks: bool = True
```
ananthsub: the config field name is really broad, so it's impossible to tell exactly which checks this disables based on the name alone

gautham-kollu: Next rev

src/megatron/bridge/training/train.py
gautham-kollu: Next rev

src/megatron/bridge/training/profiling.py
gautham-kollu: In some sense it is a bit overboard but the reason I added this was because of OmegaConfig.
If the config is converted from a raw python dictionary to OmegeConf-dict, .get() on omegaconfig-dict starts applying a lot of resolution-rules etc and ends up calling a lot of functions. 

Walltime are minimal
Screenshot 2026-02-05 at 12 45 07 AM

Screenshot 2026-02-05 at 12 44 24 AM

src/megatron/bridge/training/config.py
```diff
@@ -944,6 +950,7 @@ class LoggerConfig:
 
     timing_log_level: Literal[0, 1, 2] = 0
     """Granularity level to measure and report timing.
+    -1: To disable timing logging as the timer start from 0 and above.
```
ananthsub:
```suggestion
timing_log_level: Literal[-1, 0, 1, 2] = 0
"""Granularity level to measure and report timing.
    -1: To disable timing logging as the timer start from 0 and above.
```

src/megatron/bridge/training/profiling.py
ananthsub: out of curiosity, when is it an omegaconf during training? shouldn't it already be resolved to a dataclass?

src/megatron/bridge/training/train.py
ananthsub, line 723: won't we need the same changes for eval? https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/21379ca1e1b2d2649a803fde39eae6131cb77400/src/megatron/bridge/training/eval.py#L135-L150

src/megatron/bridge/training/profiling.py
gautham-kollu: The fields in the class could be python raw List or an OmegaConfig.ListConfig

ananthsub (approved):

PR #2241, opened 2026-02-05
Introduce refactored model builder abstractions

maanug-nv:
# What does this PR do ?

This PR refactors the `ModelProviderMixin` to split up responsibilities. 

This refactor preserves the existing level of model customizability as well as conversion compatibility, while separating model configuration and initialization/building into separate components. This architecture was determined to be more comprehensible based on several sources of feedback. 
In particular, a custom model type can define one or both of:
1. A serializable configuration object encapsulating all config settings for the model. This may be a shallow dataclass hierarchy.
2. A builder object that determines how the model is initialized for distributed training or inference using aforementioned config.

Base model architectures like GPT and Mamba will follow this design.

These new abstractions are all new classes rather than direct modifications to `ModelProviderMixin` to ease migration. Additionally, `ModelProviderMixin` relied on the inheritance from `TransformerConfig`, modified some config attributes during model building, and contained some unused code. Therefore, I think we should deprecate the `ModelProviderMixin` after integrating the new abstractions across the codebase.

The new classes are:
1. `ModelConfig` - Has some required attributes, including import path of builder. Supports serialization to and from dictionary.
2. `ModelBuilder` - Contains stub methods for implementation of (distributed) model init. Also maintains pre-wrap and post-wrap hook registration from `ModelProviderMixin` which is used by PEFT. Usage of the hooks is left up to the child class.

This PR also includes the necessary changes for the Mamba base model to support this new interface. The refactor was easier to demonstrate on Mamba than on GPT. 'mamba/mamba_builder.py' contains:
1. `MambaModelConfig` which encapsulates all config settings used to create an MCore Mamba model.
2. `MambaModelBuilder` which defines how a distributed Mamba model should be built.

Since distributed model initialization is identical for GPT and Mamba models, the logic is extracted into a separate function under 'unimodal.py', which can be called by `GPTModelBuilder` in the future. The logic is also split into helpers for readability.

I plan to upstream all new files under src/ in this PR to MegatronLM in the near future. It may also be worthwhile to push the base model configs, e.g. `MambaModelConfig` to MCore eventually use directly with the MCore model. Additionally, if `TransformerConfig` is broken up in the future, the model-specific config can remain at the top-level of the hierarchy.

Some additional notes:
- Advanced use cases may need to customize how the model wrapping/distributed model initialization is done, eg MIMO. Therefore, it makes more sense for that functionality to be defined in the ModelBuilder, so that both the single-stage building logic and distributed building logic can be overriden from the same place.
- With the functional helpers in `unimodal.py`, these abstractions should appear as a simple layer on top of existing MLM code like `get_model()` and `mamba_builders.py`.

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

## Release Notes

* **New Features**
  * Introduced a model configuration and building framework supporting distributed training with DDP/FSDP options.
  * Added Mamba model builder with configurable stack specifications.
  * Added hook system for customizing model wrapping behavior.
  * Enabled virtual pipeline staging support for models.

* **Refactor**
  * Consolidated distributed model utilities into shared module.

### DIFF
--- a/tests/unit_tests/models/common/test_unimodal.py
+++ b/tests/unit_tests/models/common/test_unimodal.py
@@ -1,0 +1,838 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from unittest.mock import MagicMock, Mock, patch
+
+import pytest
+import torch
+import torch.nn as nn
+from megatron.core.enums import ModelType
+
+from megatron.bridge.models.common.unimodal import (
+    _ddp_wrap,
+    _print_num_params,
+    _wrap_with_mp_wrapper,
+    build_virtual_pipeline_stages,
+    to_empty_if_meta_device,
+    unimodal_build_distributed_models,
+)
+
+
+# ---------------------------------------------------------------------------
+# Shared helpers
+# ---------------------------------------------------------------------------
+
+
+def _make_pg():
+    """Mock ProcessGroupCollection with dp, cp, tp, pp sub-groups."""
+    pg = Mock()
+    pg.dp.rank.return_value = 0
+    pg.cp.rank.return_value = 0
+    pg.tp.rank.return_value = 0
+    pg.pp.rank.return_value = 0
+    pg.pp.size.return_value = 1
+    return pg
+
+
+def _make_transformer_config(**kwargs):
+    cfg = Mock()
+    cfg.virtual_pipeline_model_parallel_size = None
+    cfg.init_model_with_meta_device = False
+    cfg.use_cpu_initialization = False
+    cfg.fp16 = False
+    cfg.bf16 = False
+    for k, v in kwargs.items():
+        setattr(cfg, k, v)
+    return cfg
+
+
+def _make_model_module():
+    m = Mock()
+    m.parameters.return_value = []
+    m.modules.return_value = []
+    return m
+
+
+# =============================================================================
+# Section 1 — TestToEmptyIfMetaDevice
+# =============================================================================
+
+
+class TestToEmptyIfMetaDevice:
+    """to_empty_if_meta_device() materialises meta-device parameters while leaving non-meta parameters unchanged."""
+
+    def test_meta_parameter_becomes_empty_on_target_device(self):
+        module = nn.Module()
+        module.register_parameter("weight", nn.Parameter(torch.empty(4).to("meta")))
+        result = to_empty_if_meta_device(module, device=torch.device("cpu"))
+        assert result.weight.device == torch.device("cpu")
+
+    def test_non_meta_parameter_moved_to_device(self):
+        module = nn.Module()
+        module.register_parameter("weight", nn.Parameter(torch.zeros(4)))
+        result = to_empty_if_meta_device(module, device=torch.device("cpu"))
+        assert result.weight.device == torch.device("cpu")
+
+    def test_recurse_true_applies_to_submodule_parameters(self):
+        parent = nn.Module()
+        child = nn.Module()
+        child.register_parameter("weight", nn.Parameter(torch.empty(4).to("meta")))
+        parent.add_module("child", child)
+        to_empty_if_meta_device(parent, device=torch.device("cpu"), recurse=True)
+        assert parent.child.weight.device == torch.device("cpu")
+
+    def test_recurse_false_skips_submodule_parameters(self):
+        parent = nn.Module()
+        child = nn.Module()
+        child.register_parameter("weight", nn.Parameter(torch.empty(4).to("meta")))
+        parent.add_module("child", child)
+        to_empty_if_meta_device(parent, device=torch.device("cpu"), recurse=False)
+        assert parent.child.weight.device == torch.device("meta")
+
+
+# =============================================================================
+# Section 2 — TestBuildVirtualPipelineStages
+# =============================================================================
+
+
+class TestBuildVirtualPipelineStages:
+    """build_virtual_pipeline_stages() builds one stage without VP or multiple stages with VP, setting model_type on each."""
+
+    @patch("megatron.core.pipeline_parallel.utils.is_pp_last_stage", return_value=True)
+    @patch("megatron.core.pipeline_parallel.utils.is_pp_first_stage", return_value=True)
+    @patch("megatron.core.pipeline_parallel.utils.is_vp_last_stage", return_value=True)
+    @patch("megatron.core.pipeline_parallel.utils.is_vp_first_stage", return_value=True)
+    def test_single_stage_when_pp_size_one(self, *_):
+        pg = _make_pg()
+        pg.pp.size.return_value = 1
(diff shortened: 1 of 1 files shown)


src/megatron/bridge/models/common.py
```diff
@@ -0,0 +1,441 @@
...
+        transformer_config: TransformerConfig | None = getattr(self._model_config, "transformer", None)
+
+        def find_model_attr(attr_name):
+            """Look for an attribute in both self._model_config and transformer_config."""
+            attr_value = getattr(self._model_config, attr_name, None)
+            if attr_value is None:
+                getattr(transformer_config, attr_name, None)
+            return attr_value
+
+        # TODO (@maanug): handle pre/post wrap hooks
```
yaoyu-33: hooks are needed.

src/megatron/bridge/models/common.py
```diff
@@ -0,0 +1,441 @@
...
+
+        # Get VP size from model config if available
+        vp_size = find_model_attr("virtual_pipeline_model_parallel_size")
```
yaoyu-33: why do we need find_model_attr?

src/megatron/bridge/models/common.py
```diff
@@ -0,0 +1,441 @@
...
+            pre_process: Include embedding layer
+            post_process: Include output layer
+            vp_stage: Virtual pipeline stage
+
+        Returns:
+            The constructed model
+        """
+        ...
+
+    def build_distributed_models(
```
yaoyu-33: can we move this to provider? leave builder clean, just build models

src/megatron/bridge/models/common.py
maanug-nv: implemented in f561800776be703ec8b16c38433bb3b0f8d326b0 .
It's a simpler version than `ModelProviderMixin` , since afaict only `register_pre_wrap_hook()` .

src/megatron/bridge/models/common.py
maanug-nv: i've reorganized as follows:
1. make base `build_distributed_models()` abstract and move current impl to mamba builder
2. extract mamba builder impl into a helper function (this version is specific to unimodal LLMs) so it can be used by GPT
3. take TransformerConfig as an explicit argument in the helper so we know where to find certain attributes like vp size.

src/megatron/bridge/models/common.py
maanug-nv: i don't think it makes sense to move to provider as implementation might be specific to the model, eg MIMO. 
if we moved to provider, each model type would need a custom ModelConfig, ModelBuilder, and ModelProvider. I think that's excessive complexity. 
I've split model building code into helper functions that are hopefully readable.

src/megatron/bridge/models/common/base.py
```diff
@@ -0,0 +1,359 @@
...
+
+    def composed_hook(model: list[MegatronModule]) -> list[MegatronModule]:
+        for hook in hooks:
+            model = hook(model)
+        return model
+
+    return composed_hook
+
+
+class ModelProvider(Generic[ModelT]):
```
maanug-nv: this class can probably be removed.

maanug-nv: agreed on this offline

src/megatron/bridge/models/common/base.py
```diff
@@ -0,0 +1,393 @@
...
+
+        Returns:
+            List of models (multiple for virtual pipeline parallelism)
+        """
+        builder_cls = self.model_config.get_builder_cls()
+        return builder_cls(self.model_config).build_distributed_models(
+            pg_collection,
+            wrap_with_ddp=wrap_with_ddp,
+            fp16=fp16,
+            bf16=bf16,
```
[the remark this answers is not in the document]
maanug-nv: this is true, but `ModelProvider` was removed

Phlip79 (dismissed):

src/megatron/bridge/models/common/base.py
```diff
@@ -0,0 +1,313 @@
...
+        ...
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> "Serializable":
+        """Deserialize from dictionary using _target_ to identify class."""
+        ...
+
+
+@dataclass
+class ModelConfig(abc.ABC, Serializable):
```
ananthsub: why does this need to be an ABC?

src/megatron/bridge/models/common/base.py
```diff
@@ -0,0 +1,313 @@
...
+
+    Type Parameters:
+        ModelT: The type of model this builder produces (e.g., MCoreGPTModel)
+        BuildConfigT: The type of build config this builder accepts (e.g., GPTModelBuildConfig)
+    """
+
+    def __init__(self, model_config: ModelConfig):
+        self._model_config = model_config
+        self._pre_wrap_hooks = []
+        self._post_wrap_hooks = []
```
ananthsub: how are the hooks configured by the end users, especially in the case where users want to use an already existing builder but with custom hooks attached?

src/megatron/bridge/models/common/base.py
```diff
@@ -0,0 +1,313 @@
...
+        self,
+        pg_collection: ProcessGroupCollection,
+        ddp_config: DistributedDataParallelConfig | None = None,
+        overlap_param_gather_with_optimizer_step: bool = False,
+        use_megatron_fsdp: bool = False,
+        use_torch_fsdp2: bool = False,
+        wrap_with_ddp: bool = True,
+        data_parallel_random_init: bool = True,
+        mixed_precision_wrapper: Callable[[Any, MegatronModule], MegatronModule] | None = Float16Module,
+        model_type: ModelType = ModelType.encoder_or_decoder,
```
ananthsub, line 237: i see that this is the translation for the existing setup, but how do these args generalize for the MIMO use case?

src/megatron/bridge/models/common/base.py
maanug-nv: I suppose it doesn't have to be since it has no abstract methods, but why would anyone want to initialize this base class?
maybe the one reason is because `builder` needs to be overriden to point to something. 
anyway, i don't have a strong opinion, i can remove

src/megatron/bridge/models/common/base.py
maanug-nv: very good point, i definitely missed that. need to think about this.

src/megatron/bridge/models/common/base.py
```diff
@@ -0,0 +1,313 @@
...
+from megatron.core.enums import ModelType
+from megatron.core.process_groups_config import ProcessGroupCollection
+from megatron.core.transformer import MegatronModule
+from megatron.core.transformer.module import Float16Module
+
+
+class Serializable(Protocol):
+    """Protocol for serializable configurations."""
+
+    def to_dict(self) -> dict[str, Any]:
```
yaoyu-33: minor: can we do "as_dict()"? more pythonic

src/megatron/bridge/models/common/base.py
yaoyu-33: mimo can take in pg collection and ddp config as dict for submodules? I dont want to over-design to fit mimo

maanug-nv: i think signature can be a little flexible until it's integrated into training loop, which will be in follow-up PR(s). 

@yashaswikarnati do you have any feedback on how this signature can be a bit better for MIMO without changing too much?

src/megatron/bridge/models/common/base.py
maanug-nv: moved to config as per discussion offline

src/megatron/bridge/models/common/base.py
yashaswikarnati: Sorry just coming back to this. I see that in the signature we accept only one pg_collection. do we plan to have a different signature for multimodal. 

I dont think thats overdesign  - I would say thats the bare minimum to achieve the functionality.

PR #2248, opened 2026-02-05
nemotron 3 nano examples

liding-nv:
## Summary by CodeRabbit

* **New Features**
  - Comprehensive documentation for Nemotron 3 model examples with usage guidance
  - Model conversion tools supporting bidirectional format conversion (Hugging Face ↔ Megatron) with validation
  - SLURM-based distributed training scripts for Pretraining, Supervised Fine-Tuning (SFT), and Parameter-Efficient Fine-Tuning (PEFT) with LoRA
  - Packed sequence support in training workflows

### DIFF
--- a/examples/models/nemotron_3/conversion.sh
+++ b/examples/models/nemotron_3/conversion.sh
@@ -38,5 +38,5 @@ uv run python -m torch.distributed.run --nproc_per_node=8 \
     examples/conversion/hf_megatron_roundtrip_multi_gpu.py \
     --hf-model-id $HF_MODEL_ID \
     --megatron-load-path ${WORKSPACE}/models/$MODEL_NAME/iter_0000000 \
-    --tp 2 --pp 2 \
+    --tp 2 --ep 8 \
     --trust-remote-code


scripts/training/run_recipe.py
```diff
@@ -112,17 +118,36 @@ def parse_args() -> tuple[argparse.Namespace, list[str]]:
...
+        action="store_true",
+        default=False,
+        help="Enable packed sequence training (default: False)",
+    )
+    parser.add_argument(
+        "--seq_length",
+        type=int,
+        default=None,
+        help="Sequence length for training",
+    )
```
cuichenx: can we omit `--seq_length` and override it in the command?

liding-nv: iirc you need to set both together. The second one can be easily missed in i'm not familiar with this levels of details
```
dataset.seq_length=${SEQ_LEN}
dataset.packed_sequence_specs.packed_sequence_size=${SEQ_LEN}
```

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2248/commits/3da2a2d6d81825907c68ec6408721bb843ca9ade

sleepymalc:
Wants to point out that I find the current main branch does the weight conversion between megatron-LM <-> HF fine, but the actual layer (e.g., Mamba SSM) computation is different and hence will yield different forward pass results (#2381)

examples/models/nemotron_3/conversion.sh
```diff
@@ -0,0 +1,42 @@
...
+    --hf-model $HF_MODEL_ID \
+    --megatron-path ${WORKSPACE}/models/$MODEL_NAME/iter_0000000 \
+    --hf-path ${WORKSPACE}/models/$MODEL_NAME-hf-export
+
+# Round-trip validation
+uv run python -m torch.distributed.run --nproc_per_node=8 \
+    examples/conversion/hf_megatron_roundtrip_multi_gpu.py \
+    --hf-model-id $HF_MODEL_ID \
+    --megatron-load-path ${WORKSPACE}/models/$MODEL_NAME/iter_0000000 \
+    --tp 2 --pp 2 \
```
yaoyu-33: does nano have ep?

scripts/training/run_recipe.py
```diff
@@ -112,17 +118,36 @@ def parse_args() -> tuple[argparse.Namespace, list[str]]:
         default=None,
         help="PEFT scheme to use: 'lora', 'dora', or None.",
     )
+    parser.add_argument(
+        "--packed_sequence",
```
yaoyu-33: I dont think we need to test packed seq = false situation.

scripts/training/run_recipe.py
yaoyu-33: >iirc you need to set both together. The second one can be easily missed in i'm not familiar with this levels of details

that's fine, we will handle later for this part. I dont want args here if it's possible to do cli override

scripts/training/run_recipe.py
liding-nv: removed

scripts/training/run_recipe.py
liding-nv: ok fine with me, removed those args. imo we can verify and set those in `finalize` func so no need for user to explicitly set them in cli override.

examples/models/nemotron_3/conversion.sh
liding-nv: yes it has ep.
Q: for conversion, does it matter if we set ep or not
anyway i changed to tp2pp1ep2, this is the one used in training

PR #1564, opened 2025-12-02
[MoE LoRA] fix: Optimizer checkpointing shape mismatch

HollowMan6:
# What does this PR do ?

Resolves https://github.com/volcengine/verl/issues/4303.

# Changelog

- The experts.py code in Megatron-LM set replica_id = (PP, ETP, EDP), but somehow we need to change the ETP rank here with other ids to fit.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci)in the Contributing doc for how to trigger the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

# Before your PR is "Ready for review"

**Pre checks**:

- [X] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?

If you haven't finished some of the above items you can still open "Draft" PR.

# Additional Information

- Related to https://github.com/volcengine/verl/issues/4303

<sub>✨ Presented to you with <a href="https://macaron.im/mindlab">Mind Lab</a> - A Lab for Experiential Intelligence.</sub>

PR #2252, opened 2026-02-06
[peft] fix: Add replica_id handling for dense LoRA adapters with TP > 1

yaoyu-33:
## Summary

When using LoRA adapters on dense layers (non-MoE) with TP > 1, only one TP shard was being saved during checkpointing. This caused significant training-inference loss discrepancy because other TP ranks loaded zero/uninitialized adapter weights.

## Root Cause

In `ParallelLinearAdapter.sharded_state_dict()`, there was special `replica_id` handling for expert adapters (added in PR #1564, related to verl-project/verl#4303), but dense adapters never received the equivalent fix.

For dense adapters with TP > 1:
- Each TP rank generates a ShardedTensor with the same `replica_id`
- Shards should be distinguished by `global_offset`, but during the PEFT-filtered checkpoint save, TP shards are incorrectly deduplicated
- Result: Only TP rank 0's shard is saved

## Solution

Add `replica_id` handling for dense adapters similar to expert adapters. When TP > 1, the `replica_id` is adjusted to include the TP rank, ensuring each TP shard is correctly identified and saved during PEFT-filtered checkpoint saves.

## Changes
- Add `replica_id` adjustment for dense adapters with TP > 1 in `ParallelLinearAdapter.sharded_state_dict()`
- Add unit tests for the fix covering:
  - Dense adapters with TP > 1 (replica_id correctly updated)
  - Dense adapters with TP = 1 (no change to replica_id)
  - Expert adapters (still use EP-based replica_id, not affected)

## Testing
- [x] Unit tests added
- [ ] Functional tests with multi-GPU TP > 1 (requires CI)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed replica ID assignment for adapter checkpoints when using tensor model parallel configurations, ensuring correct deduplication during checkpoint saving.

* **Tests**
  * Added comprehensive test coverage for dense and expert adapter scenarios across different parallel configurations.


tests/functional_tests/training/test_finetune_lora.py
```diff
@@ -443,3 +443,473 @@ def _setup_directories(self, base_dir, suffix=""):
             os.makedirs(lora_tensorboard_dir, exist_ok=True)
 
         return pretrain_checkpoint_dir, pretrain_tensorboard_dir, lora_checkpoint_dir, lora_tensorboard_dir
+
+
+class TestLoRAFinetuneTP2:
```
priyatham-resolve, line 453: nit: This seems to duplicate all the helper methods from `TestLoRAFinetune` above. Could share them via a base class or mixin to reduce the ~300 lines of boilerplate? Not a blocker.

priyatham-resolve (approved):
LGTM. Thanks for the fix

yaoyu-33:
Cannot reproduce, defer to later.

PR #2261, opened 2026-02-06
fix: Use nargs for `custom_bash_cmds`

ko3n1g:
# What does this PR do ?

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

## Summary by CodeRabbit

* **Refactor**
  * Updated command argument processing to accept multiple separate inputs instead of a single delimited format.

### DIFF
--- a/scripts/performance/argument_parser.py
+++ b/scripts/performance/argument_parser.py
@@ -425,7 +425,7 @@ def parse_cli_args():
         nargs="*",
         action="append",
         help="List of bash commands to execute before the main command",
-        default=[],
+        default=None,
     )
     slurm_args.add_argument(
         "--gres",


scripts/performance/argument_parser.py
```diff
@@ -422,8 +422,8 @@ def parse_cli_args():
     slurm_args.add_argument(
         "-cb",
         "--custom_bash_cmds",
-        type=list_of_strings,
-        help="Comma separated string of bash commands",
+        nargs="*",
+        help="List of bash commands to execute before the main command",
         default=[],
```
malay-nagda: Can we keep None as default or change the condition in `scripts/performance/utils/executors.py` that currently checks for `None`?

scripts/performance/utils/executors.py
```diff
@@ -127,11 +127,11 @@ def slurm_executor(
     numa_cmd = f"numactl --cpunodebind=$((SLURM_LOCALID/{numa_divisor})) --membind=$((SLURM_LOCALID/{numa_divisor}))"
     if gpu.lower() in ["b300"]:
         numa_cmd += " -C $((SLURM_LOCALID * 16)),$((SLURM_LOCALID * 16 + 1))"
-    custom_bash_cmds.append(numa_cmd)
+    custom_bash_cmds.append(f" ; {numa_cmd}")
 
     launcher = SlurmTemplate(
         template_inline=INLINE_TEMPLATE,
-        template_vars={"pre_cmds": " ; ".join(custom_bash_cmds)},
+        template_vars={"pre_cmds": " ".join(custom_bash_cmds)},
```
malay-nagda: this and line 130 only separates numa cmd from all other cmds. Wouldn't it be better if we guarantee that any and all commands passed in by a user from CLI will be executed as separate commands?

ko3n1g: I guess we can refactor the code so that we can do

`-cb bash cmd1.sh -cb bash cmd2.sh` which will result in `custom_bash_cmds: List[List[str]]` which we then can flatten out to `str ; str`.. that good?

malay-nagda: I think current implementation is cleaner. Why not keep it as is?

ko3n1g: Because `custom_bash_cmds` is a list of bash arguments (i.e. `bash hello.py` -> `["bash", "hello.py"]`) so if we split that by ` ; ` we dont preserve the intended command

malay-nagda (approved):

PR #2323, opened 2026-02-11
fix: local checkpoint integration

ananthsub:
# What does this PR do ?

This PR resolves the following issues with local checkpointing support:

1. Resume from local checkpoint crashes with TypeError

Issue: `LocalCheckpointManager.load()` returns `checkpoint_name` as a `CkptID` tuple (iteration, rank, session_id), but callers expect `checkpoint_name` as a string path. The first call in `get_checkpoint_run_config_filename` call hits os.path.join(tuple, "run_config.yaml"), which raises an error.

Resolution: Guard all three call sites explicitly against `ckpt_type=LOCAL`
-  Since local checkpoints operate within a single run, we assume the parallelism degree will not change. Therefore, we don't need to load the whole `run_config` to validate against.

2. Training crashes after local checkpoint save when most_recent_k is enabled
Issue: `save_checkpoint` unconditionally cleans up on the local checkpoint directory when `most_recent_k > -1`. 

Resolution: Local checkpointing already handles its own cleanup/retention for old checkpoints, so we can skip local checkpoints with `most_recent_k` checks

3. Training counters lost on resume from local checkpoint
Global checkpoints persist TrainState (step, consumed_train_samples, skipped_train_samples, etc.) to a separate train_state.pt file in the save finalize function. For local checkpoints, we need to to better conform to what the local checkpoint manager saves to disk.

Resolution: Include training counters in the tensor aware state dict. This places the data in the common (non-sharded) part of the tensor-aware state dict, where it is saved atomically with the checkpoint, replicated across nodes, and returned by to_state_dict() on load.

4. Wandb/mlflow callbacks fire for ephemeral local checkpoint saves
The wandb and mlflow on_save_checkpoint_success callbacks ran unconditionally for both global and local checkpoint saves. Local checkpoints are ephemeral and should not be logged to external tracking systems.
Resolution: Add ckpt_type != CheckpointType.LOCAL to the callback guard so wandb/mlflow logging only fires for global checkpoint saves.

# Changelog

- Fix local checkpointing integration

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

## Summary by CodeRabbit

* **Improvements**
  * Enhanced local checkpoint handling with improved metadata management and restoration logic
  * Optimized checkpoint cleanup and finalization processes for local checkpoints

* **Tests**
  * Added comprehensive functional tests validating local checkpoint saving, resuming, and cleanup behavior in distributed training environments

### DIFF
--- a/tests/functional_tests/training/test_nvrx_straggler.py
+++ b/tests/functional_tests/training/test_nvrx_straggler.py
@@ -20,12 +20,12 @@ This test runs the actual pretrain function with NVRx straggler detection
 enabled, using mock data and a tiny model configuration for fast testing.
 """
 
-import argparse
+import io
 import logging
-import os
-import tempfile
+import sys
 import time
 
+import pytest
 import torch
 
 from megatron.bridge.models.llama import Llama32ModelProvider1B
@@ -45,9 +45,10 @@ from megatron.bridge.training.config import (
     ValidationConfig,
 )
 from megatron.bridge.training.gpt_step import forward_step
+from megatron.bridge.training.nvrx_straggler import HAVE_NVRX
 from megatron.bridge.training.pretrain import pretrain
 from megatron.bridge.training.state import GlobalState
-from megatron.bridge.utils.common_utils import get_rank_safe, print_rank_0
+from megatron.bridge.utils.common_utils import get_rank_safe
 
 
 def create_functional_test_config(enable_nvrx: bool = True) -> ConfigContainer:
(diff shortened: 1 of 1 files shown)


dimapihtar (approved):
LGTM. Thank you!

maanug-nv:
> Since local checkpoints operate within a single run, we assume the parallelism degree will not change. 

Does inprocess restart with "warm reserves" support changing parallelisms?

ananthsub:
> Does inprocess restart with "warm reserves" support changing parallelisms?

no it assumes the same parallelism

PR #2405, opened 2026-02-17
feat: Enhance dataset loading efficiency with tensor parallelism

izikgo:
# What does this PR do?

Adds a `broadcast_data_across_tp` configuration flag that restores the TP-rank-0 broadcast data loading path removed in PR #491, allowing users on high-latency storage backends (e.g. VAST) to avoid the severe I/O contention caused by replicated loading across all TP ranks.

# Changelog

- Added `broadcast_data_across_tp: bool = False` to `DataloaderConfig`, inherited by all dataset configs.
- Updated `is_dataset_built_on_rank` to support optional `pg_collection`, falling back to global parallel state so it can be passed directly as a zero-argument callable to `BlendedMegatronDatasetBuilder`.
- Modified `pretrain_train_valid_test_datasets_provider` to conditionally pass `is_dataset_built_on_rank` (broadcast) or `lambda: True` (replicated) based on the flag.
- Modified `get_batch` in `gpt_step.py` to conditionally use `get_batch_on_this_tp_rank` (broadcast) or `get_batch_from_iterator` (replicated) based on the flag.
- Appended extra-keys broadcast to `get_batch_on_this_tp_rank` via `broadcast_object_list` to forward any non-standard batch keys (e.g. packed-sequence metadata) without modifying the existing direct tensor broadcast path.

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/491

**Context:** PR #491 replaced the TP-rank-0 broadcast data loading path with replicated loading on all ranks. This works correctly on Lustre-backed clusters but introduces severe data loading stalls on VAST (network-attached storage) at multi-node scale due to redundant I/O amplification (8x with TP=8). On 64 nodes / 512 GPUs, step times alternate between 22s (compute) and 250-400s (data stall).

**Approach:** Rather than a hard revert, this PR adds a `broadcast_data_across_tp` flag (default `False`) so users can select the optimal strategy for their storage backend. When enabled, only TP-rank-0 loads data and broadcasts to other TP ranks, reducing VAST readers by a factor of TP.

**Validation:** Benchmarked on 8x B200 over NVLink, the direct tensor broadcast adds <0.7 ms even at 256K sequence length. The extra-keys path (`broadcast_object_list`) adds negligible overhead for the common pretraining case where no extra keys are present.

## Summary by CodeRabbit

* **New Features**
  * Added `broadcast_data_across_tp` configuration flag to enable data broadcasting across tensor-parallel ranks. When activated, data is loaded on a single tensor-parallel rank and broadcast to others, providing an alternative to per-rank data loading.
  * Improved batch synchronization to broadcast additional metadata and non-fixed keys across all tensor-parallel ranks, ensuring consistent data distribution during training.


docs/training/broadcast-data-across-tp.md
```diff
@@ -0,0 +1,104 @@
+# Broadcast Data Across Tensor-Parallel Ranks
```
ananthsub, line 1: thank you for adding this documentation! I believe it'd be useful to have this be part of a general data loading page, which can later be expanded to cover different datasets supported in megatron bridge and how to plug in custom datasets, rather than a page specicially dedicated to TP broadcast vs replicated load, but I'll defer to @yaoyu-33 and @cuichenx for this

tests/benchmarks/bench_broadcast_tp.py
```diff
@@ -0,0 +1,173 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
ananthsub: is it expected for this benchmark to be run in the test suite? or do you want this as a standalone script? 

@yaoyu-33 does it make more sense to have a top-level benchmarks / microbenchmarks directory for scripts like this?

src/megatron/bridge/training/gpt_step.py
```diff
@@ -169,13 +170,19 @@ def get_batch(
...
         return None, None, None, None, None, None, None, None, None, None
 
-    batch = get_batch_from_iterator(
-        data_iterator,
-        use_mtp,
-        getattr(cfg.dataset, "skip_getting_attention_mask_from_dataset", True),
-        is_first_pp_stage=is_first,
-        is_last_pp_stage=is_last,
-    )
+    broadcast_data = getattr(cfg.dataset, "broadcast_data_across_tp", False)
```
ananthsub, line 173: since the dataloader config is shared, this setting is also exposed for vlm datasets. there needs to be handling here for `broadcast_data_across_tp` too: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/training/vlm_step.py


https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/training/llava_step.py

tests/benchmarks/bench_broadcast_tp.py
izikgo: you're right. I'll move it. I didn't intend for this to be a unittext

src/megatron/bridge/training/gpt_step.py
izikgo: added

src/megatron/bridge/data/utils.py
```diff
@@ -37,17 +40,27 @@
...
+            parallel state, which allows this function to be passed directly
+            as a zero-argument callable to ``BlendedMegatronDatasetBuilder``.
+
     Returns:
         True if the dataset should be built on the current rank, False otherwise.
     """
-    return (is_pp_first_stage(pg_collection.pp) or is_pp_last_stage(pg_collection.pp)) and (
-        pg_collection.tp.rank() == 0
+    if pg_collection is not None:
+        return (is_pp_first_stage(pg_collection.pp) or is_pp_last_stage(pg_collection.pp)) and (
```
asolergi-nv: Check if rank contains MTP layer, required to support placing MTP layers into standalone stages (Not the last PP stage)

https://github.com/NVIDIA/Megatron-LM/blob/3d1a4ba71ecc49f1a0c9480c90f819d2b00f9915/pretrain_gpt.py#L209

izikgo: added

PR #2416, opened 2026-02-17
Fix LLAMA3 LoRa TFLOPs Formula

rhmukundan:
## Release Notes

* **Improvements**
  * Enhanced FLOPs calculation to support LoRA-based model training with sequence-length-specific computation paths and validation checks.


src/megatron/bridge/data/datasets/packing_utils.py
malay-nagda, line 1: can you add unit tests for new methods in this file?

rhmukundan: With the new refactor, these changes have been reverted.

malay-nagda (approved):

PR #2470, opened 2026-02-20
feat: add on_before_data_init callback hook for MLPerf compliance

matthew-frank:
Fire on_before_data_init in setup() after model/optimizer/checkpoint are ready but before setup_data_iterators() opens any dataset files.

This gives benchmarks a hook to run JIT warmup with mock data and log MLPerf init_stop/run_start markers, ensuring no real dataset I/O occurs before run_start is recorded — a hard MLPerf benchmark requirement.

# What does this PR do ?

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

## Summary by CodeRabbit

* **New Features**
  * Added a new callback hook that fires before dataset initialization, enabling custom operations like JIT warmup and MLPerf logging during training setup.


src/megatron/bridge/training/callbacks.py
```diff
@@ -139,6 +141,15 @@ def on_train_end(self, context):
         ```
     """
 
+    def on_before_data_init(self, context: CallbackContext) -> None:
```
ananthsub:
```suggestion
    def on_data_init_start(self, context: CallbackContext) -> None:
```
only nit is in naming for consistency with other methods

matthew-frank:
I think I've addressed the change requests except that the original code rabbit review wanted a new integration test.

What we did instead was:   Updated the existing unit tests in test_callbacks.py to include on_data_init_start:                                                                                                                                                         
                                                                                                                                                                                                                                              
  - Added it to the VALID_EVENTS assertion in TestValidEvents                                                                                                                                                                                 
  - Added it to the base class no-op test in TestCallback.test_all_methods_are_no_ops

  Regarding an integration test for firing order within setup(): the existing test suite doesn't have integration tests verifying that any callback (e.g. on_train_start, on_eval_start) fires at the correct point in its respective
  function. Those are all covered only at the CallbackManager machinery level (registration, firing, ordering). Adding a setup() integration test for this one event would require mocking the entire function (initialize_megatron,
  build_tokenizer, provide_distributed_model, setup_optimizer, load_checkpoint, setup_data_iterators, etc.) and would be the only test of its kind — setting a precedent we'd then want to backfill for every other event. That seems out of
  scope for this PR.

PR #2474, opened 2026-02-21
feat: add bridge-level ADDITIONAL_FILE_PATTERNS for automatic model file preservation

liding-nv:
Add support for automatically copying model-specific files during HuggingFace export via bridge-level configuration.

### DIFF
--- a/.github/workflows/build-test-publish-wheel.yml
+++ b/.github/workflows/build-test-publish-wheel.yml
@@ -64,7 +64,7 @@ jobs:
       container-options: "--gpus all --runtime=nvidia"
     secrets:
       TWINE_USERNAME: __token__
-      TWINE_PASSWORD: ${{ (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/r')) && secrets.SVC_PYPI_TOKEN || secrets.SVC_PYPI_TEST_TOKEN }}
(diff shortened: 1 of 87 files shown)


ananthsub (commented):
IIUC the additional modeling files are only copied if `trust_remote_code=True` is specified, and this PR lifts the restriction such that callers can specify files needed from the repo regardless of that flag. 

one question is whether we expect callers to need to pass this information through each time on the AutoBridge, or if we want to configure this per bridge class so callers don't have to concern themselves with it.

if we want to configure this just once, we could do this instead:
1. add `additional_file_patterns` here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/1d65cd8a5aff5c679f71439f4905701593a9ae43/src/megatron/bridge/models/conversion/model_bridge.py#L241
2. configure this per model bridge here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/1d65cd8a5aff5c679f71439f4905701593a9ae43/src/megatron/bridge/models/nemotronh/nemotron_h_bridge.py#L71
3. read the pattern in `save_hf_pretrained`

wdyt @yaoyu-33 @cuichenx ?

ananthsub (dismissed):

chtruong814:
Overall repo still at ~86% coverage.  Moving forward with merging.

PR #2522, opened 2026-02-25
Auto-slice output vocab for faster SFT with answer_only_loss

### LINKED ISSUE #2473: SFT with answer_only_loss: auto-slice output vocab for faster training
## Problem

When fine-tuning with `answer_only_loss`, the output projection (`logits = hidden @ embedding.T`) and cross-entropy loss are computed over the **full vocabulary**, even though the model may only need to predict tokens from a small subset of the vocab. This is wasteful — the output layer matmul and softmax scale linearly with vocab size.

For example:
- **Qwen3-0.6B** has 151,936 tokens. If you fine-tune it for any-language → Hebrew translation, and Hebrew uses ~50 unique token IDs in answers, you're still multiplying hidden states by a 151,936-row matrix every step.
- **Our use case**: We combine a spoken-language byte-level tokenizer (260 tokens) with a video tokenizer (64,000 tokens) into a single 64,260-token vocabulary. During SFT (video → text), all answer tokens are byte-only (IDs 0–259). The input contains video tokens (IDs 260–64,259), so we can't resize the embedding matrix — but the output head only ever needs 260 of the 64,260 rows.

## Measured speedup

We monkey-patched the output layer during SFT to slice the shared embedding weight:

```python
def _byte_output_forward(input_, weight=None, **kwargs):
    weight = weight[:260]  # slice to answer-only vocab
    return F.linear(input_, weight), None

model.output_layer.forward = _byte_output_forward

# Also clamp labels at masked positions (video token IDs > 259
# would be out-of-bounds for the 260-dim logits, but loss_mask=0
# for those positions anyway):
original_loss = model.compute_language_model_loss
model.compute_language_model_loss = lambda labels, logits: original_loss(
    labels.clamp(max=259), logits
)
```

**Results** (60M model, SEQ=16384, packed sequences, single GPU):

| Config | Step time | TFLOP/s/GPU | Speedup |
|--------|-----------|-------------|---------|
| Full vocab (64,260) | 5.41s | 31.6 | 1.0x |
| Byte-only head (260) | 3.03s | 55.8 | **1.79x** |

Isolated benchmarks on the output layer alone show **210x** speedup on the matmul and **286x** on cross-entropy.

Even though we do not update the other embeddings at the LM head - Quality is preserved (chrF 0.0527 vs 0.0497 baseline at 1000 steps). The checkpoint is saved with the full vocab and converts to HF format normally — the slicing is training-only.

## Proposed generic solution

When `answer_only_loss` is enabled, Megatron-Bridge could automatically detect which token IDs actually appear in answers and slice the output projection accordingly:

1. **During data loading**: scan all answer spans (where `loss_mask=1`) and collect the set of unique token IDs → `active_ids` (a sorted tensor of, say, N unique IDs out of V total vocab).
2. **Patch output layer**: instead of `F.linear(hidden, weight)` with weight `[V, H]`, compute `F.linear(hidden, weight[active_ids])` with weight `[N, H]`.
3. **Remap labels**: map original label IDs to their index in `active_ids` (e.g., via a lookup table). Map all masked-position labels to 0 (or any valid index, since `loss_mask=0` anyway).

This gives `[batch, seq, N]` logits instead of `[batch, seq, V]`, and cross-entropy over N classes instead of V.

### Trade-off

This is not exactly equivalent to full fine-tuning because gradients only flow through the `active_ids` rows of the embedding matrix (for the output projection path). The input embedding still uses all V rows. In practice, for SFT where answers are a small subset of the vocab, this is unlikely to matter — and users who care can disable the optimization.

### Where this helps most

- Large vocab + small answer vocab (Qwen 151k → Hebrew 50 tokens)
- Multimodal models where input modality tokens dominate vocab but answers are text-only
- Any SFT task where the answer vocabulary is much smaller than the full tokenizer vocabulary

A flag like `auto_slice_output_vocab: true` (or even on-by-default for `answer_only_loss`) would make this transparent to users.

AmitMY:
## Summary

Closes #2473

When fine-tuning with `answer_only_loss`, the output projection (`logits = hidden @ embedding.T`) and cross-entropy loss are computed over the **full vocabulary**, even though only a small subset of tokens appear in answers. This PR adds a utility module that automatically detects which token IDs appear in answers and slices the output projection accordingly.

### API

Three functions in `megatron.bridge.training.vocab_slice`:

- **`collect_active_vocab_ids(dataset)`** — scans an SFT dataset (standard or packed) and returns a sorted tensor of unique token IDs that appear in answer positions (where `loss_mask > 0`). Uses `loss_mask` when available in the dataset item, falls back to `answer_start_idx` otherwise.
- **`install_vocab_slice(model, active_ids)`** — patches the model's `output_layer.forward` to slice the shared embedding weight to `weight[active_ids]`, and remaps labels in `compute_language_model_loss` to use indices into the active set. Validates that all active_ids are within vocab bounds.
- **`create_vocab_sliced_forward_step(active_ids)`** — wraps `forward_step` to lazily install the vocab slice. Tracks model identity so the patch is re-installed if the model changes.

### Usage

```python
from megatron.bridge.training.vocab_slice import (
    collect_active_vocab_ids,
    create_vocab_sliced_forward_step,
)

train_ds, valid_ds, test_ds = dataset_builder.build()
active_ids = collect_active_vocab_ids(train_ds)
fwd_step = create_vocab_sliced_forward_step(active_ids)
finetune(config=cfg, forward_step_func=fwd_step)
```

### Benchmark

Toy model (vocab=151,936 like Qwen 2.5, active=200 tokens, SEQ=2048, single GPU):

| Metric | Full vocab | Sliced vocab | Improvement |
|--------|-----------|-------------|-------------|
| Training step | 188.7 ms | 45.3 ms | **4.16x faster** |
| Output matmul | 16.25 ms | 0.04 ms | **370x faster** |
| Cross-entropy | 43.05 ms | 0.03 ms | **1631x faster** |
| Peak GPU memory | 13,186 MB | 3,418 MB | **74% less** |

Real-world results (60M model, vocab=64,260, active=260, SEQ=16,384, packed sequences):

| Config | Step time | TFLOP/s | Speedup |
|--------|-----------|---------|---------|
| Full vocab | 5.41s | 31.6 | 1.0x |
| Sliced (260 tokens) | 3.03s | 55.8 | **1.79x** |

Quality preserved: chrF 0.0527 vs 0.0497 baseline at 1000 steps.

### Design decisions

- **No changes to existing files** — the utility is self-contained. Users opt-in by wrapping their `forward_step`.
- **Index-based slicing** (not contiguous range) — works for any subset of the vocab, not just a prefix.
- **Label remapping via lookup table** — labels outside `active_ids` (at masked positions) map to index 0 (safe since `loss_mask=0`).
- **Lazy installation with model tracking** — the model is patched on the first `forward_step` call; re-patches if the model identity changes.
- **Bounds checking** — `install_vocab_slice` raises `ValueError` if any active_id >= vocab_size.

### Trade-off

Cross-entropy is computed over N active classes instead of V total. This means the softmax normalization differs — the model sees fewer "distractor" classes. In practice, for SFT where answers are a small subset of the vocab, this has no measurable quality impact.

## Test plan

- [x] 20 unit tests covering `collect_active_vocab_ids` (standard, packed, loss_mask, max_samples, dedup, empty), `install_vocab_slice` (shape, logit correctness, label remap, masked safety, DDP unwrap, bounds check, shared weight), `create_vocab_sliced_forward_step` (wrapping, install-once, model-change re-install, argument forwarding), and end-to-end correctness (logit match, finite loss, gradient flow)
- [x] Benchmark script (`benchmarks/vocab_slice_benchmark.py`) verifying speedup and correctness
- [x] Lint passes (`ruff check`)


AmitMY:
This paper https://arxiv.org/pdf/2603.10145
also claims that this setup would train better, because training at high vocab size is unstable

suiyoubi:
/ok to test [2146fc6](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2522/commits/2146fc6b759d2c099a2b3f181101af1705284780)[2146fc6](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2522/commits/2146fc6b759d2c099a2b3f181101af1705284780)

src/megatron/bridge/training/vocab_slice.py
```diff
@@ -0,0 +1,228 @@
...
+    remap[active_ids] = torch.arange(len(active_ids), dtype=torch.long, device=device)
+
+    n_active = len(active_ids)
+
+    # 1. Patch output_layer to slice weight to active_ids rows.
+    # runtime_gather_output is part of the Megatron ColumnParallelLinear
+    # interface but unused here since we bypass the parallel output layer.
+    def _sliced_output_forward(input_, weight=None, runtime_gather_output=None):
+        if weight is not None:
+            weight = weight[active_ids]
```
suiyoubi, line 165: I think this would break when enabling the tensor parallelism. Have you tested with TP > 1 use case ?

benchmarks/vocab_slice_benchmark.py
```diff
@@ -0,0 +1,349 @@
...
+    total_params = sum(p.numel() for p in model.parameters())
+    embed_params = model.embedding.weight.numel()
+    print(f"Total params:       {total_params:>12,}")
+    print(f"Embedding params:   {embed_params:>12,}")
+    print(f"Transformer params: {total_params - embed_params:>12,}")
+    print()
+
+    # -- Synthetic data --
+    active_ids = torch.arange(n_active, device=device)
+    remap = torch.zeros(vocab_size, dtype=torch.long, device=device)
```
suiyoubi, line 191: Can we avoid silent corruption for labels outside `active_ids`? Right now remap defaults to 0, so any unscanned-but-real target token is trained as class 0. I’d strongly prefer a sentinel like -1 plus an assertion/check on unmasked positions before loss.

src/megatron/bridge/training/vocab_slice.py
AmitMY: I did not.

benchmarks/vocab_slice_benchmark.py
AmitMY: that is just a benchmark - in practice - whatever is not being trained, is outside of the logits completely. if token id 500 is used, then we select it in the logits matrix anyway. if it is never used, it won't be there

PR #2556, opened 2026-02-26
[model] refactor: Remove individual qwen providers

yaoyu-33:
## Summary

- Refactors `Qwen3NextBridge.provider_bridge()` to use `super().provider_bridge()` + attribute setting, matching the pattern used by `Qwen3MoEBridge` and other bridges
- Removes direct dependency on `Qwen3NextModelProvider` — model configuration is now fully defined within the bridge's `provider_bridge()` method
- Adds `model_type="qwen3_next"` to the `@register_bridge` decorator for proper HF config generation

## Motivation

The old bridge bypassed the base `provider_bridge()` (which auto-maps HF config fields via `CONFIG_MAPPING`) and instead manually constructed a `Qwen3NextModelProvider` with all parameters passed as constructor arguments. This was inconsistent with how other bridges (Qwen3, Qwen3 MoE, Llama, etc.) define their models and made the bridge harder to maintain.

## Verification

Ran a comparison script that instantiates both the old (`Qwen3NextModelProvider` constructor) and new (`super().provider_bridge()` + setattr) approaches against the `Qwen/Qwen3-Next-80B-A3B-Instruct` HF config:

- **263 attributes matched, 0 diffs**
- Only difference: `max_position_embeddings` (a dead field from the old provider dataclass that had no effect on model construction)

## Test plan

- [x] Verified 263/263 provider attributes match between old and new approach
- [ ] CI unit tests pass

Made with [Cursor](https://cursor.com)

## Release Notes

* **New Features**
  * Added experimental attention variant support for Qwen3-Next models
  * Enabled gated attention and zero-centered RMSNorm features
  * Enhanced Mixture of Experts (MoE) configuration capabilities

* **Improvements**
  * Optimized model registration and configuration process

### DIFF
--- a/tests/functional_tests/launch_scripts/active/L0_Launch_models_qwen3.sh
+++ b/tests/functional_tests/launch_scripts/active/L0_Launch_models_qwen3.sh
@@ -21,6 +21,5 @@ uv run coverage run --data-file=/opt/Megatron-Bridge/.coverage --source=/opt/Meg
   -o log_cli=true -o log_cli_level=INFO -v -s -x -m "not pleasefixme" --tb=short -rA \
   tests/functional_tests/test_groups/models/qwen/test_qwen3_conversion.py \
   tests/functional_tests/test_groups/models/qwen/test_qwen3_moe_conversion.py \
-  tests/functional_tests/test_groups/models/qwen/test_qwen3_next_conversion.py \
-  tests/functional_tests/test_groups/models/qwen/test_qwen3_next_provider.py
+  tests/functional_tests/test_groups/models/qwen/test_qwen3_next_conversion.py
 coverage combine -q


examples/recipes/decentralized_pg/pretrain_qwen3_with_decentralized_pg.py
```diff
@@ -0,0 +1,709 @@
+#!/usr/bin/env python3
```
yaoyu-33: remove

examples/recipes/local_parallel_groups/pretrain_qwen3_with_local_parallel_groups.py
```diff
@@ -0,0 +1,711 @@
+#!/usr/bin/env python3
```
yaoyu-33: remove

PR #2579, opened 2026-02-27
[training] refactor: Align bridge training defaults with Megatron-LM upstream

yaoyu-33:
## Summary

Aligns Bridge training config defaults with Megatron-LM upstream to reduce behavioral divergence when running the same model/data between the two codebases.

### Changes

- **`DataloaderConfig.num_workers`**: `8` → `2` (match Megatron-LM `--num-workers` default)
- **`DataloaderConfig.persistent_workers`**: `False` → `True` with automatic fallback to `False` when `num_workers=0` (match Megatron-LM's `True if args.num_workers > 0 else False`)
- **`data_parallel_random_init`**: `True` → `False` in all model provider function signatures to match Megatron-LM's default. Affected:
  - `ModelProviderMixin.provide_distributed_model`
  - `get_model`
  - `unimodal_build_distributed_models`
  - `ModelBuilder.build_distributed_models`
  - `MambaModelBuilder.build_distributed_models`
  - `MimoModelProvider.provide_distributed_model`

### Notes

- Existing recipes that explicitly set `num_workers=8` are **unaffected** by the default change.
- The training loop path (`setup.py`) already passes `cfg.rng.data_parallel_random_init` (default `False`) explicitly, so this change only affects **direct callers** of the model provider API (e.g., standalone scripts, tests).
- `persistent_workers` is guarded at the DataLoader construction site in `samplers.py` to prevent PyTorch errors when `num_workers=0`.

## Test plan

- [ ] Verify existing pretrain recipes still work (they hardcode `num_workers=8`)
- [ ] Verify finetuning recipes still work (`finetune_utils.py` hardcodes `num_workers=1`)
- [ ] Verify standalone model provider usage matches expected behavior with new `data_parallel_random_init=False` default
- [ ] Run unit tests for data samplers

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Chores**
  * Updated default configuration for distributed model training initialization behavior across multiple model builders.
  * Adjusted dataloader defaults: reduced worker count and enabled persistent workers for improved data loading efficiency.
  * Enhanced configuration validation to support dynamic finalization of dataloader settings based on worker availability.

### DIFF
--- a/tests/unit_tests/models/mamba/test_mamba_builder.py
+++ b/tests/unit_tests/models/mamba/test_mamba_builder.py
@@ -520,6 +520,6 @@ class TestMambaModelBuilderBuildDistributedModels:
         assert args[2] is self.pg
         assert args[3] is None  # ddp_config
         assert args[7] is True  # wrap_with_ddp
-        assert args[8] is True  # data_parallel_random_init
+        assert args[8] is False  # data_parallel_random_init
         assert args[9] is Float16Module  # mixed_precision_wrapper
         assert args[11] is ModelType.encoder_or_decoder  # model_type


src/megatron/bridge/data/samplers.py
```diff
@@ -112,7 +112,7 @@ def build_pretraining_data_loader(
         num_workers=num_workers,
         pin_memory=pin_memory,
         collate_fn=collate_fn,
-        persistent_workers=persistent_workers,
+        persistent_workers=persistent_workers if num_workers > 0 else False,
```
ananthsub: this could go into a validate() function for the dataloader config dataclass so its less subtle?

yaoyu-33: cool, agree, can do that.

PR #2643, opened 2026-03-04
fix: GPTSFTChatDataset.collate_fn use pad_seq_length_to_mult instead of hardcoded 16

### LINKED ISSUE #2642: GPTSFTChatDataset.collate_fn ignores pad_seq_length_to_mult
## Description

`GPTSFTChatDataset.collate_fn` hardcodes sequence padding to a multiple of 16, ignoring the configurable `self.pad_seq_length_to_mult` attribute. This causes a crash when using variable-length sequences (`pad_to_max_length=False`) with context parallelism sizes where `2 * cp_size > 16` (e.g. CP=16 requires divisibility by 32).

The base class `GPTSFTDataset.collate_fn` (line 703) and `GPTSFTPackedDataset.collate_fn` (line 892) both correctly use `self.pad_seq_length_to_mult` — only the chat variant has the bug.

## Location

`src/megatron/bridge/data/datasets/sft.py`, line 1200:

```python
# BUGGY — hardcoded 16
max_length = min(self.max_seq_length, self._ceil_to_nearest(max_length, 16))
```

Should be:

```python
# CORRECT — uses configurable attribute
max_length = min(self.max_seq_length, self._ceil_to_nearest(max_length, self.pad_seq_length_to_mult))
```

## How to Reproduce

1. Create a `GPTSFTChatDataset` with `pad_seq_length_to_mult=32` and `pad_to_max_length=False`
2. Call `collate_fn` on a batch
3. Observe output sequence length is padded to a multiple of 16, not 32

## Expected Behavior

Output sequence length should be padded to a multiple of `pad_seq_length_to_mult` (32 in this case), consistent with the base class behavior.

shanecmoran:
## Description

`GPTSFTChatDataset.collate_fn` hardcodes sequence padding to a multiple of 16, ignoring the configurable `self.pad_seq_length_to_mult` attribute. The base class `GPTSFTDataset.collate_fn` and `GPTSFTPackedDataset.collate_fn` both correctly use `self.pad_seq_length_to_mult` — only the chat variant has the bug.

This causes failures when using variable-length sequences (`pad_to_max_length=False`) with context parallelism sizes where `2 * cp_size > 16` (e.g. CP=16 requires divisibility by 32).

## Changes

**`src/megatron/bridge/data/datasets/sft.py`** (line 1200):
```python
# Before — hardcoded 16
max_length = min(self.max_seq_length, self._ceil_to_nearest(max_length, 16))

# After — uses configurable attribute, consistent with base class
max_length = min(self.max_seq_length, self._ceil_to_nearest(max_length, self.pad_seq_length_to_mult))
```

**`tests/unit_tests/data/datasets/test_sft.py`**:
- Added `test_collate_fn_respects_pad_seq_length_to_mult` to `TestDataGPTSFTChatDataset` — constructs dataset with `pad_seq_length_to_mult=32`, calls `collate_fn`, asserts output sequence length is divisible by 32.

Fixes #2642

## Summary by CodeRabbit

* **Improvements**
  * Sequence padding mechanism updated to use configurable granularity parameters instead of hardcoded values, enabling dataset instances to specify custom padding requirements for improved data preparation flexibility.

* **Tests**
  * Added unit test to validate that sequence padding operations properly respect and enforce configured granularity settings during batch processing.

### DIFF
--- a/src/megatron/bridge/data/datasets/sft.py
+++ b/src/megatron/bridge/data/datasets/sft.py
@@ -1197,7 +1197,10 @@ class GPTSFTChatDataset(GPTSFTDataset):
         if self.pad_to_max_length:
             max_length = self.max_seq_length
         else:
-            max_length = min(self.max_seq_length, self._ceil_to_nearest(max_length, 16))
+            max_length = min(
+                self.max_seq_length,
+                self._ceil_to_nearest(max_length, max(16, self.pad_seq_length_to_mult)),
+            )
         assert max_length <= self.max_seq_length
 
         position_ids = [list(range(max_length)) for _ in batch]
--- a/tests/unit_tests/data/datasets/test_sft.py
+++ b/tests/unit_tests/data/datasets/test_sft.py
@@ -325,6 +325,52 @@ class TestDataGPTSFTChatDataset:
         ]
         dataset.collate_fn(batch)
 
+    def test_collate_fn_respects_pad_seq_length_to_mult(self, tmp_path):
+        datasets_dir = tmp_path / "datasets"
+        datasets_dir.mkdir(exist_ok=True)
+        path = str(datasets_dir / "sft.jsonl")
+        line = {"input": "hi how are you?", "output": "I'm fine, thanks."}
+        with open(path, "w") as f:
+            for _ in range(10):
+                f.write(json.dumps(line) + "\n")
+
+        tokenizer = create_mock_tokenizer()
+        dataset = GPTSFTChatDataset(
+            file_path=path,
+            tokenizer=tokenizer,
+            label_key="output",
+            prompt_template="{input}\n\n### Response:\n{output}",
+            truncation_field="output",
+            pad_seq_length_to_mult=32,
+        )
+        batch = [
+            {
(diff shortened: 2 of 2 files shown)


src/megatron/bridge/data/datasets/sft.py
```diff
@@ -1197,7 +1197,7 @@ def collate_fn(self, batch):
         if self.pad_to_max_length:
             max_length = self.max_seq_length
         else:
-            max_length = min(self.max_seq_length, self._ceil_to_nearest(max_length, 16))
+            max_length = min(self.max_seq_length, self._ceil_to_nearest(max_length, self.pad_seq_length_to_mult))
         assert max_length <= self.max_seq_length
```
yaoyu-33, line 1204: change to max(16, self.pad_seq_length_to_mult)?

we always want to pad to multiple of 16, otherwise will end up with a slower kernel

shanecmoran: Good point — updated to `max(16, self.pad_seq_length_to_mult)` so we always pad to at least 16 for kernel performance.

PR #2702, opened 2026-03-09
[ckpt] feat: use explicit checkpoint format

denys-fridman:
# What does this PR do ?

Currently, checkpoint loading logic ignores the `ckpt_format` specified by the user and silently tries to infer the format. This leads to counter-intuitive failure in loading. I mistakenly had metadata.json in FSDP folder, but a user-specified ckpt_format should still take precedence.

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

## Summary by CodeRabbit

* **New Features**
  * Added support for explicitly specifying checkpoint format during base checkpoint loading. When a format is provided, automatic format inference from the checkpoint path is bypassed, giving users greater control over checkpoint loading behavior.

### DIFF
--- a/src/megatron/bridge/training/checkpointing.py
+++ b/src/megatron/bridge/training/checkpointing.py
@@ -1711,7 +1711,7 @@ def _load_checkpoint_from_path(
         update_num_microbatches(consumed_samples=state.train_state.consumed_train_samples, verbose=True)
 
     # Load model weights
-    if not skip_load_to_model_and_opt:
+    if not skip_load_to_model_and_opt and ckpt_type != CheckpointType.FSDP_DTENSOR:
         # Handle PEFT resume for strict loading
         load_strict = strict
         is_peft_resume = (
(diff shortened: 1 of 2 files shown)


src/megatron/bridge/training/checkpointing.py
```diff
@@ -2229,9 +2230,15 @@ def _load_base_checkpoint(
...
         return None, "", False, None
 
-    # Determine the checkpoint format
+    # Determine the checkpoint format from config, warn if it differs from auto-detected.
     checkpoint_path = get_checkpoint_name(load_dir, iteration, release)
-    ckpt_format = _get_checkpoint_format(checkpoint_path)
+    inferred_ckpt_format = _get_checkpoint_format(checkpoint_path)
+    ckpt_format = ckpt_cfg.ckpt_format
+    if ckpt_format != inferred_ckpt_format:
+        print_rank_0(
```
yaoyu-33: assert equal might be a safer choice here.

If we have a infer strategy, I feel better choice to add better strategy in infer method to consider meta data issue instead of adding a config. Seems confusing and extra step.

yaoyu-33:
@denys-fridman  plz check comment

src/megatron/bridge/training/checkpointing.py
denys-fridman: @yaoyu-33  Replaced with an assert. I'm not adding a config in this MR - `ckpt_format` was originally there.

PR #2778, opened 2026-03-12
[ckpt] feat: Add auto-config for self-contained Megatron-to-…

ShiftyBlock:
## What does this PR do?
 
Export now synthesizes `config.json` from the Megatron checkpoint's `run_config.yaml` using `--hf-model` as an architecture reference instead of a weight source. This enables conversion of custom or fine-tuned checkpoints that share an architecture with a known HF model.
 
## Changelog
 
- **Default auto-config export**: `from_auto_config` is now the default export path — no flag needed
- **Per-model config reconstruction**: `megatron_to_hf_config` overrides added for fields not bijective through `CONFIG_MAPPING`
- **Roundtrip test coverage**: autoconfig roundtrip tests for all bridges listed below
- **NemotronH 8B excluded**: upstream HF modeling bug (`_init_weights` missing `_is_hf_initialized` guard) corrupts `dt_bias` on load, making roundtrip comparison unreliable — [issue filed](https://huggingface.co/nvidia/Nemotron-H-8B-Base-8K/discussions/7)
 
## Tested Bridges
 
| Model | Status |
|-------|--------|
| NemotronH 8B | ⚠️ Not tested — [upstream HF bug](https://huggingface.co/nvidia/Nemotron-H-8B-Base-8K/discussions/7) in `_init_weights` corrupts `dt_bias` on load |
| Nemotron-3-Nano (MoE) | ✅ |
| Nemotron | ✅ |
| Gemma | ✅ |
| Gemma 2 | ✅ |
| Gemma 3 | ✅ |
| Mistral | ✅ |
| Ministral3 | ✅ |
| GLM-4.5 | ✅ |
| Qwen 2.5 | ✅ |
| Qwen 3 | ✅ |
| Qwen 3 MoE | ✅ |
| MiniMax M2 | ✅ |
| DeepSeek V2 | ✅ |
| DeepSeek V3 | ✅ |
| Llama-Nemotron | ✅ |
| OLMoE | ✅ |
 
Vision models (Gemma 3 VL, GLM-4.5 V, Qwen 2.5/3.5 VL, NemotronVL) are out of scope for this PR.
 
## Usage
 
```bash
python examples/conversion/convert_checkpoints.py export \
  --hf-model meta-llama/Llama-3.2-1B \
  --megatron-path ./checkpoints/my_custom_llama \
  --hf-path ./exports/my_custom_llama_hf
```
 
## Verification
 
All roundtrip tests pass — HF→Megatron→HF with bit-exact weights and 0.0 max logit difference.
 
## Acknowledgments
 
Thanks @ethanhharrison for user feedback and for sharing his checkpoint to debug early versions of this feature.

### DIFF
--- a/.github/workflows/remove-needs-author.yml
+++ b/.github/workflows/remove-needs-author.yml
@@ -1,0 +1,56 @@
+name: Remove needs-author label
+
+on:
+  issue_comment:
+    types: [created]
+  pull_request:
+    types: [synchronize]
+
+jobs:
+  remove-label-on-comment:
+    if: github.event_name == 'issue_comment' && github.event.issue.pull_request != null
+    runs-on: ubuntu-latest
+    steps:
+      - name: Remove needs-author if author commented
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const commenter = context.payload.comment.user.login;
+            const prAuthor = context.payload.issue.user.login;
+
+            if (commenter !== prAuthor) return;
+
+            const labels = context.payload.issue.labels.map(l => l.name);
+            if (!labels.includes('needs-author')) return;
+
+            await github.rest.issues.removeLabel({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.payload.issue.number,
+              name: 'needs-author',
+            });
+
+  remove-label-on-push:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Remove needs-author if author pushed commits
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const prAuthor = context.payload.pull_request.user.login;
+            const pusher = context.payload.sender.login;
+
+            if (pusher !== prAuthor) return;
+
+            const labels = context.payload.pull_request.labels.map(l => l.name);
+            if (!labels.includes('needs-author')) return;
+
+            await github.rest.issues.removeLabel({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.payload.pull_request.number,
+              name: 'needs-author',
+            });
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -166,6 +166,10 @@ Format your commit messages and PR titles as:
 - `ckpt` - Checkpoint conversion, loading, export, and save paths
 - `peft` - PEFT methods (LoRA, adapters) and adapter export
 - `perf` - Performance optimizations and throughput improvements
+- `distill` - Knowledge distillation
+- `prune` - Pruning and sparsity
+- `quant` - Quantization (PTQ, QAT, FP8 recipes)
+- `diffusion` - Diffusion model implementations and training
 - `ci` - CI, automation, and workflow infrastructure
 - `docs` - Documentation, examples, and contributor guidance
 - `build` - Dependencies, packaging, and environment setup
@@ -188,6 +192,23 @@ Format your commit messages and PR titles as:
 [BREAKING][training] refactor: Change optimizer config structure
 [ci, build] chore: Update ruff version
 ```
+
+## 🏷️ Labeling Your PR
+
+When you create a pull request, **add labels immediately** so reviewers and CI can route it correctly. At minimum, apply:
+
+1. **One type label** — `bug`, `feature`, `docs`, or `ci`
+2. **One or more area labels** — `area:model`, `area:recipe`, `area:training`, `area:data`, `area:ckpt`, `area:peft`, `area:perf`, `area:distill`, `area:diffusion`, `area:prune`, `area:quant`, `area:build`, or `area:misc`
+3. **`docs-only`** — if the PR touches only documentation (no code changes); this skips most CI jobs
+4. **`needs-review`** — when the PR is ready for review
+5. **`needs-more-tests`** — if the change needs additional test coverage; triggers both L0 and L1 CI
+6. **`high-complexity`** — if the PR is large, touches many files, or is prone to merge conflicts
+
+Add risk labels when applicable:
+- `breaking-change` — if any public API, CLI argument, config key, or function signature changes
+- `needs-more-tests` — if the change needs additional test coverage (also triggers L1 CI)
+
+Properly labeled PRs get faster reviews and avoid sitting in the triage queue.
 
 ## 🏷️ Repository Labels and Triage
 
@@ -245,6 +266,10 @@ Use one primary area label after triage:
 | `area:ckpt` | Checkpoint conversion, loading, export, and save paths |
 | `area:peft` | PEFT methods (LoRA, adapters) and adapter export |
 | `area:perf` | Performance optimizations, kernel integration, and throughput improvements |
+| `area:distill` | Knowledge distillation |
+| `area:diffusion` | Diffusion model implementations and training |
+| `area:prune` | Pruning and sparsity |
+| `area:quant` | Quantization (PTQ, QAT, FP8 recipes) |
 | `area:build` | Dependencies, packaging, images, and environment setup |
 | `area:misc` | Cross-cutting utilities, logging, helpers, and other changes that do not fit a primary domain |
 
@@ -317,7 +342,11 @@ We use [pytest](https://docs.pytest.org/en/stable/) for writing both unit and fu
 **Unit tests** aim to test functions in isolation. They generally do not depend on artifacts like Hugging Face checkpoints or larger datasets. Exception to this is a small toy dataset consisting of tokenizers.
 Unit tests are stored at `tests/unit_tests`. Please add your test to an existing folder or create a new one if none matches.
 
+> **Preferred:** Unit tests are strongly recommended over functional tests. CI resources are limited, so every functional test slot has a real cost. Cover as much logic as possible with unit tests before reaching for a functional test.
+
 **Functional tests** are integration tests that perform model training or operate on larger artifacts. We use pytest for writing these. In some cases, it might be desired to run your test (or parts of it) in a subprocess to avoid process contamination. We use `subprocess.run` for this inside the pytest function. Please add your test into one of the predefined folders. If none of the folders matches semantically, please reach out to the `@nvidia-nemo/automation` in your PR for consultation.
+
+> **GPU limit:** Functional tests must use **at most 2 GPUs**. Do not add tests that require more than 2 GPUs — they will not fit in the CI resource budget.
 
 ### Functional Test Launcher Scripts
 
@@ -326,10 +355,10 @@ Functional tests are placed in tiered launcher scripts inside [`tests/functional
 | Tier | Prefix | Trigger | Purpose |
 |------|--------|---------|---------|
 | **L0** | `L0_Launch_*.sh` | Every PR, main push, schedule | Core smoke tests — must be fast and stable |
-| **L1** | `L1_Launch_*.sh` | Main push + schedule (not PRs) | Broader model/recipe coverage |
+| **L1** | `L1_Launch_*.sh` | Main push + schedule; PRs labeled `needs-more-tests` | Broader model/recipe coverage |
 | **L2** | `L2_Launch_*.sh` | Schedule / `workflow_dispatch` only | VL models, checkpoint conversion, heavy quantization |
 
-When adding a new launcher script, choose the appropriate tier and **also update** [`.github/workflows/cicd-main.yml`](.github/workflows/cicd-main.yml) to include it in the corresponding `cicd-functional-tests-l{0,1,2}` job matrix:
+When adding a new launcher script, always start with the **L0** tier so it runs on every PR. A maintainer will adjust the tier later if the test is too slow or better suited for nightly coverage. You must **also update** [`.github/workflows/cicd-main.yml`](.github/workflows/cicd-main.yml) to include it in the corresponding job matrix:
 
 ```yaml
 # Example: adding an L1 test
@@ -343,22 +372,40 @@ Without this step, your new launcher script will not be picked up by CI.
 We use [uv](https://docs.astral.sh/uv/) for managing dependencies. For reproducible builds, our project tracks the generated `uv.lock` file in the repository.
 On a weekly basis, the CI attempts an update of the lock file to test against upstream dependencies.
 
-New required dependencies can be added by `uv add $DEPENDENCY`.
+### Adding a New Dependency
 
-New optional dependencies can be added by `uv add --optional --extra $EXTRA $DEPENDENCY`.
+**Adding required (non-optional) dependencies is strongly discouraged** and will be strictly reviewed. Every required dependency inflates the package and container image for all downstream consumers — most of whom will not need it. Prefer **optional dependencies** under an extra group whenever possible.
+
+If your feature requires a dependency that is not already in `pyproject.toml`, **submit the dependency change as a separate PR first**. Do not bundle dependency additions with feature code — this keeps reviews focused and makes CI failures easier to diagnose.
+
+1. Add the dependency to `pyproject.toml` (either via `uv add` or by editing the file directly):
+
+```bash
+# Preferred: optional dependency under an extra group
+uv add --optional --extra $EXTRA $DEPENDENCY
+
+# Required dependency (needs strong justification — affects all downstream)
+uv add $DEPENDENCY
+```
 
 `EXTRA` refers to the subgroup of extra-dependencies to which you're adding the new dependency.
 Example: For adding a TRT-LLM specific dependency, run `uv add --optional --extra trtllm $DEPENDENCY`.
 
-Alternatively, the `pyproject.toml` file can also be modified directly.
-
-Adding a new dependency will update UV's lock-file. Please check this into your branch:
+2. Regenerate the lock file:
 
 ```bash
-git add uv.lock pyproject.toml
-git commit -m "build: Adding dependencies"
+uv lock
+```
+
+3. Commit both files and open a PR:
+
+```bash
+git add pyproject.toml uv.lock
+git commit -s -m "[build] chore: Add $DEPENDENCY"
 git push
 ```
+
+4. Once the dependency PR is merged, rebase your feature branch onto `main` and open the feature PR.
 
 ### 🧹 Linting and Formatting
 
@@ -372,9 +419,16 @@ uv run ruff format .
 Note: If `ruff` is missing, please follow the [installation](#local-workstation) guide.
 
 
-## 📄 Documentation Requirement
+## 📄 Documentation and Test Requirements
 
-**Important**: All new key features (e.g., enabling a new model, enabling a new parallelism strategy) must include documentation update (either a new doc or updating an existing one). This document update should:
+### All Features
+
+**Every feature PR must evaluate whether documentation and tests need to be added or updated.** This applies to all changes, not just large features. Before submitting a PR, check:
+
+- [ ] **Docs**: Does this change need a new doc page, or does an existing doc need updating?
+- [ ] **Tests**: Does this change need new unit or functional tests, or do existing tests need updating?
+
+For new key features (e.g., enabling a new model, enabling a new parallelism strategy), documentation is **required**. The documentation should:
 
 - Explain the motivation and purpose of the feature
 - Outline the technical approach and architecture
@@ -387,6 +441,16 @@ This ensures that all significant changes are well-thought-out and properly docu
 2. **Developer Extensibility**: Enables developers to understand the internal architecture and implementation details, making it easier to modify, extend, or adapt the code for their specific use cases
 
 Quality documentation is essential for both the usability of Megatron-Bridge and its ability to be customized by the community.
+
+### Refactoring PRs
+
+Refactoring PRs that rename symbols, move files, or change paths are high risk for creating stale references. When a refactor changes any public name or path, you **must** check whether the following need corresponding updates:
+
+- [ ] **Docs**: Any docs that reference the old names, paths, config keys, or CLI arguments
+- [ ] **Docstrings**: Docstrings in the codebase that mention the old names or paths
+- [ ] **Scripts**: Example and training scripts under `scripts/` that import or reference the old names or paths
+
+A refactor PR that renames something without updating all references will break users silently. Reviewers should verify these are addressed before approving.
 
 ## ✨ Code Quality
 
--- a/README.md
+++ b/README.md
@@ -12,13 +12,15 @@
 
 ## 📣 News
 
-- [03/11/2026] Day 0 support for [**Nemotron 3 Super**](https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/examples/models/nemotron_3)! Checkpoint conversion and SFT/LoRA recipes are available on the [`super-v3`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/super-v3) branch. Read the [blog post](https://developer.nvidia.com/blog/introducing-nemotron-3-super-an-open-hybrid-mamba-transformer-moe-for-agentic-reasoning/).
+- [03/31/2026] **Agent Skills for Megatron Bridge!** We've added a [`skills/`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/skills) directory with structured guides that AI coding agents (Cursor, Claude Code, Codex, etc.) can use to help you add model support, set up dev environments, tune performance, and more. Try them out, and PRs to improve or add new skills are very welcome!
+
+- [03/26/2026] [**Nemotron 3 Super**](https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/examples/models/nemotron_3) is now on **main**! Checkpoint conversion and SFT/LoRA recipes (120B-A12B) are available in the main branch. Read the [blog post](https://developer.nvidia.com/blog/introducing-nemotron-3-super-an-open-hybrid-mamba-transformer-moe-for-agentic-reasoning/).
 
 - [03/12/2026] **Deprecating Python 3.10 support:** We're officially dropping Python 3.10 support with the upcoming 0.4.0 release. Downstream applications must raise their lower boundary to 3.12 to stay compatible with Megatron-Bridge.
 
 - [12/16/2025] [Mind Lab](https://macaron.im/mindlab) successfully used Megatron-bridge and [VeRL](https://github.com/volcengine/verl) to trained GRPO Lora for Trillion-parameter model on 64 H800 - See their [techblog](https://macaron.im/mindlab/research/building-trillion-parameter-reasoning-rl-with-10-gpus).
 
-- [12/15/2025] Day 0 support for [NVIDIA-NeMotron-3-Nano-30B-A3B-FP8](https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8)! [Reproducible code](https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/nano-v3) and custom NGC container: [nvcr.io/nvidia/nemo:25.11.nemotron_3_nano](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/nemo?version=25.11.nemotron_3_nano)
+- [12/15/2025] Day 0 support for [NVIDIA-NeMotron-3-Nano-30B-A3B-FP8](https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8)! [Reproducible code](https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/examples/models/nemotron_3/nano) and custom NGC container: [nvcr.io/nvidia/nemo:25.11.nemotron_3_nano](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/nemo?version=25.11.nemotron_3_nano)
 
 ## Overview
 
@@ -44,6 +46,22 @@ docker run --rm -it -w /workdir -v $(pwd):/workdir \
 ```
 
 For development installation and additional details, please refer to our [Contribution guide](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md).
+
+### Megatron-Core Submodule (main & dev)
+
+Megatron Bridge pins [Megatron-Core](https://github.com/NVIDIA/Megatron-LM) as a git submodule at `3rdparty/Megatron-LM`. The repository tracks two pinned commits — one from the upstream **main** branch (default) and one from **dev** — managed by `scripts/switch_mcore.sh`.
+
+The submodule committed to the repo always points to the **main** commit. Use the **dev** commit when you need a Megatron-Core feature or fix that has not yet landed on main, or to validate forward-compatibility with upcoming MCore changes:
+
+```bash
+./scripts/switch_mcore.sh status   # Show current commit
+./scripts/switch_mcore.sh dev      # Switch to dev; then run: uv sync
+./scripts/switch_mcore.sh main     # Switch back; then run: uv sync --locked
+```
+
+> **Note:** `uv.lock` is generated against the main commit. After switching to dev, use `uv sync` (without `--locked`). After switching back to main, use `uv sync --locked`.
+
+The dev branch follows Megatron-LM's upstream [dev branch philosophy](https://github.com/NVIDIA/Megatron-LM/tree/dev) — features are experimental, follow a streamlined review process, and must graduate to stable within 6 months or be deprecated.
 
 ## ⚡ Quickstart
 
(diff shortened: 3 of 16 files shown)


davidyang05:
outdated, see revisions for history.

davidyang05:
@coderabbitai
Hi coderabbit, I realized that you actually are reading my old code. Can you take a deeper look at my new code - which I force pushed just today?

davidyang05:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> `@ShiftyBlock` Thank you for the thorough round-trip verification — this is excellent evidence for the correctness of the `--auto-config` implementation!
> 
> Here's a quick summary of what was validated:
> 
> Verification	Result
> HF → Megatron → HF (`--hf-model`) vs HF → Megatron → HF (`--auto-config`) on Nemotron-H-8B-Base-8K	✅ 311/311 tensors identical
> Megatron → HF (`--auto-config`) vs ground truth on custom 7B checkpoint	✅ 2305/2305 tensors identical
> A few notes based on the results:
> 
> 1. **Config synthesis correctness**: The fact that all tensor keys are present with no mismatches confirms that `to_hf_config()` / `to_hf_export_config()` correctly reconstructs the HF model architecture from `run_config.yaml`.
> 2. **Weight mapping fidelity**: Bit-identical tensors across both paths confirm the Megatron → HF weight mapping logic is unaffected by the new code path.
> 3. **Sharding behavior**: It's worth noting in the documentation that `--auto-config` outputs a single `model.safetensors` rather than preserving the original HF shard layout — this is expected behavior for custom models without an upstream HF repo, but users migrating from `--hf-model` should be aware of this difference.
> 
> Given this verification, the implementation looks sound. The one remaining open item from the PR checklist is unit tests — the round-trip approach you described (import via HF-to-Megatron, then compare both export paths) would be a natural fit as a regression test. Would you like me to help scaffold that test?
> 
> 🧠 Learnings used

Hi @coderabbitai, I'm not too familiar with NVIDIA's testing infrastructure. Is it feasible to spin up GPUs for the purpose of this round-trip test? Some tests I've written are mostly local which load models that use old-megatron-format and new-megatron-format from classmates in my class. I can also look into Unit Testing key parts - if you think this might be a fruitful direction.

tests/functional_tests/converter/test_nemotronh_autoconfig.py
```diff
@@ -0,0 +1,254 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+NemotronH converter round-trip and forward pass tests.
+
+Tests:
+- test_roundtrip_creates_checkpoint: HF→Megatron→HF weight round-trip (bit-exact).
```
davidyang05: @coderabbitai May I request a code review? I've added some documentation in the comments for using --auto-config. The UX has slightly changed but for very necessary reasons to prevent regressions due to user misunderstanding. Here is the output of my tests: https://pastebin.com/27rPkgc0

davidyang05:
@coderabbitai Can you review the latest version of my PR? I have attached the results of the provided test_nemotronh_autoconfig.py tests here https://pastebin.com/zV4wAezt. Documentation is listed in in the comments I added. I should have addressed all of your concerns, in addition, I revamped the CLI interface to require a huggingface model_id (currently only supporting Nemotron-3-Nano). The primary reason is to prevent silent bugs caused by **differences in the forward pass logic between Nemotron-3-Nano's and Nemotron-3-Super's modeling file**. 

Some other changes worth mentioning is the @cached_property fix for `_model_bridge` to allow overriding the `HUB_REFERENCE_REPO` value required to pull the configuration and modeling files. I would also mention that the scope is currently limited to Nemotron-3-Nano, and that the `megatron_to_hf_config` pattern validator will raise an error for non-MoE models (the use of `-` will raise a ValueError).

examples/conversion/convert_checkpoints.py
davidyang05, line 1: Requesting a code review @coderabbitai

davidyang05:
Hi @yaoyu-33, may you please take a look at my PR? Thank you very much for your time and help!

src/megatron/bridge/training/model_load_save.py
```diff
@@ -382,6 +389,11 @@ def load_megatron_model(
     model_cfg.context_parallel_size = 1
     model_cfg.expert_model_parallel_size = 1
     model_cfg.expert_tensor_parallel_size = 1
+    # Flex dispatcher requires TPxEP > 1; fall back to allgather for single-rank export
+    if getattr(model_cfg, 'moe_token_dispatcher_type', None) == 'flex':
+        model_cfg.moe_token_dispatcher_type = 'allgather'
```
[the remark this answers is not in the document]
davidyang05: great catch

src/megatron/bridge/models/nemotronh/nemotron_h_bridge.py
```diff
@@ -104,6 +110,48 @@ def provider_bridge(self, hf_pretrained: PreTrainedCausalLM) -> MambaModelProvid
 
         return provider
 
+    @classmethod
+    def megatron_to_hf_config(cls, provider) -> dict:
+        hf_cfg = super().megatron_to_hf_config(provider)
+
+        # Clean up base CONFIG_MAPPING field that don't apply to NemotronH
+        hf_cfg.pop("num_experts", None)
```
yaoyu-33: can we do a general filter? check what hf config expected and remove additonal? we can make a it util. easy to share between models

examples/conversion/convert_checkpoints.py
```diff
@@ -50,6 +50,12 @@
     --megatron-path ./checkpoints/custom_model \
     --hf-path ./exports/custom_model_hf \
     --no-progress
+
+  # Export using auto-config (no HF model needed, synthesizes config from checkpoint)
+  uv run python examples/conversion/convert_checkpoints.py export \
+    --auto-config nemotron-3-nano \
```
yaoyu-33: this design a bit ugly, we can keep the same --hf-model, once given hf-model, you are able to get the model info from model bridge registered metadata (in decorator) and follow up. I dont want to add `AUTO_CONFIG_ALIASES`, seems not necessary.

src/megatron/bridge/models/nemotronh/nemotron_h_bridge.py
```diff
@@ -54,7 +54,13 @@ class NemotronHBridge(MegatronModelBridge):
         >>> provider = bridge.to_megatron_provider()
     """
 
+    # Reference Hub repo for downloading generic modeling files (configuration_nemotron_h.py,
+    # modeling_nemotron_h.py). Used by auto_bridge.py's config-only save path.
+    HUB_REFERENCE_REPO = "nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16"
```
cuichenx: this is not general enough to the nemotron_h bridge class

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -602,7 +639,40 @@ def _filter_quant(gen):
...
+            if box_of_tensors:
+                flush_tensors(box_of_tensors, current_box_id)
+
+            index = {
+                "metadata": {"total_size": total_size},
+                "weight_map": weight_map,
+            }
+            if not (dist.is_available() and dist.is_initialized()) or dist.get_rank() == 0:
+                with open(Path(path) / "model.safetensors.index.json", "w") as f:
+                    json.dump(index, f, indent=4)
```
cuichenx: is it feasible to initialize an HF model and save the weights directly, instead of doing this roundabout way to save index.json?

examples/conversion/convert_checkpoints.py
cuichenx: +1, I don't think --auto-config requires an argument. The model class class should be inferred from the megatron model. Ideally --auto-config just tells mbridge it's ok to use megatron config to override HF config

examples/conversion/convert_checkpoints.py
davidyang05: Would like a code review @coderabbitai

src/megatron/bridge/models/nemotronh/nemotron_h_bridge.py
davidyang05: Added in the latest commit!

examples/conversion/convert_checkpoints.py
davidyang05: Added in the latest commit!

src/megatron/bridge/models/nemotronh/nemotron_h_bridge.py
davidyang05: Removed in the latest commit!

src/megatron/bridge/models/conversion/auto_bridge.py
davidyang05: I agree this is a roundabout way to save index.json. Before deciding on my current approach, I also considered using SafeTensorsStateSource that could potentially do distributed saving. However generating the shard_to_rank file from scratch requires a single-rank pass over the tensors, so this method does not yield performance improvement. I went with direct safetensors saving approach so that tensors are written exactly as they come from Megatron, which saves the model instantiation and dtype inference step. The sharding logic comes from huggingface_hub.

davidyang05:
Hi @cuichenx @yaoyu-33, I've made the changes you've discussed:
- Modified CLI so that auto-config is an optional boolean flag and reverted hf-model to be required string argument
- Add ValueError guard for models that don't override megatron_to_hf_config
- Added filter_hf_config_fields() utility 
- Use HF `split_torch_state_dict_into_shards` instead of manual index.json to save model safetensors

### Fixes
I also put back the (hybrid_override_pattern, hybrid_layer_pattern) to the CONFIG_MAPPING for nemotron_h_bridge.py, which I noticed got changed in a previous migration PR. This is to be back in line with the [Add flexible virtual pipeline parallel (fVPP) to hybrid model](https://github.com/NVIDIA/Megatron-LM/pull/3377/changes) PR.

### Testing
Tests continue to pass after changes: https://pastebin.com/U9BF9uUM

Also tagging @coderabbitai to request AI review as well.

davidyang05:
Hi @yaoyu-33, hope you're doing well. All feedback has been addressed, could the needs-author label be removed? Thank you!

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -1303,11 +1303,8 @@ def build_conversion_tasks(
         populated.
         """
 
-        # Ensure hf_pretrained has the required state structure
-        if not (hasattr(hf_pretrained, "state") and hasattr(hf_pretrained.state, "source")):
-            raise ValueError("hf_pretrained.state.source is required for weight ordering")
```
cuichenx, line 1327: this value error is now replaced with a silent failure in the non-auto config path, we should at least add a warning

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -942,23 +999,10 @@ def export_ckpt(
             ...     "./hf_exports/my_model"
             ... )
 
-            >>> # Export with specific settings
-            >>> bridge.export_ckpt(
-            ...     "./megatron_checkpoints/my_model",
-            ...     "./hf_exports/my_model",
-            ...     show_progress=False
-            ... )
-
```
cuichenx, line 951: why is this removed?

src/megatron/bridge/models/conversion/model_bridge.py
davidyang05: will do

src/megatron/bridge/models/conversion/auto_bridge.py
davidyang05: thanks for the catch - will add it back!

cuichenx (approved):
LGTM

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/commits/aa690f2b4718e290f5b8035145791d241b1354f9

tests/functional_tests/converter/test_nemotron_h_autoconfig.py
```diff
@@ -0,0 +1,216 @@
...
+        hidden sizes, and expert configurations than the reference Nano model.
+        """
+        import shutil
+
+        from transformers import AutoModelForCausalLM
+
+        from megatron.bridge import AutoBridge
+
+        hf_model_id = "shiftyblock/DAARTH-7B-A1B"
+        hf_revision = "f21eaa6c4b83e420715b9ade1bdb65366e325d44"
```
cuichenx: this may not work in our ci system. please create a toy model on the fly (check other functional tests).

Also, please create a L1_xyz.sh launch script and add the script in cicd-main.yml, otherwise this test won't run

davidyang05:
@cuichenx Thank you very much for your valuable feedback throughout your reviews. Very appreciate you for pointing me to the existing toy model testing infrastructure - this helped refactor the tests and make the code much cleaner.

### Summary of changes
- Moved auto-config tests to `test_nemotron_h_conversion.py` by leveraging existing fixtures
- Alleviated heavy GPU memory burden of loading full Nano 30B model
- **Removed** old testing harness file (`test_nemotron_h_autoconfig.py`).
 
Ready for review!

davidyang05:
Hi @cuichenx, would you mind taking a quick look at the test? Thanks for your help!

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/commits/2d274ad26b1e5be8dadd82e052312ab3af33472c

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/commits/a48bf36e2ade82b24fdb82c334c8fcf4735ca5d7

cuichenx:
unit tests are failing 

```
FAILED tests/unit_tests/models/test_auto_bridge.py::TestAutoBridge::test_save_hf_pretrained_config_only_raises - TypeError: argument of type 'Mock' is not iterable
FAILED tests/unit_tests/models/test_auto_bridge.py::TestAutoBridge::test_export_ckpt_config_only_raises - FileNotFoundError: [Errno 2] No such file or directory: 'megatron_checkpoint'
```

davidyang05:
> unit tests are failing
> 
> ```
> FAILED tests/unit_tests/models/test_auto_bridge.py::TestAutoBridge::test_save_hf_pretrained_config_only_raises - TypeError: argument of type 'Mock' is not iterable
> FAILED tests/unit_tests/models/test_auto_bridge.py::TestAutoBridge::test_export_ckpt_config_only_raises - FileNotFoundError: [Errno 2] No such file or directory: 'megatron_checkpoint'
> ```

Hi Chen, thanks for bringing it up! I'm planning to rewrite both tests to the new contract where, since we are enabling config-only export, these tests should ideally *not fail* on the code. Is that an acceptable direction to take?

examples/conversion/convert_checkpoints.py
```diff
@@ -140,16 +147,22 @@ def export_megatron_to_hf(
...
     strict: bool = True,
 ) -> None:
     """
     Export a Megatron checkpoint to HuggingFace format.
 
     Args:
+        hf_model: HuggingFace model ID or path (loads full model with weights).
         megatron_path: Directory path where the Megatron checkpoint is stored
         hf_path: Directory path where the HuggingFace model will be saved
+        auto_config: If True, synthesizes the HF config from run_config.yaml using
```
yaoyu-33: I feel we can default auto_config to True and remove the args. I cannot think of a case we need this False

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -577,29 +581,65 @@ def save_hf_pretrained(
             saves the configuration files, while weight saving is coordinated
             across all ranks.
         """
-        if not isinstance(self.hf_pretrained, PreTrainedCausalLM):
-            raise ValueError(
-                "save_hf_pretrained requires a pretrained HuggingFace model. "
-                "AutoBridge.from_hf_config() creates a config-only bridge; "
-                "use AutoBridge.from_hf_pretrained(...) instead."
-            )
+        if not isinstance(self.hf_pretrained, (PreTrainedCausalLM, PretrainedConfig)):
```
yaoyu-33, line 645: might needs to update in docs as well.

examples/conversion/convert_checkpoints.py
davidyang05: Hi Yu, this is a really sharp insight - thanks for the comment. In my first iteration, my implementation relied on the using the class level configuration.py file to fill in the defaults. But your comment about making auto_config the default made me think deeply - what would enabling it look like? Under the class level conversion, supporting models requires testing each added bridge on each individual model to see if conversion succeeds, showing generalization. With the reference model, however, this becomes a non-issue – generalization becomes cheap. I'll definitely make these changes.

old comment: This is a really interesting suggestion! Let me think about it further and get back to you. 

old thoughts: Hi Yu, not all bridges implement megatron_to_hf_config yet, so I'm a bit hesitant on making this the default.

examples/conversion/convert_checkpoints.py
```diff
@@ -174,9 +187,26 @@ def export_megatron_to_hf(
...
+    # Create bridge: auto-config (synthesize from run_config.yaml) or full pretrained
+    if auto_config:
+        from transformers import AutoConfig, PretrainedConfig
+
+        from megatron.bridge.models.conversion.utils import filter_hf_config_fields
+        from megatron.bridge.training.model_load_save import load_model_config
+
+        megatron_cfg, _ = load_model_config(str(config_files[0].parent))
+        config = AutoConfig.from_pretrained(hf_model, trust_remote_code=True)
+        bridge = AutoBridge.from_hf_config(config)
```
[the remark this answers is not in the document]
davidyang05: This strict check is intentional, as I can only confirm conversion verification against the overridden versions and have not tested the base class default. I will add this comment.

tests/functional_tests/test_groups/models/nemotronh/test_nemotron_h_conversion.py
```diff
@@ -577,3 +577,97 @@ def test_nemotron_3_nano_conversion_parallelism(
...
+        if iter_folders:
+            latest = max(iter_folders, key=lambda f: int(f.name.replace("iter_", "")))
+            megatron_iter = megatron_root_path / latest.name
+        else:
+            megatron_iter = megatron_root_path
+
+        # Megatron → HF via auto-config export
+        megatron_cfg, _ = load_model_config(str(megatron_iter))
+        config = AutoConfig.from_pretrained(local_model_path, trust_remote_code=True)
+        bridge = AutoBridge.from_hf_config(config)
```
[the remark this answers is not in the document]
davidyang05: You have a good point here and I will implement 1.

src/megatron/bridge/models/conversion/auto_bridge.py
davidyang05: added

tests/functional_tests/test_groups/models/nemotronh/test_nemotron_h_conversion.py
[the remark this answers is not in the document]
davidyang05: Implemented

examples/conversion/convert_checkpoints.py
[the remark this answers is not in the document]
davidyang05: added

examples/conversion/convert_checkpoints.py
```diff
@@ -50,6 +50,12 @@
     --megatron-path ./checkpoints/custom_model \
     --hf-path ./exports/custom_model_hf \
     --no-progress
+
+    # Export synthesizes config from checkpoint + HF model reference by default
```
davidyang05: Will remove redundant comment

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -701,7 +847,29 @@ def _filter_quant(gen):
...
         else:
-            raise ValueError("The state source is not a SafeTensorsStateSource, cannot save in streaming mode.")
+            # Config-only path: shard and write safetensors directly
+            import json
+
+            from huggingface_hub import split_torch_state_dict_into_shards
+
+            # NOTE: Collects the full state dict into CPU memory before sharding.
+            # For very large models (>70B), this may require significant host RAM.
+            state_dict = {name: tensor.contiguous().cpu() for name, tensor in generator}
```
davidyang05: change to 
if rank == 0:
        state_dict = {name: tensor.contiguous().cpu() for name, tensor in generator}
    else:
        for _ in generator:
            pass
        state_dict = None

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1295,7 +1462,7 @@ def mla_transformer_config(self) -> MLATransformerConfig:
 
         return self._create_config_from_provider(_model_provider, MLATransformerConfig)
 
-    @property
+    @cached_property
```
davidyang05: change back to property

tests/unit_tests/models/test_auto_bridge.py
```diff
@@ -569,11 +569,11 @@ def test_save_hf_pretrained(self, mock_is_init, mock_is_avail, mock_barrier, moc
                 )
 
     def test_save_hf_pretrained_config_only_raises(self):
```
davidyang05: config no longer has bridge original dict, consider removing this test

tests/unit_tests/models/test_auto_bridge.py
```diff
@@ -875,16 +875,6 @@ def test_export_ckpt_basic(self):
                     strict=False,
                 )
 
-    def test_export_ckpt_config_only_raises(self):
```
davidyang05, line 878: Need to add tests for autoconfig being the default path

davidyang05: added

davidyang05:
# Diff Summary Since Last Review

## Core Infrastructure

### auto_bridge.py
- Fix import: remove unused `property`
- Downgrade modeling-file download error from `logger.error` to `logger.warning` (expected for standard models)
- Revert `_config_only_pretrained` safetensors shim attachment (removed)
- Revert `stream_weights_megatron_to_hf` calls back to `self.hf_pretrained` (removed shim wiring)
- Remove unused `StateDict` import

### model_bridge.py (+1 line)
- Add `router_aux_loss_coef` -> `moe_aux_loss_coeff` alias to `CONFIG_MAPPING`

---

## Per-Model Bridge Fixes

### DeepSeek V2 & V3 bridges (+28 lines each, identical logic)
- Add `megatron_to_hf_config()` override: reconstruct `first_k_dense_replace` from `moe_layer_freq`, `n_shared_experts` from shared/ffn size ratio, normalize `num_nextn_predict_layers`/`n_group`/`topk_group` None->int

### Gemma2 bridge (+14 lines)
- Add `megatron_to_hf_config()`: reconstruct `query_pre_attn_scalar`, `attn_logit_softcapping`, `final_logit_softcapping`, `sliding_window`

### Gemma3 bridge (+27 lines)
- Add `megatron_to_hf_config()`: reconstruct dual `rope_theta`/`rope_local_base_freq` from tuple, `sliding_window`, `query_pre_attn_scalar` from `softmax_scale`, `rope_scaling`

### GLM45 bridge (+14 lines changed)
- `build_conversion_tasks`: guard for config-only path (no `.state.source`)
- `_uses_fused_experts`: return `True` on config-only path (GLM always uses fused experts)
- `_hf_expert_suffix`: add comment for config-only fallback

### GLM MoE mappings (+5 lines)
- `GLMExpertDownProjMapping`: flip `transpose_on_export=True` -> `False`, update docstring

### Llama Nemotron bridge (+22 lines)
- Store `rope_scaling=True` on provider during ingest
- Add `megatron_to_hf_config()`: reconstruct `rope_scaling` dict with llama3 type/factor/freq params

### MiniMax M2 bridge (+1 line changed)
- Read `router_aux_loss_coef` from HF config instead of hardcoding `1e-3`

### Nemotron bridge (-5/+1 lines)
- Replace local `squared_relu` with canonical `megatron.core.activations.squared_relu` import (fixes reverse export activation identity)

### OLMoE bridge (+4 lines)
- Add validation: raise `ValueError` if `scoring_func != "softmax"`

### Qwen3 MoE bridge (+8 lines)
- Add `megatron_to_hf_config()`: hardcode `decoder_sparse_step=1` and `norm_topk_prob=False`

---

## Test Infrastructure

### utils.py (+16 lines)
- Add debug prints for exported weight shapes and config fields

---

## Functional Tests

### deepseek test (+72 lines)
- Add V2 toy config + fixture, add V2 and V3 autoconfig roundtrip tests, tweak V3 config (`n_group`/`n_routed_experts`/`topk_group`)

### glm45 test (+25 lines)
- Fix `model_type` "glm" -> "glm4_moe" throughout, add load verification, add autoconfig roundtrip test

### llama_nemotron test (+6 lines)
- Add autoconfig roundtrip test

### ministral3 test (+14 lines)
- Add autoconfig roundtrip test

### nemotron test (+6 lines)
- Add autoconfig roundtrip test

### olmoe test (+6 lines)
- Add autoconfig roundtrip test

---

## Unit Tests

### gemma3 unit test (+34 lines)
- Add 2 tests for `megatron_to_hf_config` reverse export (4B with scaling, 1B without)

davidyang05:
Checking in here: I'll have touch ups on this done before 12pm pst tomorrow. Thanks for your consideration!

davidyang05:
## Self-contained Megatron-to-HF export via auto-config synthesis

Export now synthesizes an HF `config.json` from `run_config.yaml` in the Megatron checkpoint plus a reference HF model, instead of requiring the original HF weights at export time. This lets you export custom or pruned checkpoints that share an architecture with a known HF model.

### What changed

- **`AutoBridge.from_auto_config(megatron_path, hf_model_id)`** — new constructor that builds a config-only bridge from a Megatron checkpoint + reference HF config, without loading any weights.
- **`convert_checkpoints.py export`** now uses `from_auto_config` by default. `--hf-model` is the architecture reference, not a weight source.
- **`save_hf_pretrained` / `export_ckpt`** accept config-only bridges: writes `config.json` directly and downloads custom modeling `.py` files from the Hub when needed.
- **Per-model `megatron_to_hf_config` overrides** added for DeepSeek V2/V3, Gemma 2/3, GLM 4.5, Llama-Nemotron, NemotronH, OLMoE, and Qwen3-MoE to reconstruct fields that aren't bijective through `CONFIG_MAPPING`.
- **`conform_config_to_reference`** projects the synthesized config onto the reference key set and backfills missing keys.
- **Autoconfig roundtrip functional tests** added for all 15 supported bridges via shared `autoconfig_roundtrip()` helper in `tests/functional_tests/utils.py`.
- **L1 CI launch script** for `autoconfig_roundtrip` tests.
- **Backward-compat fixes** in `load_model_config` for `hybrid_override_pattern` rename and `pipeline_model_parallel_layout` dict→None.
- **Flex dispatcher fallback** to `allgather` when TP×EP=1 during single-rank export.

### Usage

```bash
python examples/conversion/convert_checkpoints.py export \
  --hf-model meta-llama/Llama-3.2-1B \
  --megatron-path ./checkpoints/my_custom_llama \
  --hf-path ./exports/my_custom_llama_hf

**Comment by davidyang05** [comment]
Dear @cuichenx @yaoyu-33, all reviewer feedback addressed in latest push!

File: examples/conversion/convert_checkpoints.py [hunk 19]
```diff
@@ -174,9 +176,8 @@ def export_megatron_to_hf(
 
     print(f"📋 Found configuration: {config_files[0]}")
 
-    # For demonstration, we'll create a bridge from a known config
-    # This would typically be extracted from the checkpoint metadata
-    bridge = AutoBridge.from_hf_pretrained(hf_model, trust_remote_code=True)
+    # Export always uses the config synthesized from checkpoint + HF reference.
+    bridge = AutoBridge.from_auto_config(megatron_path, hf_model)
```
cuichenx, line 180: Q: iiuc the behavior has changed from HF config as the source of truth to megatron config as source of truth. This means if someone changes the mapping megatron_to_hf_config, then export could be incorrect. Or for a new model, it's another layer of verification responsibility on the model supporter. Is my understanding correct? I would rather use a flag in this script and do `from_auto_config` only if explicitly instructed, otherwise do `from_hf_pretrained` as the default

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/commits/b70645fb14ded5f698c2e52f62131c95a842e182

cuichenx:
please run `pre-commit run --all-files` to resolve lint issues

examples/conversion/convert_checkpoints.py
davidyang05: Hi Chen, thanks for the review! I'd be more than happy to contribute an example script and user guide.
On the default behavior, your concern makes sense. This change was based on Yu's suggestion [here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778#discussion_r2978128880) to default auto_config to True. 

I think both approaches are reasonable and I don't have a strong opinion on which should be the default. One promising approach could be an opt-in at the model_bridge level, say `supports_auto_config` as an attribute, and fall back to `from_hf_pretrained` if False. Would you and Yu like to align on the preferred approach?

davidyang05:
@cuichenx linted and rebased (wasn't working with the github button)

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/commits/ab703ef34c226887bc5b102a64ed963973c4f02a

cuichenx:
Running tests. please refrain from pushing for now otherwise tests will be interrupted. Can you also check my comment in [examples/conversion/convert_checkpoints.py](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/files#diff-4b5ba14ac6f9d36c3cc22e171cbe9c98770e8c6a719fdf53037fa3fdcd7862b3)?

davidyang05:
> Running tests. please refrain from pushing for now otherwise tests will be interrupted. Can you also check my comment in [examples/conversion/convert_checkpoints.py](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/files#diff-4b5ba14ac6f9d36c3cc22e171cbe9c98770e8c6a719fdf53037fa3fdcd7862b3)?

I did reply, could you review my reply?

cuichenx:
lint is failing again 👀

davidyang05:
> lint is failing again 👀

*sigh* i ran the precommit thing. let me lint manually (turns out i forgot to commit 😢 )

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/commits/8d8803d4fafa35431ff3276d6d0913db4d8fba2a

cuichenx:
> > Running tests. please refrain from pushing for now otherwise tests will be interrupted. Can you also check my comment in [examples/conversion/convert_checkpoints.py](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/files#diff-4b5ba14ac6f9d36c3cc22e171cbe9c98770e8c6a719fdf53037fa3fdcd7862b3)?
> 
> I did reply, could you review my reply?

I don't see your reply. Is it pending?

davidyang05:
> > > Running tests. please refrain from pushing for now otherwise tests will be interrupted. Can you also check my comment in [examples/conversion/convert_checkpoints.py](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/files#diff-4b5ba14ac6f9d36c3cc22e171cbe9c98770e8c6a719fdf53037fa3fdcd7862b3)?
> > 
> > 
> > I did reply, could you review my reply?
> 
> I don't see your reply. Is it pending?

Oh... oops 😬 . Thanks a lot! Let me submit my reply via submit review.

tests/functional_tests/converter/test_nemotronh_autoconfig.py
davidyang05: will resolve after cicd

cuichenx:
Thanks I see it now!

Sorry, missed the previous message with @yaoyu-33 , let me discuss this with him and get back to you

davidyang05:
Hi @cuichenx, I looked into the failing tests and they seem to be caused by Nano fixture’s Hub dependency. The failing Nemotron-3-Nano tests error out while building `nemotron_3_nano_toy_model_path`, where `dynamic_module_utils.get_class_from_dynamic_module()` tries to fetch remote modeling code from Hugging Face.

I believe this is some CI issue as [this CI](https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/23506028407/job/68416414564) ran without issues just two days ago. 

Happy to look into making the fixture work offline if that's the right direction to take.

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/commits/7f6ab13a8e225f4e548ff983f21d618b87f10fb6

cuichenx:
> Hi @cuichenx, I looked into the failing tests and they seem to be caused by Nano fixture’s Hub dependency. The failing Nemotron-3-Nano tests error out while building `nemotron_3_nano_toy_model_path`, where `dynamic_module_utils.get_class_from_dynamic_module()` tries to fetch remote modeling code from Hugging Face.
> 
> I believe this is some CI issue as [this CI](https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/23506028407/job/68416414564) ran without issues just two days ago.
> 
> Happy to look into making the fixture work offline if that's the right direction to take.

The current PR as is is fine. We think any issue would be caught by the roundtrip verification script. 
I've retrigger the CI. This PR looks good to me

cuichenx (dismissed):

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/commits/f01c771b37d84ae2fe0a18315de6f9b09c557d60

davidyang05:
Hi @cuichenx thanks for re-triggering CI! It seems like it’s failing again for the same reason. 

My first thought was that this is caused by my code changes but it’s failing on creating the toy model itself. What confuses me a bit is that this an existing test fixture that has passed before. Would you have any intuition as to why this might be happening?

cuichenx:
> Hi @cuichenx thanks for re-triggering CI! It seems like it’s failing again for the same reason.
> 
> My first thought was that this is caused by my code changes but it’s failing on creating the toy model itself. What confuses me a bit is that this an existing test fixture that has passed before. Would you have any intuition as to why this might be happening?

don't worry about the failed tests. they are usually transient.

davidyang05:
Link to tests: https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/23770676789/job/69294859229?pr=2778

The Gemma (Gemma1) model test is failing due to the gemma_bridge.py file hardcoding the activation function as "fast_gelu" when it is supposed to be "gelu" as per the [HF config](https://huggingface.co/google/gemma-7b/blob/main/config.json).

Pushed a new commit. Verifying that I've linted:
```
(base) shiftyblock@instance-20260131-203914:~/a2/part_2/Bridge-Debug$ pre-commit run --all-files
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
ruff.....................................................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed
Disallow '_' in Markdown filenames...................(no files to check)Skipped
```

Oh, sorry, I didn't realize pushing a new commit would completely invalidate my old test results. Could you kindly rerun CI @cuichenx?

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2778/commits/94cafada8120f9c72dbca40a83647e53579a317d

cuichenx:
@ShiftyBlock we need to add nemotron 3 nano cache to the ci server, don't worry about the failing tests for now

davidyang05:
Gotcha, makes sense, thanks for checking in!

tests/functional_tests/launch_scripts/active/L1_Launch_autoconfig_roundtrip.sh
cuichenx, line 1: this test is redundant?

davidyang05: Yes if running the whole suite this test is redundant! Originally made this in case someone wanted to test the autoconfig path directly.

PR #2869, opened 2026-03-18
feat(mimo): Phase 4 - MiMo training, model/provider, data loading, heterogeneous parallelism

aroshanghias-nvd:
## Summary

Adds MiMo (Multi-Input Multi-Output) training support to Megatron-Bridge, enabling heterogeneous multi-modal model training with independent per-module parallelism.

- **MimoModelProvider**: ModuleSpec-based model construction with heterogeneous LLaVA support (language model + vision encoder with independent configs)
- **Training loop**: `pretrain_mimo` / `train_mimo` / `mimo_step` entry points for MiMo-aware training with per-module forward/backward orchestration
- **Heterogeneous parallelism**: Each module (LLM, vision encoder, etc.) can run with its own TP/PP/DP configuration on a disjoint set of ranks (`mimo_parallel_utils`)
- **Data loading**: MiMo-aware collation, dataset, and data loader dispatch routing for multi-modal inputs
- **DDP wrapping**: Per-module distributed data parallel with rank-aware grid assignment
- Megatron-LM submodule pinned to PR #3212 head
- Full unit test coverage (122 tests)

Phase 5 (checkpointing, evaluation, e2e tests) is stacked in a follow-up PR.

## Validation

- 122 unit tests passed (models/mimo, training/mimo, data/mimo)

## Stack

- **PR1 (this)**: Phase 4 — training, model, data, parallelism
- **PR2**: Phase 5 — checkpoint save/resume, evaluation, e2e tests

## Summary by CodeRabbit

* **New Features**
  * Added MIMO (Multi-Instance Model Optimization) training framework enabling heterogeneous multi-module parallelism with dedicated pretraining entry point and training pipeline
  * Enhanced data loading infrastructure for multi-module models with loss masking support

* **Refactor**
  * Reorganized model infrastructure, process group management, and distributed utilities for improved multi-module training efficiency and flexibility


src/megatron/bridge/training/pretrain_mimo.py
```diff
@@ -0,0 +1,370 @@
...
+    # Initialize num-microbatches calculator if not already set.
+    from megatron.core import num_microbatches_calculator as nmc
+
+    if nmc._GLOBAL_NUM_MICROBATCHES_CALCULATOR is None:
+        nmc.init_num_microbatches_calculator(
+            dist.get_rank(),
+            getattr(cfg.train, "rampup_batch_size", None),
+            cfg.train.global_batch_size,
+            cfg.train.micro_batch_size,
+            cfg.data_parallel_size,
```
liding-nv: iiuc `data_parallel_size` for mimo should always be 1, right? Since all ranks load the same global micro-batch and per-module DP sharding is handled by `slice_batch_for_mimo` in forward_step, the training loop always sees dp=1.

If so, we should probably just hard code this value at the beginning of pretrain_mimo. This also avoids the user having to set it manually.

src/megatron/bridge/training/pretrain_mimo.py
```diff
@@ -0,0 +1,370 @@
...
+        multimodule_pg_collection=multimodule_pg_collection,
+        multimodule_communicator=multimodule_communicator,
+        module_to_grid_tuple=module_to_grid_tuple,
+        train_data_iterator=train_data_iterator,
+        valid_data_iterator=valid_data_iterator,
+        global_state=global_state,
+    )
+
+
+def pretrain_mimo(
```
liding-nv, line 41: standard `pretrain` function in MB calls `runtime_config_update` to update configuration, and also calls `_finish_train` and `_maybe_destroy_process_group` after the training loop. 
 wondering if there is any particular reason not calling them here.

src/megatron/bridge/training/train_mimo.py
```diff
@@ -0,0 +1,456 @@
...
+        raise RuntimeError(
+            "No local ProcessGroupCollection found for this rank. "
+            "Ensure rank participation is correctly configured in MIMO infrastructure."
+        )
+
+    # Configure gradient hooks on model config
+    model_config = get_model_config(model)
+
+    # Bind custom parameters via partial(), leaving schedule-provided args unbound
+    model_config.no_sync_func = partial(multimodule_no_sync, module_to_grid_tuple=module_to_grid_tuple)
```
liding-nv: standard path uses `_update_model_config_funcs` to set all model config hooks in one place (no_sync_func, finalize_model_grads_func, grad_scale_func, etc). `train_mimo` sets these inline. might be worth having a similar `_update_mimo_model_config_funcs` to consolidate them

src/megatron/bridge/training/pretrain_mimo.py
```diff
@@ -0,0 +1,370 @@
...
+    cfg: ConfigContainer,
+    mimo_provider: "MimoModelProvider",
+    forward_step_func: Callable,
+    build_data_iterators_fn: Callable,
+    opt_config: "OptimizerConfig",
```
liding-nv: why do we take a separate `opt_config` param instead of using `cfg.optimizer`. the e2e test ends up creating two config objects (one MCore, one MB) that can diverge. we can remove opt_config and using cfg.optimizer with finalize() instead.

src/megatron/bridge/training/pretrain_mimo.py
```diff
@@ -0,0 +1,370 @@
...
+            "MimoModelConfig.module_to_grid_map must be set at model construction time. "
+            "Ensure MimoModelProvider.provide() passes module_to_grid_map for MIMO parallelism."
+        )
+    logger.info(f"Rank {dist.get_rank()}: Creating MimoOptimizer")
+
+    # Create MimoOptimizer using the factory function
+    # Note: get_mimo_optimizer needs the unwrapped MimoModel to access mimo_config and submodules
+    from megatron.core.models.mimo.optimizer import get_mimo_optimizer
+
+    optimizer = get_mimo_optimizer(unwrapped_model, opt_config)
```
liding-nv: should optimizer and scheduler be created in `setup_mimo` (similar to the standard training path)

src/megatron/bridge/training/pretrain_mimo.py
```diff
@@ -0,0 +1,370 @@
...
+    multimodule_communicator = MultiModulePipelineCommunicator(
+        mimo_infra.module_to_grid_map,
+        mimo_infra.topology,
+        model_config,
+        dim_mapping={"s": 0, "b": 1, "h": 2},  # SBH mapping - matches MimoModel output
+        module_output_ndim=mimo_infra.module_output_ndim,
+    )
+
+    # Build pg_collection for schedule
+    multimodule_pg_collection = build_pg_collection_for_schedule(mimo_infra)
```
liding-nv: `multimodule_pg_collection` and `module_to_grid_tuple` built twice, in `setup_mimo` and `train_mimo`

src/megatron/bridge/training/pretrain_mimo.py
aroshanghias-nvd: yes, global DP is always 1 in mimo. It's better to hardcode it, to not run into confusion/errors.

src/megatron/bridge/training/pretrain_mimo.py
aroshanghias-nvd: - runtime_config_update: Agreed this should be added.
- _finish_train: This depends on CheckpointManager which MIMO doesn't have yet. It should be wired up in the Phase 5 PR.
- _maybe_destroy_process_group: Same — MIMO's process group lifecycle is different (per-module grids), so cleanup should probably be addressed in Phase 5 with proper teardown logic.

Can you look into phase 5 and see if it is best to add them over there?

src/megatron/bridge/training/train_mimo.py
aroshanghias-nvd: Makes sense to consolidate them.

src/megatron/bridge/training/pretrain_mimo.py
aroshanghias-nvd: You're right — the separate opt_config was initially added because early MIMO prototyping used experimental MCore OptimizerConfig, but there's no reason it can't use cfg.optimizer directly. We can remove the opt_config parameter and use cfg.optimizer with finalize() to keep it aligned with the standard path.

src/megatron/bridge/training/pretrain_mimo.py
aroshanghias-nvd: Agreed — moving optimizer and scheduler creation into setup_mimo (and returning them in MimoSetupOutput) would better mirror the standard path's setup().

src/megatron/bridge/training/pretrain_mimo.py
aroshanghias-nvd: Good catch — this is indeed redundant.

cuichenx (commented):
let's add a functional test that covers the mimo training code path. otherwise lgtm

cuichenx (approved):

PR #2921, opened 2026-03-20
fix: write distributed coverage to fixed path in minimax_m2 tests

ko3n1g:
## Summary

- The parallelism tests in `test_minimax_m2_conversion.py` were writing coverage data to `tmp_path / ".coverage"` (a throwaway pytest temp dir), so the `.coverage.*` files were never visible to `coverage combine` at `/opt/Megatron-Bridge/.coverage`
- This caused `coverage xml` to fail with "No data to report" even though all 4 tests passed
- Fix: use the hardcoded container path `--data-file=/opt/Megatron-Bridge/.coverage`, matching how other model tests (e.g. deepseek) are written
- Remove the now-unnecessary in-process import workaround from `L0_Launch_models_minimax_m2.sh`

## Test plan

- [ ] CI `L0_Launch_models_minimax_m2` passes (all 4 tests pass + coverage collection succeeds)


yaoyu-33:
resolved in original pr

PR #2926, opened 2026-03-20
[model] fix: restore position_embedding_type and gradient_accumulation_fusion for MLA models

yaoyu-33:
## Problem

Two regressions caused a ~30% perf drop on DSV3 GB300 MXFP8 (1292 → 914 TFLOPs, March 14 → March 18), both introduced by PRs #2599 and #2739:

1. `position_embedding_type` changed from `rope` to `learned_absolute`
2. `gradient_accumulation_fusion` changed from `True` to `False`

### Fix 1: `position_embedding_type` for MLA providers (`model_bridge.py`)

PR #2599 deleted `deepseek_provider.py` which had `position_embedding_type: str = "rope"` as a class field on `DeepSeekModelProvider`. After switching to raw `MLAModelProvider`, this default was lost.

The base `provider_bridge()` only set `"rope"` when `rope_type != "yarn"` — but DSV3's HF config has `rope_scaling.type = "yarn"`, so the guard tripped and `position_embedding_type` fell through to `GPTModelProvider`'s default `"learned_absolute"`.

**Fix:** For MLA providers, `position_embedding_type` must always be `"rope"` — YaRN scaling is handled separately via `mla_rope_params` (not via `position_embedding_type`). Added `is_mla_provider` check in `provider_bridge()` to correctly handle this.

This also covers other MLA models (Kimi, etc.) going forward.

### Fix 2: `gradient_accumulation_fusion` for DeepSeek V2/V3 (`deepseek_v2/v3_bridge.py`)

PR #2739 removed the hardcoded `provider.gradient_accumulation_fusion = True` from the DeepSeek bridges to use the dynamic `can_enable_gradient_accumulation_fusion()` check. That check returns `False` if the `fused_weight_gradient_mlp_cuda` CUDA extension (from APEX) is not importable, silently disabling the fusion with no error.

**Fix:** Restore explicit `gradient_accumulation_fusion = True` in `DeepSeekV2Bridge` and `DeepSeekV3Bridge`.

## Test Plan

- [ ] DSV3 GB300 MXFP8 perf run to verify TFLOPs recover to ~1292
- [ ] Confirm config diff shows `position_embedding_type=rope` and `gradient_accumulation_fusion=True`

## Summary by CodeRabbit

* **New Features**
  * Configuration translation tool for bidirectional conversion between Megatron-LM and Megatron Bridge formats.
  * Vanilla GPT pretraining recipe with baseline defaults.
  * CLI parameter support for string-based activation functions and data paths.

* **Improvements**
  * Enhanced sequence-parallel configuration normalization.
  * Gradient accumulation fusion optimization for DeepSeek models.
  * Updated documentation and command execution examples.

* **Bug Fixes**
  * Fixed activation function and dtype string resolution in configuration pipeline.

### DIFF
--- a/src/megatron/bridge/models/conversion/model_bridge.py
+++ b/src/megatron/bridge/models/conversion/model_bridge.py
@@ -448,6 +448,7 @@ class MegatronModelBridge(MegatronPeftBridge, Generic[HFPreTrained, ModelProvide
             ModelProviderTarget: A configured model provider instance
         """
         from megatron.bridge.models.gpt_provider import GPTModelProvider
+        from megatron.bridge.models.mla_provider import MLAModelProvider
 
         hf_config = hf_pretrained.config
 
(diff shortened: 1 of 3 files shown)


yaoyu-33:
Closing in favor of clean branch with only the intended commit.

PR #2991, opened 2026-03-25
Dump the env for rank0 for slurm based runs.

sudostock:
Dumps the 'env' on job start for the first slurm rank, with some mild redactions.

Many params or cluster specifics end up in env vars, being able to quickly see what was set after the fact is very useful for debugging. Compared to the current process of manually editing the sbatch or adding 'env' to the pre_cmds for a specific job.

## Summary by CodeRabbit

* **New Features**
  * Environment variables are now automatically logged at script startup with sensitive credentials (tokens, API keys, passwords) automatically redacted to ensure security.

* **Chores**
  * Enhanced debugging capabilities with automatic environment state snapshots. Logging errors are handled gracefully without interrupting execution.

### DIFF
--- a/scripts/performance/run_script.py
+++ b/scripts/performance/run_script.py
@@ -62,12 +62,13 @@ def _dump_env_rank0() -> None:
 
 def main():
     """Main function to run the pretraining/finetuning script."""
-    _dump_env_rank0()
-
     # Parse known args and treat any unknown args as Hydra-style config overrides.
     # `argparse.parse_known_args()` returns the unknown args as a `list[str]`.
     parser = parse_cli_args()
     args, cli_overrides = parser.parse_known_args()
+
+    if args.dump_env:
+        _dump_env_rank0()
 
     recipe = get_perf_optimized_recipe(
         model_family_name=args.model_family_name,
--- a/scripts/performance/argument_parser.py
+++ b/scripts/performance/argument_parser.py
@@ -730,6 +730,12 @@ def parse_cli_args():
         default=None,
     )
     logging_args.add_argument("--save_config_filepath", type=str, help="Path to save the task configuration file")
+    logging_args.add_argument(
+        "--dump_env",
+        action="store_true",
+        help="Write environment variables to /nemo_run/env_<SLURM_JOB_ID>.log on rank 0. "
+        "Useful for post-run debugging of NCCL, CUDA, and SLURM settings.",
+    )
 
     # Config variant selection
     config_variant_args = parser.add_argument_group("Config variant arguments")


scripts/performance/run_script.py
```diff
@@ -26,10 +28,42 @@
...
+def _dump_env_rank0() -> None:
+    """Capture the container environment to /nemo_run/env_<SLURM_JOB_ID>.log on rank 0.
+
+    The file lands alongside log*.out and configs/ inside the per-run nemo_run
+    directory for easy post-run debugging.
+    """
+    if os.environ.get("SLURM_JOB_ID") is None:
+        return
+    if int(os.environ.get("SLURM_PROCID", "-1")) != 0:
+        return
```
ko3n1g, line 46: why is this only useful in Slurm environments?

scripts/performance/run_script.py
```diff
@@ -26,10 +28,42 @@
...
+                    safe_v = v.replace("\r", "\\r").replace("\n", "\\n")
+                    f.write(f"{k}={safe_v}\n")
+        logger.info(f"Environment dump written to {env_path} (mode 600)")
+    except OSError as e:
+        logger.warning(f"Failed to write environment dump to {env_path}: {e}")
 
 
 def main():
     """Main function to run the pretraining/finetuning script."""
+    _dump_env_rank0()
```
ko3n1g: I think we should give users control over this. Some stakeholders might have concerns with dumping vars to disk.

scripts/performance/run_script.py
sudostock: It's not, but I'm not familiar enough with the non-slurm logic in mbridge to make assumptions.

The primary thing I was trying to avoid is outputting the env for _every_ process as that's mostly useless noise.

Open to suggestions on how best to accommodate the other workload environments.

scripts/performance/run_script.py
sudostock: Something like a '--dump-env' flag so users can opt in?

ko3n1g: yeah that would be great!

scripts/performance/run_script.py
ko3n1g: i see, that makes sense now. Thanks!

scripts/performance/run_script.py
sudostock: Done!

scripts/performance/run_script.py
```diff
@@ -26,10 +28,42 @@
...
+    directory for easy post-run debugging.
+    """
+    if os.environ.get("SLURM_JOB_ID") is None:
+        return
+    if int(os.environ.get("SLURM_PROCID", "-1")) != 0:
+        return
+    job_id = os.environ["SLURM_JOB_ID"]
+    env_path = f"/nemo_run/env_{job_id}.log"
+    try:
+        fd = os.open(env_path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
```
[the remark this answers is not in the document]
sudostock: These values are set by slurm. If slurm is outputting malformed values we've got worse problems.

PR #3010, opened 2026-03-27
ci: Remove per-unit-test timeout and do not uv sync during test

chtruong814:
# What does this PR do ?

ci: Remove per-unit-test timeout and do not uv sync during test

Tests seem to be flaky due to:
* Unit tests going beyond the 2s timeout. It's happening enough right now where this is causing inconsistent runs. Let's just rely on the overall test timeout.
  * https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/23609509257/job/68762568760
  * https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/23663793529/job/68949672272
* Timeouts during uv sync on test runs. These are random failures during a `uv run` where MBridge and MCore are re-syncing. This should not be necessary since it should be pre-installed in the docker build, so we avoid the code mount when running the tests.
  * https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/23644399947/job/68877723519
* Timeout due to one of the dataset tests. Claude code suggested capping the cpu workers during the test. The problem may be that the test sees all the cpus on the entire node but that is not what is dedicated to the pod, so it may hit this hang on resources.
  * https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/23569098639/job/68630616370 

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

## Summary by CodeRabbit

* **Chores**
  * Updated container environment configuration for test execution.
  * Increased test timeout duration to improve test reliability.
  * Modified dataset test initialization parameters.

### DIFF
--- a/.github/actions/test-template/action.yml
+++ b/.github/actions/test-template/action.yml
@@ -115,9 +115,6 @@ runs:
         #!/bin/bash
         docker container rm -f nemo_container_${{ github.run_id }} || true
 
-        # Only mount Megatron-Bridge/src directory to avoid overwriting the Megatron-LM submodule.
-        # The container will have built Megatron-LM including the datasets_cpp helper file already.
-        # Additionally, the container may be built with a different Megatron-LM commit as part of MCore's downstream testing.
         docker run \
           --rm \
           -d \


ko3n1g:
@chtruong814 instead of continuously increasing the threshold, can we perhaps ask the team if there’s room for speed up?

chtruong814:
@ko3n1g yeah for sure.  But we're hindered now with flaky CI.  Please cut an issue and raise it to the team.  I feel something else is the problem though with this being flaky, maybe with how many tests are being ran in the suite and some other interaction?  If it passes some times, then it's not an easy solve.

chtruong814:
I need to fix that one Data test. 'megatron.core.datasets.helpers_cpp' is missing for some reason.

.github/actions/test-template/action.yml
```diff
@@ -141,8 +130,9 @@ runs:
           --env HF_HOME=/home/TestData/HF_HOME \
           --env NEMO_HOME=/home/TestData/nemo_home \
           --env RUN_ID=${{ github.run_id }} \
+          --env UV_NO_SYNC=1 \
           --workdir /opt/Megatron-Bridge \
-          --volume ${{ github.workspace }}/Megatron-Bridge:/opt/Megatron-Bridge \
+          --volume ${{ github.workspace }}/Megatron-Bridge/src:/opt/Megatron-Bridge/src \
```
chtruong814: @ko3n1g Do we need to actually mount Megatron-Bridge at all if the container is always built specific to each workflow?

ko3n1g: No this isn’t required anymore. We had a “code-less” container for a while that required custom mount + uv install (for MCore’s dataset compiler). But now we could remove both mount and uv install

chtruong814: ok.  I'm removing it.  I'm also going to remove the timeout per unit test.  I hit a failure.  I think we need to see this entirely fail and maybe it'll help us see whether there's a problem somewhere else.  
https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/23663793529/job/68949672272?pr=3010

chtruong814:
This L2 test failure is on main
https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/23669870595/job/68961730313?pr=3010

Debugging that separately.  Unless there's other concerns, I'd recommend merging this PR.

PR #3031, opened 2026-03-30
Fix DSV3 gb200 NVFP4 NaN grad norm and improve mappings

dingqingy-nv:
## Summary
- Update DeepSeek V3 GB200 recompute strategy from `mla_up_proj` to `mlp` for higher throughput and resolve NaN grad
- Add `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` and `NCCL_GRAPH_REGISTER=0` env vars for DeepSeek V3 GB200 pretrain in `PerfEnvPlugin` for better memory usage

## Test plan
- [x] Verified DSV3 GB200 NVFP4 V2 (20 steps, no NaN grad norms) in 26.04 RC3 container 
- [ ] Verifying on internal CI.

## Summary by CodeRabbit

* **Chores**
  * Updated DeepSeek model recomputation settings for improved performance optimization
  * Extended GPU support configuration to include gb200 hardware with environment-specific optimizations

### DIFF
--- a/.github/actions/test-template/action.yml
+++ b/.github/actions/test-template/action.yml
@@ -69,21 +69,6 @@ runs:
         path: ${{ github.workspace }}/Megatron-Bridge
         submodules: recursive
 
-    - name: Update MCore submodule (if triggered from MCore)
-      if: ${{ github.event.inputs.mcore_commit != '' }}
-      shell: bash
-      working-directory: ${{ github.workspace }}/Megatron-Bridge
-      run: |
-        echo "🔄 Updating MCore submodule to commit: ${{ github.event.inputs.mcore_commit }}"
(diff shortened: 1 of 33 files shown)


scripts/performance/perf_plugins.py
```diff
@@ -271,9 +271,11 @@ def _set_model_specific_environment_variables(
             model_family_name in ["deepseek"]
             and model_recipe_name in ["deepseek_v3"]
             and train_task == "pretrain"
-            and gpu in ["h100"]
+            and gpu in ["h100", "gb200"]
         ):
             executor.env_vars["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+            if gpu in ["gb200"]:
+                executor.env_vars["NCCL_GRAPH_REGISTER"] = "0"
```
malay-nagda, line 278: @dingqingy-nv I think it would be good to add this var when CG is enabled (if condition to check)?

PR #3034, opened 2026-03-30
add container_env to SlurmExecutor for sbatch workloads

sudostock:
NeMo Run automatically sets --container-env for srun workloads but not sbatch. This means env var overrides that conflict with container defaults are silently dropped for sbatch jobs.

Changes:
  - Set container_env on SlurmExecutor using PERF_ENV_VARS keys, ensuring performance-critical vars always override container defaults
  - Ports perf_env change from #2847 to avoid bug depending on which merges first.
  - ~~Add container_env parameter to slurm_executor() for callers that need additional keys~~
  - ~~Add --container_env CLI flag to setup_experiment.py for key-only forwarding when the value is already present in the job environment~~

Not including a dedicated flag for '--container-env' for now as most cases should be covered by '-E and -ce' can revisit later.

Background: 
Slurm by default does `--export=ALL` which passes the entire env at submission time into the job env. Pyxis/Enroot does the same thing. **Except** when the container already has that variable set in which case it keeps the container version set. Unless the user tells it to override with `--container-env` flag.

Currently there is a gap if any of the values set in PERF_ENV_VARs ever already exists in the container they will silently be dropped. Additionally there's no easy way to override something like `NCCL_DEBUG=INFO` which is set 'WARN' in the container. 

I proposed fixing this directly in Nemo Run ([#394](https://github.com/NVIDIA-NeMo/Run/pull/394)) to fix the incongruity between the srun and sbatch ops. In offline discussions the recommendation was to fix it in NeMo/Mbridge instead.

## Release Notes

* **New Features**
  * Users can now specify environment variables via command-line that should override container image defaults when running performance experiments, enabling better customization of execution environments.

* **Tests**
  * Added comprehensive test coverage for environment variable propagation in container execution workflows, ensuring proper handling, deduplication, and consistent ordering of variables.

### DIFF
--- a/tests/unit_tests/scripts/performance/test_executors.py
+++ b/tests/unit_tests/scripts/performance/test_executors.py
@@ -19,6 +19,7 @@ from pathlib import Path
 
 import pytest
 
+
 # scripts/performance is not an installed package; add it to sys.path so we
 # can import ``utils.executors`` the same way the scripts themselves do.
 _PERF_SCRIPTS_DIR = Path(__file__).resolve().parents[4] / "scripts" / "performance"
@@ -40,8 +41,12 @@ if HAS_NEMO_RUN:
 def test_container_env_includes_perf_vars(tmp_path):
     """PERF_ENV_VARS keys must appear in container_env so they override container defaults."""
     executor = slurm_executor(
-        gpu="h100", account="test", partition="test",
-        log_dir=str(tmp_path), nodes=1, num_gpus_per_node=8,
+        gpu="h100",
+        account="test",
+        partition="test",
+        log_dir=str(tmp_path),
+        nodes=1,
+        num_gpus_per_node=8,
     )
     assert executor.container_env is not None, "container_env is None — was the field removed from the executor?"
     missing = set(PERF_ENV_VARS) - set(executor.container_env)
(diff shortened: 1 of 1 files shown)


sudostock:
I also added the perf_env dict copy change from #2847 to prevent a bug depending on order they get merged and to pre-empt the claude review flag every time PERF_ENV_VAR is mutated in a change.

ko3n1g:
@malay-nagda can you take a look?

scripts/performance/utils/executors.py
```diff
@@ -143,7 +146,8 @@ def slurm_executor(
         gres=gres,
         container_image=container_image,
         container_mounts=mounts,
-        env_vars=PERF_ENV_VARS,
+        env_vars=perf_env,
+        container_env=sorted(set(perf_env.keys()) | set(container_env)),
```
malay-nagda: Is it possible to just set `container_env=perf_env` here instead of adding an extra user arg?
Users can specify vars to override via `custom_env_vars` and they will still be picked up by `container_env` here?

sudostock: Yes, as the primary bug is lack of `container_env=perf_env`. The main downside to not having the standalone flag is for scripting.

Using -E or -ce plus the `container_env=perf_env` would handle one of the main pain points today with something like NCCL_DEBUG.

I'm fine with removing the flag, we can revisit if the lack of flag causes us enough friction.

sudostock: Okay I removed the container-env flag which simplified this a bit.

PR #2602, opened 2026-03-01
[model] feat: add MiniMax-M2 MoE bridge

yaoyu-33:
## Summary

Add Megatron Bridge for [MiniMaxAI/MiniMax-M2](https://huggingface.co/MiniMaxAI/MiniMax-M2), a sparse MoE model with 256 experts, top-8 sigmoid routing, and expert bias correction.

### Bridge (`minimax_m2_bridge.py`)
- Config mapping with manual `rotary_percent` calculation (`rotary_dim / head_dim`)
- Per-expert weight mapping using `block_sparse_moe` prefix (w1/w2/w3 format)
- Sigmoid routing with `e_score_correction_bias` buffer mapping
- QK layernorm intentionally disabled (full-dim vs per-head mismatch — documented TODO)
- MoE settings: grouped gemm, alltoall dispatcher, aux_loss load balancing

### Examples
- `conversion.sh` — single/multi-GPU round-trip and checkpoint import/export
- `inference.sh` — text generation with TP support
- `verify_toy_model.py` — creates a toy model and runs forward-pass comparison via `compare.py`

### Tests
- `test_minimax_m2_conversion.py` — toy model creation, single-GPU round-trip, TP=2 and PP=2 parallelism tests

### Bug fix in `compare.py`
- Truncate Megatron logits to HF vocab size before comparison, fixing shape mismatch when Megatron pads vocab for GPU kernel efficiency

## Verification
- 1-GPU forward pass: cosine similarity 0.999990 ✅
- EP=2 forward pass: cosine similarity 0.999990 ✅
- EP=2 weight round-trip: all 43 weights match ✅
- TP=2+EP=2 weight round-trip: all weights match ✅

## Known Limitations
- QK layernorm weights (`q_norm`, `k_norm`) are dropped — MiniMax-M2 uses full-dimension QK norm while Megatron uses per-head. Acceptable for fine-tuning.
- MTP (Multi-Token Prediction) modules are not mapped.

## Test plan
- [ ] Run `verify_toy_model.py` on 1-GPU, TP=2, EP=2
- [ ] Run `test_minimax_m2_conversion.py` with 2+ GPUs
- [ ] Verify weight round-trip with TP=2+EP=2

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added MiniMax-M2 model support with conversion and inference capabilities.
  * Added toy model verification script for MiniMax-M2 conversion testing.

* **Bug Fixes**
[...]

PR #3163, opened 2026-04-05
[model] feat: add MiMo-V2-Flash model support

beccohov:
# What does this PR do ?

Add Megatron Bridge support for [MiMo-V2-Flash (Xiaomi)](https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash), a 309B / ~15B active parameter LLM with hybrid attention, fine-grained MoE, value scaling, asymmetric V head dims, and Multi-Token Prediction.  

# Changelog

- Add MiMoV2FlashBridge with FP8 block-wise dequantization (supports non-uniform block sizes). Return fp32 weights to allow internal type cast.
- Add MiMoV2FlashModelProvider with dual-base RoPE, per-layer KV head switching, and asymmetric V head dim                                                                                                                                                                                          
- Add custom MiMoV2FlashSelfAttention that rebuilds linear_qkv/linear_proj for V head dim ≠ K head dim (128 vs 192)                                                                                                                                                                                 
- Add custom MiMoV2FlashTEDotProductAttention with per-layer sliding window, attention sink bias, and TE k/v channel support                                                                                                                                                                        
- Add MiMoV2FlashQKVMapping for asymmetric QKV merge/split during checkpoint conversion                                                                                                                                                                                                             
- Add MTP support: dense MLP spec (not MoE), SWA attention, with MTP layer count auto-detected from safetensor keys                                                                                                                                                                                 
- Add TP/CP assertions for known limitations (TP ≤ min KV groups, CP unsupported due to TE learnable softmax + CP). I decided not to replicate kv heads because it's not efficient.
- Add unit tests: provider bridge config mapping, MTP detection, config round-trip, mapping registry coverage, QKV round-trip, FP8 dequant, weight loading hooks                                                                                                                                    
- Validated: per-layer fp32 numerics match HF reference, end-to-end generation produces correct output, parallelism tested (TP, EP, SP, TP+EP combinations). Also tested generation on 8GPUs.

# GitHub Actions CI

The CI requires approval from an NVIDIA developer to run for external contributors. 

**Pre checks**:

# Additional Information

 - Related to https://github.com/NVIDIA/Megatron-LM/issues/2976


beccohov:
Drafted this WIP PR for MiMoV2 Flash support according to [this issue](https://github.com/NVIDIA/Megatron-LM/issues/2976)

beccohov:
@sbhavani Hey,
MiMo-V2-Flash uses asymmetric attention head dimensions: `Q/K` use `head_dim=192 `but `V` uses `v_head_dim=128`. This is currently not supported by MCore's standard SelfAttention for two reasons:
                                                                                                                                                                                                                                                                                                        
  1. `linear_proj` input size — `Attention.__init__` sizes `linear_proj` input as `kv_channels * num_attention_heads = 192 * 64 = 12288` ([here](https://github.com/NVIDIA/Megatron-LM/blob/17a67b9a97fb11a75933fd7f76ad76e1ac98a53d/megatron/core/transformer/attention.py#L270)), but the correct size is `v_head_dim * num_attention_heads = 128 * 64 = 8192`.                                                                                           
  2. QKV split in forward pass — `get_query_key_value_tensors` splits the fused QKV output using `hidden_size_per_attention_head (= kv_channels = 192)` for both `K` and `V`, so `V` would be extracted with the wrong size ([here](https://github.com/NVIDIA/Megatron-LM/blob/17a67b9a97fb11a75933fd7f76ad76e1ac98a53d/megatron/core/transformer/attention.py#L1458)).
                                                                                                                                                                                                                                                                                                        
Should I expose support similar to MLA but in standard SelfAttention for decoupled `V` size in megatron? Alternatively, is there a recommended pattern for bridging models with asymmetric `V` dims without modifying MCore?

yaoyu-33:
> Should I expose support similar to MLA but in standard SelfAttention for decoupled V size in megatron? Alternatively, is there a recommended pattern for bridging models with asymmetric V dims without modifying MCore?

@beccohov : usually we patch the config / implementation / fwd in bridge directly and use custom layer spec to specify the customized version of it.

Here's the GitHub link to the OLMoE provider file:
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/models/olmoe/olmoe_provider.py
Layer spec builder — L41-45
Custom OLMoESelfAttention class — L98-188

beccohov:
Hey, @yaoyu-33, I have a few questions before I mark this PR as "ready for review":
1. Currently my implementation of MiMo-V2-Flash asserts CP because TE lacks a backend for CP + learnable softmax (attention sink bias). SWA layers use softmax_type="learnable" which is refused by all TE backends when CP is enabled. However, seems like it is relatively easy to fix in TE ([here](https://github.com/NVIDIA/TransformerEngine/blob/2dd31bb849e83cce51c7d169db883862063d3a95/transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py#L1854) we'll go via MLA branch and we have v.shape wrong because of GQA). I tried to patch it in provider, but there are lot's of TE assertions that strictly control ability of hacks. Is asserting acceptable, or should we implement a workaround ?
  2. The HF config has attention_value_scale=0.707, but the released HF modeling code does not apply it in the forward pass (i.e. the released weights rescale V back). We currently skip it too (matching HF behavior). For training, however, I think we should re-enable the scale (i.e. to do value = value * 0.707) to match MiMo-v2 Flash fully. What do you think ? This is basically minor fix and I think it plays role only for large scale pretraining with fp8. For finetunes I think it's better to skip this scaling, because naive implementation would consume more memory on activations.

  3. Should `mimo_v2_flash/` be merged into the existing `mimo/` directory? The existing `mimo/ `contains both the original Xiaomi MiMo bridge (Qwen2-based) and the unrelated multimodal MIMO provider. I kept it separate to avoid confusion, but happy to merge if preferred.

sbhavani:
@beccohov I'd recommend creating an issue in TE to track CP support. I think the incomplete CP coverage for non-vanilla softmax should be fixed eventually.

beccohov:
[Created](https://github.com/NVIDIA/TransformerEngine/issues/2868) issue in TE.
Apart from this issue, what about `v_scale` and directory ?

beccohov:
The TE issue with CP is fixed in TE2.15.

beccohov:
@sbhavani hey, would it be possible to have review please?
Thanks!

sbhavani:
CC @snowmanwwg

beccohov:
@sbhavani the label is removed automatically anyways for some reason

sbhavani:
@beccohov thanks! we need to fix that automation bug

CC @yaoyu-33

src/megatron/bridge/models/mimo_v2_flash/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
cuichenx:
```suggestion
# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```

src/megatron/bridge/models/mimo_v2_flash/mimo_v2_flash_provider.py
```diff
@@ -0,0 +1,401 @@
...
+    # Per-layer KV heads (full attention vs SWA layers)
+    full_attn_num_query_groups: int = 4
+    swa_num_query_groups: int = 8
+
+    # Asymmetric V head dimension
+    v_head_dim: int = 128
+
+    # Attention sink bias
+    add_swa_attention_sink_bias: bool = True
+    add_full_attention_sink_bias: bool = False
```
cuichenx: is there an existing model where these values are different from this default? If not we should write these as default values in the model building blocks. Ideally we can remove the model specific provider if not necessary

cuichenx (commented):
Could you add a readme page and runs scripts following example here? https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/main/examples/models/minimax_m2

Can you also add a functional test following this example https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2602/changes

src/megatron/bridge/models/mimo_v2_flash/mimo_v2_flash_provider.py
```diff
@@ -0,0 +1,401 @@
...
+
+TEDotProductAttention, _ = safe_import_from("megatron.core.extensions.transformer_engine", "TEDotProductAttention")
+TELayerNormColumnParallelLinear, _ = safe_import_from(
+    "megatron.core.extensions.transformer_engine", "TELayerNormColumnParallelLinear"
+)
+TERowParallelLinear, _ = safe_import_from("megatron.core.extensions.transformer_engine", "TERowParallelLinear")
+SplitAlongDim, _ = safe_import_from("megatron.core.extensions.transformer_engine", "SplitAlongDim")
+
+
+class MiMoV2FlashRotaryEmbedding(RotaryEmbedding):
```
cuichenx: could you move these model blocks to a modeling_mimo_v2_flash.py file?

src/megatron/bridge/models/mimo_v2_flash/__init__.py
beccohov: Thanks, changed in all places

beccohov:
@cuichenx I believe I've addressed all your comments — would appreciate a re-review!

examples/models/mimo_v2_flash/slurm_inference.sh
```diff
@@ -0,0 +1,101 @@
...
+CMD="${CMD}uv run --no-sync python examples/conversion/hf_to_megatron_generate_text.py"
+CMD="$CMD --hf_model_path $HF_MODEL_ID"
+CMD="$CMD --prompt '$PROMPT'"
+CMD="$CMD --max_new_tokens $MAX_NEW_TOKENS"
+CMD="$CMD --tp $TP --ep $EP"
+CMD="$CMD --trust-remote-code"
+
+echo "Executing: $CMD"
+
+$SRUN_CMD bash -c "cd $WORKDIR && $CMD"
```
yaoyu-33, line 97: This should propagate the `srun` failure. As written, a non-zero generation exit falls through to the final `echo` statements, so the Slurm job can report success even when inference failed. Please add `set -euo pipefail` or capture/check `$?` here like `slurm_conversion.sh` does.

src/megatron/bridge/models/mimo_v2_flash/mimo_v2_flash_bridge.py
```diff
@@ -0,0 +1,368 @@
...
+        }
+
+
+def _dequant_fp8_blockwise(weight: torch.Tensor, scale_inv: torch.Tensor) -> torch.Tensor:
+    """Block-wise FP8 dequantization: out = fp8_val * scale_inv."""
+    M, N = weight.shape
+    sM, sN = scale_inv.shape
+    bM, bN = M // sM, N // sN
+    scale_full = scale_inv.repeat_interleave(bM, dim=0).repeat_interleave(bN, dim=1)
+    return weight.float() * scale_full
```
yaoyu-33: For the FP8 checkpoint this returns a float32 dequantized tensor, while the provider later runs bf16. That doubles temporary conversion memory versus bf16 and is especially risky for the full MiMo-V2-Flash weights. Please cast the dequantized result back to `torch.bfloat16` before returning.

yaoyu-33:
Inference coherence check on cluster for MiMo-V2-Flash passed with the PR-locked uv environment.

- PR SHA: `36fff757a5d9b7cf740c0eb16c9eeb68dac5df9b`
- Successful job: `11912655` (`pr3163-mimo-uv`), 2 nodes / 16 GPUs, `TP=1 PP=1 EP=16`, `MAX_NEW_TOKENS=64`
- Logs: `pr3163_mimo_v2_flash_20260519/logs/inference_uvsync_11912655.log`, env log `pr3163_mimo_v2_flash_20260519/logs/env_uvsync_11912655.log`

Prompts / outputs:

```text
Prompt: What is artificial intelligence?
Generated: What is artificial intelligence? Artificial intelligence (AI) is a branch of computer science that aims to create systems capable of performing tasks that normally require human intelligence. These tasks include learning, reasoning, problem-solving, perception, and language understanding. AI can be classified into two main types: narrow AI, which is designed for specific tasks (like voice assistants
```

```text
Prompt: Explain why the sky appears blue in two sentences.
Generated: Explain why the sky appears blue in two sentences.The sky appears blue because sunlight is scattered by air molecules in the atmosphere, and shorter blue wavelengths are scattered more efficiently than longer red wavelengths. This Rayleigh scattering directs blue light in all directions, making the sky appear blue to our eyes from the ground.<|im_end|>
```

Verdict: coherent for this short sanity run. No runtime errors, OOMs, NaNs, empty output, or repeated-special-token loops in the successful job. The first output is cut off by the 64-token cap but stays on prompt; the second output follows the requested two-sentence explanation.

yaoyu-33:
@beccohov  can you take a look at the final 2 comments above. Otherwise we are good to go. thanks for the effort and patient work again.

beccohov:
Thanks, @yaoyu-33 !
Will take a careful look soon!

src/megatron/bridge/models/mimo_v2_flash/mimo_v2_flash_bridge.py
beccohov: Done. I actually initially wanted to avoid fp32->bf16->fp32 if someone will be initializing master copy of weights for finetuning

examples/models/mimo_v2_flash/slurm_inference.sh
beccohov: Done!

beccohov:
@yaoyu-33 I think all your comments are addressed! Thanks!

yaoyu-33 (approved):

yaoyu-33 (dismissed):

kamran-nvidia:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/commit/fe2a8b3556564d8f23fc628d3c71e05855c0aac8

kamran-nvidia:
@beccohov Please address the CI failure: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3163/checks?check_run_id=77247324548

beccohov:
@kamran-nvidia Done!

kamran-nvidia:
/ok to test [18c3acc](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3163/commits/18c3accd8e0ca9544af39ade45b0ae1d018a53e3)

kamran-nvidia (dismissed):

cuichenx:
code cov failed because mimo functional test was run on H100 without a launcher under h100/active/. Could you add a launcher?

beccohov:
@cuichenx ah, apparently the launcher logic has changed since I started this PR and now there are separate h100/b200. I should have watched references, sorry! Now added h100 version.

cuichenx:
no worries, thanks!

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3163/commits/72c35724f83cf1f6c3260f3100c056deaba26c7f

cuichenx (dismissed):

beccohov:
@cuichenx @yaoyu-33 — three quick questions about the CI failure and the fixes I'm proposing. Could you comment each one so that I'll understand the strategy for CI fix?

1. Offline HF cache. 
The CI failure on `test_toy_model_creation` is `AutoConfig.from_pretrained("XiaomiMiMo/MiMo-V2-Flash", trust_remote_code=True)` hitting `HF_HUB_OFFLINE=1`. Other `trust_remote_code` models in the repo (`kimi_vl`, `nemotron_h`, etc.) pass CI because they're pre-cached in `/home/TestData/HF_HOME` via `cache-hf-model.yml`. The current workflow in `main` uses `AutoModel.from_pretrained(MODEL_ID)` without `trust_remote_code=True`, which seems to break for `trust_remote_code` repos. I saw @yaoyu-33's PR #3235 (closed without merge) that added a `snapshot_download`-based fix for exactly this case. But then it was closed probably due to security reasons. What's the recommended path here — should we reopen / re-land that PR first, or is there another mechanism I'm missing for caching `trust_remote_code` repos (or maybe add miso code in repo folder?)? Either way, I don't have `workflow_dispatch` permission, so I'd need one of you to actually trigger the cache for `XiaomiMiMo/MiMo-V2-Flash`.

2. `tests/functional_tests/utils.py` fix. 
While debugging locally I traced a regression in `autoconfig_roundtrip`. Originally (PR #2778, 2026-04-01) `from_auto_config` had `trust_remote_code=True` hardcoded internally, so `utils.py` calling `from_auto_config(...)` with no kwarg worked for all models including custom ones. 5 days later commit `0e6b05a1` parameterized it with `trust_remote_code: bool = False` as the default — likely a security-driven choice — but the existing caller in `utils.py` was never updated. The result: `autoconfig_roundtrip(..., trust_remote_code=True)` silently passes `False` to `from_auto_config`, which fails on custom models with `ValueError: ... contains custom code ...`. `nemotron_h`'s Nemotron-3-Nano hit this and was marked `@pleasefixme`. My fix is a one-line forward; backward-compatible since 16/17 callers pass the default `False`. OK to land this, or would you prefer a different approach?

3. `python` → `sys.executable` in the subprocess test. 
The parallelism roundtrip subprocess in my test currently invokes bare `"python"`. CI's `uv run` rigs `PATH` so it works, but it's fragile (broke for me locally where `python` resolved to a different venv without `coverage`). `test_minimax_m2_conversion.py` uses `sys.executable` already. Should I then align with that pattern?


Thanks!

kamran-nvidia:
@beccohov We are working on caching this model on CI server. I will re-run the failed CI test, when the fix is in place. No action needed from your side. Thanks.

beccohov:
@kamran-nvidia can you please share the PR / something where I can track the progress, so that I can understand when I'll be able to finish this PR?

kamran-nvidia:
@JRD971000 Can you help @beccohov please? Thanks.

beccohov:
Afaiu, the tests failed due to (2) from my comment [here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3163#issuecomment-4528199303). So what should I do? Should I add `trust_remote_code` ?
cc @yaoyu-33 @kamran-nvidia

yaoyu-33:
@beccohov  nvm, I will force merge this. It's ci caching issue.

yaoyu-33:
will follow up add back the functional tests after caching issue resolved.

Eisenhower:
Hi @beccohov @yaoyu-33 — small follow-up while reading this merged work.

`MiMoV2FlashTEDotProductAttention.__init__` reads `attention_value_scale` from config and stores it as `self._attention_value_scale` ([modeling_mimo_v2_flash.py#L256](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/models/mimo_v2_flash/modeling_mimo_v2_flash.py#L256)), but the value is never consumed in `forward` ([L270-L271](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/models/mimo_v2_flash/modeling_mimo_v2_flash.py#L270-L271)). For the released `XiaomiMiMo/MiMo-V2-Flash` checkpoint (`attention_value_scale = 0.707`) this silently scales the attention output by ~1.414× at every layer relative to the reference implementations.

The scale can't be folded into `softmax_scale` (softmax is non-linear) and `linear_proj.weight` is mapped 1:1 from HF `o_proj.weight`, so it has to land on V (or equivalently on the attention output). Both upstream references do this on V before the kernel:

- HF: https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash/blob/main/modeling_mimo_v2_flash.py — `value_states = value_states * self.v_scale`
- vLLM: https://github.com/vllm-project/vllm/blob/e68988a24807c9dfb2bf6936eb17425ce7812c5f/vllm/model_executor/models/mimo_v2.py#L327-L329 — `v = v * self.v_scale`

I opened a two-line follow-up: #4155. Would appreciate a quick look when you have a moment — happy to add a focused unit test if that helps the review.

beccohov:
Hi @Eisenhower,
thanks for your comment. Actually, the reason why I did not scale is that HF weights seem to be already static scaled in advance (V weights). When I was implementing this I noticed that by quality degradation.
So you don't need to scale this to reproduce results if you load from retrained model because it was already trained with suitable magnitudes.
However, for training dynamics you may want to scale (this will affect gradients magnitude for V and also if you retrain from scratch this may change magnitude of attention scores). But naive scaling will increase memory consumption (and this would be notable on this large model) so ideally you need to implement this scaling to avoid memory overhead.

So, one way is no upscale back HF weights while loading them and then use this scaling in training. This is especially valuable when you train from scratch.
Another way is to skip this scaling completely (as in my implementation) and keep in mind that gradients for V would be slightly different. Since we use init from pertained model, I believe this won't significantly change model convergence (since we use smaller lr and model is more stable for post-training). But will save you memory (if you don't use full AC).

Eisenhower:
Thanks for the context. I think the HF implementation you checked was likely the old version. I raised this issue with the Xiaomi MiMo team afterwards, and they have since updated the HF modeling code to apply `attention_value_scale` on `value_states` in forward. vLLM also applies the same scale before attention.

I agree the extra allocation from a naive `value = value * scale` is worth optimizing. My concern is mainly RL consistency: rollout usually follows the vLLM/HF inference path, while training/logprob/KL are computed by the actor path. If rollout applies the scale but training does not, the two forward paths are no longer equivalent. For that reason I would prefer Megatron-Bridge to match the HF/vLLM semantics first, and then optimize the implementation or fold the scale only if we can keep the two paths numerically equivalent.

Eisenhower:
@beccohov Thanks for the context. I think the HF implementation you checked was likely the old version. I raised this issue with the Xiaomi MiMo team afterwards, and they have since updated the HF modeling code to apply `attention_value_scale` on `value_states` in forward. vLLM also applies the same scale before attention.

I agree the extra allocation from a naive `value = value * scale` is worth optimizing. My concern is mainly RL consistency: rollout usually follows the vLLM/HF inference path, while training/logprob/KL are computed by the actor path. If rollout applies the scale but training does not, the two forward paths are no longer equivalent. For that reason I would prefer Megatron-Bridge to match the HF/vLLM semantics first, and then optimize the implementation or fold the scale only if we can keep the two paths numerically equivalent.

beccohov:
@Eisenhower Sure, if HF weights will be updated then the situation changes. You can then add a follow-up PR with this small change!

PR #3162, opened 2026-04-05
[model] fix: fix gpt-oss down_proj weight handling

yuki-97:
## Problem

NeMo-RL's `grpo-gptoss-20b-8n8g-megatron` regressed with `train/gen_kl_error ≈ 13` (expected < 0.002) and `train/reward ≈ 0` from step 0.

## Root cause

`down_proj` was being double-transposed, causing both Megatron and vLLM to receive incorrect weights:

  import:  W_HF [in, out]  →  W_HF.T [out, in]   (stored in Megatron) ✗
  export:  W_HF.T          →  W_HF   [in, out]    (sent to vLLM)       ✗

Megatron's forward pass requires `down_proj` in `[in, out]` layout, and vLLM expects `[out, in]`. The import transpose caused Megatron to compute with transposed weights (high gen_kl_error), while the export transpose on top made vLLM also receive the wrong layout (zero reward).

## Fix

Remove the import transpose from `maybe_modify_loaded_hf_weight` and restore the export transpose in `GPTOSSMLPDownProjMapping.megatron_to_hf`:

  import:  W_HF [in, out]  →  W_HF   [in, out]   (stored in Megatron) ✓
  export:  W_HF             →  W_HF.T [out, in]   (sent to vLLM)       ✓

## Testing

Fixes NeMo-RL's `grpo-gptoss-20b-8n8g-megatron` nightly test.

## Release Notes

* **Refactor**
  * Improved efficiency of model weight loading during import by reorganizing transposition logic, enhancing system reliability without affecting user-facing functionality.

PR #3271, opened 2026-04-10
[model] fix: fix gpt oss export

yuki-97:
## Summary
Move the transpose of `down_proj` for GPT-OSS expert weights to NeMo-RL side instead of bridge side to avoid mismatch layout when saving checkpoint.

**History:**
- #3162 fixed a double-transpose bug: both import and export were transposing `down_proj`. The import-side transpose was removed, leaving only the export transpose in `GPTOSSMLPDownProjMapping.megatron_to_hf`. This helps fixing Megatron `[in, out]` layout and vLLM `[out, in]` layout.
- However, the intermediate `[out, in]` layout produced by the bridge transpose corrupted saved checkpoints (`save_hf_pretrained` wrote weights that didn't match the original HF `[in, out]` layout).
- #3250 papered over this with a post-export generator wrapper that un-transposed before saving — correct but "transpose then un-transpose".

**Root cause:** The transpose is a NeMo-RL refit concern, not a bridge concern. The bridge's `megatron_to_hf` should always produce standard HF layout.

## Changes
- Reverts #3250 (`transpose_on_export` adaptive wrapper in `GPTOSSMLPDownProjMapping`).
- Removes the `down_proj` transpose from `GPTOSSMLPDownProjMapping.megatron_to_hf`. The export path now returns weights in the correct HF `[in, out]` layout for checkpoint saving.

## Testing
Validated. See validate steps and results in https://github.com/NVIDIA-NeMo/RL/pull/2249.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined weight export logic by removing model-specific conditional handling across conversion utilities and bridge modules.

* **Chores**
  * Updated documentation to reflect simplified weight handling approach.

### DIFF
--- a/examples/conversion/hf_megatron_roundtrip_multi_gpu.py
+++ b/examples/conversion/hf_megatron_roundtrip_multi_gpu.py
@@ -53,7 +53,6 @@ from rich.table import Table
 from megatron.bridge import AutoBridge
 from megatron.bridge.models.decorators import torchrun_main
 from megatron.bridge.models.hf_pretrained.utils import is_safe_repo
-from megatron.bridge.utils.common_utils import fix_gpt_oss_export_transpose, get_hf_model_type
 
 
 HF_MODEL_ID = "meta-llama/Llama-3.2-1B"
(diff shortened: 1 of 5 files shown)


cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3271/commits/c4d5effa3c178543179982f7087e47a41b06677b

yaoyu-33 (approved):

PR #2803, opened 2026-03-14
[recipe][perf] feat: unified flat performance recipes for all model families

yaoyu-33:
## Summary

Consolidate performance benchmark recipes from the old two-level pipeline (`scripts/performance/configs/`) into self-contained flat recipes under `src/megatron/bridge/recipes/<family>/<model>_perf.py`.

Each perf recipe function is fully self-contained: it calls the base pretrain/SFT recipe, overrides parallelism and precision inline, then calls `_benchmark_common()`. This eliminates the old indirection through `WorkloadBaseConfig` + `set_workload_base_configs` + `get_perf_optimized_recipe`.

### Models covered
- **Llama 3 / 3.1** — pretrain, SFT, LoRA (8B/70B/405B × H100/B200/B300/GB200/GB300)
- **DeepSeek V3** — pretrain V1+V2 (H100/B200/B300/GB200/GB300)
- **Qwen3 MoE** — pretrain (30B-A3B / 235B-A22B)
- **Qwen3-VL** — pretrain (30B-A3B / 235B-A22B)
- **Kimi K2** — pretrain (H100/B200/GB200/GB300)
- **NemotronH** — pretrain (Nemotron 3 Nano + NemotronH 56B)
- **GPT-OSS** — pretrain (120B)

### Key changes
- Add `_benchmark_common()` helper in `common.py` for shared perf overrides (train_iters, timing, `te_rng_tracker` derived from `cuda_graph_impl`)
- Add `_perf_precision()` helper for bf16/fp8_cs/fp8_mx/nvfp4 mixed precision configs
- Add Qwen3-VL pretrain base recipes (`qwen3_vl_30b_a3b_pretrain_config`, `qwen3_vl_235b_a22b_pretrain_config`) delegating to text-only configs
- Remove 10 DeepSeek V3 recipes that have no legacy equivalent (fp8mx/nvfp4 for H100, nvfp4 for B200/B300/GB200 — unverifiable)
- Update `run_script.py` to support new flat recipe discovery path
- Add comprehensive equivalence test covering all 245 old↔new recipe pairs

### Verification
All 245 recipe pairs pass config parity equivalence test on the cluster:
```
245 passed, 0 failed in 157.61s
```

## Test plan
- [x] `tests/unit_tests/recipes/test_all_perf_equivalence.py` — 245/245 pass (old vs new config parity)
- [x] `tests/unit_tests/recipes/llama/test_llama3_perf.py` — Llama3 perf recipe unit tests
- [ ] CI unit tests
[...]

PR #3284, opened 2026-04-12
feat(perf): switch llama3 70B GB200/GB300 MXFP8 from 3D parallel to FSDP

dingqingy-nv:
## Summary
- Switch llama3 70B GB200 MXFP8 V2 config from TP=2/PP=4/VPP=5 to FSDP with MBS=2 and 45 layers CPU offloading
- Switch llama3 70B GB300 MXFP8 V2 config from TP=1/PP=4 to FSDP with MBS=2 and 20 layers CPU offloading


yaoyu-33:
Closing because this modifies the retired legacy performance-config layout. The flat performance recipes landed in #2803 and the legacy configs were removed by #4623. Any FSDP retuning should be re-benchmarked against the current recipe and dependencies.

PR #3290, opened 2026-04-12
[ckpt] fix: Prevent int32 overflow for high train sample counts.

BlueCrescent:
# What does this PR do ?

Previously, when creating the TrainState for checkpointing int32 was used as type for encoding the number of consumed samples and related values. For longer trainings, torch will detect this as potential overflow and crash with `RuntimeError: value cannot be converted to type int32 without overflow`. We saw this crash while checkpointing step 700,000 while training with sequence length 8192 and batch size of 3072, so at 2,150,400,000 (> 2^31 - 1 = 2,147,483,648) samples.

# Changelog

- Switched from int32 to int64 for all values using this type in TrainState.

## Bug Fixes
* Improved integer precision in training state serialization. Counter values are now stored as 64-bit integers instead of 32-bit, preventing potential overflow issues during extended training runs. This ensures reliable checkpoint persistence and stability when handling large training step and sample count values.

### DIFF
--- a/src/megatron/bridge/training/state.py
+++ b/src/megatron/bridge/training/state.py
@@ -70,10 +70,10 @@ class TrainState(Stateful):
             their corresponding tensor representations.
         """
         return {
-            "step": torch.tensor(self.step, dtype=torch.int32),
-            "consumed_train_samples": torch.tensor(self.consumed_train_samples, dtype=torch.int32),
-            "skipped_train_samples": torch.tensor(self.skipped_train_samples, dtype=torch.int32),
-            "consumed_valid_samples": torch.tensor(self.consumed_valid_samples, dtype=torch.int32),
+            "step": torch.tensor(self.step, dtype=torch.int64),
+            "consumed_train_samples": torch.tensor(self.consumed_train_samples, dtype=torch.int64),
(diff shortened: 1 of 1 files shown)


BlueCrescent:
Hmm, with this change I was able to write out the iteration 700,000 checkpoint. But when warmstarting from it I saw cuda OOM twice and `iteration 700004: Unexpected result nan (message='found NaN in local forward loss calculation')` once.
Need to investigate further.

BlueCrescent:
Seems like the OOM or NaN issues were some other instability in the checkpoint unrelated to this issue/fix.

yaoyu-33:
fixed in https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3312, thanks for pr.

pruprakash:
## QA RCCA Analysis

### 1. Fix Reference
- **PR**: #3290 - [ckpt] fix: Prevent int32 overflow for high train sample counts (NOT MERGED)

### 2. Root Cause
Same issue as #3312 - int32 overflow in TrainState counters for long training runs.

### 3. Status
- PR #3290 was **NOT MERGED** (state: CLOSED)
- This appears to be a duplicate of #3312 which was merged with the same fix

### 4. Conclusion
**Verdict**: DUPLICATE - Same fix merged in #3312. This PR was closed without merging.

PR #3311, opened 2026-04-13
perf(nsys): reduce CPU-side overhead in profiling defaults

dingqingy-nv:
## Summary
- Override nemo_run's default `nsys_extra_args` to disable CPU context-switch tracing (`--cpuctxsw=none`), backtrace collection (`--backtrace=none`), and remove CUDA graph node tracing (`--cuda-graph-trace=node`)
- These flags eliminate unnecessary CPU-side activity during GPU profiling, reducing profiling overhead
- Applied to both `NsysPlugin` implementations (recipes + perf scripts) and the interactive launch script

## Changes vs nemo_run defaults

| Flag | nemo_run default | After this PR |
|------|-----------------|---------------|
| `--cpuctxsw` | `process-tree` (nsys default) | `none` |
| `--backtrace` | `auto` (nsys default) | `none` |
| `--cuda-graph-trace` | `node` | removed |

Since `-s none` (disable CPU sampling) is already set by nemo_run, `--backtrace=none` is defensive — there are no CPU samples to attach backtraces to. `--cpuctxsw=none` is the main win, stopping OS thread scheduling trace collection that was active by default.

## Test plan
- [ ] Verify nsys profiling still produces valid `.nsys-rep` files
- [ ] Confirm NVTX ranges and CUDA kernels are still captured
- [ ] Check that `nsys_extra_args` user override still works (user args take precedence)

### DIFF
--- a/scripts/performance/perf_plugins.py
+++ b/scripts/performance/perf_plugins.py
@@ -121,12 +121,19 @@ class NsysPlugin(Plugin):
         if self.nsys_trace is not None:
             launcher.nsys_trace = self.nsys_trace
 
-        # Combine default extra args with user-provided extra args
+        # Override nemo_run defaults to reduce CPU-side collection overhead
+        launcher.nsys_extra_args = [
+            "--force-overwrite=true",
+            "--capture-range=cudaProfilerApi",
+            "--capture-range-end=stop",
+            "--cuda-event-trace=false",
+            "--cpuctxsw=none",
+            "--backtrace=none",
+        ]
+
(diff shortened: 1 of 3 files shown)


dingqingy-nv:
Let's treat this as an test PR. If getting less profiling overhead, we can raise a PR directly to nemo run and close this

yaoyu-33:
Closing in favor of #4669, which implements the nsys behavior on the current performance harness and allows user arguments to override launcher defaults generally instead of hard-coding a specific default set.

PR #3350, opened 2026-04-16
[perf] fix: use direct assignment for NCCL env vars when nccl_ub enabled

dingqingy-nv:
## Summary
- Fixes review feedback from #3305 / #3283
- `os.environ.setdefault()` does not override pre-existing values, so clusters that default `NCCL_NVLS_ENABLE=0` in their container environment would silently break `nccl_ub` when enabled via recipe config
- Switches to direct assignment (`os.environ["KEY"] = "1"`), consistent with `setup_experiment.py` which already uses `dict.update()` for the same env vars

## Test plan
- [ ] CI passes
- [ ] Verify NCCL env vars are correctly set when `recipe.ddp.nccl_ub=True`, even on clusters with `NCCL_NVLS_ENABLE=0` default

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated distributed training configuration to enforce NCCL environment settings when UB optimization is enabled, now overriding existing cluster settings for consistent performance behavior.


malay-nagda (approved):

pruprakash:
## QA RCCA Analysis

### 1. Fix Reference
- **PR**: #3350 - [perf] fix: use direct assignment for NCCL env vars when nccl_ub enabled

### 2. Root Cause
`os.environ.setdefault()` does not override pre-existing values, so clusters that default `NCCL_NVLS_ENABLE=0` in their container environment would silently break `nccl_ub` when enabled via recipe config. The env vars were not being set correctly.

### 3. Trigger Configuration
- Using `recipe.ddp.nccl_ub=True`
- Running on clusters with `NCCL_NVLS_ENABLE=0` default in container

### 4. Nature of the Bug
**Classification**: CODE BUG - Silent performance degradation due to incorrect env var setting

### 5. Existing Test Coverage
**In Fix PR**: NO - No test files

**In NMFW Tests**: Not specifically covering NCCL env var setting

### 6. Coverage Assessment
| Fix PR tests | NO | N/A |

### 7. New Regression Test
**NOT ADDED** - This is an environment variable handling fix. Testing would require specific cluster configurations.

### 8. Conclusion
**Verdict**: PARTIAL COVERAGE - No dedicated test, but fix is straightforward env var assignment change.

PR #3364, opened 2026-04-16
[training] feat: enable fp4_param_gather in MixedPrecisionConfig

dingqingy-nv:
## Summary
- Add \`fp4_param\` and \`fp4_param_gather\` fields to \`MixedPrecisionConfig\`, mirroring the existing \`fp8_param\` / \`fp8_param_gather\` pair (sync in \`__setattr__\`, default in \`finalize\`).
- Enable \`fp4_param_gather=True\` by default in \`bf16_with_nvfp4_mixed()\` so NVFP4 runs pick up the new packed-uint8 parameter all-gather path introduced by MCore [PR #4005](https://github.com/NVIDIA/Megatron-LM/pull/4005).

## Context
MCore's \`fp4_param_gather\` packs NVFP4 model weights as uint8 (2 values per byte) in the DDP param buffer and all-gathers them in that packed form; the grad buffer stays full-width BF16. Both \`DistributedDataParallelConfig.fp4_param_gather\` and \`TransformerConfig.fp4_param\` must be set for the feature to apply end-to-end, which is why we add both Bridge fields and keep them synced (same pattern MCore itself uses at \`megatron/training/arguments.py:1732\`).

## Validation
Measured on DeepSeek-V3 256×GB300 pretrain (NVFP4 V2 config, GBS=4096, PP=2/VPP=8/EP=32, \`recompute_modules=[\"mla_up_proj\"]\`):

| Metric (Rank 0 / PP stage 0) | Baseline (no fp4pg) | This PR (fp4pg on) | Δ |
|---|---:|---:|---:|
| \`mem-max-allocated\` (iter 2) | 235.15 GB | 202.75 GB | **−32.40 GB** |
| \`mem-max-reserved\` (iter 2) | 249.25 GB | 214.40 GB | **−34.85 GB** |

Loss trajectory over iters 3–6 tracks the baseline within ≤0.2 nat (within noise). No per-iter throughput regression observed on this workload; the win is memory headroom, which unlocks dropping recompute in subsequent tuning.

## Test plan
- [x] Existing \`bf16_with_nvfp4_mixed\` recipe consumers pick up \`fp4_param_gather=True\` automatically and propagate via \`update_config_with_precision_overrides\` to both \`DistributedDataParallelConfig.fp4_param_gather\` and \`TransformerConfig.fp4_param\` (verified in training log).
- [x] DeepSeek-V3 256×GB300 pretrain run completes iter 1+ without crash, loss trajectory matches prior NVFP4 run, memory drops as shown above.
- [ ] Add a unit test covering the new sync/finalize logic for the \`fp4_param\` / \`fp4_param_gather\` pair.

## Summary by CodeRabbit

* **New Features**
  * Added FP4 mixed precision configuration options to control parameter gathering behavior during training.

* **Updates**
  * Modified the default FP4 recipe configuration to enable parameter gathering for optimized training behavior.

### DIFF
--- a/scripts/performance/configs/nemotronh/nemotron_3_llm_pretrain.py
+++ b/scripts/performance/configs/nemotronh/nemotron_3_llm_pretrain.py
@@ -23,6 +23,7 @@ from utils.utils import get_workload_base_config
 from megatron.bridge.recipes.nemotronh.nemotron_3_nano import nemotron_3_nano_pretrain_config
 from megatron.bridge.recipes.nemotronh.nemotron_3_super import nemotron_3_super_pretrain_config
 from megatron.bridge.training.config import ConfigContainer
+from megatron.bridge.training.mixed_precision import nemotron_3_super_bf16_with_nvfp4_mixed
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


sanandaraj5597:
LGTM.

sanandaraj5597 (approved):

PR #3370, opened 2026-04-17
[ci] feat: use AWS ephemeral runners for external contributors

ko3n1g:
## Summary

Routes external contributors to isolated, ephemeral runners while NVIDIA maintainers keep the persistent ones, leveraging the SSO membership check already built into the pre-flight workflow.

**Changes:**
- Passes `nemo-ci-aws-gpu-x2` / `nemo-ci-aws-gpu-x2-ephemeral` as `default_runner_prefix` / `non_nvidia_runner_prefix` inputs to the pre-flight workflow
- Uses `needs.pre-flight.outputs.runner_prefix` in all GPU jobs (`cicd-container-build`, unit tests, L0/L1/L2, flaky) — the pre-flight already handles SSO membership and emits the correct runner based on contributor status
- Removes the redundant `is-not-external-contributor` job and the copied `check-nvidia-sso-membership` composite action

## Example

\`\`\`yaml
pre-flight:
  uses: NVIDIA-NeMo/FW-CI-templates/.github/workflows/_cicd_preflight.yml@v0.80.1
  with:
    default_runner_prefix: nemo-ci-aws-gpu-x2
    non_nvidia_runner_prefix: nemo-ci-aws-gpu-x2-ephemeral
\`\`\`

## Test plan
- [ ] Trigger CI from an NVIDIA-member PR — verify `nemo-ci-aws-gpu-x2` runner is selected
- [ ] Trigger CI from an external contributor PR — verify `nemo-ci-aws-gpu-x2-ephemeral` runner is selected


.github/actions/check-nvidia-sso-membership/action.yml
```diff
@@ -0,0 +1,139 @@
+name: 'Check NVIDIA SSO Membership'
```
chtruong814: We shouldn't need to copy this.  We already check this as part of the pre-flight.

https://github.com/NVIDIA-NeMo/FW-CI-templates/blob/main/.github/workflows/_cicd_preflight.yml#L211

.github/workflows/cicd-main.yml
chtruong814, line 75: The pre-flight handles identifying whether it's a internal vs external contributor.  And it also provides the output for non_nvidia runner prefix. So hopefully we just need to update the runner prefix from Azure to the ephemeral prefix.  And we also need to taint the nodes at end of workflow unless @theothermike has handled that automatically.

chtruong814 (commented):
Thanks for working on this. But sorry I was not clear but most of the external vs internal distinction is already handled.  But we just haven't tested if the 2 x gpu ephemeral runners work.

ko3n1g:
not required

PR #3407, opened 2026-04-18
ci: migrate CI to AWS ephemeral runners for external contributors

ko3n1g:
the nemo-ci ephemeral runner's name do not allow for easy size-swap. So for now we keep x2 hardcoded and make it dynamic with follow-up PRs

---

## Summary

Moves Megatron-Bridge CI fully onto AWS infrastructure and routes external contributors to ephemeral runners:

- **Internal (NVIDIA SSO):** `nemo-ci-aws-gpu-x2`
- **External contributors:** `nemo-ci-aws-gpu-x2-ephemeral`
- **Container registry:** `766267172432.dkr.ecr.us-east-1.amazonaws.com` (both paths)
- **Test data path:** `/mnt/datadrive/TestData/nemo-fw/TestData` (both paths)

All Azure-specific code (Azure CLI install, `azure/login`, ACR login, `AZURE_*` secrets, `nemo-ci` environment gate) is removed since it is now unreachable.

The `pre-flight` job (`_cicd_preflight.yml`) continues to check SSO membership and emit `registry`, `runner_prefix`, and `test_data_path` outputs — downstream jobs are unchanged.

`cleanup-taint-node` runs after all tests and taints the ephemeral node for recycling. The ephemeral check is done at the step level via `runner.name` (instead of `needs.pre-flight.outputs.runner_prefix` in the job `if:`), because GH Actions reusable workflow free-form string outputs are not reliably accessible in job-level `if:` conditions.

## Example

An external contributor opening a PR will have their jobs routed to:
```
runs-on: nemo-ci-aws-gpu-x2-ephemeral
container-image: 766267172432.dkr.ecr.us-east-1.amazonaws.com/megatron-bridge:<sha>
test-data-path: /mnt/datadrive/TestData/nemo-fw/TestData
```

Fixed `cleanup-taint-node` being incorrectly skipped: `contains(needs.pre-flight.outputs.runner_prefix, 'ephemeral')` in the job-level `if:` evaluates to `false` because GH Actions does not reliably expose free-form string outputs from reusable workflows in `if:` conditions. The check is now done at the step level using `runner.name`, which is always available at execution time.

</details>


.github/workflows/cicd-main.yml
```diff
@@ -66,12 +66,12 @@ jobs:
   pre-flight:
     uses: NVIDIA-NeMo/FW-CI-templates/.github/workflows/_cicd_preflight.yml@v0.80.1
     with:
-      default_runner_prefix: ${{ vars.DEFAULT_RUNNER_PREFIX }}
-      non_nvidia_runner_prefix: ${{ vars.NON_NVIDIA_RUNNER_PREFIX }}
-      default_test_data_path: ${{ vars.DEFAULT_TEST_DATA_PATH }}
-      non_nvidia_test_data_path: ${{ vars.NON_NVIDIA_TEST_DATA_PATH }}
-      default_registry: ${{ vars.DEFAULT_CONTAINER_REGISTRY }}
-      non_nvidia_registry: ${{ vars.NON_NVIDIA_CONTAINER_REGISTRY }}
+      default_runner_prefix: nemo-ci-aws-gpu-x2-ephemeral
```
chtruong814: Thanks for fixing this.  Let's change the default runner back to non-ephemeral.

.github/workflows/cicd-main.yml
```diff
@@ -590,14 +551,10 @@ jobs:
           script: ${{ matrix.script }}
           timeout: ${{ fromJSON(matrix.timeout || '30') }}
           is_unit_test: "false"
-          has-azure-credentials: ${{ contains(needs.pre-flight.outputs.registry, 'azure') }}
-          azure-client-id: ${{ secrets.AZURE_CLIENT_ID }}
-          azure-tenant-id: ${{ secrets.AZURE_TENANT_ID }}
-          azure-subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
```
chtruong814, line 596: Great to move on from the legacy runners

chtruong814 (commented):
Thank you!  Let's revert the default runner back to non-ephemeral.

chtruong814 (approved):

PR #3408, opened 2026-04-18
[PEFT] feat: Support not sharing expert adapters

HollowMan6:
# What does this PR do ?

Support not sharing expert adapters while keeping MoE expert linears share one adapter across all local experts on the EP rank by default.

# Changelog

- Make it configurable so that we can create one adapter per local expert
- Also make it compatible with normalize moe lora

# GitHub Actions CI

**Pre checks**:

- [X] Did you write any new necessary tests?
- [X] Did you add or update any necessary documentation?

# Additional Information

- Related to # (issue)

## Release Notes

* **New Features**
  - Added support for per-local-expert adapters in mixture-of-experts models with configurable sharing behavior.
  - Enhanced adapter export functionality with improved state conversion and automatic rank inference.

* **Bug Fixes**
  - Improved distributed synchronization handling for adapter export operations.
  - Fixed expert dimension alignment under normalization settings.

* **Tests**
  - Extended test coverage for grouped expert adapter scenarios and expert tensor-parallel operations.


src/megatron/bridge/peft/utils.py
```diff
@@ -732,3 +742,404 @@ def sharded_state_dict(
...
+        """Gather a tensor across expert TP ranks by concatenating its last dimension."""
+
+        expert_tp_size = parallel_state.get_expert_tensor_parallel_world_size() or self.config.expert_tensor_parallel_size
+        if expert_tp_size == 1:
+            return tensor
+        gathered = [torch.empty_like(tensor) for _ in range(expert_tp_size)]
+        torch.distributed.all_gather(
+            gathered,
+            tensor,
+            group=parallel_state.get_expert_tensor_parallel_group(),
```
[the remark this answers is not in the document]
HollowMan6: fixed

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -579,23 +580,32 @@ def save_hf_adapter(
...
+        adapter_state, module_adapter_keys, target_parameters = convert_adapter_weights_to_peft_state(
+            raw_adapter_weights,
+        )
+        rank_pattern = infer_rank_pattern_from_adapter_weights(
+            raw_adapter_weights,
+            default_rank=getattr(peft_config, "dim", 32),
+        )
 
-        is_rank0 = not (dist.is_available() and dist.is_initialized()) or dist.get_rank() == 0
+        is_rank0 = not dist.is_initialized() or dist.get_rank() == 0
```
cuichenx, line 608: why is this removed? is this for cpu merge?

src/megatron/bridge/peft/lora.py
```diff
@@ -112,6 +121,27 @@ def _get_effective_dim(self, module: nn.Module, is_expert: bool) -> int:
             )
         return self.dim // topk
 
+    def _align_expert_dim_for_tp(
```
cuichenx: can this be shared?

src/megatron/bridge/peft/adapter_wrapper.py
```diff
@@ -156,6 +156,11 @@ def base_linear_forward(
 
         return linear_output, bias, layernorm_output
 
+    def adapter_forward(self, adapter: nn.Module, x: torch.Tensor, *args: Any, **kwargs: Any) -> torch.Tensor:
```
cuichenx, line 159: what's the purpose of this rewrite?

src/megatron/bridge/peft/utils.py
```diff
@@ -732,3 +742,414 @@ def sharded_state_dict(
...
+            )
+        if self.dropout_position == "pre":
+            x = self.dropout(x)
+
+        expert_tp_size = (
+            parallel_state.get_expert_tensor_parallel_world_size() or self.config.expert_tensor_parallel_size
+        )
+        outputs = []
+        start = 0
+        for expert_idx, split_size in enumerate(expert_splits):
```
cuichenx, line 1294: have you tested the performance of this for loop? can we make use of the grouped gemm kernel (TEGroupedMLP) instead of the for loop?

src/megatron/bridge/peft/lora.py
HollowMan6: Thanks for pointing this out, I have just moved these helper functions to utils.py

src/megatron/bridge/models/conversion/auto_bridge.py
HollowMan6: No, it's not related to CPU merge, just a cleanup. `dist.is_initialized()` will already return `False` when the c10d backend is unavailable, as it's a python based check and checks GroupMember.WORLD is not None https://github.com/pytorch/pytorch/blob/5eac6610b488e05cb1cfa6a15f2b298d69d5404f/torch/distributed/distributed_c10d.py#L1350-L1352  so dist.is_available() should be redundant here.

src/megatron/bridge/peft/adapter_wrapper.py
HollowMan6: `GroupedExpertLinearAdapter` will need `m_splits` / `tokens_per_expert` from kwargs to dispatch per local expert. I think this can be a good abstraction to have an additional method that specialize this instead of inlining `self.adapter_abc(x, *args, **kwargs)`, which gives subclasses a single override point, but of course we can change this in the inline way, but in any case we will need to pass `*args, **kwargs` for all the call sites now.

src/megatron/bridge/peft/utils.py
HollowMan6: Not really, it will definitely be slower than the grouped gemm kernel, ~~we can leave this as future work if this is needed (added a comment for this), as this will need more work than the current way. We need an all-gather between linear_in and linear_out when expert-TP > 1, which will need to be inserted between the two grouped GEMMs.~~ I added the implementation to use torch grouped_mm, ~~maybe we can also switch to TEGroupedMLP when we detect that te is installed later.~~ and TEGroupedMLP is also added and will be the first piority to use when available.

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3408/commits/45309c0ca5dcd1321742a5ce70350e043f918530

cuichenx (dismissed):
LGTM

HollowMan6:
fixed the test cases error.

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3408/commits/30c71cfe8731271fbd04fb939de578523a60a69b

weijiac0619:
@HollowMan6 could you fix the failed test? Thanks

HollowMan6:
@weijiac0619 should now be fixed, would you mind triggering the CI again?

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3408/commits/ac85b38a36f0f927bd8f443910d063d55d7fb48c

cuichenx (approved):

PR #3428, opened 2026-04-20
[ckpt] fix: handle multi-owner PP tensors in MLA/MTP broadcast during weight export

zhenyanikishkina:
# What does this PR do?
Fixes a deadlock / crash during Megatron-to-HuggingFace weight export and weight sync (`export_hf_weights`, `stream_weights_megatron_to_hf`) when using pipeline parallelism (PP > 1) with architectures that place the same weight tensor on multiple PP stages.

This issue occurs when the same tensor is legitimately owned by more than one PP stage. This is confirmed for MTP (Multi-Token Prediction) models that replicate `embedding.word_embeddings.weight` onto the MTP stage. Models with tied embeddings (`share_embeddings_and_output_weights=True`) could also trigger this pattern, although the existing `output_layer` filtering in `build_conversion_tasks` already handles the standard tied-embedding case.

# Changelog
- `broadcast_from_pp_rank`: replace `raise ValueError` on duplicate PP owners with deterministic first-owner selection. MLA/MTP weights and tied parameters legitimately exist on multiple PP ranks.
- `broadcast_obj_from_pp_rank`: add `break` to select the first owning rank instead of the last. Remove dead-code duplicate `src_rank is None` check.
- Update docstring to reflect new behavior (no longer raises on multi-owner).
- Add three unit tests: multi-owner tensor broadcast, multi-owner with `cache_key` path, and multi-owner object broadcast.

# Motivation
The old `all_gather_object` → `raise ValueError` → `torch.distributed.broadcast` pattern is unsafe: when only some PP ranks raise the exception (after completing `all_gather_object`), the remaining ranks block forever on `broadcast`, causing a **deadlock**.

# Reproduction
**Confirmed failing configuration:**
- DeepSeek-V3-family 10B MoE model with MLA and MTP
- PP=2, TP=2, EP=2, megatron-bridge v0.3.1
- `share_embeddings_and_output_weights=False`, `tie_word_embeddings=false`
- Error occurs during `update_weights` (actor→rollout weight sync via `export_hf_weights` → `stream_weights_megatron_to_hf`)
megatron_workers.py:739   rollout_mode() → export_hf_weights
→ model_bridge.py:953     stream_weights_megatron_to_hf → megatron_to_hf
→ param_mapping.py:345    broadcast_from_pp_rank
→ ValueError: Tensor exists on more than one PP rank. Found on ranks 0 and 1.

# GitHub Actions CI
Tests added in `tests/unit_tests/models/test_param_mapping.py`:
- `test_broadcast_from_pp_rank_multi_owner`
- `test_broadcast_from_pp_rank_multi_owner_with_cache`
- `test_broadcast_obj_from_pp_rank_multi_owner`
- [ ] Does the PR affect components that are optional to install?
  
  
# Additional Information
- The existing tied-embedding logic in `model_bridge.py` (`build_conversion_tasks` output_layer filter + `lm_head` clone emission) remains intact. Removing that overlapping mechanism can be deferred to a separate PR as discussed.
- Related commit a5ad8be (yuya/nemo-rl-use branch) addresses the `stream_weights_megatron_to_hf` side but not the `param_mapping.py` broadcast side.


zhenyanikishkina:
This is a confirmed bug affecting production workloads with PP>1 (both tied embeddings and MLA architectures). Would appreciate a review when possible. cc @yaoyu-33

yaoyu-33:
@zhenyanikishkina 

Thanks for the fix! Quick framing question before we merge.

The tied-embedding case is already handled at the AutoBridge stream layer in `model_bridge.py` — `build_conversion_tasks` filters `output_layer.*` when `share_embeddings_and_output_weights=True`, and `stream_weights_megatron_to_hf` clones `embed_tokens.weight` out as `lm_head.weight`. For a standard Llama/Qwen-style tied model, the broadcast should only ever see one owner. In your failed case with embedding, do you know why it escaped the filtering in existing logic?


Where this PR is genuinely valuable is the second trigger you mention: MLA / DeepSeek-V3 (and likely MTP, which also replicates `embedding.word_embeddings.weight` onto the `mtp_process` stage). Those duplicates aren't named `output_layer`, so the existing filter can't catch them. The collective-safety argument also stands on its own — the old `all_gather_object` → `raise` → `broadcast` pattern deadlocks any rank that doesn't hit the raise.


A few asks:

1. Could you reframe the PR description/title around the MLA (and ideally MTP) case as the primary motivation? It's not really a tied-embed fix.
2. If this generic first-owner approach works, we should probably **delete** the tied-embed-specific logic in `model_bridge.py` (the `output_layer` filter + `lm_head` clone emission, including the `TODO(yuya): fix this hard coded naming`) in the same PR — otherwise we have two overlapping mechanisms. (WE CAN DEFER TO ANOTHER PR.)
3. Add one test that exercises the `cache_key` path (`cache_key=str(self.hf_param)`), since that's how every real mapping calls it.

zhenyanikishkina:
@yaoyu-33 Thanks for the review!
**On the tied-embedding filtering question:** You're right — the standard tied-embedding case is already handled by `build_conversion_tasks`. In our failing run, `share_embeddings_and_output_weights=False` and `tie_word_embeddings=false`.

**Addressing your asks:**
1. **Reframed PR title/description** — updated to focus on MLA/MTP as the primary motivation. Title is now `[ckpt] fix: handle multi-owner PP tensors in MLA/MTP broadcast during weight export`.
3. **Added `cache_key` test** — new test `test_broadcast_from_pp_rank_multi_owner_with_cache` exercises the `cache_key=str(self.hf_param)` path, verifies that the second call skips `all_gather_object` and reuses the cached spec.

yaoyu-33 (dismissed):

zhenyanikishkina:
Looks like CI is currently blocked on manual test deployment approval (`cicd-wait-in-queue` shows “test needs approval to start deploying changes”).
Could someone with the required permissions please approve the test deployment so the remaining checks can proceed?

@yaoyu-33 @chtruong814

zhenyanikishkina:
The latest push only fixes the test mock setup. The code diff is unchanged from the previous follow-up.

Could you please rerun CI for d25d834, re-approve, and re-enable auto-merge if everything looks good?

@yaoyu-33 @chtruong814

zhenyanikishkina:
`Launch_Unit_Tests_Core` passed on d25d834, so the new unit tests are fine. The remaining failures seem unrelated to this PR: many L0/L1/L2 jobs are failing with the same `exit 127`, which looks like a CI issue rather than a regression from this change. Could someone please rerun or check the failing matrix jobs?

@yaoyu-33 @chtruong814

sajadn:
@zhenyanikishkina I restarted the failed CI tests!

zhenyanikishkina:
`Launch_Unit_Tests_Core` passed for this fix, so the added unit tests look fine. The remaining failures appear CI-related rather than caused by this PR; could you please re-approve once the rerun completes?
 
 @yaoyu-33 @chtruong814

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3428/commits/3b5d70a4005ea36e9a174cbee97a871fb285ba66

cuichenx (approved):

PR #3486, opened 2026-04-22
[data] fix: raise on missing {% generation %} in chat template

### LINKED ISSUE #3359: [bug] SFT chat loss masking silently falls back to all-ones when tokenizer chat_template lacks {% generation %}
## Summary

When running SFT on chat-format data with `use_hf_tokenizer_chat_template=True`, the loss mask is built via HF's `apply_chat_template(..., return_assistant_tokens_mask=True)`, which requires the Jinja chat template to wrap assistant content in `{% generation %}...{% endgeneration %}`. If the template does **not** contain that tag, Bridge silently falls back to `loss_mask = [1] * len(input_ids)` — i.e. loss is computed on the **entire** conversation (system + user + assistant), with no warning, no error, and no log line.

Most upstream HF tokenizers (Nemotron-3, Llama-3, Qwen2.5, Mistral, etc.) do **not** include `{% generation %}` in their default `chat_template`, so this is the default-case behavior for a large majority of users. The training loss curve looks plausible, which makes the bug easy to miss.

## Where it happens

`src/megatron/bridge/data/datasets/utils.py::_chat_preprocess` (around L933–L950):

```python
GENERATION_REGEX: Pattern[str] = re.compile(r"\{%-?\s+generation\s+-?%\}")

# assistant mask only works if chat template has generation keyword
template_has_generation_kwd = GENERATION_REGEX.search(tokenizer.chat_template) is not None

tokenized_chat = tokenizer.apply_chat_template(
    chat,
    tools=tools,
    tokenize=True,
    return_dict=True,
    return_assistant_tokens_mask=template_has_generation_kwd,
)

input_ids = tokenized_chat.get("input_ids")
if template_has_generation_kwd:
    mask = tokenized_chat["assistant_masks"]
else:
    mask = [1] * len(input_ids)   # <-- silent fallback, loss computed on user/system too
```

The only existing warning on this path is for truncation (`sft.py:1221`), which is unrelated.

The current behavior is also pinned by a unit test (`tests/unit_tests/data/datasets/test_chat_template.py::test_chat_preprocess_without_generation_keyword`), which asserts the all-ones fallback as expected behavior.

## Repro

1. Pick any HF tokenizer whose `chat_template` does not contain `{% generation %}` (e.g. `nvidia/Nemotron-3-...`, `meta-llama/Meta-Llama-3-8B-Instruct`, `Qwen/Qwen2.5-7B-Instruct`).
2. Run SFT with `use_hf_tokenizer_chat_template=True` on any `messages`-format dataset.
3. Observe that no warning is logged, and the effective `loss_mask` is all-ones (can be verified by inspecting a batch).

## Impact

- Silent correctness regression: SFT silently degrades to causal LM on the whole conversation.
- No log / warning / error signals this to the user.
- Affects all models whose upstream tokenizer does not ship with `{% generation %}` — which is the common case today.

### Stage 1 — stop the silent failure (small PR)

- When `use_hf_tokenizer_chat_template=True` and `GENERATION_REGEX.search(tokenizer.chat_template)` is `None`:
  - Raise `ValueError` by default, with a message explaining (a) why HF `return_assistant_tokens_mask` fails, and (b) the available remediations (patch the template, use the legacy special-tokens path, or explicit opt-in).
  - Add an explicit opt-in flag (e.g. `allow_full_sequence_loss: bool = False` or `loss_mask_strategy: Literal[\"assistant_only\", \"all_ones\"]`) for users who truly want full-sequence loss.
- Update `test_chat_preprocess_without_generation_keyword` accordingly.

### Stage 2 — best-effort fallback for templates without `{% generation %}` (medium PR)

Important caveat — there is **no** fully-correct, template-agnostic way to rebuild assistant loss masks without `{% generation %}`. Any approach that re-tokenizes the conversation in pieces and tries to align to the full tokenization is best-effort. Concretely:

- **Do not** implement the naive ''tokenize each turn's content in isolation and concatenate'' approach. This is exactly the same failure mode as the VLM substring-search code (leading-space / BPE boundary drift, dropped role headers and EOS, etc.) and produces wrong `input_ids` on top of a wrong mask.
- **Do** implement incremental `apply_chat_template` **prefix differencing**:
  ```python
  full_ids = apply_chat_template(messages, tokenize=True)
  mask = [0] * len(full_ids)
  for i, turn in enumerate(messages):
      ids_before = apply_chat_template(
          messages[:i],
          tokenize=True,
          add_generation_prompt=(turn[''role''] == ''assistant''),
      )
      ids_upto = apply_chat_template(messages[:i+1], tokenize=True)
      if turn[''role''] == ''assistant'':
          mask[len(ids_before):len(ids_upto)] = [1] * (len(ids_upto) - len(ids_before))
  ```
  This operates at the template level, not at BPE-boundary level, so it is strictly better than substring search. But it relies on the template being **token-prefix-monotonic** — an empirical property, not a theorem. Known ways this breaks:
  - Templates with a `{% if loop.last %}` branch that emits a different suffix for the final message.
  - Conditional BOS / outer wrappers that appear only at certain lengths.
  - Context-dependent separators (user→user vs user→assistant differ).
  - `clean_up_tokenization_spaces` eating trailing whitespace and merging on re-tokenization.
  - Tool-calling templates that only render `tools` when at least one user message is present.
- **Always** gate the fallback with a correctness checksum: re-tokenize the full conversation, compare against the prefix-diff reconstruction, and **raise** on drift. Never silently ship a drifted mask.
- Keep `{% generation %}` as the **preferred** primary path. The fallback is a safety net, not the recommended configuration.

A truly correct alternative, for completeness:
- **Offset-mapping over rendered text**: render the template with `tokenize=False`, locate assistant content spans in the rendered string (regex against role markers), then map character spans to token spans via `return_offsets_mapping=True`. Correctness of this depends on recognizing the template''s role markers, so it''s essentially a per-template adapter — not a universal drop-in. TRL''s `DataCollatorForCompletionOnlyLM(response_template=...)` is the low-rent version of this idea.

### Stage 3 — docs, first-party model support, CI guardrails

- Document the `{% generation %}` requirement, the three remediations, and the Stage 2 fallback''s known failure modes in `docs/training/` (there is currently no mention of `use_hf_tokenizer_chat_template` or the generation-keyword requirement anywhere in `docs/`).
- For first-party supported models (Nemotron family in particular), ship a pre-patched `chat_template` alongside the recipe, or a small helper that patches the tokenizer in-place.
- Add a ''template compatibility check'' utility: given a `(tokenizer, sample conversations)` pair, verify (a) `{% generation %}` present OR (b) prefix-monotonic checksum passes. Ship a small whitelist of tokenizers known to be safe on the Stage 2 fallback.
- CI smoke test: pick 2–3 common tokenizers (Llama-3, Qwen2.5, Nemotron-3), run `_chat_preprocess`, and assert that non-assistant tokens have `loss_mask == 0`.

## Workarounds (until fixed)

In priority order:

1. **Patch the chat template** (HF-official recommendation, lowest maintenance, aligned with Bridge''s current code path): copy the tokenizer''s `chat_template`, wrap assistant content blocks with `{% generation %}...{% endgeneration %}`, and set `tokenizer.chat_template = patched` (or point at a local tokenizer directory with the patched template). This is the only approach guaranteed to be correct.
2. **Preprocess offline**: build `{input_ids, loss_mask}` yourself and feed a packed dataset — bypasses `_chat_preprocess` entirely. If you use a prefix-differencing helper here, still validate with a full-tokenization checksum.
3. **Do not** leave `use_hf_tokenizer_chat_template=True` on an unpatched tokenizer; the loss curve will look fine while training on user/system tokens.

## Related files

- `src/megatron/bridge/data/datasets/utils.py` (`_chat_preprocess`, `GENERATION_REGEX`)
- `src/megatron/bridge/data/datasets/sft.py` (`use_hf_tokenizer_chat_template` wiring)
- `tests/unit_tests/data/datasets/test_chat_template.py`

## Related

- A sibling issue for the VLM / Energon side will be filed separately — those paths use substring search (`create_multiturn_loss_mask_by_search`, `find_pattern_indices`) instead of `{% generation %}`, and have their own silent-failure modes (per-turn miss silent, only warns on full-zero mask, and in `energon/hf_encoder_task_encoder.py` not even that).

yaoyu-33:
## Summary
- Raise `ValueError` in `_chat_preprocess` when `use_hf_tokenizer_chat_template=True` and the tokenizer's `chat_template` lacks a `{% generation %}` block
- Previously this silently fell back to `loss_mask = [1] * len(input_ids)`, computing loss on the entire conversation (system + user + assistant) with no warning
- This affects most upstream HF tokenizers (Llama-3, Qwen2.5, Nemotron, etc.) whose default templates don't include `{% generation %}`
- The error message includes actionable remediation steps: patch the template or use the legacy preprocessing path
- Updated unit tests to expect the `ValueError` and use valid templates with `{% generation %}` where needed

Closes #3359 (Stage 1)

## Test plan
- [ ] `uv run python -m pytest tests/unit_tests/data/datasets/test_chat_template.py -v` passes
- [ ] `test_chat_preprocess_without_generation_keyword` verifies `ValueError` is raised
- [ ] `test_context_answer_split_no_mask_raises` verifies `ValueError` is raised
- [ ] All other chat template tests pass with valid `{% generation %}` templates


cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3486/commits/9260da53d5a2abde793d2fad4da92ba64689722a

cuichenx (approved):
changes look good. curious what tests would fail

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3486/commits/b2dcf49ccedbc445ab494fc2611b4b7d65eea89d

PR #3529, opened 2026-04-26
perf(fix): accumulate per-microbatch FLOPS metadata for accurate…

### LINKED ISSUE #3498: [Bug] VLM TFLOPS calculation is inaccurate: uses config seq_length instead of actual FA sequence length, and omits ViT encoder FLOPs
### Problem

The current TFLOPS calculation in flop_utils.py has two significant issues when training Vision-Language Models (e.g., Qwen3.5-VL):

### Issue 1: LLM FLOPS uses configured seq_length instead of actual FA sequence length
During VLM SFT training, preprocess_packed_seqs / pack_or_pad_batch_sequences dynamically pads sequences to the actual content length (rounded up to divisible_by), not to cfg.model.seq_length. For example, with cfg.model.seq_length=4096, the actual FA sequence length observed is typically ~1280 (e.g., 950 vision tokens after spatial merge + ~330 text tokens).

However, num_floating_point_operations() unconditionally uses cfg.model.seq_length for all computations, including:

- The outer multiplier: batch_size * cfg.model.seq_length
- Attention core FLOPs (which scale quadratically with sequence length)

This causes the reported TFLOPS to be significantly overestimated (potentially 3-10x higher than reality depending on the actual-to-configured length ratio).

### Issue 2: ViT encoder FLOPs are completely missing

The FLOPS calculation only accounts for the LLM (language model) portion. For VLM models, the Vision Transformer (ViT) encoder contributes a non-trivial fraction of total compute:

- ViT transformer layers with bidirectional full attention (not causal → 4hs² per attention core vs 2hs² for causal)
- GELU MLP layers
- Patch Merger (spatial merge 2×2 + 2-layer MLP projection from ViT hidden dim to LLM hidden dim)

For Qwen3.5-VL-9B with a typical 864×1152 input image (3888 patches), ViT accounts for ~15-18% of total training FLOPS — this is entirely omitted.

### Minimal repro

```shell
1. Train any Qwen VL model (e.g., Qwen3.5-VL-9B) with SFT
2. Set cfg.model.seq_length = 4096
3.Observe that actual FA sequence length is ~1280 (visible via FA debug logs)
4.The reported TFLOPS will be ~3x higher than expected because the calculation assumes seq_length=4096
```

### Expected behavior

- TFLOPS calculation should use the actual sequence length that FA processes, not the configured maximum
- ViT encoder FLOPS should be included in the total TFLOPS metric for VLM models

### Affected area

area:perf

### Regression?

No

### Environment

Run on H20  Qwen3.5-9B.
H20 FP16 FLOPS ~=150 . But in the log  **_GPU utilization: 242.1MODEL_TFLOP/s/GPU_**

 

### Logs

```shell
Step Time : 3.81s GPU utilization: 242.1MODEL_TFLOP/s/GPU
 [2026-04-21 02:40:03] iteration        3/      60 | consumed samples:           96 | elapsed time per iteration (ms): 3813.3 | learning rate: 7.500000E-08 | global batch size:    32 | lm loss: 1.315767E+01 | mtp_1 loss: 1.332609E+01 | loss scale: 1.0 | grad norm: 221.443 | number of skipped iterations:   0 | number of nan iterations:   0 |
Step Time : 4.31s GPU utilization: 214.4MODEL_TFLOP/s/GPU
```

SophusDavid:
perf(training): improve VLM TFLOPS calculation with actual sequence length and ViT

## What does this PR do ?

Fixes inaccurate TFLOPS reporting for VLM training where the reported FLOPS was based on the statically configured `cfg.model.seq_length` instead of the dynamically padded sequence lengths that FlashAttention actually processes. This led to significant overestimation (up to ~2.6x for typical VLM SFT scenarios with short sequences + images).

The PR introduces a per-micro-batch accumulator pattern that captures the real padded sequence length and vision patch count from each `forward_step` invocation, scales them by data-parallel size for global estimation, and feeds them into an extended `num_floating_point_operations()` API.

Closes #3498

## Changelog

- **vlm_step.py**
  - Accumulate `_flops_seqlen_sum` and `_flops_seqlen_sq_sum` across micro-batches using `tokens.shape[1]` (actual padded length).
  - Accumulate `_flops_vision_patches` from `visual_inputs.image_grid_thw` / `video_grid_thw`.
  - Add `"total_tokens"` to `packed_seq_params` for downstream visibility.

- **train.py**
  - Reset `_flops_*` accumulators to 0 before each `wrapped_train_step` call.
  - After the step, read accumulated per-rank totals, scale by `dp_size` for global FLOPS, and pass effective `seq_length` to `training_log`.
  - Remove the earlier `seq_length` return value from `train_step` (reverting to original 8-element return signature).

- **flop_utils.py**
  - Add `vit_flops()` helper for Vision Transformer encoder FLOPS (bidirectional attention + GELU MLP + patch merger).
  - Extend `num_floating_point_operations()` signature with optional `seqlen_sum`, `seqlen_squared_sum`, and `num_vision_patches` parameters.
  - Add `_compute_vit_flops()` that derives per-image patch count from total batch patches for correct quadratic attention scaling.

- **train_utils.py**
  - `training_log()` now accepts `seq_length` parameter and uses accumulated `_flops_vision_patches` (scaled by `data_parallel_size`) for consistent per-log FLOPS recomputation.

## Before your PR is "Ready for review"

**Pre checks**:

  - Added `test_flops_calculation.py` with 13 verification tests covering VLM and non-VLM scenarios.

## Additional Information

- Related to #3498
- Verified on H20(BF16 148TFLOPS) Qwen3.5-VL-9B dense model SFT: old method reported ~199 TFLOP (seq=4096, no ViT), new method reports ~77 TFLOP (actual seq=1280 + ViT=15.5 TFLOP).


src/megatron/bridge/training/utils/flop_utils.py
```diff
@@ -26,8 +26,98 @@
 _lora_seq_stats_cache: dict = {}
 
 
-def num_floating_point_operations(cfg: ConfigContainer, batch_size: int = 1):
-    """Return the number of floating point operations"""
+def vit_flops(
+    batch_size,
```
yaoyu-33: why not still take in whole config + batch size? The kwargs list might change later, the signiture change will break compatibility, I would just pass in the config here.

yaoyu-33:
## Bug: `seqlen_squared_sum` plumbing is dead — attention FLOPs underestimated for variable-length batches

In `src/megatron/bridge/training/utils/flop_utils.py`, `effective_seq_length_sq` is computed at lines 116-119 from `seqlen_squared_sum`, but is **never used** anywhere else in the function.

The attention core term in both the MLA branch (lines 480-486) and the MHA/GQA / SWA branch (lines 526-538) is written as `... * effective_seq_length / 2 ...`, and the outer expression multiplies by `seqlen_sum` (line 618). The product is:

```
seqlen_sum × effective_seq_length / 2  =  (Σ Lᵢ) × mean(Lᵢ) / 2  =  batch_size × mean(L)² / 2
```

But the physically correct attention cost is `Σ Lᵢ² / 2 = batch_size × mean(L²) / 2`. By Jensen's inequality `mean(L)² ≤ mean(L²)`, so whenever sequence lengths actually vary across the micro-batches accumulated in `vlm_step.py` (lines 439-440) — exactly the case this PR adds plumbing for — the attention FLOPs are systematically **under-counted**. For fixed-length batches the two are equal, which is why a homogeneous-batch test wouldn't surface this.

### Suggested fix

In the core-attention terms, replace `effective_seq_length` with `effective_seq_length_sq / effective_seq_length` (equivalently `seqlen_squared_sum / seqlen_sum`). After the outer multiplication by `seqlen_sum`, the attention contribution becomes `seqlen_squared_sum × (...) / 2`, which is what `seqlen_squared_sum` was added to enable.

Concretely:

- MLA branch (lines 480, 486): `effective_seq_length` → `effective_seq_length_sq / effective_seq_length`
- MHA/GQA `full_core` (line 538) and SWA `full_core` (line 526): same substitution; `swa_core` (line 527) stays as-is since SWA attention is linear in seq_len once the window saturates.

Worth adding a unit test that feeds heterogeneous `(seqlen_sum, seqlen_squared_sum)` and asserts the attention term scales with Σ L², not (Σ L)²/B.

src/megatron/bridge/training/utils/flop_utils.py
SophusDavid: You are right . Done in new patch.

src/megatron/bridge/training/utils/flop_utils.py
```diff
@@ -26,8 +26,98 @@
...
+    )
+
+    return (transformer_flops_val + merger_flops_val) * batch_size * 3  # 3x for training (fwd + bwd)
+
+
+def num_floating_point_operations(
+    cfg: ConfigContainer,
+    batch_size: int = 1,
+    seqlen_sum: int = None,
+    seqlen_squared_sum: int = None,
```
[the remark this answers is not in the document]
SophusDavid: Done.

src/megatron/bridge/training/utils/flop_utils.py
```diff
@@ -26,8 +26,98 @@
...
+    if seqlen_sum is not None and batch_size > 0:
+        effective_seq_length = seqlen_sum / batch_size
+    else:
+        effective_seq_length = cfg.model.seq_length
+        seqlen_sum = batch_size * cfg.model.seq_length
+
+    if seqlen_squared_sum is not None and batch_size > 0:
+        effective_seq_length_sq = seqlen_squared_sum / batch_size
+    else:
+        effective_seq_length_sq = effective_seq_length**2
```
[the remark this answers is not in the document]
SophusDavid: Applied the suggested substitution.  I introduced a single compensating factor named for its role
instead of its algebraic form:

```python
# flop_utils.py, after resolving seqlen_sum / effective_seq_length
if seqlen_squared_sum is not None and seqlen_sum > 0:
    core_attn_seq_factor = seqlen_squared_sum / seqlen_sum   # ≡ avg(s²) / avg(s)
else:
    core_attn_seq_factor = effective_seq_length              # legacy fallback
```
Mathematically identical to the suggestion (effective_seq_length_sq / effective_seq_length = seqlen_squared_sum / seqlen_sum). Applied to all quadratic call-sites.

Add some tests and all new tests pass under ```uv run python -m pytesttests/unit_tests/training/utils/test_flop_util`s.py -v ```.

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3529/commits/f5d422bf4b6d9e7614d783c4c0fb1244743b9f07

yaoyu-33 (dismissed):

yaoyu-33:
@SophusDavid : please run precommit locally to resolve lint issues

weijiac0619:
@SophusDavid could you fix the failed test? Thank you!

SophusDavid:
> @SophusDavid : please run precommit locally to resolve lint issues

Ok!  resolved in commit [dfc2fd4](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3529/commits/dfc2fd42297dea0ed25c7f77d6e2833bafc3266e)

meatybobby:
@SophusDavid: could you verify if the conflict with main resolved and rerun precommit?

SophusDavid:
> /ok to test [91e73a0](https://github.com/NVIDIA-NeMo/Megatron-Bridge/commit/91e73a0b4b394aaab3170d2c468cf0d3b2e6952d)

Apologies — I didn't notice the UT job was failing on this PR earlier. I've now run the relevant tests locally and pushed fixes:

> tests/unit_tests/training/utils/test_flop_utils.py
> tests/unit_tests/training/utils/test_train_utils.py
> tests/unit_tests/training/test_train.py

While investigating, I also found that this patch was computing seqlen_sum / seqlen_squared_sum / num_vision_patches incorrectly under VPP (interleaved pipeline): forward_step_func is invoked once per virtual-stage per microbatch, so the accumulators were inflated by vp_size, while num_floating_point_operations() already covers all layers of the full model. This caused TFLOP/s/GPU to be reported vp_size× too high (e.g. 3× with vp_size=3).
Fixed by dividing the accumulators by vp_size in both train.py and train_utils.py before scaling to DP. Also hardened the accumulator reads against MagicMock test doubles (isinstance(..., int) guard) so the existing mock-based unit tests keep working.

aroshanghias-nvd:
/ok to test [0369c27](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3529/commits/0369c274945fee3630ad0b07a143f70dc35708f1)

src/megatron/bridge/training/utils/flop_utils.py
```diff
@@ -524,39 +632,49 @@ def transformer_flops():
...
+
+        Note: num_vision_patches is the *total* patches across the batch.
+        ViT attention is per-image (not cross-image), so we convert to
+        per-image patch count before invoking ``vit_flops`` to get the
+        correct quadratic attention scaling. ``vit_flops`` itself returns
+        0 when ``cfg.model.vision_config`` is absent.
+        """
+        if num_vision_patches <= 0:
+            return 0
+        patches_per_image = num_vision_patches / batch_size if batch_size > 0 else num_vision_patches
```
cuichenx, line 676: this has an implicit assumption that batch size == num images?

cuichenx:
merging this first, comment above can be fixed in a followup PR

cuichenx (approved):

PR #3530, opened 2026-04-26
Update Llama3 70B LoRA GB200 BF16 perf config

rhmukundan:


yaoyu-33:
Closing because this modifies the retired legacy performance-config layout. The flat performance recipes landed in #2803 and the legacy configs were removed by #4623; current LoRA performance tuning should be applied to the maintained recipe.

PR #3621, opened 2026-05-01
[misc] refactor: flatten skills layout, add perf- prefix, and when_to_use frontmatter

ko3n1g:

Flattens the skills directory layout and enriches skill frontmatters to make agent routing more reliable.

## Changes

**Flatten perf-techniques into `skills/`** (`0636e750`)
All performance technique skills (previously nested under `skills/perf-techniques/`) are moved to top-level entries under `skills/`. This makes the layout consistent with other skills and simplifies `@`-path references.

**Add `perf-` prefix to performance technique skills** (`7ef7ab37`)
Renamed all former perf-technique skills (`activation-recompute`, `cpu-offloading`, etc.) to `perf-activation-recompute`, `perf-cpu-offloading`, etc. so they are visually grouped and unambiguous alongside workflow skills.

**Replace co-located `card.yaml` refs with `@`-file paths** (`dffe28c7`)
Skill bodies that referenced `card.yaml` via relative paths now use the `@skills/<name>/card.yaml` notation, consistent with how docs are referenced.

**Add `when_to_use` field to all skill frontmatters** (`ba3bdcf4`)
Every SKILL.md gains a `when_to_use:` YAML key alongside `name` and `description`. This makes invocation triggers machine-readable and separate from the free-form description, and lets agents decide which skill to load without reading the full body.

## Example

```yaml
---
name: perf-memory-tuning
description: Techniques for reducing peak GPU memory in Megatron Bridge — ...
when_to_use: When hitting GPU OOM, diagnosing memory fragmentation, or systematically reducing peak activation or optimizer memory.
---
```

</details>


docs/training/activation-recomputation.md
```diff
@@ -193,5 +193,5 @@ up-projection modules than to enable blanket full recomputation immediately.
 
 - [docs/training/cuda-graphs.md](cuda-graphs.md)
 - [docs/training/moe-optimization.md](moe-optimization.md)
-- [skills/perf-techniques/activation-recompute/SKILL.md](../skills/perf-techniques/activation-recompute/SKILL.md) — per-module cost/savings data, measured results
```
chtruong814, line 196: were these paths just wrong?

chtruong814 (approved):
Just had a clarification question.  If that's good, then I'm good.

docs/training/activation-recomputation.md
ko3n1g: Yeah, skills are only identified as first-class citizens when they are flat inside skills/. Nested dirs makes them anonymous markdowns and not “skills”. This impacts discoverability and context window compression

cuichenx (approved):
LGTM

PR #3647, opened 2026-05-04
[training] feat: forward MoE/MTP metrics to MLFlow and Comet

### LINKED ISSUE #2989: track_moe_metrics() does not forward MoE metrics to Comet ML
## Summary

PR #2910 added excellent Comet ML support to Megatron-Bridge with 18+ metric call sites in `training_log()` (thank you for that!). However, MoE-specific metrics (load balancing loss, z-loss, etc.) are not forwarded to Comet ML because `track_moe_metrics()` in Megatron-LM's `moe_utils.py` only writes to TensorBoard and W&B writers.

## Current Behavior

In `train_utils.py`, `training_log()` calls `track_moe_metrics()` which accepts `writer` (TensorBoard) and `wandb_writer` parameters but has no `comet_logger` parameter. The MoE metrics are computed and reduced correctly but only written to TB/W&B:

```python
track_moe_metrics(
    loss_scale=moe_loss_scale,
    iteration=iteration,
    writer=writer,
    wandb_writer=wandb_writer,
    total_loss_dict=total_loss_dict,
    ...
)
```

The `comet_logger` is available in the same scope but not used.

## Expected Behavior

MoE metrics (`load_balancing_loss`, `seq_load_balancing_loss`, `global_load_balancing_loss`, `z_loss`) should be forwarded to Comet ML alongside TB/W&B, matching how all other metrics in `training_log()` are dispatched to all three logging backends.

## Environment

- Megatron-Bridge: `combined/all-fixes` branch (post PR #2910)
- Megatron-LM: commit `f26190677`
- Container: `nvcr.io/nvidia/nemo:26.02.nemotron_3_super`

lonexreb:
## Summary

MCore's `track_moe_metrics` and `track_mtp_metrics` only forward metrics to TensorBoard and W&B. Users wiring up Comet (or MLFlow) never see MoE auxiliary losses (`load_balancing_loss`, `seq_load_balancing_loss`, `global_load_balancing_loss`, `z_loss`) or MTP per-layer losses on those backends.

Per [maintainer guidance on the issue](https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/2989) — *\"Megatron-Bridge side can monkey-patch `track_moe_metrics` to avoid a cross-repo dependency\"* — wrap the TB writer with a small `SummaryWriter`-shaped adapter that fans out every `add_scalar(name, value, iteration)` call to MLFlow and Comet using the **same per-step value** TB receives. W&B is unaffected — the underlying MCore functions still receive `wandb_writer` directly.

Refs #2989.

## Why a wrapper, not `total_loss_dict`

The original report tried reading from `total_loss_dict` for the Comet path. That's wrong because `total_loss_dict` accumulates with `+=` across iterations, so values monotonically grow (the issue records 1.24 → 6.14 instead of the correct 1.20–1.24 per-step range). The wrapper approach captures the exact per-step averaged value that TB receives (`loss_list.sum() / num_moe_layers`) without any further bookkeeping.

## Implementation

`src/megatron/bridge/training/utils/train_utils.py`:

- New `_MoeMetricFanoutWriter` adapter — `add_scalar(name, value, iteration)` forwards to TB (if any), MLFlow (`log_metrics(..., step=iteration)`), and Comet (`log_metrics(..., step=iteration)`).
- New `_build_moe_metric_writer(tb_writer, comet_logger, mlflow_logger)` factory:
  - Returns the real TB writer unchanged when neither Comet nor MLFlow is wired up — **zero overhead, no behavior change**.
  - Returns the wrapper when at least one of Comet / MLFlow is wired up. The wrapper is returned even if `tb_writer is None`, which is required to surface MoE / MTP metrics in Comet / MLFlow when the user hasn't enabled TensorBoard.
- Wired into both call sites:
  - `track_moe_metrics(..., writer=moe_metric_writer, ...)`
  - `MTPLossLoggingHelper.track_mtp_metrics(..., mtp_metric_writer, ...)`

Tensor sanitation: 0-d torch tensors are converted to Python scalars with `.item()` before being handed to MLFlow / Comet — TB tolerates 0-d tensors, MLFlow / Comet do not. The TB writer still receives the original value untouched.

Per-layer logging (`per_layer_logging=True`) is naturally handled — each `writer.add_scalar(f\"moe/{name}_layer_{i}\", ...)` call inside `track_moe_metrics` fans out individually.

## Test plan

- [x] `python3 -m ast` parse of changed files
- [x] `ruff check` clean on changed files (auto-fixed import order)
- [x] `ruff format --check` clean on changed files
- [x] 8 new unit tests under `TestMoeMetricFanoutWriter`:
  - `_build_moe_metric_writer` returns the original writer when no Comet / MLFlow → bypass contract
  - returns wrapper when only Comet present
  - returns wrapper when only MLFlow present
  - `add_scalar` fans out to TB + Comet + MLFlow when all three are present
  - `add_scalar` works when TB writer is None
  - 0-d tensors sanitized to Python floats for Comet / MLFlow but TB receives the tensor untouched
  - Plain Python scalars pass through unchanged
  - Per-layer loop logs each layer individually with the right step
- [ ] CI: existing `test_moe_logging*` tests continue to pass — those tests do not provide a `comet_logger` / `mlflow_logger`, so the helper returns the original writer unchanged (the existing assertions on `writer.add_scalar` are unaffected).
- [ ] CI: L0 / L1 functional tests on H100 / GB200.
- [ ] Manual: with `mlflow_experiment` or `comet_experiment` configured and a MoE model, observe `load_balancing_loss` (and friends) appearing in MLFlow / Comet at the same per-step values as TensorBoard.

## Risk

Low.
- No public API change.
- `_build_moe_metric_writer` returns the original writer object when no fanout target is configured — bit-for-bit equivalent to the previous code path.
- The wrapper only adds `log_metrics(...)` calls on the rank that already has those loggers configured (rank N-1 only).
- W&B path is untouched — `wandb_writer` is still passed directly to MCore.
- No change to `total_loss_dict` semantics.

## Notes for reviewers

- An alternative was forking the helper signature in MCore (`comet_logger` parameter), but the maintainer explicitly requested a Bridge-side workaround to avoid the cross-repo dependency.
- The wrapper deliberately mimics the full `add_scalar(name, value, iteration)` shape rather than monkey-patching `track_moe_metrics`, so future MCore changes that add new metric emissions through the same writer interface get fanned out automatically.


cuichenx (approved):
LGTM

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3647/commits/6de675062ba6cf565874a3dd6bb672afa9c77542

PR #3650, opened 2026-05-04
[training, test] feat: add unit tests for pg_utils helpers

lonexreb:
## Summary

`src/megatron/bridge/training/utils/pg_utils.py` contains two key helpers used across the training stack but had **no direct unit-test coverage**:

- `get_pg_collection(model)` — accessor that returns a model's `pg_collection`, with a documented MPU fallback.
- `DistTrainProcessGroupCollection` — process-group wrapper used by the dist-train multi-module pipeline.

Existing references in [`tests/unit_tests/training/test_decentralized_pg.py`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/tests/unit_tests/training/test_decentralized_pg.py) only `isinstance`-check the result — they don't exercise the constructor's field-copy logic, the language-model branching, or `get_pg_collection`'s control flow.

This PR adds **10 focused unit tests** for both helpers. Tests-only — no production code changes.

### `get_pg_collection` (4 tests)

| Test | Path covered |
|---|---|
| `test_returns_pg_collection_attached_to_single_model` | Happy path: model has `pg_collection` → return it |
| `test_uses_first_chunk_when_passed_a_list` | List-of-chunks path (used when models wrap in lists for VPP) → `model[0]` is referenced |
| `test_falls_back_to_mpu_when_pg_collection_attribute_missing` | The documented contract: `get_attr_wrapped_model` raises with the exact `"couldn't find attribute pg_collection"` substring → `ProcessGroupCollection.use_mpu_process_groups()` is invoked |
| `test_reraises_runtime_error_when_message_does_not_match` | An unrelated `RuntimeError` propagates unchanged AND the MPU fallback is **not** invoked |

### `DistTrainProcessGroupCollection` (6 tests)

| Test | Behavior locked in |
|---|---|
| `test_inherits_process_group_collection` | Subclass relationship (so callers can `isinstance`) |
| `test_copies_set_fields_from_source_collection` | All set fields (`tp`, `pp`, `cp`) on the source are copied to the wrapper |
| `test_unset_fields_default_to_none` | Fields never set on the source default to `None` on the wrapper (constructor uses `getattr(..., None)`) |
| `test_no_language_model_by_default` | `language_model_module_name=None` keeps `has_language_model()` False and `language_model_collection` None |
| `test_language_model_attaches_source_collection` | `language_model_module_name="llm"` flips `has_language_model()` True and exposes the source as `language_model_collection` |
| `test_get_language_model_cp_size_returns_lm_cp_size` | Returns `cp.size()` on the LM collection |
| `test_get_language_model_cp_size_raises_without_lm` | Raises `ValueError("No language model specified")` when no LM is configured |

## Why this matters

- The MPU fallback behavior in `get_pg_collection` is brittle: it depends on a substring match against `RuntimeError`'s message. If MCore ever changes the error message, this fallback silently breaks. A test that asserts the substring contract makes that breakage loud.
- `DistTrainProcessGroupCollection`'s constructor uses `getattr(..., None)` for every field — without coverage, regressions that change defaults (e.g., raising vs returning None) would slip through.

## Test plan

- [x] `python3 -m ast` parse clean
- [x] `ruff check` clean
- [x] `ruff format --check` clean
- [ ] CI: `cicd-unit-tests-core` picks up the new module automatically (lives under `tests/unit_tests/training/utils/`)

## Risk

Zero — tests only. No production code, no public API touched. The contracts under test have been stable since the file was added; this just locks them in so future refactors fail fast if behavior drifts.

## Notes for reviewers

- The MPU fallback test patches `ProcessGroupCollection.use_mpu_process_groups` to a sentinel return rather than letting it touch `parallel_state`, so the test runs cleanly without distributed init.
- The wrapper test uses bare `ProcessGroupCollection()` (all fields are `init=False`) plus `setattr` for the fields we actually care about — keeps the test surface small and resilient to ProcessGroupCollection field additions in MCore.


cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3650/commits/b0bd1e23219c30c2eada575e57d15325d7b937ab

cuichenx (approved):

PR #3665, opened 2026-05-04
[data] fix: pad chat tensors and loss_mask in pre_pad_dataset (#2610)

### LINKED ISSUE #2610: tokenize_dataset crashes with TypeError when chat=True and pad_seq_to_mult > 1
**Describe the bug**

`tokenize_dataset()` in `packed_sequence.py` fails when `chat=True` and `pad_seq_to_mult > 1`:

1. **Tensor/list mismatch**: `_chat_preprocess` returns `torch.LongTensor`/`torch.BoolTensor`, but `pre_pad_dataset` concatenates with plain lists (`val + [pad_id] * ...`), raising `TypeError`.

2. **Missing `loss_mask` padding**: `pre_pad_dataset` pads `input_ids` and `context_ids` but not `loss_mask`. Sequences with different original lengths can round to the same padded `input_ids` length, so `create_hist` groups them together — but their `loss_mask` arrays differ in length, causing `np.array()` in `fill_packing_strategy` to fail with `ValueError: inhomogeneous shape`.

**Steps/Code to reproduce bug**

```python
from megatron.bridge.data.builders.finetuning_dataset import FinetuningDatasetBuilder
from megatron.bridge.data.datasets.packed_sequence import PackedSequenceSpecs

builder = FinetuningDatasetBuilder(
    dataset_root="path/to/jsonl",
    tokenizer=tokenizer,
    seq_length=221184,
    packed_sequence_specs=PackedSequenceSpecs(
        packed_sequence_size=221184,
        tokenizer_model_name="qwen3-14b",
        pad_seq_to_mult=8,
    ),
    dataset_kwargs={"chat": True, "use_hf_tokenizer_chat_template": True},
)
builder.prepare_packed_data()
```

**Expected behavior**

Tokenization and packing complete without error when using chat datasets with `pad_seq_to_mult > 1`.

**Additional context**

- `GPTSFTDataset` (non-chat) is unaffected — it returns plain lists and does not include `loss_mask` in its output dict.
- `pad_seq_to_mult=1` is unaffected — the padding block is skipped entirely.

lonexreb:
## Summary

Fixes #2610 — `tokenize_dataset` raises when packing chat datasets with `pad_seq_to_mult > 1`.

Two distinct bugs in `pre_pad_dataset` (inside `src/megatron/bridge/data/datasets/packed_sequence.py`) interacted to break this path:

1. **Tensor / list mismatch.** `_chat_preprocess` (and the legacy `_preprocess`) return `torch.LongTensor` / `torch.BoolTensor`, but the padding logic concatenated them with a Python list (`val + [pad_id] * N`), which raises `TypeError` for tensors.
2. **Missing `loss_mask` padding.** Only `input_ids` and `context_ids` were padded. `loss_mask` kept its original length, so when `create_hist` grouped two items by their *padded* `input_ids` length, `fill_packing_strategy` later hit `ValueError: setting an array element with a sequence ... inhomogeneous shape` on `np.array([x['loss_mask'] for x in per_seq_data])`.

The failure surfaces only when `chat=True` *and* `pad_seq_to_mult > 1` — `pad_seq_to_mult == 1` and the non-chat `GPTSFTDataset` path are unaffected.

## Fix

`src/megatron/bridge/data/datasets/packed_sequence.py:121-167`

- Normalize each value to a Python list via `tolist()` before concatenation, so tensor and list inputs share the same padding path.
- Pad `input_ids` and `context_ids` with `pad_id`; pad `loss_mask` with `False` so padded positions never contribute to the training loss.
- Iterate from a fixed `_PAD_VALUES` map keyed `{input_ids, context_ids, loss_mask}` and skip keys missing from `data` — preserves the non-chat path which has no `loss_mask`.

The diff is +44 / -13 in the source file; the public function signature is unchanged.

## Tests

`tests/unit_tests/data/datasets/test_packed_chat_padding.py` (new, 233 LoC) covers the full bug surface:

| Test | What it pins |
|---|---|
| `test_torch_tensor_inputs_do_not_raise` | Bug 1: tensor inputs no longer raise `TypeError` during padding |
| `test_loss_mask_is_padded_to_input_ids_length` | Bug 2: bin-mates with different original lengths now share equal padded length |
| `test_loss_mask_padding_uses_false` | Padded loss_mask positions are `False` (no spurious loss on pad) |
| `test_context_ids_padded_with_pad_id` | `context_ids` continues to pad with `pad_id`, like `input_ids` |
| `test_non_chat_list_inputs_still_padded` | Non-chat path (lists, no `loss_mask`) still works; we don't invent `loss_mask` |
| `test_pad_seq_to_mult_one_skips_padding` | `pad_seq_to_mult=1` bypass is preserved (tensors pass through unchanged) |
| `test_create_hist_then_fill_packing_strategy_no_error` | End-to-end round-trip through `create_hist` → `fill_packing_strategy` that previously raised `ValueError: inhomogeneous shape` |

Implementation notes for reviewers:
- All tests use `unittest.mock` to stub `create_sft_dataset`, so they run on CPU and don't need a real tokenizer or HF Hub access.
- Sequence lengths are tiny (3-8 tokens) and `max_seq_length=64`, keeping the suite well under the L0 budget.
- The end-to-end test deliberately constructs two items of original length 3 and 5 — both round to `max_length_to_pad=8` and land in the same `create_hist` bucket. Before the fix, this is the exact configuration that triggered the inhomogeneous-shape error reported in #2610.

## Test plan

- [x] `ruff check` and `ruff format --check` pass on both files
- [x] Standalone simulation of the padding logic verified all invariants pre-CI (cannot run pytest locally on macOS without the Megatron-LM submodule + GPU deps; CI will validate)
- [ ] CI L0 unit tests — runs `tests/unit_tests/data/datasets/test_packed_chat_padding.py` automatically (test discovery is path-based)

## Risk

Zero production risk for users who don't combine `chat=True` with `pad_seq_to_mult > 1` — those code paths are untouched. For the affected combination, behavior changes from \"raises `TypeError` / `ValueError`\" to \"produces a correctly padded packed dataset\" — no silent behavior change.

## Notes for reviewers

- I considered limiting the change to converting tensors to lists in the existing `{input_ids, context_ids}` set, but that would leave Bug 2 unfixed: bin-mates with mismatched `loss_mask` lengths still crash inside `fill_packing_strategy`. Both fixes are required for the reported repro.
- `loss_mask` pad value: `False` is consistent with how `_collate_item` treats the padding region (`pad_id=0`) inside `GPTSFTChatDataset.collate_fn` (`src/megatron/bridge/data/datasets/sft.py:1261`), and with the non-chat `_build_loss_mask` path which uses `0.0`/`1.0` floats.


cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3665/commits/8572aa6d368604d5ba201026b8c913daf2219da2

asolergi-nv (commented):
Thanks for the fix! The implementation looks good to me. The CI failure is just an off-by-2 in `test_create_hist_then_fill_packing_strategy_no_error`: both samples pad to length 9, and `fill_packing_strategy` extends the stored length-9 lists (not `seq_len=8`), so the pack ends up at `9+9=18` tokens. Could you flip the last two assertions from `== 16` to `== 18` (and tweak the comment)?

lonexreb:
Hi @asolergi-nv, thanks for the careful CI triage on May 7 — the off-by-2 you flagged in `test_create_hist_then_fill_packing_strategy_no_error` is addressed in [a0c0f08a](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3665/commits/a0c0f08a197610f5a8dbf6e4d4077660da635de6) (pushed May 8 — looks like the `waiting-on-customer` label and CI never refreshed against the new SHA, so the update may have slipped past).

Confirming the diagnosis exactly as you described:

- `pre_pad_dataset` pads each item's `input_ids` to `max_length_to_pad + 1 = 9`.
- `create_hist` keys items by `len(input_ids) - 1 = 8` (the label-aligned length).
- `fill_packing_strategy` iterates per histogram bucket but writes the **full** padded `input_ids` (length 9) into the packed output — not the bucket key.
- For the `[8, 8]` assignment in this test, the packed pair is therefore `9 + 9 = 18`. `loss_mask` is rolled via `x[1:] + [False]`, preserving the same length.

The two assertions are now `== 18`, with a comment explaining the histogram-key vs stored-tensor distinction so the next reader doesn't have to re-derive it. Production code is unchanged — only the test expectation was wrong.

Would you mind kicking CI on `a0c0f08a197610f5a8dbf6e4d4077660da635de6` (`/ok to test a0c0f08a197610f5a8dbf6e4d4077660da635de6`) and flipping the label back to `needs-review`? Happy to rebase if it would help.

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3665/commits/a0c0f08a197610f5a8dbf6e4d4077660da635de6

yaoyu-33:
we are entirely refactor dataset part, wouldn't need to get this in for now. thanks for contribution.

lonexreb:
Thanks @yaoyu-33, that makes sense — a full refactor of the dataset path is the right call over an incremental patch here. For anyone landing on this later: the specific crash this PR targeted (issue #2610, `TypeError`/ragged-array on `chat=True` + `pad_seq_to_mult > 1`) was already resolved on `main` by #4056, so no coverage is lost by closing this. Happy to help validate the new dataset path once it lands. 🙏

PR #3673, opened 2026-05-04
[model, ckpt, docs] fix: support HF→Megatron conversion under decentralized PGs

cuichenx:
## Summary

Two of three `examples/decentralized_pg/` examples failed on `nvcr.io/nvidian/nemo:26.04.rc7`. The bug in `pretrain_qwen3_with_decentralized_pg.py` is the load-bearing one: with `use_decentralized_pg=True`, Megatron-Core's `mpu` globals are never populated — process groups live only inside the user's `ProcessGroupCollection`. Two layers of the bridge were still reading those globals and asserted at HF→Megatron conversion time.

- `model_bridge.build_conversion_tasks` and `_broadcast_shared_embeddings` called `parallel_state.get_pipeline_model_parallel_*` / `get_embedding_group` unconditionally.
- `MegatronParamMapping.__init__` snapshotted `mpu` groups at registry construction; in the decentralized path the snapshot was all `None`, so `tp_size` collapsed to `world_size` and shape calculations were wrong.

This PR resolves PP/EP/embedding groups from the model's `pg_collection` (set by `LanguageModule`) with a fallback to `parallel_state` for the legacy MPU path, and adds `MegatronParamMapping.set_process_groups_from_pg_collection` so the bridge re-installs the user-supplied groups onto every mapping before running tasks. The four `parallel_state.*` call sites the bug report cited (and one EP sister site uncovered while testing) now go through small `_get_*` helpers.

Also refresh the two `decentralized_pg/` examples that the bug report cited — they had drifted from the recipe and training APIs:

- `pretrain_qwen3_simple.py`: `qwen3_4b_pretrain_config()` no longer takes any args; apply overrides on the returned config.
- `pretrain_qwen3_with_decentralized_pg.py`: `train()` takes `checkpoint_manager` (not `checkpointing_context`); add `share_embeddings_and_output_weights=False` to skirt the known MCore-globals issue during checkpoint save; `load_weights=False` since the demo deliberately downsizes the model and uses `NullTokenizer`.

## Test plan

Verified end-to-end on 4×H100 against the bug report's reproduction commands:

- [x] **Simple example** — exercises the fix end-to-end (HF Qwen3-4B weights loaded through the decentralized PG path):
  \`\`\`
  uv run python -m torch.distributed.run --nproc_per_node=4 \\
    examples/decentralized_pg/pretrain_qwen3_simple.py
  \`\`\`
  100 iters with TP=2 PP=2; lm loss 6.17 → 0.02.
- [x] **Manual example** — exercises the manual \`HyperCommGrid\` PG creation API:
  \`\`\`
  uv run python -m torch.distributed.run --nproc_per_node=4 \\
    examples/decentralized_pg/pretrain_qwen3_with_decentralized_pg.py \\
    --tp-size 2 --pp-size 2 --train-iters 5 --global-batch-size 4 --micro-batch-size 1
  \`\`\`
  5 iters with TP=2 PP=2 plus checkpoint save → "Training complete!".
- [x] Existing unit tests pass for the touched paths: \`tests/unit_tests/models/test_param_mapping.py\`, \`test_mapping_registry.py\`, \`test_auto_bridge.py\` (133 passed; one pre-existing failure on \`test_export_adapter_weights\` is unrelated and reproduces on \`main\`), \`test_qwen3_bridge.py\` (23 passed).
- [x] Cherry-pick to \`r0.4.0\` was prepared and applied (see linked PR); not a clean cherry-pick — see notes there.

### DIFF
--- a/examples/decentralized_pg/pretrain_qwen3_with_decentralized_pg.py
+++ b/examples/decentralized_pg/pretrain_qwen3_with_decentralized_pg.py
@@ -70,6 +70,7 @@ from megatron.core.utils import get_pg_rank
 from megatron.bridge.data.loaders import setup_data_iterators
 from megatron.bridge.data.utils import get_dataset_provider
 from megatron.bridge.models import AutoBridge
+from megatron.bridge.training.checkpointing import DefaultCheckpointManager
 from megatron.bridge.training.config import (
     CheckpointConfig,
     ConfigContainer,
@@ -436,7 +437,11 @@ def run_training(args: argparse.Namespace, pg_collection: ProcessGroupCollection
     # expect pg_collection to be passed explicitly rather than reading from mpu.
 
     bridge = AutoBridge.from_hf_pretrained("Qwen/Qwen3-4B")
-    model_cfg = bridge.to_megatron_provider()
+    # This demo trains a downsized Qwen3-4B (e.g. ``--num-layers 4``) on mock
+    # data with ``NullTokenizer``, so the architecture deliberately does not
+    # match the real Qwen3-4B HF checkpoint. Skip weight loading; the model is
+    # initialized randomly, which is what the manual-PG demo needs.
+    model_cfg = bridge.to_megatron_provider(load_weights=False)
 
     # Parallelism - must match what we used to create pg_collection
     model_cfg.tensor_model_parallel_size = args.tp_size
@@ -451,6 +456,9 @@ def run_training(args: argparse.Namespace, pg_collection: ProcessGroupCollection
     model_cfg.attention_softmax_in_fp32 = True
     model_cfg.make_vocab_size_divisible_by = 128
     model_cfg.vocab_size = None
+    # Known issue: tied embeddings rely on Megatron-Core globals during
+    # checkpoint save, which are not populated in the decentralized PG path.
+    model_cfg.share_embeddings_and_output_weights = False
 
     train_cfg = TrainingConfig(
         train_iters=args.train_iters,
@@ -612,6 +620,7 @@ def run_training(args: argparse.Namespace, pg_collection: ProcessGroupCollection
     print_rank_0("")
 
     # Run the training loop
+    checkpoint_manager = DefaultCheckpointManager(checkpoint_config=cfg.checkpoint)
     train(
         forward_step_func=forward_step,
         model=model,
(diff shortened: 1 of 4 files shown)


examples/decentralized_pg/pretrain_qwen3_with_decentralized_pg.py
```diff
@@ -451,6 +456,9 @@ def run_training(args: argparse.Namespace, pg_collection: ProcessGroupCollection
     model_cfg.attention_softmax_in_fp32 = True
     model_cfg.make_vocab_size_divisible_by = 128
     model_cfg.vocab_size = None
+    # Known issue: tied embeddings rely on Megatron-Core globals during
+    # checkpoint save, which are not populated in the decentralized PG path.
+    model_cfg.share_embeddings_and_output_weights = False
```
cuichenx, line 461: Not sure what was handled before this error surfaced. 

MCore handles tying during sharded_state_dict(): when both PP first and last stages own a copy of the weight, the save path needs a DP-with-CP group to coordinate which rank actually writes. So tie_embeddings_and_output_weights_state_dict reads parallel_state.get_data_parallel_rank(with_context_parallel=True) — which asserts in the decentralized PG path because that global was never populated

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3673/commits/743fb6a7b5412118056795a93b84142f90682cb6

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3673/commits/fca303ba0924fe84f3ebead5e583137634ed44c6

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3673/commits/28fc9b2e8e1f185443dbc5cef92c8e00dd5fafc4

examples/decentralized_pg/pretrain_qwen3_with_decentralized_pg.py
yashaswikarnati: thx for flagging, as you mentioned, probably needs fix in MLM, may be we should open an issue for someone to get to it

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -167,6 +168,92 @@ def megatron_to_hf(
...
+    ``parallel_state``.
+    """
+    if megatron_model is None:
+        return None
+    models_list = megatron_model if isinstance(megatron_model, list) else [megatron_model]
+    if not models_list:
+        return None
+    unwrapped = unwrap_model(models_list[0])
+    # VLM wrappers expose the language model on ``.language_model``; the
+    # ``pg_collection`` lives on the inner ``LanguageModule``.
```
yashaswikarnati, line 191: I think this if ok for now, but not a fan or just language model having PG collection as encoders can have their own mapping to.. just wanted to flag... may be for follow up refactoring

yashaswikarnati (approved):

PR #3676, opened 2026-05-04
fix(test): remove stale duplicate test_export_adapter_weights

cuichenx:
## Summary

- Commit e5ac8aed (#2644) added a corrected \`test_export_adapter_weights\` that properly patches \`_model_bridge\` via \`PropertyMock\`, but left the old version in the same class.
- Python silently discards the first definition and runs the last, so the stale broken version was executing.
- The old test patched a non-existent module-level function (\`model_bridge.stream_adapter_weights_megatron_to_hf\`) and failed with \`No implementation found for type 'Mock'\` from the plum dispatch system.
- Fix: remove the 36-line stale duplicate; the correct version is retained.

## Test plan

- [x] \`uv run pytest tests/unit_tests/models/test_auto_bridge.py::TestAutoBridge::test_export_adapter_weights\` passes

### DIFF
--- a/tests/unit_tests/models/test_auto_bridge.py
+++ b/tests/unit_tests/models/test_auto_bridge.py
@@ -771,42 +771,6 @@ class TestAutoBridge:
                         show_progress=False,
                     )
 
-    def test_export_adapter_weights(self):
-        """Test exporting adapter weights from Megatron to HF format."""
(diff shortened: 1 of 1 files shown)


cuichenx:
force merging to fix main

PR #3744, opened 2026-05-08
[training] feat: port mcore PR #4687 MoE memory estimator fix

### LINKED ISSUE #4050: [QUESTION] MoE layer theoretical memory calculation needs to account for ETP/EDP.
Why are the effects of Expert Tensor Parallel (ETP) and Expert Data Parallel (EDP) not considered in the MoE layer weight and optimizer memory computation?

cuichenx:
## Summary

Ports the MoE-aware theoretical memory estimator from upstream **[NVIDIA/Megatron-LM#4687](https://github.com/NVIDIA/Megatron-LM/pull/4687)** (which fixes [NVIDIA/Megatron-LM#4050](https://github.com/NVIDIA/Megatron-LM/issues/4050)) into Bridge's local `theoretical_memory_utils.py`.

Bridge maintains its own copy of this estimator (it takes a `ConfigContainer`, not Megatron's argparse `args`), so the fix doesn't propagate automatically when the submodule is bumped — it has to be downstreamed by hand.

## Why

The previous Bridge formula divided **all** transformer-layer parameters — including routed experts — by `tensor_model_parallel_size`. Routed experts are actually sharded by `expert_tensor_parallel_size * expert_model_parallel_size`, with distributed-optimizer state sized by the expert data-parallel domain. The old formula silently under-reports MoE memory whenever `ETP * EP != TP`, which is the common case for Mixtral / DeepSeek / Qwen3-MoE / GPT-OSS recipes.

The new formula splits parameters into three buckets and applies the correct divisor to each:

| Bucket | Examples | Divisor |
|---|---|---|
| TP-sharded | attention, dense MLP, shared experts | TP |
| Replicated | layernorms, router gate, shared-expert gate | 1 |
| Routed experts | MoE FFN per expert | ETP × EP |

Distributed-optimizer state for routed experts uses the expert DP domain (`world_size / (ETP × EP × PP)`).

Also picks up several upstream features the prior Bridge version was missing:
- MoE layer patterns (`int` and `list` form of `moe_layer_freq`)
- Shared experts with optional gate (`moe_shared_expert_gate`)
- Multi-latent attention (DeepSeek-style `q_lora_rank` / `kv_lora_rank` / split `qk_head_dim` + RoPE term)
- Multi-Token Prediction (MTP) blocks
- Active-vs-total parameter counts in verbose output

The estimator is informational only — it does not affect training correctness or actual memory allocation. Users sizing jobs from the printed memory estimate will get accurate numbers for MoE recipes after this lands.

## Why draft

Upstream **PR #4687 is open, not merged**. The implementation can still change in review. This PR holds the port until #4687 merges, then we'll rebase to match the final upstream form before un-drafting.

## Test plan

- [x] Pre-commit (ruff + format + trailing-ws) — clean
- [x] Five new unit tests in `tests/unit_tests/training/utils/test_theoretical_memory_utils.py`:
  - Dense model returns positive memory
  - EP > 1 reduces per-rank routed-expert weight (vs. EP = 1)
  - Distributed optimizer state shrinks as DP grows
  - List form of `moe_layer_freq` is honored
  - Invalid `moe_layer_freq` type raises `TypeError`
- [ ] CI: `cicd-unit-tests-core` picks up the new module

## Checklist

- [x] Sign-off (DCO)
- [x] Copyright year 2026 on the production file
- [x] Function signature unchanged (still `compute_weight_and_optimizer_memory(config: ConfigContainer, verbose: bool = False) -> float`)
- [x] No new dependencies
- [x] Bridge `ConfigContainer` already exposes every needed field (verified against `mcore` `TransformerConfig` + `ModelParallelConfig`)
- [ ] Re-sync against final form of upstream #4687 before un-drafting

## Refs

- Upstream PR: NVIDIA/Megatron-LM#4687
- Upstream issue: NVIDIA/Megatron-LM#4050

### DIFF
--- a/src/megatron/bridge/training/utils/theoretical_memory_utils.py
+++ b/src/megatron/bridge/training/utils/theoretical_memory_utils.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -29,10 +29,16 @@ NUM_BYTES_IN_MEGABYTE: int = 1024 * 1024
 def compute_weight_and_optimizer_memory(config: ConfigContainer, verbose: bool = False) -> float:
     """Compute theoretical memory footprint for model weights and optimizer states.
 
-    Calculates the number of parameters for the model based on the configuration,
-    determines the number of parameters on the most loaded shard considering
-    pipeline and tensor parallelism, and estimates the memory needed based on
-    bytes per parameter (considering precision and optimizer type).
+    Splits parameters into three buckets — regular tensor-parallel-sharded,
+    replicated, and routed-expert — and applies the correct sharding factor
+    per bucket. Routed experts are sharded by ``expert_tensor_parallel_size *
+    expert_model_parallel_size``, with distributed-optimizer state sized by the
+    expert data-parallel domain. Other parameters use the regular TP and DP
+    domains. Supports MoE with optional shared experts, multi-latent attention
+    (DeepSeek), and Multi-Token Prediction (MTP) blocks.
+
+    Ported from ``megatron/training/theoretical_memory_usage.py`` in
+    Megatron-LM (NVIDIA/Megatron-LM PR #4687).
 
     Args:
         config (ConfigContainer): The main configuration container.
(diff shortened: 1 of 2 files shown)


src/megatron/bridge/training/utils/theoretical_memory_utils.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
cuichenx, line 1: don't modify this

PR #3746, opened 2026-05-08
[ckpt] fix: Fix qwen3_5vl Megatron-FSDP CP and TP conversion

conver334:
# What does this PR do ?

This PR fixes Megatron-FSDP issues found while validating Qwen3.5-35B-A3B models with TP/CP/EP enabled.

1. Qwen VL with CP requires `calculate_per_token_loss`
When running Qwen VL with `context_parallel_size > 1`, model construction failed with:

```bash
  assert self.config.calculate_per_token_loss, (
  AssertionError: Qwen3-VL model only supports context parallelism with calculate_per_token_loss enabled
```

2. Megatron-FSDP DTensor TP loading used the wrong output shape

 HF-to-Megatron-FSDP loading failed on language_model.output_layer.weight, which maps from HF lm_head.weight. It is handled through ColumnParallelMapping for LinearCrossEntropyModule.

```bash
  [HF->Megatron DTensor shape mismatch] ...
  param_name=language_model.output_layer.weight ...
  module_type=LinearCrossEntropyModule ... target_shape=(124160, 2048) ...
  converted_shape=(62080, 2048)
  RuntimeError: The size of tensor a (128647168) must match the size of tensor b (1507328)
  ```

3. Qwen3 MoE fused expert gate/up weights is wrong.

```bash
  bridge.load_hf_weights(megatron_model)
  ...
  param_mapping.py", line 2496, in hf_to_megatron
      expert_weight = _align_expert_weight_to_shape(expert_weight, gate_up_full_shape, "gate_up")
  ...
  ValueError: Unexpected gate_up shape (1024, 2048); expected (2048, 2048).
```

# Changelog
  - Set CP loss default into the Qwen VL provider finalize() methods:
      - if context_parallel_size > 1, set calculate_per_token_loss = True.
  - Fix DTensor TP scatter output shapes in parameter mappings:
      - use target_param.orig_param.shape for Megatron-FSDP DTensor params;
  - Fix fused gated expert DTensor shape handling so expert weights are aligned using the DTensor shape semantics.

# GitHub Actions CI

Local validation: Megatron-FSDP HF load/export roundtrip validation:

- Qwen3.5-35B-A3B:
  - `TP=2 CP=1 EP=4`, `TP=1 CP=1 EP=8`, `TP=2 CP=1 EP=1`,`TP=4 CP=1 EP=2`,`TP=2 CP=2 EP=2` passed
- Qwen2.5-7B:
  - `TP=2 CP=1 EP=1` passed
- Qwen3-30B-A3B:
  - `TP=2 CP=1 EP=4` passed

**Pre checks**:

- [] Did you write any new necessary tests?
  - No new unit test changes are included in this PR. Existing `test_param_mapping.py` coverage was rerun successfully, and the fix was validated with Megatron-FSDP HF load/export roundtrip jobs across dense and MoE Qwen models.
- [] Did you add or update any necessary documentation?
  - No documentation update is required; this fixes conversion behavior and updates conversion examples only.
- [] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - No new optional dependency or import path is introduced.

# Additional Information

- Related issue: N/A
- Branch head: `64bcc85b Fix Megatron-FSDP DTensor TP loading`
- Notes: `hf_to_megatron_fsdp_generate_text.py` still does not fully validate CP generation for Qwen3.5. Enabling `calculate_per_token_loss` gets past model construction, but the current generation path needs separate CP+MTP-aware handling. This PR keeps the CP-required config setting in the example and does not add a local generation workaround.


conver334:
/ok to test [64bcc85](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3746/commits/64bcc85b2b56bcde85f749415d09b2a83d918680)

examples/conversion/mfsdp/hf_fsdp_roundtrip.py
```diff
@@ -73,6 +73,8 @@ def _configure_model_provider(model_provider, tp: int, cp: int, ep: int) -> None
     model_provider.tensor_model_parallel_size = tp
     model_provider.context_parallel_size = cp
     model_provider.expert_model_parallel_size = ep
+    if cp > 1:
+        model_provider.calculate_per_token_loss = True
```
yaoyu-33: this should not be here, can be added in MBridge's TransformerConfigs finalize - but how does it needed here in conversion / generate?

src/megatron/bridge/models/conversion/param_mapping.py
```diff
@@ -914,10 +936,7 @@ def hf_to_megatron(
         else:
             splits = None
 
-        if isinstance(target_param, DTensor):
-            output_shape = [target_param.shape[0] // self.tp_size, *target_param.shape[1:]]
-        else:
-            output_shape = target_param.shape
+        output_shape = None if isinstance(target_param, DTensor) else target_param.shape
```
yaoyu-33: Reverting this hunk avoids one broadcast per column-parallel param. The old formula is correct under the same DTensor-shape semantic the new `FusedGatedExpertMapping` branch assumes.

```suggestion
        if isinstance(target_param, DTensor):
            output_shape = (target_param.shape[0] // self.tp_size, *target_param.shape[1:])
        else:
            output_shape = target_param.shape
```

src/megatron/bridge/models/conversion/param_mapping.py
```diff
@@ -1023,10 +1042,7 @@ def hf_to_megatron(
         else:
             splits = None
 
-        if isinstance(target_param, DTensor) and hf_weights.ndim != 1:
-            output_shape = [target_param.shape[0], target_param.shape[1] // self.tp_size, *target_param.shape[2:]]
-        else:
-            output_shape = target_param.shape
+        output_shape = None if isinstance(target_param, DTensor) else target_param.shape
```
yaoyu-33: Same as the `ColumnParallelMapping` comment — the prior DTensor branch is correct and avoids a per-param broadcast.

```suggestion
        if isinstance(target_param, DTensor) and hf_weights.ndim != 1:
            output_shape = (target_param.shape[0], target_param.shape[1] // self.tp_size, *target_param.shape[2:])
        else:
            output_shape = target_param.shape
```

yaoyu-33 (commented):
Thanks for tracking down the Qwen3 MoE fused-expert loading bug — the `FusedGatedExpertMapping` conditional is clearly the load-bearing fix. A design question on the broader change before approving:

**Cost of the broadcast**

`_broadcast_tp_split_shape` fires one `broadcast_object_list` per parallel parameter on every HF→Megatron load. For a large model that's thousands of small synchronous broadcasts on the TP group. Is this needed for the non-fused mappings?

Walking through each call site:

- `ColumnParallelMapping`, `RowParallelMapping`, `GatedMLPMapping`: the prior formulas (`target_param.shape[parallel_dim] // tp_size`) are the correct local TP shard *if* DTensor reports the unsharded logical shape — which is exactly the assumption the new `FusedGatedExpertMapping` branch relies on (`gate_full_shape = (target_shape[0] // 2, target_shape[1])`, no `* tp_size`). So either DTensor reports unsharded logical shapes (in which case the old formulas were fine and the broadcast is unnecessary), or it doesn't (in which case the new FusedGatedExpert branch is wrong). Both can't be true simultaneously.
- `FusedGatedExpertMapping`: the old code double-shifted by `tp` when `target_shape` was already `[2*I, H]`. The new conditional is exactly right and should stay.

**Suggestion**

1. Keep the `FusedGatedExpertMapping` conditional — that's the real fix.
2. Revert `ColumnParallelMapping`, `RowParallelMapping`, `GatedMLPMapping` to the previous DTensor branches.
3. Drop `_broadcast_tp_split_shape` and the `Optional[output_shape]` plumbing entirely.

If there's a specific non-fused param where `target_param.shape // tp_size` actually fails under Megatron-FSDP, could you share the placement spec and observed shape? That would either justify the broadcast or point to a cleaner fix (e.g., normalizing the logical shape at DTensor-wrap time).

If a broadcast is truly unavoidable, consider one batched `broadcast_object_list({name: shape, ...})` at the start of the HF→Megatron load instead of one-per-param.

**Unrelated change**

The `calculate_per_token_loss = True` when `cp > 1` in the two example scripts isn't part of the DTensor shape fix. Worth splitting out or at least calling out separately in the changelog so the bisect story stays clean.

examples/conversion/mfsdp/hf_fsdp_roundtrip.py
conver334: Thanks!! 
I moved it to qwen3_vl finalize.
I update the PR description with bug details.
Because Qwen3-VL model only supports context parallelism with calculate_per_token_loss enabled.

src/megatron/bridge/models/conversion/param_mapping.py
conver334: Thank you so much! Great suggestions! Made modifications according to your design.
1. Keep the FusedGatedExpertMapping conditional 
2. Revert ColumnParallelMapping, RowParallelMapping, GatedMLPMapping to the previous DTensor branches. The broadcast can be avoidable. Instead of using calculations to obtain the shape of the output, the unique attribute `target_param.orig_param.shape` of Dtensor is employed
3. Drop _broadcast_tp_split_shape and the Optional[output_shape] plumbing entirely.


About Unrelated change: You are right. I actually fixed three issues in this PR. I updated the PR description to explain them. They were all encountered during my benchmarking of Qwen3.5-35B-A3B, so I changed PR's title.

src/megatron/bridge/models/conversion/param_mapping.py
conver334: Done.

conver334:
/ok to test [970fce7](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3746/commits/970fce768fa8b41bfd99e4be743efaf884a1a989)

conver334:
/ok to test [1507f31](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3746/commits/1507f318bc8f13f4bed493b7445c2a0e8bb527e8)

yaoyu-33 (approved):

PR #3704, opened 2026-05-06
fix(build): restore rdma-core v60 + bump NVSHMEM 3.6.5 for multi-node B200 DeepEP

ko3n1g:
<details><summary>Claude summary</summary>

## What

Restore the build-time invariants the multi-node B200 DeepEP path needs after the pytorch:26.02 → 26.04 base bump (#3637) and the aws_ofi_builder removal (#3638). Two surgical changes to `docker/Dockerfile.fw_base`:

1. **Bring back rdma-core v60 user-space stack.** Add a thin `rdma_core_v60` Docker stage that installs just the rdma-core .debs bundled with the EFA installer 1.46.0 for UBUNTU2404 — no `/opt/amazon`, no aws-ofi-nccl rebuild, just the v60 user-space libraries DeepEP needs.
2. **Bump `nvidia-nvshmem-cu13` pip pin from 3.4.5 → 3.6.5** in the DeepEP install stage so the device bitcode and the system NVSHMEM host runtime are at the same version.

## Why

After #3637 + #3638 the failing pipeline 50394913 had three Container-axis regressions in the same window:

- `gpu-util-improvements` bucket — Container probe 50412231 returned `same-error` +13.38% (deliberate)
- `cuda-oom` on `nemotron_3_nano_8gpu_h100_fp8_cs_perf` — bisect MR !2300 confirmed Container axis (probe 1 `same-error`, probe 2 `healed`)
- `deepep-timeout` on `deepseek_v3_256gpu_b200_fp8_mx_v2_perf` and `qwen3_next_80b_a3b_b200_fp8_mx_50steps_perf` — `RuntimeError: DeepEP error: timeout (dispatch CPU)`, `moe_recv_expert_counter[*]: -1` (QP rendezvous never happens)

### rdma-core v60 → v59.1 downgrade

| Container | `ibverbs-providers` / `libibverbs1` | `rdma-core` |
|---|---|---|
| Anchor `nemofw-nightly.50083097` (pytorch:26.02 + aws_ofi_builder) | **60.0-1** | **60.0-1** |
| Failing `nemofw-nightly.50390086` (pytorch:26.04, no aws_ofi_builder) | 59.1-1 | 50.0-2ubuntu0.2 |
| Raw `pytorch:26.02-py3` base | 56.0-1 | — |
| Raw `pytorch:26.04-py3` base | 59.1-1 | — |

[...]

PR #3750, opened 2026-05-08
build: pin flashinfer-python + flashinfer-cubin in pyproject.toml

ko3n1g:

## Problem

After the `Dockerfile.fw_base` refactor in #3702 / #3704, the megatron-bridge image
ships with mismatched `flashinfer` and `flashinfer-cubin` packages. At import time:

```
RuntimeError: flashinfer-cubin (0.6.8.post1) != flashinfer (0.6.10.post1) version mismatch
```

`flashinfer-cubin 0.6.8.post1` is preinstalled in the `pytorch:26.04-py3` base image.
`uv sync --all-extras --all-groups` in `Dockerfile.ci` later upgrades `flashinfer` to
`0.6.10.post1` via `flashinfer-python`'s transitive deps, but does not touch the
preinstalled cubin — so they drift apart and fail the version-match check on import.

This crashed every B200/H100/GB200 test case that imports flashinfer at startup,
including `nemotronh_4b_8gpu_b200` (job `dl/joc/nemo-ci/-/jobs/313723593`).

## Fix

Add a single RUN step in `Dockerfile.ci` after `uv sync` that pins `flashinfer-cubin`
to whatever version `flashinfer` resolved to. Defensive — no-op if `flashinfer` is
not installed:

```dockerfile
RUN FLASHINFER_VER=$(uv pip show flashinfer 2>/dev/null | awk '/^Version:/{print $2}') && \
    if [ -n "$FLASHINFER_VER" ]; then \
        uv pip install --no-deps "flashinfer-cubin==${FLASHINFER_VER}"; \
    fi
```

`--no-deps` keeps the rest of the locked environment untouched.

## Test plan

- [x] Local pull + import test on `pipe.50656285-x86`: `flashinfer-python 0.6.8.post1`
      installed as a single bundled package; `import flashinfer` succeeds; the
      import-time version-skew check is now satisfied by construction.
- [x] Targeted re-run of the originally failing test: `nemotronh_4b_8gpu_b200`
      passes (verification pipeline `dl/joc/nemo-ci/-/pipelines/50656285`,
      job `dl/joc/nemo-ci/-/jobs/313907344`).
- [x] Full L0 suite re-run on x86 clusters using the fixed image: previously-broken
      `nemotronh_*`, `llama3_8b_*`, etc. all pass on `eos` / `prenyx`
      (`dl/joc/nemo-ci/-/pipelines/50661890`). Remaining failures are unrelated
      pre-existing container packaging issues (missing `deep_ep` / `triton_kernels`)
      that will be tracked separately.

</details>


pyproject.toml
```diff
@@ -100,6 +100,8 @@ dependencies = [
     "mlflow>=3.9.0",    # To address CVE-2025-15031
     "comet-ml>=3.50.0",
     "torch>=2.6.0",
+    "flashinfer-python==0.6.10.post1",  # keep flashinfer-python and flashinfer-cubin in lockstep:
+    "flashinfer-cubin==0.6.10.post1",   # the runtime version-match check fails otherwise
```
chtruong814: Isn't the current version in the container 0.6.8.post1?  I just pinned this in Export-Deploy.

https://github.com/NVIDIA-NeMo/Export-Deploy/pull/679

What shoudl we use?

ko3n1g: lets use what you used in ED

chtruong814 (approved):

PR #3752, opened 2026-05-08
Remove qwen3_asr import because of noisy docstring errors

filaretov:
# What does this PR do ?

Remove an import that causes noisy and misleading `[ERROR]` messages.


filaretov:
Closing in favor of #3763

PR #3757, opened 2026-05-08
expandable_segments:True for GB200 405B NVFP4 256x gpu scale.

rsalagame-nvidia:
# What does this PR do ?

expandable_segments:True for GB200 405B NVFP4 256x gpu scale.


bdubauski (approved):
LGTM

PR #3778, opened 2026-05-11
[model, ckpt] feat: Add on-the-fly dequantization utilities for HF checkpoints

cuichenx:
## Summary

This PR adds on-the-fly dequantization support for FP8 DeepSeek-V3 Hugging Face checkpoints during Bridge import. DeepSeek-V3 publishes FP8 safetensor weights with per-block `*_scale_inv` tensors; users can already follow the documented preprocessing path to create a BF16 checkpoint for Bridge. This change adds the convenience path where Bridge can consume the FP8 checkpoint directly and dequantize each imported tensor before the standard mapping code runs.

As requested in review, the model-specific quantization helpers have been consolidated into `megatron.bridge.models.conversion.quantization_utils` instead of adding another DeepSeek-local helper. The shared utility now covers the checkpoint quantization formats currently used by model bridges:

- FP8 blockwise dequantization with 128x128 `*_scale_inv` blocks, used by DeepSeek-V3 and MiniMax-M2.
- FP8 scalar/broadcast scale dequantization, used by Ministral3.
- MXFP4 dequantization, used by GPT-OSS.
- INT4 pack/unpack helpers, used by Kimi VL.

Existing model bridge behavior is preserved, with compatibility aliases kept for helper names that tests or external callers already referenced.

## Validation

- DSV3 parity run comparing Bridge output to Hugging Face output:
  - mean cosine: `0.999171078`
  - min cosine: `0.997706413`
  - max cosine: `0.999795198`
  - top1 agreement: `8/8`
- `uvx ruff check ...` on touched files.
- `uvx ruff format --check ...` on touched files.
- `python -m compileall ...` on touched files.
- CPU torch smoke test for FP8 blockwise, MXFP4, and INT4 utilities.

Full local pytest was not run because the workstation checkout has no torch/pytest/transformers environment and the repo `.venv` is root-owned/broken for `uv run`.

## Test plan

- Added `tests/unit_tests/models/test_quantization_utils.py` for the shared utility coverage:
  - FP8 blockwise dequantization with distinct per-block scales.
  - FP8 passthrough and fallback cast behavior.
  - FP8 scalar/broadcast scale dequantization.
  - MXFP4 low/high nibble decode order.
  - INT4 quantize/dequantize shape and dtype behavior.
- Existing DeepSeek and MiniMax tests keep coverage for their bridge-level `maybe_modify_loaded_hf_weight` paths.

## Labels

`area:model` · `area:ckpt` · `needs-review`


src/megatron/bridge/models/deepseek/deepseek_v3_bridge.py
```diff
@@ -36,6 +36,29 @@
     HAVE_TE = False
 
 
+_FP8_BLOCK_SIZE = 128
+
+
+def _dequant_fp8_blockwise(weight: torch.Tensor, scale_inv: torch.Tensor) -> torch.Tensor:
```
yaoyu-33: can we move all dequat/ quat utils to one place?

cuichenx: good idea, done

yaoyu-33 (approved):

PR #3804, opened 2026-05-13
Add option to disable MLFlow artifact logging

nathan-az:
# What does this PR do ?

Unlike other loggers, MLFlow's artifact logging uploads over http rather than creating a link / storing a path, which is outrageously slow for large checkpoints. Some users may still want this, but this gives the option to disable it (in my case for example, checkpoints are on lustre/nvme and sync'd in the background as-needed to avoid blocking the training process).

# Changelog

- Adds an arg to toggle MLFlow artifact logging which is the only action in the finalize currently.

### DIFF
--- a/src/megatron/bridge/training/checkpointing.py
+++ b/src/megatron/bridge/training/checkpointing.py
@@ -1215,6 +1215,8 @@ def save_checkpoint(
             )
 
         def mlflow_finalize_fn() -> None:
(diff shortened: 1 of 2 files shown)


yaoyu-33 (approved):

cuichenx:
tests passed, code coverage is missing just one line

PR #3825, opened 2026-05-14
[training, ckpt] feat: Save HF weights alongside native Megatron checkpoints

dancingpipi:
# What does this PR do?

Improves HuggingFace mixed-format checkpointing for Megatron Bridge training.

When `save_weight_format="hf"` is enabled, training still saves a complete native Megatron checkpoint under each `iter_*/` directory. HuggingFace-format weights are exported as an additional artifact under `iter_*/hf/`.

This keeps resume training simple and aligned with the normal Megatron checkpoint flow, while still producing HF-format model or adapter artifacts for downstream use.

Important behavior:

- `checkpoint.load` is for native Megatron checkpoints only.
- Full-model HuggingFace directories are supported for initialization through `checkpoint.pretrained_checkpoint`, not for resume through `checkpoint.load`.
- `iter_*/hf/` is an export artifact and is ignored by parent-directory resume.
- PEFT HF adapter exports are output artifacts, not Megatron PEFT resume checkpoints.

HF export reuses the existing Bridge conversion path. Even when Megatron async checkpointing is enabled, the HF export itself is synchronous because it may participate in distributed collectives.

# Changelog

* Checkpointing (`checkpointing.py`)
    - `save_weight_format="hf"` now preserves the complete native Megatron checkpoint at `iter_*/`, including model weights, optimizer state, scheduler state, RNG state, rerun state, and metadata.
    - HF weights are exported as an extra artifact under `iter_*/hf/`.
    - Parent save-dir resume follows the native Megatron checkpoint path and ignores the extra `iter_*/hf/` export.
    - `checkpoint.load` now rejects HuggingFace-format directories. Resume training should use a complete native Megatron checkpoint.
    - Full-model HF initialization remains supported through `checkpoint.pretrained_checkpoint`.
    - HF initialization resets training progress to step 0 and does not load optimizer, scheduler, RNG, or training-state metadata.
    - PEFT HF export writes adapter artifacts under `iter_*/hf/` using the Bridge adapter export path.
    - HF adapter-only directories are treated as export artifacts, not Megatron PEFT resume checkpoints.
    - When Megatron async save is enabled, the Megatron checkpoint save is scheduled before the synchronous HF export so checkpoint I/O can overlap with HF conversion where possible.
    - Added INFO-level logging around full HF model export and PEFT adapter export.

* Config (`config.py`)
    - Documents `save_weight_format="hf"` as an additional HF export, not a replacement for native Megatron checkpointing.
    - Documents HF source resolution order.
    - Adds validation for unsupported `save_weight_format="hf"` combinations.

* Setup (`setup.py`)
    - Clarifies checkpoint loading semantics:
        - `checkpoint.load` resumes from native Megatron checkpoints.
        - `checkpoint.pretrained_checkpoint` may point at a full HF model directory for initialization.

* Checkpoint utils (`checkpoint_utils.py`)
    - Adds detection for HuggingFace PEFT adapter-only directories.
    - Keeps full HF model directories and adapter-only directories distinguishable for checkpoint resolution.

* Tests
    - Adds coverage that HF export does not replace or rewrite the native Megatron checkpoint.
    - Adds coverage that parent save-dir resume ignores `iter_*/hf/`.
    - Adds coverage that `checkpoint.load` rejects HF-format directories.
    - Adds coverage that Megatron async save is scheduled before HF export.
    - Adds coverage for full-HF initialization through the HF load helper.
    - Adds coverage that HF adapter-only directories are not accepted as Megatron PEFT resume checkpoints.
    - Adds coverage for HF adapter-only directory detection.
    - Adds config validation coverage for `save_weight_format`.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci) in the Contributing doc for how to trigger CI. An NVIDIA developer will need to approve and trigger CI for external contributors.

**Pre checks**:

- [x] Reviewer: Does the PR have correct import guards for all optional libraries?

# Additional Information

- Related to # (issue)


dancingpipi:
@cuichenx This PR is ready to review.

src/megatron/bridge/training/config.py
```diff
@@ -653,6 +653,45 @@ class CheckpointConfig(MTrainCheckpointConfig):
     dist_ckpt_workers: int = 1
     """Specifies the number of distributed checkpoint workers for asynchronous saving."""
 
+    save_weight_format: Literal["megatron", "hf"] = "megatron"
```
cuichenx: since "hf" saves both megatron and hf, I think it's more intuitive to make this a boolean called somethng like "also_save_hf_ckpt"

src/megatron/bridge/training/utils/checkpoint_utils.py
```diff
@@ -160,6 +160,75 @@ def get_checkpoint_tracker_filename(checkpoints_path: str) -> str:
...
+
+    Args:
+        path: Filesystem path to check (may be ``None``).
+
+    Returns:
+        True when ``path`` looks like a HuggingFace model/checkpoint directory.
+    """
+    if path is None:
+        return False
+    if not file_exists(join_paths(path, "config.json")):
```
yaoyu-33, line 230: this seems too hacky?

cuichenx (commented):
Thanks for putting this together. Can I ask what is your intended use case of saving HF checkpoints during training? Did you consider an offline conversion job that runs asynchronously in the background as new checkpoints are generated? 

`_save_hf_weights` runs synchronously on every rank, on the critical path between dist_checkpointing.save and schedule_async_save. async_save only covers the Megatron finalize — the HF export blocks training (collectives + barriers in bridge.export_hf_weights). For a 70B model this is minutes per save. The comment "scheduled before HF export so I/O can overlap" reads as overlap, but the
   code looks inline-blocking — which is intended?

src/megatron/bridge/training/utils/checkpoint_utils.py
yaoyu-33: please check @cuichenx 's  comment about use case first.

dancingpipi:
> Thanks for putting this together. Can I ask what is your intended use case of saving HF checkpoints during training? Did you consider an offline conversion job that runs asynchronously in the background as new checkpoints are generated?
> 
> `_save_hf_weights` runs synchronously on every rank, on the critical path between dist_checkpointing.save and schedule_async_save. async_save only covers the Megatron finalize — the HF export blocks training (collectives + barriers in bridge.export_hf_weights). For a 70B model this is minutes per save. The comment "scheduled before HF export so I/O can overlap" reads as overlap, but the code looks inline-blocking — which is intended?

@cuichenx Thanks for your review!
Your reading of the code is correct: `_save_hf_weights` is synchronous and on the training critical path. `async_save=True` only schedules/finalizes the native Megatron checkpoint asynchronously. The HF export still blocks because `bridge.export_hf_weights()` participates in distributed collectives and barriers, so moving it to a background thread from the training process would be unsafe unless we redesign the export flow around a separate process/job.
The intended use case is optional “sidecar” HF export during training for workflows that need periodic HF artifacts immediately after a training save, e.g. quick downstream eval/inference, PEFT adapter publishing, or environments where users do not want to run a separate conversion job after training. The native Megatron checkpoint remains the source of truth for resume.

src/megatron/bridge/training/config.py
dancingpipi: I’ll rename this to a boolean such as `also_save_hf_ckpt` / `also_save_hf_checkpoint` and update the related validation, tests, docs, and log messages accordingly. That should make the user-facing config match the actual behavior more directly.

src/megatron/bridge/training/utils/checkpoint_utils.py
dancingpipi: > please check [@cuichenx](https://github.com/cuichenx) 's comment about use case first.

The intent here is only a lightweight local-path check, not a full HF compatibility validator.

src/megatron/bridge/training/checkpointing.py
```diff
@@ -769,6 +777,235 @@ def create_checkpoint_manager(checkpoint_config: CheckpointConfig) -> Checkpoint
...
+        hf_id = getattr(model_cfg, "hf_model_id", None)
+        if hf_id:
+            return hf_id
+    tokenizer_cfg = getattr(cfg, "tokenizer", None)
+    if tokenizer_cfg is not None and getattr(tokenizer_cfg, "tokenizer_model", None):
+        return tokenizer_cfg.tokenizer_model
+    return None
+
+
+_AUTO_BRIDGE_CACHE: dict[tuple[str, bool], Any] = {}
```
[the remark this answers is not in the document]
dancingpipi: The key includes `hf_source_path`, so switching to a different source path will not reuse the wrong bridge. But I agree that module-level cache state should be resettable for tests and long-running multi-stage processes, especially if the same local path is reused with different contents. I’ll add `_clear_auto_bridge_cache()` and use it in tests where needed.

cuichenx:
> @cuichenx Thanks for your review! Your reading of the code is correct: `_save_hf_weights` is synchronous and on the training critical path. `async_save=True` only schedules/finalizes the native Megatron checkpoint asynchronously. The HF export still blocks because `bridge.export_hf_weights()` participates in distributed collectives and barriers, so moving it to a background thread from the training process would be unsafe unless we redesign the export flow around a separate process/job. The intended use case is optional “sidecar” HF export during training for workflows that need periodic HF artifacts immediately after a training save, e.g. quick downstream eval/inference, PEFT adapter publishing, or environments where users do not want to run a separate conversion job after training. The native Megatron checkpoint remains the source of truth for resume.

Got it. If you can put down a warning that this feature would slow down checkpoint saving considerably and should only be used for small models, that would be great. For larger models, the recommended path is to run a background async job that greps for new checkpoints.

src/megatron/bridge/training/checkpointing.py
```diff
@@ -969,10 +1209,19 @@ def save_checkpoint(
                         use_megatron_fsdp=cfg.ddp.use_megatron_fsdp,
                     )
 
-    # Apply PEFT filtering to save adapter-only checkpoints
-    if cfg.peft is not None:
+    # Apply PEFT filtering to save adapter-only checkpoints.  In HF sidecar export mode,
+    # keep the Megatron checkpoint complete and write the PEFT adapter as the
+    # extra HF artifact under ``iter_*/hf/``.
+    if cfg.peft is not None and not ckpt_cfg.also_save_hf_checkpoint:
         state_dict = apply_peft_adapter_filter_to_state_dict(state_dict, cfg.peft)
```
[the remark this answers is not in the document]
cuichenx: @dancingpipi please check this, verify PEFT is not broken

cuichenx:
can you make sure `also_save_hf_checkpoint=True` works with PEFT and async save (or prevent such combinations in the config)

dancingpipi:
> > @cuichenx Thanks for your review! Your reading of the code is correct: `_save_hf_weights` is synchronous and on the training critical path. `async_save=True` only schedules/finalizes the native Megatron checkpoint asynchronously. The HF export still blocks because `bridge.export_hf_weights()` participates in distributed collectives and barriers, so moving it to a background thread from the training process would be unsafe unless we redesign the export flow around a separate process/job. The intended use case is optional “sidecar” HF export during training for workflows that need periodic HF artifacts immediately after a training save, e.g. quick downstream eval/inference, PEFT adapter publishing, or environments where users do not want to run a separate conversion job after training. The native Megatron checkpoint remains the source of truth for resume.
> 
> Got it. If you can put down a warning that this feature would slow down checkpoint saving considerably and should only be used for small models, that would be great. For larger models, the recommended path is to run a background async job that greps for new checkpoints.

Added a config warning and a rank-0 runtime warning before full-model HF export.

dancingpipi:
> can you make sure `also_save_hf_checkpoint=True` works with PEFT and async save (or prevent such combinations in the config)

For PEFT training, I reverted the behavior so `apply_peft_adapter_filter_to_state_dict()` is called regardless of whether `also_save_hf_checkpoint` is enabled. This keeps the native Megatron checkpoint behavior consistent with the original PEFT save path. I also added a unit test covering `PEFT + also_save_hf_checkpoint=True + async_save=True`.

tests/unit_tests/training/utils/test_checkpoint_utils.py
```diff
@@ -815,6 +816,18 @@ def test_is_iteration_dir_nonexistent(self):
...
+        """A raw HF model directory should not be treated as a Megatron iteration directory."""
+        hf_dir = tmp_path / "hf_model"
+        hf_dir.mkdir()
+        (hf_dir / "config.json").touch()
+        (hf_dir / "model.safetensors").touch()
+
+        assert is_hf_checkpoint_dir(str(hf_dir)) is True
+        assert is_checkpoint_iteration_directory(str(hf_dir)) is False
+        assert checkpoint_exists(str(hf_dir)) is False
+        assert checkpoint_exists(str(hf_dir)) or is_hf_checkpoint_dir(str(hf_dir))
```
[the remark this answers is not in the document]
dancingpipi: Added the negative test.

src/megatron/bridge/training/checkpointing.py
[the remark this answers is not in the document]
dancingpipi: For PEFT training, I reverted the behavior so apply_peft_adapter_filter_to_state_dict() is called regardless of whether also_save_hf_checkpoint is enabled. This keeps the native Megatron checkpoint behavior consistent with the original PEFT save path. I also added a unit test covering PEFT + also_save_hf_checkpoint=True + async_save=True.

src/megatron/bridge/training/checkpointing.py
```diff
@@ -769,6 +777,244 @@ def create_checkpoint_manager(checkpoint_config: CheckpointConfig) -> Checkpoint
...
+    from megatron.bridge.models.conversion.auto_bridge import AutoBridge
+
+    source = hf_source or _resolve_hf_source(cfg)
+    if not source:
+        raise ValueError(
+            "also_save_hf_checkpoint=True requires an HF source to template config/tokenizer files. "
+            "Set cfg.checkpoint.hf_source_path, cfg.model.hf_model_id "
+            "(via AutoBridge / recipe metadata), or cfg.tokenizer.tokenizer_model when it points at "
+            "an HF model id."
+        )
```
cuichenx, line 835: can we move this check to finalize() so that this error is caught before any training step runs

cuichenx (commented):
One minor comment. Also, could you add a functional test for this feature (maybe in test_pretrain_resume.py?)

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3825/commits/97e2303ad045dcd7c170a8e40a072ed69ccfce19

src/megatron/bridge/training/checkpointing.py
dancingpipi: Added an early config validation for this.

cuichenx (dismissed):
LGTM

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3825/commits/5d2d456de53fddd01c76f878700689c311192b9e

dancingpipi:
@cuichenx  CI exposed a test-only issue where broad filesystem mocks caused native Megatron checkpoint load tests to be misclassified as HF directories; I’ve fixed the affected tests by explicitly mocking is_hf_checkpoint_dir=False. Could you please take another quick look?

cuichenx (approved):

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3825/commits/d247265c4147930cb20896c0dfc6fff1196bea2b

PR #3827, opened 2026-05-14
remove unused nemotron_h model providers

liding-nv:
Removed `nemotron_h_provider.py` that is no longer used


cuichenx (approved):

cuichenx (changes requested):
Never mind, i was reading stale code

cuichenx (approved):

PR #3829, opened 2026-05-14
Upgrade transformers to 5.8.1

cuichenx:
## Summary
- Pin `transformers` to `5.8.1` in `pyproject.toml`
- Regenerate the `uv.lock` entry for `transformers` 5.8.1
- This bump is needed for DeepSeek v4 support
- Update Transformers 5.8.1-sensitive conversion test fixtures for GLM-4.5V, Kimi K2.5 VL, and NemotronH toy checkpoints

## Why The Test Fixtures Changed
- `test_glm_45v_conversion.py`: the test only needs tokenizer artifacts that can be loaded and re-saved during conversion. Transformers 5.8.1 is stricter about tokenizer backends, so the fixture now writes a deterministic minimal fast tokenizer instead of depending on optional `tiktoken`/`sentencepiece` support for the full GLM tokenizer.
- `test_kimi_k25_vl_conversion.py`: the local toy checkpoint is loaded with `trust_remote_code=True`. Under Transformers 5.8.1, local dynamic-module loading needs the fixture to provide every custom Python module used by the checkpoint; the Kimi modeling file also has a transitive relative import, so the fixture makes that dependency visible from the top-level modeling file before loading from the local checkpoint. It also saves a minimal fast tokenizer for the same tokenizer-backend reason as GLM.
- `test_nemotron_h_conversion.py`: the NemotronH local toy checkpoints also use `trust_remote_code=True`. The fixture now copies all required custom modeling/configuration modules from the HF repo, and resets the temporary `transformers_modules` cache between tests so one local toy checkpoint cannot affect the next dynamic-module import.

## Validation
- `uv lock --check`
- `git diff --check`
- `python3 -m py_compile` on the changed test files
- `tests/functional_tests/test_groups/models/glm_vl` -> `4 passed, 1 skipped`
- `tests/functional_tests/test_groups/models/kimi_vl` -> `2 passed`
- NemotronH Nano + Super toy model creation tests -> `2 passed`


yaoyu-33 (dismissed):

cuichenx:
tests passed https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/25887611922/job/76101928024?pr=3829

cuichenx:
waiting for https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3833 to merge first

chtruong814 (dismissed):

PR #3830, opened 2026-05-14
[Cherry pick] fix: Handle bracketed entries inside mixed Slurm nodelists

sudostock:
Cherry Pick https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3807 for upcoming llmb release.

_parse_slurm_nodelist took the bracket branch whenever "[" appeared anywhere in the nodelist, then split on the first "[" — so a nodelist like "nodeA,nodeB[001,003],nodeC" produced "nodeA,nodeB001" instead of "nodeA".


bdubauski (approved):
LGTM

PR #3323, opened 2026-04-14
[main] qwen-vl THD packed-sequence support and fixes

DAISY-gh:
## Summary
- Add Qwen-VL THD packed-sequence support path and related runtime/config fixes in Megatron-Bridge.
- Update Qwen-VL recipe/task-encoder/vlm-step flow for THD batch assembly and MRoPE/padding-mask handling.
- Megatron-LM dependency context: this work is aligned with upstream Open PR https://github.com/NVIDIA/Megatron-LM/pull/2645.
image
curves of SFT on cord_v2 dataset based on Qwen3.5 35BA3B.

## Release Notes

* **New Features**
  * Added batch-level sequence packing support for improved training efficiency in data processing pipelines.
  * Introduced new Qwen3.5-VL training recipes with Energon dataset integration and configurable packing options.

* **Improvements**
  * Enhanced robustness by making optional dependencies (modelopt, diffusion modules) gracefully degrade when unavailable.
  * Improved multi-modal model training with better position computation and token accounting for mixture-of-experts layers.
  * Extended data provider configuration with new shuffling and packing buffer parameters for flexible batch preparation.

PR #2645, opened 2026-03-04
[DFM migration] Wan features verifications

huvunvidia:
…o partial convergence test run

# What does this PR do ?

Add a one line overview of what this PR aims to accomplish.

# Changelog

- Add specific line by line info of high level changes in this PR.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci)in the Contributing doc for how to trigger the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

# Before your PR is "Ready for review"

**Pre checks**:

- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?

If you haven't finished some of the above items you can still open "Draft" PR.

# Additional Information

- Related to # (issue)

PR #3838, opened 2026-05-15
[model, perf] feat: real THD packing in qwen3_vl_step

cuichenx:
## Summary

Make ``dataset.pack_sequences_in_batch=True`` actually skip pad-attention FLOPs on the
``qwen3_vl_step`` path (HF datasets such as cord_v2). The current implementation only
relayouts BSHD into a flat row and emits ``cu_seqlens = [0, T, 2T, …, BT]`` — every
segment looks fully real to flash-attn / GDN, so attention still computes the full
padded tile per row.

This PR rewires the packing branch in ``forward_step`` to:

1. Capture real per-sample lengths from the dataset's 2D ``(B, T)`` attention mask
   *before* ``pad_or_truncate_attn_to_len`` strips it (that utility only accepts 4D
   Megatron-style masks).
2. Concatenate per-sample real content into a single ``(1, total_padded)`` THD row.
3. Emit ``cu_seqlens_q/kv`` (real boundaries) distinct from ``cu_seqlens_q/kv_padded``
   (kernel stride into the flat tensor) — this is what lets the attention kernel skip
   the pad tail inside each segment.
4. Pass ``rope_cu_seqlens`` and ``moe_padding_mask`` so the model's per-sub-seq MRoPE
   resets at segment boundaries and the MoE router doesn't count align-pad as real.

Two new helpers, no changes to public APIs or other models.

### Empirical results

Qwen3.5-VL 2B SFT on cord_v2, 1 node × 8 H100, seq_length=2048, mcore main @ d167123d
(includes NVIDIA/Megatron-LM#2645 GDN packed-seq support).

| config                                | PACK=0 (BSHD) | PACK=1 (THD)  |
|---------------------------------------|---------------|---------------|
| MBS=2 GBS=16, force-pad to seq_length | 363 ms / iter | 342 ms / iter |
| MBS=4 GBS=32, force-pad to seq_length | **OOM**       | 505 ms / iter |

Loss curves match within bf16 noise. The bigger win is that THD enables micro-batch
sizes that BSHD cannot fit under PP>1 / EP>1 force-padding — at MBS=4 we get +44%
samples/sec over the best BSHD-fit baseline.

### Dependencies — DO NOT MERGE BEFORE

* NVIDIA-NeMo/Megatron-Bridge#3323 — adds the ``rope_cu_seqlens`` / ``moe_padding_mask``
  kwargs to ``Qwen3VLModel.forward`` and the matching plumbing through
  ``text_model.py`` / ``transformer_block.py``. Without it those two kwargs would be
  swallowed by ``**kwargs`` and MRoPE / MoE-router would silently lose per-segment
  semantics.
* NVIDIA/Megatron-LM#2645 — adds packed-seq support to ``gated_delta_net``. Qwen3.5-VL
  is a hybrid attention/GDN architecture; before #2645 GDN explicitly raised
  ``NotImplementedError(\"GDN does not support packed sequence for now.\")``.
* A subsequent recipe change (e.g. NVIDIA-NeMo/Megatron-Bridge#3481) is needed for
  users to opt in via ``pack_sequences_in_batch=True``. This PR is a no-op when that
  flag is left at its default ``False``.

### Usage

To exercise the packing path on HF VLM datasets the user must also flip
``dataset.skip_getting_attention_mask_from_dataset=False`` so the real attention mask
reaches ``forward_step``. The recipe-side default for that flag will be addressed
separately. A fallback path that infers real lengths from the trailing zero-pad is
provided for completeness, but only matches in the (uncommon) case where the
collator's ``tokenizer.pad_token_id`` is 0.

## Test plan

- [ ] L0 CI green
- [ ] Reviewer to run Qwen3.5-VL 2B SFT with ``dataset.pack_sequences_in_batch=True`` +
      ``dataset.skip_getting_attention_mask_from_dataset=False`` (requires #3323
      merged for full correctness).

### DIFF
--- a/src/megatron/bridge/models/qwen_vl/qwen3_vl_step.py
+++ b/src/megatron/bridge/models/qwen_vl/qwen3_vl_step.py
@@ -163,9 +163,11 @@ def pack_or_pad_batch_sequences(
     seq_length: int = None,
 ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, PackedSeqParams]:
     """
-    Pad or truncate the batch sequences to the target length, and build packed sequences.
-    If is_qwen3vl, return bshd tokens for be compatible with qwen3vl model.
-    Otherwise, return thd tokens and packed sequences.
+    Right-pad BSHD sequences to a common length and build a bookkeeping PackedSeqParams.
+
+    The returned ``packed_seq_params`` carries uniform `target_len` per segment and is only used
+    as a fallback. Real-length-aware packing (the path that delivers pad-skip FLOP savings) is
+    handled separately by ``_pack_bshd_to_thd`` in ``forward_step``.
     """
 
     batch_size, cur_len = tokens.shape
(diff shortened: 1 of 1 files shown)


liangxuZhang:
Great work, may I ask why the original solution used the BSHD layout in Qwen3 VL training. Is there any difference in the Qwen3 VL model?

cuichenx:
> Great work, may I ask why the original solution used the BSHD layout in Qwen3 VL training. Is there any difference in the Qwen3 VL model?

qwen3 vl step was created for VERL to use. VERL required the input to the forward pass to be in BSHD format so the packing was done inside the forward pass

zhongbozhu:
Let's also figure out if HybridEP can be used after removing the padding.

src/megatron/bridge/models/qwen_vl/qwen3_vl_step.py
```diff
@@ -207,6 +209,111 @@ def pack_or_pad_batch_sequences(
...
+        # Megatron causal masks are (B|1, 1, T, T) bool with True=masked. The diagonal row tells
+        # us which key positions are "real": True on the diagonal means a fully masked-out row.
+        keep_row = ~attention_mask[:, 0, :, 0].bool()
+        if keep_row.size(0) == 1:
+            keep_row = keep_row.expand(attention_mask.size(0) if attention_mask.size(0) > 1 else 1, -1)
+        return keep_row.to(torch.int32).sum(dim=-1)
+    raise ValueError(f"Unsupported attention_mask rank: {attention_mask.dim()}")
+
+
+def _pack_bshd_to_thd(
```
zhongbozhu, line 230: maybe we should make this optimization optional? since it affects the feasibility of HybridEP for moe models.

zhongbozhu:
Will CP>1 be supported with this PR, looks like it's gonna have some impact on CP.

cuichenx:
Closing as superseded by the newer collator-owned/in-batch packing architecture and native Energon Qwen-VL packing in #5285. The old step-owned implementation also retains an unresolved HybridEP interaction.

PR #3927, opened 2026-05-21
[data] fix: derive assistant loss mask for chat templates lacking {% generation %} (#2598)

### LINKED ISSUE #2598: Hardcoded GENERATION_REGEX logic (with return_assistant_tokens_mask) causes incorret masks production for chat examples in _chat_preprocess function
**Describe the bug**

The function [_chat_preprocess](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/383b610dca1fe553346b37fc0172f54917b02069/src/megatron/bridge/data/datasets/utils.py#L881) is used to apply chat template to the hf chat examples.

The examples are in the format of messages. To correctly produce the mask and train only on assistant messages, the user/system/tool messages need to be masked (mask=0). The function logic is:
```
template_has_generation_kwd = GENERATION_REGEX.search(tokenizer.chat_template) is not None

tokenized_chat = tokenizer.apply_chat_template(
      chat,
      tools=tools,
      tokenize=True,
      return_dict=True,
      return_assistant_tokens_mask=template_has_generation_kwd,
  )

input_ids = tokenized_chat.get("input_ids")
if template_has_generation_kwd:
    mask = tokenized_chat["assistant_masks"]
else:
    mask = [1] * len(input_ids)
```

The issue here is that not all tokenizers has `{% generation %}` keyword, for example Nemotron 3 Nano model doesn't, which means we will get default `attention_mask` (all 1's). Such mask is not the best for SFT, cause it forces the model to be trained on every user/system/tool message.

As far as I understand this issue is only for hf chats, cause for not hf chats there is `_preprocess` function from the same file, that uses `_mask_targets` function to correctly mask messages.

**Expected behavior**

Maybe remove the ` return_assistant_tokens_mask` parameter, it doesn't help with the modern chat templates and apply own logic of masking.

lonexreb:
## Summary

Restores a working SFT path for tokenizers whose chat_template does not
mark assistant content with `{% generation %}...{% endgeneration %}`
(e.g. **Nemotron-3-Nano**, several Llama variants).

Closes #2598.

## Context

`apply_chat_template(return_assistant_tokens_mask=True)` only produces a
useful mask when the template wraps assistant content in `{% generation %}`
blocks. Without them, the HF API silently returns an all-ones mask — SFT
would then learn on system, user, and tool tokens, which is wrong.

Prior behavior: `_chat_preprocess` raised a hard `ValueError` for these
tokenizers (correct but leaves users blocked).

This PR: derives the assistant-only mask via incremental prefix rendering,
so affected tokenizers train correctly with no template changes required.

## Approach

For each assistant message at index `i`:

1. Render `chat[:i]` with `add_generation_prompt=True` — ends with the
   assistant-header tokens (e.g. `<|im_start|>assistant\n`).
2. Render `chat[:i + 1]` — same prefix plus assistant content + closing
   tokens (e.g. `<|im_end|>`).
3. Mark `mask[len(prefix) : len(with_assistant)] = 1`.

Role-header tokens stay in the prefix (correctly masked out); the closing
turn token is included in the loss, matching the convention HF uses for
`{% generation %}`-aware templates.

Templates that are not incrementally renderable (e.g. running-summary
templates) cannot be handled this way — the helper raises a clear
`ValueError` instead of producing a wrong mask, so the caller can still
fix the template upstream.

## Files changed

- `src/megatron/bridge/data/datasets/utils.py`
  - New helper: `_derive_assistant_mask_via_prefix_diff(chat, tokenizer, tools, full_ids) -> list[int]`.
  - `_chat_preprocess` branches on `template_has_generation_kwd`: fast
    path uses the native HF mask; fallback uses the new helper.
- `tests/unit_tests/data/datasets/test_chat_template.py`
  - New `TestDeriveAssistantMaskViaPrefixDiff` class — 7 tests:
    - single-turn marks only the assistant span
    - multi-turn marks each assistant span independently
    - system prompt is never marked
    - tool message between assistant turns is not marked
    - non-incremental template raises `ValueError`
    - no assistant messages → all-zero mask
    - tool schemas are forwarded to `apply_chat_template`
  - `test_chat_preprocess_without_generation_keyword_uses_prefix_diff_fallback`
    replaces the prior ValueError-only test with an integration check on
    `input_ids` / `loss_mask` / `context_ids` / `answer_ids`.

## Test plan

- [x] New unit tests pass under hand-traced execution of the helper.
- [x] Existing `{% generation %}` fast-path tests untouched (no behavior
      change on that branch).
- [x] Syntax + py_compile on edited files.
- [ ] CI: please run on NVIDIA runners (`/ok to test <commit-SHA>`).

### DIFF
--- a/src/megatron/bridge/data/datasets/utils.py
+++ b/src/megatron/bridge/data/datasets/utils.py
@@ -879,6 +879,101 @@ def _convert_to_openai_messages(source: dict) -> list[dict]:
     return chat
 
 
+def _derive_assistant_mask_via_prefix_diff(
+    chat: list[dict],
+    tokenizer: Any,
+    tools: Optional[list[Any]],
+    full_ids: list[int],
+) -> list[int]:
+    """Derive an assistant-only loss mask without relying on `{% generation %}` markers.
+
+    When a tokenizer's chat_template lacks `{% generation %}...{% endgeneration %}` blocks,
+    `apply_chat_template(..., return_assistant_tokens_mask=True)` silently returns an all-ones
+    mask, which would cause SFT loss to be computed on every token (system / user / tool /
+    assistant). This helper instead reconstructs the mask by re-rendering message prefixes:
+
+      1. For each assistant message at index ``i``, render ``chat[:i]`` with
+         ``add_generation_prompt=True`` — this ends with the assistant-header tokens
+         (e.g. ``<|im_start|>assistant\\n``).
+      2. Render ``chat[:i + 1]`` — this is the same prefix plus the assistant content
+         and any closing tokens (e.g. ``<|im_end|>``).
+      3. Mark ``mask[len(prefix) : len(with_assistant)]`` as ``1``. Role-header tokens
+         remain in the prefix and are therefore masked out; the closing turn token is
+         included in the loss, matching the convention used by templates that ship
+         ``{% generation %}`` blocks.
+
+    The function validates that each incremental rendering extends the previous one and
+    matches the full conversation rendering. Templates that fail this property (e.g. ones
+    that emit running summaries) are not safely incrementally renderable, and a
+    ``ValueError`` is raised so the caller can fix the template instead of training on a
+    wrong mask.
+
+    Args:
+        chat: Conversation in OpenAI ``messages`` format
+            (``[{"role": ..., "content": ...}, ...]``).
+        tokenizer: A HuggingFace ``PreTrainedTokenizerBase`` (or compatible) exposing
+            ``apply_chat_template``.
+        tools: Optional tool schema list forwarded to ``apply_chat_template``.
+        full_ids: Token IDs produced by rendering the entire ``chat``; used to size the
+            mask and to validate prefix consistency.
+
+    Returns:
+        A list of ``int`` (0 / 1) with the same length as ``full_ids``, where ``1`` marks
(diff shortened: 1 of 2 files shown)


lonexreb:
Friendly ping — this PR is still ready for review.

Quick recap for any reviewer picking it up cold:

- **Scope**: one new helper (`_derive_assistant_mask_via_prefix_diff`) + one branch in `_chat_preprocess`. Production diff is +95 lines in a single file.
- **Behavior change**: tokenizers whose chat_template lacks `{% generation %}` (e.g. Nemotron-3-Nano, several Llama variants) previously raised a hard `ValueError`. They now train correctly via prefix-diff masking. Templates with `{% generation %}` keep using the existing HF fast path — no behavior change there.
- **Failure mode preserved**: chat templates that are not incrementally renderable (e.g. running-summary templates) still raise a clear `ValueError` so the caller can patch upstream rather than silently train on a wrong mask.
- **Tests**: 7 new unit tests in `TestDeriveAssistantMaskViaPrefixDiff` plus an updated integration test. All `MagicMock`-based — no live HF Hub calls, no GPU, no CI cost beyond unit-tests-core.

Branch is `MERGEABLE` against current `main`; the files I touch haven't changed upstream since the branch was opened. Happy to rebase or split if it would speed review.

yaoyu-33:
will fix in another pr from us, wip.

PR #3979, opened 2026-05-26
[model] test: guard Qwen3-VL MTP CP/SP embedding locality

LoganVegnaSHOP:
## Summary

Current `main` changed the Qwen3-VL MTP data flow in #4532: dense and packed context-parallel input IDs are now localized before entering the text model. That supersedes this PR’s original inner CP-shard implementation; retaining it would shard MTP embeddings twice.

This update merges current `main`, preserves the text model’s SP-only embedding wrapper, and adds regression coverage for both sides of the corrected contract.

Tracking issue: #4328. Production behavior was superseded by #4532.

## Changes

- Add an outer-model dense-CP regression proving input IDs, MRoPE positions, labels, and loss masks use the same zigzag rank-local partition before the language/MTP model.
- Add a CP+SP text-model regression proving already-local MTP embeddings keep their sequence length and receive exactly one SP scatter through the explicit TP group.
- Preserve current-main packed-CP and padded-boundary behavior.

## Validation

- `10 passed`: complete Qwen3-VL text-model forward unit file, dense-CP producer handoff, packed-CP one-partition handoff, and full-vs-local split helper.
- Three independent semantic/test/comment review rounds completed; final round found no actionable code or test issues.

The historical `L1_Launch_recipes_qwen` failure did not reach recipe execution. One distributed test-data fixture rank received a GitHub release API `401 Bad credentials`; the other rank downloaded successfully. Exact-SHA CI is being rerun for this update.

## End-to-end context

The contributor’s earlier CP=4/SP=true MTP smoke completed 32 iterations with finite, decreasing LM and MTP losses and no skipped or NaN iterations. The final PR now guards the current-main implementation that provides this locality contract rather than restoring the superseded inner shard.


src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/text_model.py
```diff
@@ -175,18 +176,32 @@ def forward(
...
+        # MTP calls self.embedding directly (bypassing the manual CP shard + SP scatter
+        # that model.py applies to the combined VL embeddings before passing them into
+        # the decoder). When CP > 1 or SP is on, the MTP embedding output therefore has
+        # a different shape than hidden_states and downstream concatenation in
+        # MultiTokenPredictionLayer fails. Mirror model.py's pre-decoder shape
+        # transformations on this implicit MTP embedding path.
+        #
         # We write to self.__dict__ directly to bypass nn.Module.__setattr__'s type
         # check, which rejects non-Module values for registered child modules.
+        cp_size = self.pg_collection.cp.size()
```
yaoyu-33: can you use "get_pg_size" and "get_pg_rank" util from utils. it handles corner case better.

yaoyu-33:
@LoganVegnaSHOP  are you able to take a look at comment? also did you run any loss verifications?

LoganVegnaSHOP:
Update after additional ablations:

I opened a tracking issue for this specific MTP CP/SP embedding-ordering bug:

- #4328

I think this PR should stay open rather than be closed. It still appears to be the right upstream fix: apply the CP shard on the Qwen3-VL MTP embedding path before SP scatter so the MTP path mirrors the main language path.

Additional validation since opening the PR:

1. We corrected our text-only SFT data to use the Qwen3-VL / omni conversation dataset path with `qwen3_vl_step`. The earlier pure-text HF SFT collator path can independently produce dynamic-padding boundary failures when mixed with `qwen3_vl_step`, so the latest validation avoids that mixed path.

2. Dense 27B, no MTP, CP=4/SP=true smoke succeeded on the VLM/omni text-only path with the generic dynamic-padding tail monkeypatches disabled. This showed the model/topology/dataset path is OK without MTP.

3. Dense 27B, MTP enabled, CP=4/SP=true smoke failed without the MTP CP/SP ordering fix, with the expected concat sequence mismatch.

4. Dense 27B, MTP enabled, CP=4/SP=true smoke succeeded with the MTP CP/SP ordering fix and without the generic dynamic-padding tail monkeypatches:

```text
seq_length: 131072
context_parallel_size: 4
tensor_model_parallel_size: 4
pipeline_model_parallel_size: 2
sequence_parallel: true
mtp_num_layers: 1
step_func: qwen3_vl_step
```

Success markers:

```text
Starting training loop at iteration 0
iteration       10/      32 | lm loss: 1.262255E+01 | mtp_1 loss: 9.164709E+00 | grad norm: 113.018 | skipped: 0 | nan: 0
iteration       20/      32 | lm loss: 1.072900E+01 | mtp_1 loss: 8.349698E+00 | grad norm: 72.442 | skipped: 0 | nan: 0
iteration       30/      32 | lm loss: 1.017748E+01 | mtp_1 loss: 8.303581E+00 | grad norm: 65.152 | skipped: 0 | nan: 0
successfully saved checkpoint from iteration      32
[after training is done]
```

So for loss verification: both `lm loss` and `mtp_1 loss` were logged, finite, and decreasing over the 32-iteration smoke; `number of skipped iterations: 0`, `number of nan iterations: 0`.

I have not yet added the unit test suggested by the review comment. That still seems like the main missing piece before this is ready to merge.

LoganVegnaSHOP:
Addressed the outstanding review/comment items:

- Rebased this PR onto current upstream `main`.
- Added unit coverage in `test_text_model_forward.py` for the MTP embedding wrapper:
  - CP shard happens before SP scatter for `mtp_process=True`, `cp_size > 1`, `sequence_parallel=True`, `packed_seq_params is None`.
  - CP-only MTP still triggers the wrapper when `sequence_parallel=False`.
  - Packed sequence params skip the CP shard, matching the guard in `model.py`.
- Updated the PR description with the latest 128k Qwen3.6 smoke/loss validation and linked tracking issue #4328.

Local checks run:

```text
python3 -m py_compile tests/unit_tests/models/qwen_vl/modelling_qwen3_vl/test_text_model_forward.py src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/text_model.py
ruff check tests/unit_tests/models/qwen_vl/modelling_qwen3_vl/test_text_model_forward.py src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/text_model.py
```

Both passed.

I could not run the focused pytest locally on macOS because importing Megatron-Core in this checkout requires `triton`:

```text
ModuleNotFoundError: No module named 'triton'
```

The added tests are mocked CPU unit tests, so they should run in the Linux CI environment.

yaoyu-33 (dismissed):

gautham-kollu:
Container build failing

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/text_model.py
```diff
@@ -175,18 +176,32 @@ def forward(
...
         _shadow_embedding = False
-        if self.mtp_process and self.config.sequence_parallel:
+        if self.mtp_process and (self.config.sequence_parallel or cp_size > 1):
             _original_embedding = self.embedding
+            _cp_rank = self.pg_collection.cp.rank()
+            _apply_cp_shard = cp_size > 1 and packed_seq_params is None
+            _apply_sp_scatter = self.config.sequence_parallel
 
             def _sp_scatter_embedding(input_ids, position_ids):
```
HowardZorn, line 225: Perhaps this closure should have a different name?

suiyoubi:
@LoganVegnaSHOP could you take a look at the broken qwen recipe test ?

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/text_model.py
yaoyu-33: Resolved by the current-main merge. #4532 now localizes context-parallel input IDs before they enter the Qwen3-VL text model, so applying another CP shard here would double-shard the MTP embedding. The CP size/rank lookup has therefore been removed entirely, making `get_pg_size`/`get_pg_rank` unnecessary on this path. Added producer- and consumer-side regression coverage.

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/text_model.py
yaoyu-33: Resolved by the current-main merge. Because #4532 localizes CP input IDs upstream, the remaining shadow closure performs only sequence-parallel scatter again, so `_sp_scatter_embedding` accurately describes its behavior. The new CP+SP regression verifies one scatter at the unchanged rank-local sequence length.

yaoyu-33:
Closing as superseded by #4532, which merged the production Qwen3-VL context-parallel input-localization fix into main. After refreshing this branch against current main, the original inner text-model CP shard would now split MTP inputs a second time. That implementation was therefore dropped, leaving only regression tests in this PR. The latest exact-SHA CI, including the Qwen model and recipe jobs, is green. Thank you @LoganVegnaSHOP for the detailed reproduction, ablations, and loss validation; they were valuable in confirming the underlying contract. Any additional regression coverage can be proposed separately against current main.

PR #3984, opened 2026-05-27
[data] fix: prepare HF dataset before packing; preflight zstd in DCLM decompress

liding-nv:
- `scripts/training/pack_sft_data.py`: call builder.prepare_data() for HFDatasetBuilder before `prepare_packed_data()`. Without this the dataset cache is empty and packing fails with FileNotFoundError when reading `training.jsonl`.                                                     
  - `tutorials/data/dclm/decompress.py`: check that the zstd binary is on PATH before launching the worker pool. Without this preflight the `ProcessPoolExecutor.map()` iterator is discarded, so worker `FileNotFoundError` exceptions are swallowed and the script prints "Files were successfully decompressed in 0.0 minutes." and exits 0, a silent failure that looks like success.

### DIFF
--- a/scripts/training/pack_sft_data.py
+++ b/scripts/training/pack_sft_data.py
@@ -86,6 +86,14 @@ def main() -> None:
             if field.name not in dataloader_field_names
         },
     )
+
+    # For HF datasets, download + apply process_example_fn → training.jsonl. The
+    # packer in prepare_packed_data() reads that JSONL; without this step the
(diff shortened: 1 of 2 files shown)


cuichenx (approved):

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/3984/commits/82192129470c47afa1ff1262d06fb82945638d98

PR #3923, opened 2026-05-21
[recipe] feat: Add DeepSeek-V4 training smoke configs

Meirtz:
# What does this PR do ?

Adds minimal DeepSeek-V4 training smoke configs for exercising selected Megatron Core DSv4 architecture paths through Megatron Bridge. These configs are intended for short validation runs, not full production pretraining recipes.

# Changelog

- Add `deepseek_v4_tiny_pretrain_config` for a 1-layer DSv4 hybrid-attention smoke path.
- Add `deepseek_v4_flash_proxy_pretrain_config` for DSv4 hybrid attention with mHC and hash-routed MoE.
- Add `deepseek_v4_flash_mtp_proxy_pretrain_config` for DSv4 hybrid attention with mHC and MTP.
- Export the new recipe helpers and add unit and functional smoke coverage.
- Pin the Megatron Core dev submodule to a commit that contains the required DSv4 primitives.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci) in the Contributing doc for how to trigger the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

Local validation:

- `uv run --no-sync --with pre-commit pre-commit run --all-files`
- Focused GB200 recipe smoke validation for the DSv4 tiny, mHC/hash-MoE proxy, and mHC/MTP proxy paths.

# Before your PR is "Ready for review"

**Pre checks**:

- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [x] Did you add or update any necessary documentation? Not needed for these smoke-only recipe configs.
- [x] Does the PR affect components that are optional to install? No.

# Additional Information

This PR intentionally limits the scope to smoke configs and tests. It does not claim full DeepSeek-V4 training support.

The recipe configs currently depend on the Megatron Core dev pin because the required DSv4 primitives are not yet fully available in the Megatron Core main pin.

PR #3893, opened 2026-05-19
[recipe] feat: Add DeepSeek-V4-Flash pretraining recipes

weijiac0619:
# What does this PR do?

Adds pretraining recipes for DeepSeek-V4-Flash on Blackwell, plus a Slurm
launcher example, unit tests, and a functional test. The base recipe targets
TP=1 / PP=4 / EP=8 with selective activation recompute, an MTP-aware pipeline
layout, and BF16. Two variants extend the base for the two supported
optimizer + precision combinations.

# Changelog

- `src/megatron/bridge/recipes/deepseek/deepseek_v4.py` (new): three pretrain
  configs and a pipeline-layout helper.
  - `deepseek_v4_flash_pretrain_config()` — BF16 base; TP=1, PP=4, EP=8,
    selective recompute (`moe_act`, `mhc`), MTP placed on the last PP stage
    via `pipeline_model_parallel_layout`.
  - `deepseek_v4_flash_pretrain_mxfp8_config()` — Adam + MXFP8 training,
    BF16 MTP / validation eval, `quant_recipe` selects MXFP8 for TE linears.
  - `deepseek_v4_flash_pretrain_muon_config()` — Muon optimizer + BF16,
    non-layer-wise dispatch.
  - `set_deepseek_v4_pipeline_model_parallel_layout()` helper builds the
    even decoder layout with MTP and loss on the last PP rank.
- `src/megatron/bridge/recipes/deepseek/__init__.py`: re-export the new
  configs and helper.
- `examples/models/deepseek_v4/README.md`: document the new recipes and
  Slurm launcher.
- `examples/models/deepseek_v4/slurm_pretrain.sh`: Slurm sbatch script for
  the new recipes.
- `tests/unit_tests/recipes/test_deepseek_recipes.py`: extend coverage to
  DSv4 configs.
- `tests/functional_tests/test_groups/recipes/test_deepseek_recipes_pretrain.py`
  (new): L0/L1 pretrain smoke for DSv4 variants.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci) in the Contributing doc for how to trigger the CI.

# Before your PR is "Ready for review"

**Pre checks**:

- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
[...]

PR #4003, opened 2026-05-27
[model] fix: default dsa_indexer_loss_coeff to a non-None value in DeepSeek-V4 bridge

Meirtz:
# What does this PR do?

`DeepSeekV4Bridge.provider_bridge()` does not set `provider.dsa_indexer_loss_coeff`,
leaving it at the `MLATransformerConfig` default of `None`. The DSv4 attention path
reads the field via:

```python
# 3rdparty/Megatron-LM/.../experimental_attention_variant/csa.py
indexer_loss_coeff = getattr(self.config, 'dsa_indexer_loss_coeff', 0.0)
```

The intended `0.0` fallback never fires because the dataclass field always exists —
`getattr` returns `None`. Downstream
`3rdparty/Megatron-LM/.../experimental_attention_variant/dsa.py` then raises:

```
TypeError: unsupported operand type(s) for *: 'Tensor' and 'NoneType'
```

at `indexer_loss = kl_div * loss_coeff`.

Every existing DSv4 caller already works around this:
- `src/megatron/bridge/recipes/deepseek/deepseek_v4.py` sets `cfg.model.dsa_indexer_loss_coeff = 0.01`.
- `verl-project/verl#6473` sets `++override_transformer_config.dsa_indexer_loss_coeff=0.0`
  on both actor and ref configs of the GRPO example.

Install a safe default in the bridge itself, mirroring the sibling GLM5 DSA bridge
(`src/megatron/bridge/models/glm_moe_dsa/glm5_bridge.py`) which sets the same field
to `0.001` for the same reason. Recipe and external overrides continue to win as
before; this only affects callers that never set the field.

# Tests

No new tests. Mirrors the GLM5 sibling-bridge precedent (`glm5_bridge.py` sets
`dsa_indexer_loss_coeff = 0.001` without a dedicated unit test). Regression coverage
is implicit:
- Any caller going through `bridge.to_megatron_provider()` without overriding the
  field hits the DSA-active code path and would have failed loudly before this fix.
- The DSv4 recipe functional smoke (#3923) exercises the bridge end-to-end.

# AI attribution

This change was developed with AI-assisted coding (Claude). The author has reviewed
every changed line and verified the bug against the MCore `csa.py` / `dsa.py`
source paths cited above.

### DIFF
--- a/src/megatron/bridge/models/deepseek/deepseek_v4_bridge.py
+++ b/src/megatron/bridge/models/deepseek/deepseek_v4_bridge.py
@@ -402,6 +402,7 @@ class DeepSeekV4Bridge(MegatronModelBridge):
         provider.dsa_indexer_n_heads = hf_config.index_n_heads  # 64
         provider.dsa_indexer_head_dim = hf_config.index_head_dim  # 128
         provider.dsa_indexer_topk = hf_config.index_topk  # 512
+        provider.dsa_indexer_loss_coeff = 0.0
 
         # ---- Hyper-Connections (mHC) ----
         provider.enable_hyper_connections = True


yaoyu-33:
@weijiac0619  plz check.

weijiac0619:
> @weijiac0619 plz check.

in the training recipe PR I'm working on, model.dsa_indexer_loss_coeff value is set explicitly.

weijiac0619:
Thanks for the fix. Could we set the bridge default to `0.0` instead of `0.001`? For the current DSv4 recipe work we are keeping indexer loss disabled by default (`dsa_indexer_loss_coeff=0.0`) because enabling indexer loss still hits fused DSA/CUTLASS issues on our GB200 stack. A bridge-level default of `0.001` would silently enable indexer loss for callers that do not explicitly override it. Then staged/paper-training configs can still opt into `>0` explicitly.

Meirtz:
Thanks @weijiac0619, that makes sense — switched the default to 0.0 and pushed. The fix only needs it to be non-None so callers don't hit the `Tensor * None` TypeError, and disabled-by-default is the safer choice on the current stack. Recipes can still opt into >0 explicitly, as yours does.

Meirtz:
Closing this. The official recipes (#3893, merged) now set `dsa_indexer_loss_coeff = 0.0` explicitly in every DSv4 recipe, and verl sets it via override, so the real consumer paths are already covered. A bridge-level default also cuts against the repo's "no arbitrary config defaults" guidance — recipe-level explicit is the better home for it. Thanks @weijiac0619 and @yaoyu-33.

PR #4012, opened 2026-05-28
feat: add MLPerf v6.0 parity recipes for Llama3 8B and Llama2 70B LoRA

rsalagame-nvidia:
## Summary

Adds Megatron-Bridge performance configs and recipes for two MLPerf v6.0 reference-parity workloads consumed by the companion LLMB integration MR.

### Llama 3.1 8B MLPerf parity (GB200/GB300)

- New `LLAMA3_8B_PRETRAIN_CONFIG_<GB200|GB300>_<FP8_CS|FP8_MX|NVFP4>_MLPERF_V{1,2,3}` variants in `scripts/performance/configs/llama/llama3_workload_base_configs.py`. Shapes match the MLPerf v6.0 NVIDIA reference (8 GPU → V1, 16–64 GPU → V2, 128 GPU → V3) with `cuda_graph_impl="local"` + `cuda_graph_scope="full_iteration"`.
- New `set_llama3_mlperf_parity_overrides()` helper in `scripts/performance/configs/llama/llama3_llm_pretrain.py` applying universal parity knobs (`use_transformer_engine_op_fuser`, `fused_single_qkv_rope`, `tp_only_amax_red`, `use_te_rng_tracker`, cleared `cpu_offloading`, 768 MiB `comm_overlap.bucket_size`) plus the NVFP4 FP8-attention overlay.

### Llama 2 70B LoRA finetune (MLPerf v6.0 LoRA workload)

- New `scripts/performance/configs/llama/llama2_llm_finetune.py` + `llama2_workload_base_configs.py` with `LLAMA2_70B_LORA_CONFIG_<GB200|GB300>_<BF16|FP8_CS|NVFP4>_V1` baselines and `_MLPERF_V{1,2,3}` variants.
- New `llama2_70b_lora_config()` recipe in `src/megatron/bridge/recipes/llama/llama2.py` matching the MLPerf v6.0 LoRA-70B reference (rank=16, alpha=32, seq=8192, target=`["linear_proj", "linear_qkv"]`, default TP=8 PP=1).

## Test plan

- Llama 3.1 8B MLPerf parity: validated on GB200/GB300 at scales {8, 16, 32, 64, 128} in FP8 and NVFP4 via the companion LLMB launcher.
- Llama 2 70B LoRA: end-to-end validation **pending** — last cluster fire OOM'd at FP8→BF16 dequantize on checkpoint load. Tracking the fix; will mark this PR ready when green.

## Follow-up PRs

- Llama 3.1 70B and 405B MLPerf-tagged variants (separate PRs once validated).


src/megatron/bridge/recipes/llama/llama2.py
```diff
@@ -118,3 +120,113 @@ def llama2_7b_pretrain_config() -> ConfigContainer:
...
+    if isinstance(peft_scheme, str) and peft_scheme.lower() in ["lora", "dora"]:
+        peft_cfg.dim = 16
+        peft_cfg.alpha = 32
+        # MLPerf v6.0 ref applies LoRA to linear_proj + linear_qkv only.
+        if hasattr(peft_cfg, "target_modules"):
+            peft_cfg.target_modules = ["linear_proj", "linear_qkv"]
+
+    # Packed sequence settings
+    if cfg.model.context_parallel_size > 1:
+        cfg.dataset.packed_sequence_specs.pad_seq_to_mult = cfg.model.context_parallel_size * 2
```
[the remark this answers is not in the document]
malay-nagda: @rsalagame-nvidia can you resolve this?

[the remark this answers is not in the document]
yaoyu-33: @rsalagame-nvidia ping again.

scripts/performance/utils/overrides.py
```diff
@@ -87,9 +88,21 @@ def _set_megatron_fsdp_overrides(recipe: ConfigContainer, use_megatron_fsdp: boo
...
+    # a pretrained base checkpoint, so init on real device (70B fits per-rank on GB200/GB300).
+    recipe.model.init_model_with_meta_device = False
     recipe.model.gradient_accumulation_fusion = True
 
+    # Finetuning with context parallelism requires per-token loss (Megatron asserts this in
+    # training/config.py: "When finetuning with CP>1, calculate_per_token_loss must be True").
+    # The LoRA nvfp4 variant runs CP=2, so enable it here; average_in_collective is already
+    # forced False above, which is the required pairing for per-token loss.
+    if getattr(recipe.model, "context_parallel_size", 1) > 1:
+        recipe.model.calculate_per_token_loss = True
```
malay-nagda: in the current state, this knob will be set also for pre-training FSDP benchmarks. Isn't it only valid for fine-tuning (according to comment on line 99)?

scripts/performance/utils/overrides.py
```diff
@@ -87,9 +88,21 @@ def _set_megatron_fsdp_overrides(recipe: ConfigContainer, use_megatron_fsdp: boo
     # average_in_collective is not supported with Megatron FSDP
     recipe.ddp.average_in_collective = False
 
-    recipe.model.init_model_with_meta_device = True
+    # Meta-device init is incompatible with loading a legacy torch_dist (ShardedTensor)
+    # base checkpoint: torch DCP's meta-device load path only supports DTensor and
+    # raises "Found unsupported type ShardedTensor for meta device loading." This FSDP
+    # path is LoRA-only (pretrain leaves use_megatron_fsdp unset), and LoRA always loads
+    # a pretrained base checkpoint, so init on real device (70B fits per-rank on GB200/GB300).
+    recipe.model.init_model_with_meta_device = False
```
malay-nagda: Is this also supposed to be for fine-tuning only? Currently it will be set for pre-training too?

scripts/performance/perf_plugins.py
```diff
@@ -257,6 +257,12 @@ def _set_num_cuda_device_max_connections(
         moe_flex_dispatcher_backend: str,
         gpu_sm100_or_newer: bool,
     ):
+        if "CUDA_DEVICE_MAX_CONNECTIONS" in executor.env_vars:
+            logger.info(
+                f"Respecting explicit CUDA_DEVICE_MAX_CONNECTIONS="
+                f"{executor.env_vars['CUDA_DEVICE_MAX_CONNECTIONS']}"
+            )
+            return
```
malay-nagda, line 270: Is this block needed for when user overrides via CLI or something else?

src/megatron/bridge/recipes/llama/llama2.py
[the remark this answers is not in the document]
rsalagame-nvidia: Good catch — fixed in c6a862a5. The recipe-level pad_seq_to_mult was computed against the default CP=1 and never re-derived after set_workload_base_configs() overrode CP. I now re-derive pad_seq_to_mult = CP * 2 after the base configs finalize CP, in both llama2_70b_lora_config_gb200 and _gb300. Added a smoke test asserting pad_seq_to_mult == CP*2 for CP>1. In practice the gov_report path masked this (it rebuilds the dataset config with the correct padding), but the public recipe API was latently wrong for any direct CP>1 use — this closes that. @claude confirm

scripts/performance/utils/overrides.py
rsalagame-nvidia: Agreed, this was leaking into pretraining. Gated on recipe.peft is not None in c6a862a5 so it only applies to finetuning. Verified safe: the Megatron assertion that requires calculate_per_token_loss with CP>1 only fires for FinetuningDatasetConfig (config.py:1316), so pretrain CP>1 doesn't need it. Confirmed perf-neutral for the validated 8B numbers (it's a loss-reduction detail, not throughput).
@claude look at this.

scripts/performance/utils/overrides.py
rsalagame-nvidia: Same fix — now gated on recipe.peft is not None (c6a862a5). This knob is only needed for LoRA, where meta-device init conflicts with loading the ShardedTensor base checkpoint. Pretraining has no such base ckpt, so it now correctly keeps meta-device init (faster/lower-mem startup). Init-only, so no effect on the validated throughput. @claude look at this

scripts/performance/perf_plugins.py
rsalagame-nvidia: Yes — kept it, and added a clarifying comment in c6a862a5. The early-return lets an explicitly-set CUDA_DEVICE_MAX_CONNECTIONS (via CLI --env, or the MLPerf launch.sh which pins it to 1) survive the sm100+ heuristic that would otherwise force it to 32 and break parity with the v6.0 reference. Without the guard the parity bundle silently regresses. @claude look at this.

malay-nagda:
/claude review. Also, look at the comments and responses in the PR.

PR #4050, opened 2026-05-29
[training] fix: Handle FSDP checkpoint load and null finetune iterator

yaoyu-33:
## Summary
- handle `data_iterator=None` in finetuning batch preparation for ranks without local data
- unwrap raw `MegatronFSDP` before building `torch_dist` sharded state dict templates
- add focused unit coverage for both paths

## NVBug status
| NVBug | `origin/main` status | PR status | Notes |
| --- | --- | --- | --- |
| 6193300 | Not fixed | Fixed | `torch_dist` checkpoint template generation could call `sharded_state_dict()` on raw `MegatronFSDP`; this PR uses the wrapped module for that template while preserving wrapper loading. |
| 6181306 | Not fixed | Fixed | `prepare_finetuning_batch(None, ...)` now returns `(None, default_seq_length)` for ranks without a data iterator. |

## Validation
- `uv run pre-commit run --all-files` attempted; blocked before hooks by local dependency wheel compatibility on this workstation.
- `uv run --active --no-sync pre-commit run --all-files` passed.
- `uv run --active --no-sync python -m py_compile src/megatron/bridge/data/finetuning.py src/megatron/bridge/training/checkpointing.py tests/unit_tests/data/test_finetuning.py tests/unit_tests/training/test_checkpointing.py` passed.
- Focused local smoke for `prepare_finetuning_batch(None, ...)` passed.
- Focused local source-level smoke for the `MegatronFSDP` unwrap branch passed.

No full test suite was run.


yaoyu-33:
Closing per updated triage: FSDP checkpoints and regular torch_dist/5D checkpoints are not interchangeable, so this should not be treated as a Bridge code bug. We need to update the owning instructions/example flow instead of adding a compatibility workaround here.

PR #4052, opened 2026-05-29
[model] fix: Handle nested tied weight keys

yaoyu-33:
## Summary

Fixes the remaining Transformers compatibility regression for custom model code that returns nested tied-weight key containers.

- Add a shared compatibility helper that flattens string, mapping, and nested iterable tied-weight key containers before Bridge checks them.
- Add focused unit coverage for nested tied-weight key handling.

## NVBug Status

| NVBug | Status |
| --- | --- |
| 6178877 | Fixed by flattening nested tied-weight key containers before checking key membership. |
| 6178912 | Not handled in this PR. The referenced HF repository contains full checkpoint shards, so this PR no longer adds GLM config-only export handling. |
| 6178878 | Not handled in this PR. The legacy WAN recipe alias change was removed after review. |

## Validation

- `uv run --no-sync pre-commit run --files src/megatron/bridge/models/conversion/transformers_compat.py tests/unit_tests/models/conversion/test_transformers_compat.py src/megatron/bridge/models/glm/glm45_bridge.py src/megatron/bridge/models/glm_vl/glm_45v_bridge.py tests/unit_tests/models/glm/test_glm45_bridge.py tests/unit_tests/models/glm_vl/test_glm_45v_bridge.py`
- Direct Transformers 5.6.0 tied-weight repro: nested key lists flatten successfully.

Attempted focused pytest earlier with `uv run python -m pytest ...`; the local workstation blocked before test collection because the pinned `nvidia-resiliency-ext` wheel is unavailable for the local platform tag. No full test suite was run.

## Notes

- No changes under `3rdparty/Megatron-LM/`.
- Public text is sanitized; no internal paths, hosts, accounts, or job IDs are included.


src/megatron/bridge/diffusion/recipes/wan/__init__.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from megatron.bridge.diffusion.recipes.wan.wan import (
+    wan_1_3B_pretrain_config,
```
yaoyu-33: instead of do a "wan_1_3B_pretrain_config" ask in nvbug to use the correct config name. revert please

yaoyu-33:
Closing per updated triage. Bridge does not require or instruct creating a synthetic Kimi-K2 toy checkpoint for this model collection path, and the reported failure is in the Transformers/custom-model  handling for that synthetic toy model. This should not be fixed as a Megatron Bridge source bug.

yaoyu-33:
Clarification for the close reason: Bridge does not require or document creating a synthetic Kimi-K2 toy model for this workflow. The failure is in Transformers/custom Kimi-K2 model save_pretrained handling for that synthetic toy model, so this should be treated as a test-harness or upstream compatibility issue, not a Megatron Bridge source bug.

PR #4066, opened 2026-05-29
fix: correct gradient normalization for heterogeneous MIMO data parallelism (encoder DP != LLM DP)

kamran-nvidia:
# What does this PR do ?

Fixes incorrect gradient scaling in `finalize_model_grads_multimodule` when the
encoder and LLM modules use different data-parallel (DP) sizes (heterogeneous
MIMO parallelism). Previously, encoder gradients were too small by a factor of
`encoder_dp / llm_dp`, corrupting the global grad norm and the encoder's
effective learning signal.

# Changelog

- Add `_get_dp_size_from_grid()` helper that reads the DP size from a grid's
  `shape`/`dim_names` metadata instead of process groups, so it is valid on
  **all** ranks — including those outside a given module's grid.
- Rewrite `finalize_model_grads_multimodule` to handle the two loss-normalization
  modes separately:
  - **`calculate_per_token_loss=True` (`num_tokens` provided):**
    1. Run gradient all-reduce per module, passing `num_tokens` only to the LLM
       so `_finalize_model_grads` can PP-broadcast + DP-all-reduce the per-rank
       counts into the correct global token total. Encoder ranks pass
       `num_tokens=None` to skip the (otherwise broken) normalization.
    2. Broadcast the correct global total from the LLM's last PP rank to encoder
       ranks.
    3. Scale encoder gradients by `1 / global_total`.
    - Note: we must **not** globally broadcast `num_tokens` before
      `_finalize_model_grads`, since that would overwrite each rank's distinct
      count and the subsequent DP all-reduce would sum `dp_size` identical
      copies.
  - **`calculate_per_token_loss=False`:** after DDP finalization, compensate
    encoder gradients by `encoder_dp / llm_dp` to correct for the mean being
    taken over fewer samples on encoder ranks.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci) in the Contributing doc for how to trigger the CI.

**Pre checks**:

# Additional Information

- Related to # (issue)


src/megatron/bridge/training/megatron_mimo_parallel_utils.py
```diff
@@ -213,18 +230,85 @@ def finalize_model_grads_multimodule(
...
+    llm_grid = infra.module_to_grid_map.get(MIMO_LANGUAGE_MODULE_KEY)
+    llm_dp = _get_dp_size_from_grid(llm_grid) if llm_grid is not None else 1
+
+    def _find_module(grid):
+        for mn, mg in infra.module_to_grid_map.items():
+            if mg is grid:
+                return mn, infra.pg_collections.get(mn)
+        return None, None
+
+    if num_tokens is not None and llm_grid is not None:
```
yashaswikarnati, line 242: should the if condition be based on calcualte per token loss ?

kamran-nvidia: `num_tokens is not None` is exactly `calculate_per_token_loss` here. MCore's schedule (the only caller) passes `total_num_tokens if config.calculate_per_token_loss else None` (schedules.py:745/1986/2395), so the two are equivalent by construction.

We key off `num_tokens` rather than the flag because this function matches the schedule's call signature `(model, num_tokens, pg_collection, force_all_reduce=...)` and never receives `config`, and because upstream `finalize_model_grads` gates its per-token scaling on the same `num_tokens is not None` check (finalize_model_grads.py:546). Keeping the identical convention keeps the MIMO override consistent with the single-module path.

Happy to add a one-line comment making that explicit.

yashaswikarnati (dismissed):

kamran-nvidia:
/ok to test [4025e12](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4066/commits/4025e12b59a04266f7717adbcfa5c055ddb65dcf)

kamran-nvidia:
/ok to test [bd758ea](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4066/commits/bd758eabb1a48ff8448c7842ce190cae10418c2f)

yashaswikarnati (approved):

PR #4101, opened 2026-06-01
Fix SQuAD dataset name with newer Hugging Face Hub version

rhmukundan:
## Summary

This fixes SQuAD dataset loading for environments using newer `huggingface-hub` versions.

The existing config used the legacy dataset id `squad`, which can fail with newer HF URI parsing because the Hub expects canonical `namespace/name` dataset repo IDs. For SQuAD, the canonical ID is `rajpurkar/squad`.


yaoyu-33:
Closing during release-branch cleanup because this PR targets the retired llmb-r0.4.1 line (0.4 / 26.04 or earlier). If the change is still needed, please port it to a currently supported target branch in a fresh PR.

PR #4125, opened 2026-06-03
feat(training): return eval losses and add pure eval example

### LINKED ISSUE #2033: Support pure eval loop given a model and ConfigContainer
**Is your feature request related to a problem? Please describe.**

For various ModelOpt optimizations (quantization, pruning), we need to run forward-only inference on a small calibration dataset. Currently we have to write a bit hacky way of doing this, but would like to have this natively supported in megatron bridge via a API similar to Nemo2's `llm.validate` API.

**Describe the solution you'd like**

A simpler version of `megatron.bridge.pretraining.eval.evaluate_and_print_results` where we dont have to manually construct and pass state, forward_step, data_iterator, etc.:

```python
bridge = AutoBridge.from_hf_pretrained("Qwen/Qwen3-0.6B")
provider = bridge.to_megatron_provider()
provider.finalize()
model = provider.provide_distributed_model(wrap_with_ddp=False)

cfg = ConfigContainer(
    model=provider,
    train=TrainingConfig(
        micro_batch_size=1,
        global_batch_size=32,
        train_iters=32,
        eval_iters=32,
        skip_train=True,
    ),
    dataset=HFDatasetConfig(...),  # Alternatively GPTDatasetConfig
    tokenizer=TokenizerConfig(
        tokenizer_type="HuggingFaceTokenizer",
        tokenizer_model="Qwen/Qwen3-0.6B",
    ),
    # Unused - Currently still need to set them else errors out
    optimizer=OptimizerConfig(optimizer="adam", lr=1e-4, use_distributed_optimizer=False),
    scheduler=SchedulerConfig(lr_decay_style="constant"),
    logger=LoggerConfig(),
    checkpoint=CheckpointConfig(),
)

lm_loss = simpler_eval(model, cfg)
```

**Describe alternatives you've considered**

A bit hacky approach (let me know if this can already be simplified)

```python
bridge = AutoBridge.from_hf_pretrained("Qwen/Qwen3-0.6B")
provider = bridge.to_megatron_provider()
provider.finalize()
model = provider.provide_distributed_model(wrap_with_ddp=False)

cfg = ConfigContainer(
    model=provider,
    train=TrainingConfig(
        micro_batch_size=1,
        global_batch_size=32,
        train_iters=32,
        eval_iters=32,
        skip_train=True,
    ),
    dataset=HFDatasetConfig(,  # Alternatively GPTDatasetConfig
        dataset_name="cnn_dailymail",
        dataset_dict=DatasetDict({"train": ["sample1", "sample2", ...]}),
        process_example_fn=lambda example, tokenizer: {"input": example, "output": ""},
        seq_length=1024,
        dataloader_type="batch",
        num_workers=1,
        do_validation=False,
        do_test=False,
        val_proportion=None,
        split_val_from_train=False,
        rewrite=False,
    )
    tokenizer=TokenizerConfig(
        tokenizer_type="HuggingFaceTokenizer",
        tokenizer_model="Qwen/Qwen3-0.6B",
    ),
    # Unused - Currently still need to set them else errors out
    optimizer=OptimizerConfig(optimizer="adam", lr=1e-4, use_distributed_optimizer=False),
    scheduler=SchedulerConfig(lr_decay_style="constant"),
    logger=LoggerConfig(),
    checkpoint=CheckpointConfig(),
)


# Current Hacky Approach
from megatron.bridge.training.eval import evaluate_and_print_results

runtime_config_update(cfg)

state = GlobalState()
state.cfg = cfg

dataset_provider = get_dataset_provider(cfg.dataset)

def _train_valid_test_datasets_provider(
    train_val_test_num_samples: tuple, dataset_cfg: HFDatasetConfig
):
    return dataset_provider(train_val_test_num_samples, dataset_cfg, tokenizer=state.tokenizer)

train_data_iterator, _, _ = setup_data_iterators(
    cfg=cfg,
    train_state=state.train_state,
    model_length=len(model),
    train_valid_test_datasets_provider=_train_valid_test_datasets_provider,
    dp_group=get_data_parallel_group(),
)

evaluate_and_print_results( # Have this function also return validation loss
    state,
    prefix="iteration 1",
    forward_step_func=forward_step,
    data_iterator=train_data_iterator,
    model=model,
    config=cfg,
    verbose=True,
    write_to_tensorboard=False,
)
```

Happy to contribute above approach as an API if it looks right or if you have a better approach

**Additional context**

yaoyu-33:
## Summary
- Keep the public path on the existing `evaluate_and_print_results()` helper: it still performs the same logging/printing side effects and now returns the averaged loss dictionary when evaluation completes.
- Remove the separate `evaluate_model()` wrapper from this branch so there is no second eval entry point.
- Add `examples/evaluation/pure_eval.py`, a runnable minimal Bridge example that builds a small GPT model, mock data iterator, and calls `evaluate_and_print_results()` directly.
- Keep focused mock unit coverage for the existing helper's return-value behavior.

## Usage
```bash
uv run python -m torch.distributed.run --nproc_per_node=2 examples/evaluation/pure_eval.py --tp-size 2
```

Programmatic callers can use the existing helper directly:

```python
losses = evaluate_and_print_results(...)
```

The example uses the standard Bridge setup path to create the model, mock dataset iterator, and process groups, then calls `evaluate_and_print_results()`. Workflows that already have a Megatron model, `ConfigContainer`, and iterator can use the same existing helper call.

## Tests / Validation
- `ruff check src/megatron/bridge/training/eval.py tests/unit_tests/training/test_eval.py examples/evaluation/pure_eval.py`
- `ruff format --check src/megatron/bridge/training/eval.py tests/unit_tests/training/test_eval.py examples/evaluation/pure_eval.py`
- `python3 -m py_compile src/megatron/bridge/training/eval.py tests/unit_tests/training/test_eval.py examples/evaluation/pure_eval.py`
- `git diff --check origin/main...HEAD`
- `pre-commit run --all-files`
- Internal 2-GPU validation on pushed head `e1528cc`:

```bash
uv run --no-sync python -m torch.distributed.run --nproc_per_node=2 \
  examples/evaluation/pure_eval.py --tp-size 2 --eval-iters 1 \
  --seq-length 64 --hidden-size 64 --num-attention-heads 4 \
  --global-batch-size 2 --micro-batch-size 1
```

Result: passed. The run initialized tensor model parallel size 2, evaluated one iteration, and returned `{'lm loss': 10.390985488891602}`.

Not run as passing local evidence:
- `uv run pre-commit run --all-files` is blocked on this workstation before hooks run because the pinned `nvidia-resiliency-ext==0.6.0` wheel is not available for the local Linux platform baseline.
- Runtime pytest was not used as authoritative local evidence; runtime validation was done with the 2-GPU example smoke above.

## Notes
- The example intentionally keeps using the existing setup path, which still constructs optimizer/scheduler objects because current setup/config validation expects them. This PR does not refactor those broader requirements.

Fixes #2033


yaoyu-33:
For the pure-eval example, I think we should keep calling the existing `evaluate_and_print_results()` directly rather than adding a new eval API.

The train/validation step fields here are not required by `evaluate_and_print_results()` itself. They are compatibility shims because the example still uses the standard Bridge `runtime_config_update()` + `setup()` path:

- `validation.eval_iters` is the real eval length and is consumed by the eval loop.
- `validation.eval_interval=1` makes setup allocate/build a validation iterator; otherwise valid sample count is zero and the example falls back to the train iterator.
- `train.train_iters=1` is currently required by `TrainingConfig.finalize()` and by data-loader sample accounting.
- the small scheduler values keep the full training setup/scheduler validation legal for this dummy one-iteration config.

So I would either keep these with a short comment saying they are full-setup shims, or follow up with a separate eval-only setup path that bypasses optimizer/scheduler/training sample accounting. For this PR, the comment is probably the lower-risk cleanup.

PR #4079, opened 2026-05-29
perf(llama): B300/B200 pretrain recipe defaults + 8B/405B aliases to …

rsalagame-nvidia:
Defaults the b300/b200 llama3_70b and llama31_405b pretrain builders to config_variant="v2", and aliases the 8B/405B B300/B200 V1 (+ 405B V2) presets to their GB300/GB200 counterparts.

PR #4126, opened 2026-06-03
perf(llama): B300/B200 pretrain recipe defaults + 8B/405B aliases to GB-counterparts

rsalagame-nvidia:
Backport of #4079 to `r0.5.0`.

Defaults the b300/b200 llama3_70b and llama31_405b pretrain builders to `config_variant="v2"`, and aliases the 8B/405B B300/B200 V1 (+ 405B V2) presets to their GB300/GB200 counterparts.

Cherry-picked cleanly from the `main` PR; the four touched config files are identical between `main` and `r0.5.0`.


scripts/performance/configs/llama/llama31_workload_base_configs.py
```diff
@@ -114,80 +114,20 @@
     virtual_pipeline_model_parallel_size=8,
     global_batch_size=64,
     cuda_graph_impl="none",
-    cuda_graph_scope=[],
+    cuda_graph_scope="full_iteration",
```
malay-nagda, line 117: This also affects LLAMA31_405B_PRETRAIN_CONFIG_GB200_NVFP4_V2 which is the main tracked config. Can you update the golden values for LLAMA31_405B_PRETRAIN_CONFIG_GB200_NVFP4_V2?

malay-nagda (changes requested):

Issue #4083, opened 2026-05-30
_megatron_global_param_names_all_pp_ranks only gathers across PP, silently drops EP/ETP-only weights from saved HF index


dinhxuanvu:
## Summary

`MegatronModelBridge._megatron_global_param_names_all_pp_ranks`
([model_bridge.py:785-829](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/models/conversion/model_bridge.py#L785-L829))
gathers HF parameter names across the **PP group only**:

```python
pp_group = parallel_state.get_pipeline_model_parallel_group()
gathered_global_param_names = [None] * pp_group.size()
torch.distributed.all_gather_object(gathered_global_param_names,
                                    global_param_names, group=pp_group)
```

When `EP > 1` or `ETP > 1`, names that live only on non-EP-0 / non-ETP-0
ranks never enter the master list and are silently dropped from the
saved HF directory's `model.safetensors.index.json`. No warning, no
error, no failed assertion at save time.

## Reproduction

Hit on a 512-GPU Nemotron-3 Super 120B SFT run with
`(TP=4, PP=8, CP=2, EP=8, ETP=1)` on Megatron-Bridge `0.5.0+8382dc34`
(commit `8382dc34`). The saved HF index is missing **1204 of 42683**
tensors vs. the upstream `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16`
reference:

| Group | Missing | Sample |
|---|---|---|
| MoE experts on `backbone.layers.87` | 156 | `backbone.layers.87.mixer.experts.434.down_proj.weight` |
| MTP layers | 1040 | `mtp.layers.1.mixer.experts.0.down_proj.weight` |
| Head / final-layer | 8 | `lm_head.weight`, `backbone.norm_f.weight` |

Missing tensors are exactly those owned by rank-sets split across
EP/ETP that PP rank 0 doesn't see.

**2-GPU repro** (no 120B specifics needed):

- `TP=PP=CP=1, EP=2`, any MoE model with `num_moe_experts >= 2`
- `TP=PP=CP=1, EP=1, ETP=2`, any MoE model with TP-shardable expert
  weights

The existing
[`test_multi_gpu_conversion.py`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/tests/functional_tests/test_groups/converter/test_multi_gpu_conversion.py#L27-L33)
[...]

PR #3562, opened 2026-04-28
DeepSeek V4 Bridge

weijiac0619:
## Summary
- Adds DeepSeek V4 bridge support using native Hugging Face DeepSeek V4 configs.
- Adds FP8/MXFP4 checkpoint import/export mappings, docs/examples, and focused coverage.
- Pins the MCore dev commit with DeepSeek V4 Part 3 support and refreshes `uv.lock`.

## Validation
- Full HSG EP8 import and export roundtrip completed.
- `pre-commit run --all-files` passed.
- `uv lock --check` and locked `uv sync` dry-runs passed.

PR #4130, opened 2026-06-03
fix: gather conversion param names across WORLD with superset assertion

dinhxuanvu:
## Summary

This is a defensive hardening of the parameter-name gather used during Megatron to HuggingFace conversion. It is independent of any prior weight-correctness fix and does not touch checkpoint state handling.

`_megatron_global_param_names_all_pp_ranks` previously gathered parameter names across only the pipeline-parallel group. Parameter names can be sharded across multiple axes (pipeline, expert, and expert-tensor parallel), so a pipeline-only gather can silently miss expert names that live on other expert/expert-tensor ranks, producing a truncated global name set.

## Changes

- Gather parameter names across the full process group instead of the pipeline-parallel group, then union the per-rank name sets. Ordering is preserved by the existing sort, so the cross-rank synchronization behavior is unchanged.
- Add a post-gather superset assertion: the gathered global set must contain every name each rank contributed locally. A regression that drops names now fails loudly with a clear message instead of silently truncating the conversion.
- Add EP, PP+EP, and EP+ETP parametrizations to `test_multi_gpu_conversion.py` using a small Qwen3 MoE toy model, following the existing parametrization and subprocess harness. Each case asserts the saved on-disk tensor set covers the expected expert weights, exercising the gather path across expert/expert-tensor ranks.

## Test plan

- Local: ruff check and ruff-format pass on the changed files; `pytest --collect-only` collects all parametrizations (including the new EP/ETP cases) without errors.
- The new multi-GPU MoE cases require a real multi-GPU cluster and run under the existing GPU-only functional test markers.

### DIFF
--- a/src/megatron/bridge/models/conversion/model_bridge.py
+++ b/src/megatron/bridge/models/conversion/model_bridge.py
@@ -781,7 +781,6 @@ class MegatronModelBridge(
             return self._cached_param_names
 
         # Compute the result
-        pp_group = _get_pp_group(megatron_model)
         model_config = unwrap_model(megatron_model)[0].config
         global_param_names = []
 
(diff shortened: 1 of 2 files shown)


yaoyu-33:
Thanks for the contribution and for following up with the current-main repro result.

Given the update in #4083, the original issue no longer reproduces on current `main`, and the root cause appears to have been the saved-index construction path rather than `_megatron_global_param_names_all_pp_ranks`. That path should already be covered by #3562 and #3905.

For the gather path specifically, we currently try to leverage the symmetry that each rank hosts the same number of experts: the global-name list only needs the local expert slots, and the expert mappings fan those slots out across EP ranks when materializing the HF weights. This keeps the synchronization narrower than a full WORLD gather. Switching this helper to WORLD gather is a reasonable defensive hardening direction, but it also increases communication and changes a path where we do not currently have an active reproducer.

So I think we should defer this change for now rather than merge it immediately. If a new repro shows that the current symmetry-based path still drops names, we can reopen this direction with that concrete failure case and decide whether WORLD gather, PP ∪ EP ∪ ETP gather, or a narrower assertion is the right fix.

Thanks again for the report and PR.

Issue #2391, opened 2026-02-16
Support num_epochs in TrainingConfig


eladg9:
**Is your feature request related to a problem? Please describe.**

The current implementation relies on iteration-based training (train_iters). While precise for step-counting, it lacks a native mechanism to ensure training completes exactly at the end of a full data pass. Manually calculating the total steps required to cover a dataset is prone to errors and requires constant recalculation whenever the global_batch_size or dataset composition changes.

**Describe the solution you'd like**

I am requesting the addition of a --num_epochs parameter. This feature should:
Automatically calculate the equivalent train_iters based on the total sample count of the loaded dataset and the global_batch_size.
Support fractional values (e.g., 0.5 epochs).

**Describe alternatives you've considered**

The current workaround involves manual calculation using the following logic:
```
\text{train\_iters} = \frac{\text{total\_tokens\_to\_train}}{\text{batch\_size} \times \text{seq\_len}}
```
which is very inaccurate

---

Thanks!

maanug-nv:
Hi @eladg9 . Is this request for SFT or pretraining? Typically for pretraining, LLM datasets are too large to consider training multiple times. Usually we just choose a target number of tokens and set the iterations based on that, as in the alternative you mentioned. Setting `train_samples` is also supported instead of iterations.

yaoyu-33:
We can potentially add to some SFT dataset. Some dataset we cannot infer total samples in the dataset. Which training dataset you are using can help us a bit.

muyihao:
@maanug-nv Hi, I'm interested in working on this. May I have this issue assigned to me? Thank you!

PR #4147, opened 2026-06-04
feat(training): support epoch-based training duration

muyihao:
# What does this PR do ?

Add epoch-based training duration configuration for finite finetuning datasets.

# Changelog

- Add `TrainingConfig.num_epochs` with fractional epoch support.
- Calculate `train_iters` from the loaded training dataset size and global batch size.
- Build and cache finite finetuning datasets before optimizer and scheduler initialization.
- Preserve incomplete final global batches for epoch-based training using the `batch` dataloader.
- Validate mutual exclusivity between `train_iters`, `train_samples`, and `num_epochs`.
- Add tests for epoch-based configuration, fractional epochs, dataset restrictions, and dataloader behavior.
- Document epoch-based training configuration and limitations.

# GitHub Actions CI

**Pre checks**:

- [x] Does the PR affect components that are optional to install? No
  - [x] Reviewer: Does the PR have correct import guards for all optional libraries? N/A

# Additional Information

- Related to #2391
- Python 3.12 syntax checks and staged diff checks passed.
- Full tests and pre-commit were not run locally because `uv` was unavailable.


muyihao:
@yaoyu-33 Hi, this PR addresses #2391 by adding `num_epochs` support for finite finetuning datasets. Could you please help review it when you have time? Thanks!

src/megatron/bridge/training/config.py
```diff
@@ -1523,6 +1544,22 @@ def _validate_training_scheduler_compatibility(self) -> None:
                 "Can only specify one of lr_warmup_fraction or lr_warmup_iters"
             )
 
+    def resolve_num_epochs(self, train_dataset_size: int) -> None:
```
yaoyu-33: make it private

src/megatron/bridge/training/setup.py
```diff
@@ -203,6 +203,22 @@ def setup(
 
             cfg.dataset.token_dtype_code = 4 if vocab_size > numpy.iinfo(numpy.uint16).max + 1 else 8
 
+    if "tokenizer" in inspect.signature(train_valid_test_datasets_provider).parameters:
```
yaoyu-33: can this part put in validate along with original data setup part? also these 2 conditions of signiture checks can be put in into if cfg.train.num_epochs is not None:

muyihao: Good point. I moved the config-only validation there, but the dataset build itself cannot live in validate() because it opens/builds data and has side effects.

For num_epochs, we need len(train_ds) before optimizer/scheduler setup so train_iters is known before the scheduler is created. The original data setup block runs too late for that.

For the signature checks, I scoped them inside the num_epochs path for the early dataset build, but kept the normal binding before setup_data_iterators() as well. Non-num_epochs runs still need tokenizer/pg_collection passed to the dataset provider.

muyihao:
@yaoyu-33 hi，this pr is ready for review

yaoyu-33 (commented):
Review notes from the epoch-based training pass:

1. Padded final batches currently duplicate real samples with nonzero loss.

`loaders.py` enables `drop_last=False` for `num_epochs + dataloader_type="batch"`, which makes the batch sampler pad the final global batch with `-1`. The SFT dataset maps `-1` to the last real example and only marks it as autogenerated metadata, while loss masking still treats it as a normal sample. That means the final partial global batch trains on duplicated last samples instead of neutral padding.

Suggested fix: padded/autogenerated samples should have zero `loss_mask`, or the sampler/collate path should use a neutral padding sample. Please add a regression test where dataset size is not divisible by global batch size and assert padded rows do not contribute to loss.

2. Datasets smaller than one global batch are still rejected.

`build_train_valid_test_data_loaders()` still raises when `len(train_ds) < global_batch_size` for non-external dataloaders. That blocks `num_epochs=1` on a small finite dataset even though the PR's epoch mode is explicitly trying to preserve/pad incomplete final global batches.

Suggested fix: skip or relax this guard for `num_epochs + dataloader_type="batch"`, then add a test with `len(train_ds) < global_batch_size`.

3. `num_epochs` with `dataloader_type="single"` looks under-specified.

`_resolve_num_epochs()` always uses `ceil(dataset_size / global_batch_size)`, but the docs/tests allow `single` dataloader while saying it keeps drop-last behavior. For a non-divisible dataset, this can compute more train steps than the single-pass iterator can provide.

Suggested fix: either restrict epoch-based training to `batch` initially, or define/test the exact `single` semantics.

muyihao:
@yaoyu-33 Thanks for the detailed review. I pushed an update addressing these points:

1. Padded final batches no longer contribute to loss. Autogenerated SFT padding samples now produce zero loss_mask, and I added a regression test that uses the batch sampler with a dataset size not divisible by global_batch_size to verify padded rows do not contribute to loss.

2. num_epochs with dataloader_type="batch" now allows datasets smaller than one global batch, since the final global batch is padded and kept. I added a test for len(train_ds) < global_batch_size.

3. I restricted num_epochs to dataloader_type="batch" for now. The single/cyclic semantics are under-specified because train_iters is computed from global batches, while those dataloaders keep their existing incomplete-batch behavior. The docs and validation now make this explicit, and tests cover the rejection.

I also picked up Claude’s minor suggestions around finite-length error handling and the num_epochs validation message.

yaoyu-33 (dismissed):

src/megatron/bridge/data/datasets/sft.py
```diff
@@ -703,9 +703,15 @@ def _collate_item(self, item, max_length, pad_id):
         item = [x + [pad_id] * (max_length - len(x)) for x in item]
         return item
 
+    def _is_autogenerated(self, processed_example):
```
yaoyu-33, line 707: I dont think we want to keep this. We dont have "__AUTOGENERATED__"

src/megatron/bridge/data/loaders.py
```diff
@@ -228,8 +263,14 @@ def build_train_valid_test_data_loaders(
         cfg=cfg, build_train_valid_test_datasets_provider=build_train_valid_test_datasets_provider
     )
 
+    keep_partial_final_global_batch = cfg.train.num_epochs is not None and cfg.dataset.dataloader_type == "batch"
```
yaoyu-33: can we just replace `keep_partial_final_global_batch` with `drop_last`. drop last itself is clear enough.

yaoyu-33:
@muyihao  thanks for contribution, added few more comments. overall lgtm now.

src/megatron/bridge/data/loaders.py
```diff
@@ -164,6 +164,38 @@ def build_train_valid_test_datasets(
...
+        raise ValueError('num_epochs is currently supported only with dataloader_type="batch"')
+
+    train_ds, valid_ds, test_ds = build_train_valid_test_datasets_provider([0, 0, 0], cfg.dataset)
+    if train_ds is None:
+        raise ValueError("num_epochs requires a training dataset")
+
+    try:
+        train_dataset_size = len(train_ds)
+    except (TypeError, NotImplementedError) as error:
+    try:
```
yaoyu-33, line 189: This leaves the `except` block empty and immediately starts a second `try`, so `loaders.py` is syntactically invalid. A minimal compile check fails with `IndentationError` at this line, which means CI will fail during import/collection before reaching the new tests. Please collapse this to a single `try/except` around `len(train_ds)`.

src/megatron/bridge/data/loaders.py
```diff
@@ -164,6 +164,38 @@ def build_train_valid_test_datasets(
...
+        train_dataset_size = len(train_ds)
+    except (TypeError, NotImplementedError) as error:
+        raise ValueError("num_epochs requires a training dataset with a finite length") from error
+    return train_ds, valid_ds, test_ds
```
yaoyu-33, line 195: After validating `len(train_ds)`, this path returns without calling `cfg._resolve_num_epochs(train_dataset_size)`. That means `cfg.train.train_iters` remains `None` before optimizer/scheduler setup, despite the docstring and tests expecting this helper to resolve epoch-based training iterations. Please call `cfg._resolve_num_epochs(train_dataset_size)` before returning the cached datasets.

yaoyu-33 (changes requested):
Reviewed the current head with the review-pr workflow plus the testing/CI guidance. I’m requesting changes for two blockers in `src/megatron/bridge/data/loaders.py`. I also noticed the PR is currently marked conflicting with `main`, but the comments below are independent of the merge conflict.

Validation run: `python3 -m py_compile src/megatron/bridge/data/loaders.py` fails with `IndentationError: expected an indented block (loaders.py, line 192)`. I did not run broader tests because the module cannot parse at the current head.

muyihao:
> Reviewed the current head with the review-pr workflow plus the testing/CI guidance. I’m requesting changes for two blockers in `src/megatron/bridge/data/loaders.py`. I also noticed the PR is currently marked conflicting with `main`, but the comments below are independent of the merge conflict.
> 
> Validation run: `python3 -m py_compile src/megatron/bridge/data/loaders.py` fails with `IndentationError: expected an indented block (loaders.py, line 192)`. I did not run broader tests because the module cannot parse at the current head.

Good catch! 😅 Looks like those two slipped in via the claude[bot] suggestion in 79d0ad18 — reverted in d10f481a, so py_compile is happy again and _resolve_num_epochs is back where it belongs. Will rebase onto main next.

src/megatron/bridge/data/datasets/sft.py
muyihao: \_\_AUTOGENERATED\_\_ actually predates this PR. It's added in GPTSFTDataset.__getitem__ (sft.py:466-475) to tag rows that MegatronPretrainingBatchSampler produces when pad_samples_to_global_batch_size=True substitutes -1 for missing indices in the final partial batch. I reused that existing tag in _build_loss_mask to satisfy your earlier ask that padded rows must not contribute to loss.

Happy to go either direction here — which do you prefer?
1. Move the "this row is padding" signal out of dataset metadata into the sampler/collate path, and zero loss_mask there instead.
2. Drop the zero-loss-on-padding guarantee for num_epochs + dataloader_type="batch" (and remove the regression test you asked for).
Either works on my end — let me know which fits the direction you want for __AUTOGENERATED__ overall. 🙂

muyihao:
@yaoyu-33 hi, could you please help re-trigger the CI tests when you get a chance? Thanks!

suiyoubi:
/ok to test [6ad57cd](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4147/commits/6ad57cdcb2f82d977b74ea53dd0d0b4d110e0461)

muyihao:
@suiyoubi cicd tests have passed. I’ve committed changes to resolve the conflicts; could you please retrigger the test

suiyoubi:
/ok to test [0dadc71](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4147/commits/0dadc7144860b329a44bfd364c96525ca12100ad)

suiyoubi:
/ok to test [36fed3f](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4147/commits/36fed3fb765b39b6f63522d2abda2828b409882b)

muyihao:
@suiyoubi  Rebased onto latest main and restored the upstreamed `ModelConfig` import in setup.py that a prior merge had dropped. Could you re-run CI?

suiyoubi:
/ok to test [00911ae](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4147/commits/00911ae59e6b878a895c2d636cfda6574eb9c87e)

yaoyu-33 (approved):

PR #4163, opened 2026-06-04
[perf] Disable PCT binding by default

rhmukundan:
Flip the default for PCT (Priority Core Turbo) fixed-core CPU binding to disabled, in both the --enable_pct_binding CLI arg and the slurm executor signature.

Binding is now opt-in: pass --enable_pct_binding True on systems where Intel Granite Rapids PCT is configured. Mbridge's existing per-workload logic in setup_experiment.py is unchanged, so it still auto-disables binding for the b300 model/precision combos where it is known to hurt. This avoids applying core pinning on systems where PCT is not set up, without requiring recipes to disable it manually.

### DIFF
--- a/scripts/performance/argument_parser.py
+++ b/scripts/performance/argument_parser.py
@@ -494,9 +494,9 @@ def parse_cli_args():
     slurm_args.add_argument(
         "--enable_pct_binding",
         type=bool_arg,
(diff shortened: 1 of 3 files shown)


sudostock:
I like solving this on launch.sh side better now that we have discussed it more.

PR #4169, opened 2026-06-05
fix(data): anchor VLM assistant loss masks to chat templates

yaoyu-33:
## Summary

- prefer HF `return_assistant_tokens_mask` when the active chat template exposes generation blocks
- add Qwen/ChatML boundary masking from assistant role markers through `<|im_end|>`
- add a generic template-diff fallback for non-ChatML templates before legacy text search
- cover repeated user/assistant text, multi-turn, empty assistant, leading-newline, and structured-content cases

## Stack

Bottom of the current data stack: #4169 -> #4307 -> #4315. The earlier #4166 dependency has already merged into `main`.

## Validation

- `uv run python -m pytest tests/unit_tests/data/test_vlm_processing.py -v`
- Additional private tokenizer smoke across supported/newer VLM recipe IDs: Qwen2.5-VL, Qwen3-VL, Qwen3.5-VL, Gemma4-VL, Ministral3, Kimi K2.5, and Nemotron VL. Gemma3-VL load was blocked by gated model access; Ministral3 empty-assistant rendering is rejected by its template.


yaoyu-33:
Parity check summary for current head `69ce12dc0fe651abe266f88a0f44c9df0879e1ef`:

- Compared Bridge `build_assistant_loss_mask()` against HF `apply_chat_template(..., return_assistant_tokens_mask=True)` on Qwen generation-template refs for single-turn, multi-turn, and tool-like assistant content. Bridge primary masks matched HF `assistant_masks` exactly.
- Compared Bridge shifted labels/loss mask against TRL SFT label masking semantics. Bridge shifted labels matched the TRL shifted-left target semantics, and the shifted loss mask matched HF assistant masks shifted by one token.
- Checked the boundary fallback with Qwen ChatML role boundaries. It matched HF assistant masks exactly; assistant `<think>` / `<tool_call>` text contributes to loss, while system/user tool-like text does not.
- Cross-checked LLaMA-Factory Qwen3 behavior. Exact `input_ids` match for the single-turn case. Multi-turn and explicit tool-like cases differ because LLaMA-Factory renders the Qwen3 template slightly differently, but the masking semantics are aligned: source/user/system text is masked out and assistant target text contributes to loss.

No Bridge code change was needed from this parity pass. The one caveat is that stock Qwen templates without `{% generation %}` return all-zero assistant masks, so the HF comparison used generation-template refs where assistant masks are actually available.

cuichenx (approved):

Issue #4195, opened 2026-06-07
[bug] Gemma 4 global-attention RoPE inv_freq has wrong shape vs HF proportional RoPE


jeje-soy-michu:
### Problem

Gemma 4's `full_attention` (global) layers use `rope_type="proportional"` in the HF
config. HF's `_compute_proportional_rope_parameters` builds `inv_freq` of length
`head_dim // 2` — `int(head_dim * partial_rotary_factor) // 2` real frequencies
followed by zero-padding — so `rotate_half` pairs head-dim lanes `(i, i + head_dim//2)`.

`Gemma4RotaryEmbedding` (`src/megatron/bridge/models/gemma/gemma4_provider.py`) instead
passes `rotary_percent=global_rotary_percent (0.25)` to the parent `RotaryEmbedding` and
builds `inv_freq` with only the 64 real frequencies and **no zero-pad**. The 64 frequency
*values* match HF bit-for-bit, so value-level checks pass — but the **shape (64 vs 256) is
wrong**, so once a forward pass hits `rotate_half` the head-dim lanes are paired
incorrectly, producing per-global-layer cosine drift that compounds into measurable
last-token logit divergence vs HF reference.

### Scope — affects all Gemma 4 variants, not just one

`Gemma4RotaryEmbedding` is instantiated in a single place (`Gemma4ModelProvider.provide()`),
and every variant routes through it:

- **MoE `gemma-4-26B-A4B`** 
- **Dense `gemma-4-31B`**
- **Text / VL**

All shipped variants set `global_rotary_percent = 0.25` (from the config's
`partial_rotary_factor`), so all hit the buggy partial path. Only the **global /
`full_attention` layers** are affected; local/sliding layers use a separate full-rotary
`rope_local` (`rope_type="default"`) and are correct.

### Minimal repro

```shell
"""Megatron Bridge Gemma4 global RoPE inv_freq shape != HF. Deps: torch, transformers>=5.6 (CPU-only)."""
import torch
from transformers.modeling_rope_utils import _compute_proportional_rope_parameters
from transformers.models.gemma4.configuration_gemma4 import Gemma4TextConfig

HEAD_DIM, THETA, PRF = 512, 1_000_000.0, 0.25  # global ("full_attention") params, shared by MoE/Dense/VL

cfg = Gemma4TextConfig()
cfg.head_dim, cfg.partial_rotary_factor = HEAD_DIM, PRF
cfg.rope_parameters = {
[...]

PR #4148, opened 2026-06-04
feat(model): add Gemma-4 E4B support (layer spec, checkpoint loader, parity check)

DOGEUNNKIM:
# What does this PR do?

- Adds full Gemma-4 E4B (3.8B dense text) support to Megatron-Bridge,
  consolidating all Gemma4-specific code that was previously scattered in
  Megatron-LM core (see MCore PR: https://github.com/NVIDIA/Megatron-LM/pull/5090).
- Introduces `Gemma4DenseProvider` — a Bridge-native model provider that works
  against **clean Megatron-Core** (no Gemma4-specific CLI args or
  `TransformerConfig` fields required in MCore).
- Separates the text-only path (`Gemma4Bridge` / `Gemma4ForCausalLM`) from the
  VL path (`Gemma4VLBridge` / `Gemma4ForConditionalGeneration`); `gemma_vl`
  imports from `gemma`, not vice versa.
- Provides HF → Megatron checkpoint conversion, logit parity verification, and
  training launch scripts.

# Changelog

#### `src/megatron/bridge/models/gemma/modeling_gemma4.py`
Core Gemma-4 architecture:
- **`Gemma4TransformerLayer`** — 4-norm residual structure, Phase 4 PLE residual
  block (`gelu(gate(h)) × per_layer_input → proj → norm → residual`), per-layer
  scalar, optional MoE block.
- **`Gemma4SelfAttention`** — GQA with per-head `q_norm`/`k_norm`,
  heterogeneous head_dim (sliding=256, global=512), shared-KV forwarding.
- **`Gemma4DenseRotaryEmbedding`** — Dual-theta RoPE: sliding layers use
  θ=10 000, global layers use θ=1 000 000 with partial rotation (factor=0.25).
- **`wire_gemma4_kv_sharing()`** — Post-construction wiring of shared-KV source
  references for the last `num_kv_shared_layers` layers.

#### `src/megatron/bridge/models/gemma/gemma4_provider.py`
- **`Gemma4DenseProvider`** — All-in-one dataclass provider for clean MCore:
  builds `GPTModel`, attaches dual RoPE, PLE modules, and wires KV sharing.
  `activation_func` uses `fast_gelu` (registered as `"gelu_pytorch_tanh"` in
  `ACTIVATION_FUNC_MAP`) so `megatron_to_hf_activation` can match it by
  identity — avoids `"Unsupported activation function"` on HF export.

#### `src/megatron/bridge/models/gemma_vl/modeling_gemma4_vl.py`
[...]

PR #4196, opened 2026-06-07
[model] fix: Match HF proportional RoPE shape in Gemma 4 global attention

jeje-soy-michu:
# What does this PR do ?

Fixes the global-attention rotary embedding in `Gemma4RotaryEmbedding` so its
`inv_freq` matches Hugging Face's proportional RoPE layout byte-for-byte. This
affects **all Gemma 4 variants** (MoE 26B-A4B, Dense 31B, and VL), since they all
build the global RoPE through `Gemma4ModelProvider.provide()`.

# Changelog

- `gemma4_provider.py`: pass `rotary_percent=1.0` to `RotaryEmbedding.__init__`
  (was `global_rotary_percent=0.25`) so the base builds machinery on the full `head_dim`.
- `gemma4_provider.py`: construct `inv_freq` as `cat(64 real freqs, zeros(head_dim//2 - 64))`
  so `rotate_half` pairs lanes `(i, i + head_dim//2)` exactly as HF does.
- `tests/unit_tests/models/gemma/test_gemma4_provider.py`: add
  `TestGemma4RotaryEmbeddingProportionalShape` asserting the global `inv_freq` length
  (`head_dim//2`), the real/zero-pad split (64 + 192), and the proportional-denominator
  values. Verified to fail on the pre-fix code.

# Scope

- Single shared code path: `Gemma4RotaryEmbedding` is instantiated only in
  `Gemma4ModelProvider.provide()`. All variants route through it:
  - MoE `gemma-4-26B-A4B`
  - Dense `gemma-4-31B`
  - Text / VL
- Only **global / `full_attention`** layers are affected; local sliding layers use a
  separate full-rotary `rope_local` (`rope_type="default"`) and are unchanged.

# Reference (HF official)

Reference implementation: `_compute_proportional_rope_parameters` in 🤗 transformers
`src/transformers/modeling_rope_utils.py`:

```python
rope_angles = int(rope_proportion * head_dim // 2)        # 64
inv_freq_rotated = 1.0 / (base ** (arange(0, ...) / head_dim))
nope_angles = head_dim // 2 - rope_angles                 # 192
inv_freq = cat([inv_freq_rotated, zeros(nope_angles)])    # length head_dim//2 = 256
```

Docstring: *"proportional RoPE will always return an encoding that is the size of
`head_dim`."* Mainline produced length 64 instead of 256.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci)
in the Contributing doc for how to trigger the CI. A Nvidia developer will need to approve
and trigger the CI for external contributors.

**Pre checks**:

- [x] Did you write any new necessary tests? (inv_freq-shape regression test)
- [x] Did you add or update any necessary documentation? (n/a)
- [x] Does the PR affect components that are optional to install? (No)

# Additional Information

- Related to #4195

### DIFF
--- a/src/megatron/bridge/models/gemma/gemma4_provider.py
+++ b/src/megatron/bridge/models/gemma/gemma4_provider.py
@@ -653,17 +653,22 @@ class Gemma4RotaryEmbedding(RotaryEmbedding):
         super().__init__(
             kv_channels=global_kv_channels,
             rotary_base=rotary_base,
-            rotary_percent=global_rotary_percent,
+            rotary_percent=1.0,
             **global_kwargs,
         )
 
         # Fix global inv_freq to match HF's proportional RoPE formula.
         # HF proportional: inv_freq = 1/(base^(arange / head_dim)) not 1/(base^(arange / dim))
         # where dim = int(head_dim * percent) and head_dim = global_kv_channels
+        # HF also zero-pads inv_freq to length head_dim//2 so rotate_half pairs (i, i+head_dim//2).
         dim = int(global_kv_channels * global_rotary_percent)  # 128
         device = self.inv_freq.device
-        self.inv_freq = 1.0 / (
-            rotary_base ** (torch.arange(0, dim, 2, dtype=torch.float32, device=device) / global_kv_channels)
+        self.inv_freq = torch.cat(
+            [
+                1.0
+                / (rotary_base ** (torch.arange(0, dim, 2, dtype=torch.float32, device=device) / global_kv_channels)),
+                torch.zeros(global_kv_channels // 2 - dim // 2, dtype=torch.float32, device=device),
+            ]
         )
 
         # Local RoPE: full rotary with low theta
--- a/tests/unit_tests/models/gemma/test_gemma4_provider.py
+++ b/tests/unit_tests/models/gemma/test_gemma4_provider.py
@@ -17,7 +17,7 @@
 import pytest
 import torch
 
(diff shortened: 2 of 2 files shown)


src/megatron/bridge/models/gemma/gemma4_provider.py
```diff
@@ -653,17 +653,22 @@ def __init__(
         super().__init__(
             kv_channels=global_kv_channels,
             rotary_base=rotary_base,
-            rotary_percent=global_rotary_percent,
```
yaoyu-33, line 656: why is this change though?

jeje-soy-michu: `rotary_percent` is only used by the parent to compute the dim of inv_freq but the forward path derives the rotary width purely from `self.inv_freq`'s length. Since we overwrite `self.inv_freq` a few lines later the value passed here doesn't affect the final RoPE.

I set it to 1.0 (vs the previous 0.25) only so the parent's transient inv_freq is built at the full head_dim and stays consistent with the length we then assign. The actual fix is the torch.cat([real_freqs, zeros(...)]). Happy to add a clarifying comment, or revert this line to global_rotary_percent since it's functionally a no-op once inv_freq is overwritten.

jeje-soy-michu:
No longer needed, the fix was already merged https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4148

PR #4212, opened 2026-06-08
[model] fix: guard GLM fused expert detection

yaoyu-33:
## Summary
- Guard GLM 4.5 and GLM 4.5V fused expert layout detection for config-only bridges.
- Detect fused expert checkpoint layouts with `state.source.has_glob()` instead of enumerating all HF keys.
- Default config-only GLM expert mappings to the current unfused HF layout while preserving fused mapping support when checkpoint state indicates fused keys.

## Testing
- `UV_CACHE_DIR=/tmp/uv-cache-mbridge-precommit-tool uv tool run pre-commit run --all-files`
- `uv sync --group test && uv run --no-sync python -m pytest tests/unit_tests/models/glm/test_glm45_bridge.py tests/unit_tests/models/glm_vl/test_glm_45v_bridge.py -q` (43 passed)

### DIFF
--- a/src/megatron/bridge/models/glm/glm45_bridge.py
+++ b/src/megatron/bridge/models/glm/glm45_bridge.py
@@ -301,36 +301,23 @@ class GLM45Bridge(MegatronModelBridge):
 
         return MegatronMappingRegistry(*mapping_list)
 
-    def _hf_source_and_keys(self):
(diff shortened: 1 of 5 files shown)


cuichenx (commented):
Looks like the L1 conversion test needs to be updated too L1_Launch_models_glm.sh?

PR #4220, opened 2026-06-09
fix: Fix Gemma3-VL HF↔Megatron conversion for transformers ≥ 5.8

suiyoubi:
## Summary

Fixes conversion bugs that surface when running Gemma3-VL conversion with transformers 5.8+
, all rooted in a structural change to `SiglipVisionModel`
introduced in that release.

- **Bug 1 (import — wrong weight mapping):** `ReplicatedMapping` used `hf_param="vision_tower.**"`
  but the HF checkpoint stores vision weights under `vision_tower.vision_model.**`. In transformers
  ≤ 5.6, `SiglipVisionModel` wrapped a `SiglipVisionTransformer` as `self.vision_model`, making
  Megatron's `vision_tower` params naturally carry the `vision_model.*` prefix — accidentally
  matching the checkpoint. In 5.8 the wrapper was removed (flat layout), breaking the accidental
  alignment and causing 11,800+ "no mapping found" warnings. Fix: corrected `hf_param` to
  `"vision_tower.vision_model.**"` and unwrap the transformer in `Gemma3VLModel.__init__` via
  `getattr(_raw_vision, "vision_model", _raw_vision)` so Megatron always stores flat params
  regardless of transformers version.

- **Bug 2 (export — phantom `head.probe` key):** transformers 5.8 added `use_head` logic to
  `SiglipVisionModel`: when `vision_use_head` is absent from config it defaults to `True`, adding
  a `SiglipMultiheadAttentionPoolingHead`. Since `vision_use_head=False` is not part of the
  default `SiglipVisionConfig` schema it is silently dropped after OmegaConf round-trips
  `run_config.yaml`, causing the reconstructed model during export to include `head.probe` in
  its sharded state dict — a key absent from the checkpoint → `KeyError` in
  `_validate_global_shapes`. Fix: set `vision_config.vision_use_head = False` in
  `provider_bridge()` so it is persisted to YAML, with defense-in-depth in
  `Gemma3VLModel.__init__`.

### DIFF
--- a/src/megatron/bridge/models/gemma_vl/gemma3_vl_bridge.py
+++ b/src/megatron/bridge/models/gemma_vl/gemma3_vl_bridge.py
@@ -80,7 +80,10 @@ class Gemma3VLBridge(MegatronModelBridge):
         provider.autocast_dtype = torch.bfloat16
         provider.make_vocab_size_divisible_by = 128
 
-        # Vision configuration
+        # Vision configuration — always disable the pooling head so the checkpoint
+        # and the reconstructed model are consistent (vision_use_head is absent in
+        # the default SiglipVisionConfig, which would otherwise default to True).
+        vision_config.vision_use_head = False
         provider.vision_config = vision_config
         provider.mm_tokens_per_image = hf_config.mm_tokens_per_image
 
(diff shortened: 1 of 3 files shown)


suiyoubi:
/ok to test [b791f87](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4220/commits/b791f87ac3cd4935394855b2cd31273d460a96a1)

cuichenx (approved):

PR #4221, opened 2026-06-09
feat(conversion): support distributed adapter export

cuichenx:
## Summary
- add dtype control for Hugging Face adapter export
- support excluding Megatron adapter base prefixes before HF mapping lookup, used for MTP adapter exclusions
- add a distributed GPU path for exporting large PEFT checkpoints with matching TP/PP/EP/ETP settings
- document the CPU and distributed GPU adapter export workflows

## Blast Radius / Test Assessment
- Touches shared conversion and PEFT adapter export APIs, so this is broader than a model-only example change.
- L0/unit coverage is needed for adapter export and LoRA task construction.
- L1/L2 functional coverage is not added in this PR; a manual large-checkpoint adapter export is the appropriate higher-level validation.

## Validation
- `uv run --no-sync pre-commit run --files examples/conversion/adapter/export_adapter.py src/megatron/bridge/models/conversion/auto_bridge.py src/megatron/bridge/models/conversion/model_bridge.py src/megatron/bridge/models/conversion/peft_bridge.py tests/unit_tests/models/test_adapter_export.py tests/unit_tests/models/test_model_bridge_lora.py examples/conversion/adapter/README.md`
- `uv run --no-sync python -m py_compile examples/conversion/adapter/export_adapter.py src/megatron/bridge/models/conversion/auto_bridge.py src/megatron/bridge/models/conversion/model_bridge.py src/megatron/bridge/models/conversion/peft_bridge.py tests/unit_tests/models/test_adapter_export.py tests/unit_tests/models/test_model_bridge_lora.py`

Focused pytest was attempted with `uv run python -m pytest tests/unit_tests/models/test_adapter_export.py tests/unit_tests/models/test_model_bridge_lora.py -q`, but local dependency resolution failed before test collection because `nvidia-resiliency-ext==0.6.0` has no compatible wheel for this host platform.


src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -133,6 +133,32 @@ def _drop_readonly_config_properties(
     return {key: value for key, value in config_dict.items() if key not in readonly_properties}
 
 
+_ADAPTER_EXPORT_DTYPE_ALIASES: dict[str, torch.dtype] = {
+    "bf16": torch.bfloat16,
```
yaoyu-33: i think we have such utils and mapping somewhere already.

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1332,6 +1378,8 @@ def export_adapter_ckpt(
         peft_checkpoint: str | Path,
         output_path: str | Path,
         show_progress: bool = True,
+        dtype: str | torch.dtype = torch.float32,
```
yaoyu-33: why dtype needed? why different than regular model config?

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1332,6 +1378,8 @@ def export_adapter_ckpt(
...
+        exclude_adapter_base_prefixes: Iterable[str] | None = None,
```
yaoyu-33, line 1351: can you explain a bit more of why this is needed, I tend to not add too many kwargs if can handle implicitly.

src/megatron/bridge/models/conversion/auto_bridge.py
cuichenx: good catch, will replace

src/megatron/bridge/models/conversion/auto_bridge.py
cuichenx: let me remove it

src/megatron/bridge/models/conversion/auto_bridge.py
cuichenx: This is needed because we add LoRA adapters to MTP layers but there's no MTP implementation in HF (hence no MTP adapter implementation). MTP adapters are still useful if lora is merged back to the model, so we keep that but only exclude MTP adapters during export.

cuichenx: This is explicit rather than implicit because silently dropping adapter weights by model-specific convention could hide mistakes I think

cuichenx: I don't have strong opinions here. We can also make it implicit and maybe print out ignoring any layer with "mtp" in the name

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -757,9 +789,15 @@ def save_hf_adapter(
         if dist.is_initialized():
             dist.barrier()
 
+        output_dtype = _normalize_adapter_export_dtype(output_dtype)
         raw_adapter_weights = [
-            HFWeightTuple(exported_weight.param_name, exported_weight.weight.clone().float())
-            for exported_weight in self.export_adapter_weights(model, cpu=True, show_progress=show_progress)
+            HFWeightTuple(exported_weight.param_name, exported_weight.weight.clone().to(output_dtype))
```
yaoyu-33: does this need detach? again why this output_dtype needed, not just follow main param's precision?

cuichenx: removed

yaoyu-33 (approved):

PR #4233, opened 2026-06-09
ci: smoke-import installed packages in docker build stages

ko3n1g:

## Background / motivation

- The 26.06 FW container shipped `apache-tvm-ffi 0.1.12rc0` (a pre-release). Its `tvm_ffi` registry is API-incompatible with the TVM that `tilelang 0.1.8` vendors, so `import tilelang` — and therefore `import mamba_ssm` (which eagerly imports `Mamba3 → tilelang`) — crashes with `AttributeError: attribute '__dict__' of 'type' objects is not writable`.
- `pip check` passed: the version range was nominally satisfied. It validates declared ranges but never imports, so an installed-but-unimportable package ships undetected.

## What changed

- **`docker/common/import_check.py`** — stdlib-only smoke importer: enumerates installed distributions, resolves each one's top-level modules, imports each in an isolated subprocess, exits non-zero on any failure.
- **`docker/common/import_check_skip.txt`** — justified exemptions for modules that cannot import at build time (no GPU/driver/host libs).
- **One `RUN` gate per Dockerfile** — `Dockerfile.fw_base`, `Dockerfile.ci`, `Dockerfile.fw_final`.

## Details

- Runs via `--mount=type=bind` so the script + skip-file are not baked into the image layer.
- Subprocess-per-import isolates a segfault/hang (per-import timeout) so one bad module can't abort the sweep.
- Module discovery derives names from each distribution's recorded files (not stale `top_level.txt`), strips ABI tags from extension filenames, and keeps only valid identifiers — no spurious candidates.
- **Skip-list policy**: a Dockerfile `RUN` builds without a GPU, so `deep_ep`, `deep_ep_cpp`, `hybrid_ep_cpp`, `torch_tensorrt` (need `libcuda`/driver), `mfusepy` (needs `libfuse`), and `rec2idx` (legacy DALI/mxnet shim) are exempted with justification. Any **other** import error fails the build.

```mermaid
flowchart TD
  b[Dockerfile.fw_base] --> gb[RUN import_check.py]
  c[Dockerfile.ci] --> gc[RUN import_check.py]
  f[Dockerfile.fw_final] --> gf[RUN import_check.py]
  gb --> q{any import fails?}
  gc --> q
  gf --> q
  q -- yes --> x[build fails]
  q -- no --> ok[image OK]
```

## Tested

Validated the **committed** script + skip-file against the broken `26.06.rc5` image:

- Broken venv (`apache-tvm-ffi 0.1.12rc0`):
  ```
  FAIL  mamba_ssm
  FAIL  tilelang
  import check: 604 ok, 2 failed, 6 skipped, 606 total   # exit 1 -> build fails
  ```
- Fixed venv (`apache-tvm-ffi==0.1.11`) + skip-file:
  ```
  import check: 606 ok, 0 failed, 6 skipped, 606 total   # exit 0 -> build passes
  ```

> **Draft note:** the skip-list is seeded from the `fw_final` package set. `fw_base`/`ci` have different package sets and may surface a few additional build-time-only (GPU/host-lib) imports; those will be appended to the skip-list as the first CI build reports them.

</details>

### DIFF
--- a/docker/Dockerfile.ci
+++ b/docker/Dockerfile.ci
@@ -111,3 +111,17 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     uv cache prune ${UV_CACHE_PRUNE_ARGS}
 
 COPY --chmod=644 . /opt/Megatron-Bridge
+
+##############################################################################
+##
+## Verify the environment imports
+##
+##############################################################################
+
+# Fail the build if any installed package in /opt/venv cannot be imported. A
+# version/ABI skew can leave a package installed yet unimportable, which
+# `pip check` does not catch. The build has no GPU, so modules that need a
+# driver or host library at import are exempted in import_check_skip.txt.
+RUN --mount=type=bind,source=docker/common/import_check.py,target=/opt/import_check.py \
+    --mount=type=bind,source=docker/common/import_check_skip.txt,target=/opt/import_check_skip.txt \
+    python /opt/import_check.py --jobs 16 --skip-file /opt/import_check_skip.txt
--- a/docker/Dockerfile.fw_base
+++ b/docker/Dockerfile.fw_base
@@ -147,3 +147,17 @@ RUN --mount=type=bind,from=vllm_wheel,source=/src/vllm/,target=/tmp/vllm/ \
 FROM ${FW_BASE_FINAL} AS nemo_fw_base_final
 
 WORKDIR /opt
+
+##############################################################################
+##
+## Verify the environment imports
+##
+##############################################################################
+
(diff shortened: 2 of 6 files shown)


ko3n1g:
Added a second commit pinning `apache-tvm-ffi==0.1.11` in `docker/common/fw_pyproject.toml`. This is the root-cause fix: the FW container's `uv sync` is unlocked with `prerelease = "allow"`, which floated the dep to the broken `0.1.12rc0`. The pin is required for the `fw_final` import-check gate to pass (without it, the gate correctly fails on `import tilelang`/`import mamba_ssm`).

thomasdhc (approved):

PR #4235, opened 2026-06-09
feat(ckpt): complete Energon dataloader checkpointing save/restore

### LINKED ISSUE #3993: [feature] Complete Energon dataloader checkpointing: expose `dataloader_save` and implement restore
### User problem

Megatron-Bridge can **save** Energon `SavableDataLoader` state during checkpointing, but the feature is only half-wired and the corresponding **restore** is missing entirely. As a result, resuming a run reloads weights and optimizer correctly but the Energon dataloader restarts from an arbitrary position in the stream causing sample overlap/skipping and non-reproducible data.

### Desired outcome

This issue asks for end-to-end Energon dataloader checkpointing: a discoverable save config and an automatic restore on resume. 

### Alternatives considered

We've implemented this downstream as a training callback and verified it produces bit-identical resume; we'd like the same behavior supported natively in Bridge.

## Current state (v0.4.0)

There are three gaps:

### 1. `dataloader_save` is undiscoverable

`maybe_save_dataloader_state` (in `training/checkpointing.py`) reads its target path via:

```python
getattr(cfg.dataset, "dataloader_save", None)
```

But `dataloader_save` is not a declared field on `DatasetProvider`, `DataloaderConfig`, or any config dataclass. So saving silently no-ops unless a user happens to know they must monkey-set the attribute on their provider. Nothing surfaces this in the config schema or docs.

### 2. `EnergonProvider.build_datasets` strips `save_state`

`maybe_save_dataloader_state` calls `train_iterator.iterable.save_state()`. For `dataloader_type="external"`, `RerunDataIterator` wraps whatever `build_datasets` returns. But `EnergonProvider.build_datasets` returns:

```python
return (
    iter(dataset.train_dataloader()),   # <- iter() strips the EnergonDataloader interface
)
```

`iter(EnergonDataloader)` returns a bare iterator without `save_state`, so even when `dataloader_save` *is* set, the save raises:

```
RuntimeError: Could not find a save_state for the train_iterator of type
<class 'megatron.core.rerun_state_machine.RerunDataIterator'>
```

Returning the `EnergonDataloader` directly (it's already iterable, with `__next__`/`__iter__`) fixes this.

### 3. Restore is not implemented

`maybe_save_dataloader_state` writes per-DP-rank files to `{dataloader_save}/iter_{step}/train_dataloader_dprank{NN}.pt`, but nothing ever reads them back. There is no `maybe_load_dataloader_state`, `_load_checkpoint_from_path` does not restore dataloader state, `EnergonDataloader` has `save_state` but no `restore_state`, and `CheckpointLoadContext` / `CallbackContext` don't expose the train data iterator. (Upstream Megatron-LM has the restore in `examples/multimodal/dataloader_provider.py`, but it was never ported into Bridge's core checkpoint path.)

## Requested changes

1. **Declare `dataloader_save` (and a `dataloader_load`/auto-derive) as real config fields** on the dataset provider / `DataloaderConfig`, so it's discoverable and documented rather than a `getattr` convention.

2. **Fix `EnergonProvider.build_datasets`** to return the `EnergonDataloader` objects directly (drop the `iter(...)` wrap) so `save_state` survives through `RerunDataIterator.iterable`.

3. **Add `EnergonDataloader.restore_state(state)`** symmetric to `save_state`:
   ```python
   def restore_state(self, state):
       self._dataloader.restore_state_rank(state)
       self._iter = iter(cyclic_iter(self._dataloader))  # rebuild so the active iterator picks up restored state
   ```

4. **Add `maybe_load_dataloader_state(...)`** symmetric to `maybe_save_dataloader_state`, called from the checkpoint-load path after data iterators are built. It should locate `{load}/.../iter_{restored_step}/train_dataloader_dprank{dp_rank}.pt`, `torch.load` it, and call `restore_state` on the train iterator's underlying `EnergonDataloader`. Missing file → warn and continue (allows resuming pre-feature checkpoints).

## Determinism caveat worth documenting

For restore to succeed, the task encoder must be **deterministic per-`restore_key`**. Energon's restore replays in-flight / packing-buffer samples via `restore_sample(restore_key)` and re-runs them through the task encoder; if re-encoding diverges (e.g. RNG-dependent augmentation/placement with `@stateless(restore_seeds=False)`), restore raises `Unexpected skip sample during restoration`. We hit this and resolved it by setting `restore_seeds=True` on our task encoders. It would help to surface this requirement in the Energon-integration docs.

## What we did downstream (reference implementation)

We implemented all of the above in a downstream training callback (`on_train_start`): it reads `{cfg.checkpoint.load}/energon/iter_{step:07d}/train_dataloader_dprank{NN:03d}.pt`, reaches the `EnergonDataloader` via the provider, and calls `restore_state_rank`. We worked around gap nr. 3's missing `CallbackContext` data-iterator access by stashing the dataloader reference on the provider during `build_datasets`.

Verified on Qwen3.5-VL-2B resumed from step 100 reproduces the original data stream **bit-for-bit** with identical `lm loss`.

We're happy to upstream this as a PR if the maintainers agree on the approach (config fields + `build_datasets` fix + `restore_state` + `maybe_load_dataloader_state`).

## Environment

- Megatron-Bridge v0.4.0 (via Nemo image v26.04)
- Megatron-Energon v7.3.0
- Multi-modal training (Qwen3.5-VL-2B), `dataloader_type="external"`, Energon sequence packing enabled

### Affected area

area:ckpt

### Urgency / use case

Important but not blocking

### Extra context

_No response_

rob-luke:
# What does this PR do ?

Completes end-to-end Energon dataloader checkpointing: makes the save path a discoverable config field and implements the previously-missing restore, so a resumed run continues over the same data stream instead of restarting from an arbitrary position.

# Changelog

- Declare `dataloader_save` and `dataloader_load` as real fields on `DataloaderConfig`, replacing the undocumented  `getattr(cfg.dataset, "dataloader_save", None)` convention so the save path is   discoverable in the config schema.
- Fix `EnergonProvider.build_datasets` to return the train-split  `EnergonDataloader` un-wrapped (drop the `iter(...)`), so the downstream  `RerunDataIterator.iterable` retains `save_state`/`restore_state`.
- Add `EnergonDataloader.restore_state(state)`, symmetric to `save_state`:   restores the underlying loader and rebuilds the cyclic iterator.
- Add `maybe_load_dataloader_state(...)`, called from `setup()` after the data   iterators are built. The load path is `dataloader_load`, falling back to   `checkpoint.load` when unset; a missing state file is tolerated (warns and   continues) so checkpoints saved without dataloader state still resume.
- Tests: `EnergonDataloader.restore_state`; `build_datasets` train-split   preservation; and the `maybe_load_dataloader_state` branch matrix (no path /   no iterator / non-Energon iterable / non-first-rank / missing file / restore).

# GitHub Actions CI

External contributor — a NVIDIA developer will need to approve and trigger CI via `/ok to test <sha>`.

**Pre checks**:

# Additional Information

- Closes #3993


zsombor-kaiko:
Thanks for picking this up @rob-luke. The shape (config fields + un-wrap train split + `restore_state` + `maybe_load_dataloader_state` in setup) matches what we run downstream. Few notes:

1. Probably a maintainer can verify this but I think there's an issue with the restore gating. Specifically, `maybe_load_dataloader_state` early-returns unless `pp.rank()==0 and tp.rank()==0`, mirroring `maybe_save_dataloader_state`. However, the two aren't symmetric:
  - Save gates to `(pp0,tp0)` to avoid writing redundant files as the per-DP-rank state is identical across all model-parallel ranks in a DP group, so writing it once is enough.
  - Load has to restore every rank that pulls from the iterator. In the case of [`qwen3_vl_step.get_batch`]( https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/models/qwen_vl/qwen3_vl_step.py#L126), all tp and pp ranks seem to pull from the iterator. 

2. I think the dataloader checkpointing could be a default behavior. I cannot really imagine a scenario when a user would want to save mcore checkpoints but restart the dataloaders upon loading that mcore checkpoint. In the current state, the user would always need to specify explicitly a path if he wanted energon checkpointing. Therefore, I would suggest defaulting `dataloader_save` to `{mcore_ckpt_path}/energon` and same for `dataloader_load`. 

3. I would personally also fail loudly if checkpoint is not found rather than just printing `f"no dataloader checkpoint at {data_state_load_path}; dataloader starts fresh"` (those prints are hard to spot among all other logs) but I understand it would also fail previously saved checkpoints which lacked the energon checkpoint. As a middle ground, we could pass through if the `energon` subfolder is completely missing from the mcore checkpoint but fail if the subfolder exists but a file can't be loaded. This would catch issues if someone saves a training with `DP=n` but tries to load with `DP=m`. 

4. Another thing to note in `restore_state` / `maybe_load_dataloader_state` (and the feature docs?) is the requirement of `restore_key` determinism. See the [issue](https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/3993) for why this is needed.

rob-luke:
Thanks for the feedback @zsombor-kaiko 

It sounds like you have considered this issue carefully. If you have a working version I'd be happy to close this PR in favor of your complete version? Or I can make the changes here and tag you.

zsombor-kaiko:
@rob-luke I had this issue on my todos and was literally jumping on it when I saw your PR 😄

The current implementation I have is only a downstream callback whose functionality is essentially equivalent to this PR, with the differences that I highlighted in my previous comment. Since I can't push to your branch directly, the most straightforward path is if you apply those changes in this PR (they should be really straightforward to fix). Though it would be good if a maintainer could also share their opinions before you do those changes. Maybe @yaoyu-33?

src/megatron/bridge/training/checkpointing.py
```diff
@@ -1633,6 +1633,53 @@ def maybe_save_dataloader_state(
...
+        return
+    # VPP yields a list of iterators; the save side handles only the single-iterator case.
+    if isinstance(train_iterator, list):
+        return
+    iterable = getattr(train_iterator, "iterable", None)
+    if iterable is None or not hasattr(iterable, "restore_state"):
+        return
+
+    # Only the ranks that actually load data (pp/tp rank 0) restore, one file per DP rank.
+    is_first_rank = (pg_collection.pp.rank() == 0) and (pg_collection.tp.rank() == 0)
```
yaoyu-33, line 1673: seems not fully accurate, I feel we have some utils in megatron core utils? can you maybe check. tp / pp is not enough basically. we have cp for example

rob-luke:
Thanks @zsombor-kaiko and @yaoyu-33 for the feedback. I will implement your feedback, I've also found another issue locally that I want to fix. As such, I've marked this as draft and will ping you when its ready for review again (ETA early next week).

rob-luke:
@zsombor-kaiko you had a great improved approach in https://github.com/kaiko-ai/Megatron-Bridge/pull/3, and I love the detailed validation. Will you push this change upstream to this Nvidia repo? Or would you prefer to push to this PR that I wrote (I tried to give you access)? Or would you like me to fold these changes in to the existing PR?

I think your work is great, so I am very happy to close my PR and have you open one. But happy to do whatever you prefer.

rob-luke:
Closing in favor of https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4521

PR #3464, opened 2026-04-22
Explicit error when the number of training samples < global batch size

OlegSudakov:
# What does this PR do ?

When `len(train_dataset) < global_batch_size`, training currently hangs silently instead of failing. This PR adds an explicit precondition check in `build_train_valid_test_data_loaders` that raises a `RuntimeError`, avoiding a hang.

# Changelog

- `src/megatron/bridge/data/loaders.py`: after building the train dataset, raise `RuntimeError` if `len(train_ds) < cfg.train.global_batch_size`, reporting both values in the message.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci)in the Contributing doc for how to trigger the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

# Before your PR is "Ready for review"

**Pre checks**:

- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?

If you haven't finished some of the above items you can still open "Draft" PR.

# Additional Information

- Related to # (issue)

PR #4258, opened 2026-06-09
[data] fix: Handle empty PP-rank finetuning data

cuichenx:
## Summary
- Guard the train-dataset size check so it only runs on ranks that actually build `train_ds`
- Let batch finetuning pass through `data_iterator=None` with the default sequence length
- Add regression coverage for both `train_ds=None` and `data_iterator=None`

## Context
NVBug 6181306 reproduces in `nvcr.io/nvidian/nemo:26.06.rc5` with GPT-OSS SFT at `TP=2 PP=2 EP=4`. Ranks on pipeline stage 1 receive `train_ds=None`, which is valid for non-data-loading PP ranks.

PR #3464 added the explicit undersized-dataset RuntimeError, but the new condition calls `len(train_ds)` before checking whether `train_ds` exists on the current rank. That turns the intended check into `TypeError: object of type 'NoneType' has no len()` on PP ranks without a local dataset.

This preserves #3464 behavior for real datasets while keeping the existing `None` dataset contract used by `build_pretraining_data_loader()`.

## Validation
- `git diff --check`
- `python -m py_compile src/megatron/bridge/data/loaders.py src/megatron/bridge/data/finetuning.py tests/unit_tests/data/test_finetuning.py tests/functional_tests/test_groups/data/test_loaders.py`
- Blocked locally: `uv run python -m pytest tests/unit_tests/data/test_finetuning.py tests/functional_tests/test_groups/data/test_loaders.py::TestDataLoaders::test_build_train_valid_test_data_loaders_allows_none_train_dataset` fails during dependency resolution because `nvidia-resiliency-ext==0.6.0` has no wheel for this workstation platform (`manylinux_2_31_x86_64`).
- Blocked locally: `uv run pre-commit run --all-files` fails for the same dependency-resolution/platform issue.

### DIFF
--- a/src/megatron/bridge/data/finetuning.py
+++ b/src/megatron/bridge/data/finetuning.py
@@ -90,11 +90,11 @@ def split_batch_into_microbatches(
 
 
 def prepare_finetuning_batch(
-    data_iterator: Iterator,
+    data_iterator: Iterator | None,
     num_microbatches: int,
     default_seq_length: int,
     seq_key: str = "tokens",
-) -> tuple[Iterator, int]:
+) -> tuple[Iterator | None, int]:
     """Prepare a finetuning batch by getting global batch and splitting into microbatches.
 
     This function handles the finetuning-specific data flow:
@@ -125,6 +125,9 @@ def prepare_finetuning_batch(
         >>> batch1 = next(microbatch_iter)
         >>> batch1['tokens'].shape  # torch.Size([4, 240])
     """
+    if data_iterator is None:
+        return data_iterator, default_seq_length
+
     # Get full global batch from dataloader
     global_batch = next(data_iterator)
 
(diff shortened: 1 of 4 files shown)


yaoyu-33 (approved):

gautham-kollu:
Retrying L0_Launch_models_nemotron_omni jobs

```
8:10:17.7297510Z INFO:tests.functional_tests.conftest:Test data downloaded successfully.
2026-06-10T18:10:17.7299315Z [32mINFO    [0m tests.functional_tests.conftest:conftest.py:45 Test data downloaded successfully.
2026-06-10T19:00:07.3029284Z ##[error] Test timed out after 50 minutes. Please reach out to the PIC.
2026-06-10T19:00:07.3110991Z exit_code=124
2026-06-10T19:00:07.3113219Z ##[endgroup]
```

PR #4261, opened 2026-06-09
fix(training): route MTP inputs from model chunk

yaoyu-33:
## Summary
- route MTP batch fields from the current `GPTModel` chunk instead of inferring from PP/VPP layout rank state
- keep the layout/rank helper as a fallback for direct `get_batch(..., include_mtp_inputs=None)` callers
- add unit coverage for VPP stages where only the model chunk owns the MTP block

## Why
GitLab job 336688788 failed in DeepSeek V3 GB200 MXFP8 perf with PP=4, VPP=4, `mtp_num_layers=1`, and a layout ending in `mtp,loss`. The failing ranks reached MTP embedding with `input_ids=None`, producing:

```text
TypeError: embedding(): argument indices (position 2) must be Tensor, not NoneType
```

PR 4208 inferred whether to include MTP inputs from pipeline layout and the global virtual pipeline rank. During interleaved VPP, Bridge does not get the active `vp_stage` in `forward_step_func`, while the MCore model chunk already carries the correct `mtp_process` flag. Using the chunk flag avoids routing token ids to the wrong stage.

## Validation
- `uv run --no-sync ruff check src/megatron/bridge/training/gpt_step.py tests/unit_tests/training/test_gpt_step.py`
- `uv run --no-sync ruff format --check src/megatron/bridge/training/gpt_step.py tests/unit_tests/training/test_gpt_step.py`
- cw toy smoke, job 12683919: PP=4, VPP=2, layout `Et|t|t|t|t|t|t|mL`, `mtp_num_layers=1`, 2 train iterations completed with `mtp_1 loss`, exit 0

Local full `uv run python -m pytest ...` was blocked by local dependency resolution for `nvidia-resiliency-ext==0.6.0` on the host platform; the cw smoke ran inside the standard training container.

### DIFF
--- a/src/megatron/bridge/data/iterator_utils.py
+++ b/src/megatron/bridge/data/iterator_utils.py
@@ -14,7 +14,6 @@
 
 """Iterator utilities for handling virtual pipeline parallelism."""
 
-import queue
 from typing import Iterator, TypeVar, Union
 
 
@@ -29,8 +28,8 @@ def make_data_iterator_list(
     With interleaved/virtual pipeline parallelism, Megatron expects a list of one data
     iterator per model chunk. Each model chunk independently gets data from its data
     iterator, so we need to interact with the data iterator multiple times for each
-    microbatch step. Instead of incorporating this logic into the data loader, we cache
-    the iterator's output to the first model chunk and reuse it in the other model chunks.
+    microbatch step. Instead of incorporating this logic into the data loader, we lazily
+    cache the source iterator by microbatch index and expose one cache view per model chunk.
 
     Args:
         model: List of model chunks (when virtual PP is used) or single-element list
@@ -39,8 +38,8 @@ def make_data_iterator_list(
     Returns:
         If model has only 1 chunk: returns the iterator as-is
         If model has multiple chunks: returns a list of iterators with caching behavior
-            - First iterator in list consumes from data_iterator and caches values
-            - Remaining iterators are proxies that read from the cache
+            - Each iterator can advance independently
+            - All iterators see the same microbatch sequence regardless of scheduler order
 
     Example:
         >>> # With virtual PP size = 2 (model has 2 chunks)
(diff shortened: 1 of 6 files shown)


yaoyu-33:
Validation summary for the minimized cache/VPP fix at `836b8a86`:

- Reproduced the old queue-cache failure in VPP training:
  - `PP=2, VPP=2, TP=1, layers=4, GBS=2`, mock GPT pretrain, bf16, seq=128, 3 train iters.
  - Old cache fails before any loss is produced with `_queue.Empty` from the proxy iterator.
  - This matches the failure mode the new `SharedCache` avoids: a later VP/model chunk can request its batch before chunk 0 has populated the queue.
- New cache passes multiple VPP training smokes using the same recipe/config family:
  - `PP=2, VPP=2, layers=4, GBS=2`: losses are exactly repeatable: `1.039034E+01`, `1.039359E+01`, `1.039239E+01`.
  - `PP=2, VPP=2, layers=8, GBS=4`: losses are exactly repeatable: `1.039246E+01`, `1.040035E+01`, `1.038797E+01`.
  - `PP=2, VPP=2, layers=4, GBS=4`: same-node back-to-back runs are exactly repeatable: `1.039144E+01`, `1.040887E+01`, `1.038595E+01`.
- Also reran the MTP/VPP smoke shape that motivated this PR:
  - `PP=4, VPP=2`, MTP custom layout with MTP on the final stage, 2 train iters.
  - Completed successfully and logged `mtp_1 loss` instead of hitting the old `NoneType` embedding input failure.

Timing note: I do not see evidence of a step-time regression from this cache change. The short 3-iteration smokes are dominated by first-iteration warmup. For the same-node `PP=2,VPP=2,layers=4,GBS=4` repeat, step times were:

- run 1: `5549.9 ms`, `45.7 ms`, `41.3 ms`
- run 2: `5323.2 ms`, `43.3 ms`, `38.4 ms`

The steady-state iteration deltas are a few milliseconds and look like normal runtime noise for these tiny smokes. Old cache has no meaningful before timing for the VPP failure shapes because it fails before producing losses.

cuichenx (approved):
LGTM

PR #4131, opened 2026-06-03
[recipe,docs,test] feat: add DeepSeek-V4-Flash SFT recipes, launcher and tests

Meirtz:
## What

Adds **full SFT support for DeepSeek-V4-Flash** (≈285B / 13B-active MoE + MLA + MTP) to Megatron-Bridge, validated end to end on **8×GB300 (32 GPU, TP1/PP4/EP8)** with the real `deepseek-ai/DeepSeek-V4-Flash` weights.

Megatron-Core is **unchanged** — the shipped SFT path uses only already-merged DSv4 prerequisites.

## Contents

- **Recipes** (`recipes/deepseek/deepseek_v4.py`):
  - `deepseek_v4_flash_sft_config` — MTP on
  - `deepseek_v4_flash_no_mtp_sft_config` — MTP off (drops the MTP layer; trims `csa_compress_ratios` back to `num_layers`)
  - Both are **hardware-agnostic** (Adam / bf16 / SBHD; fused mHC on Blackwell, see README). One recipe set runs unchanged on Hopper and Blackwell — there is no separate "blackwell" SFT recipe.
- **Launcher** (`examples/models/deepseek_v4/slurm_sft.sh`): end-to-end HF→Megatron import + full SFT, recipe selected by `MTP=on|off` (`HARDWARE=blackwell|hopper` only sets node topology). Mirrors the existing `examples/models/gpt_oss/slurm_sft.sh`.
- **Tests**:
  - Unit build tests (`tests/unit_tests/recipes/test_deepseek_recipes.py`) — no GPU/HF/network; assert the SFT recipes build, are hardware-agnostic, and that `no_mtp` disables MTP.
  - Toy-model functional finetune smoke (`tests/functional_tests/test_groups/recipes/test_deepseek_recipes_finetune.py`) — 2-layer toy, 5 finetune iters; wired into the gb200 + h100 **L0** launch scripts (skips unless the DSv4 toy model + mcore DSv4 prereqs are present).
- **Docs** (`examples/models/deepseek_v4/README.md`): SFT usage + a Status/Blockers table. The launcher uses a generic `CONTAINER_IMAGE` (any image with the DSv4 prerequisites), mirroring `examples/models/gpt_oss/slurm_sft.sh`.

## Validation (8×GB300, real weights)

| Path | Result |
|------|--------|
| HF→Megatron import (FP8/MXFP4→bf16) + full SFT, **MTP on** | `lm loss` decreasing, no NaN |
[...]

PR #4271, opened 2026-06-10
fix(model): keep full rotary_percent for DeepSeek-V4 MLA rope

Meirtz:
## What

One-line fix: force `rotary_percent = 1.0` in the DeepSeek-V4 bridge, plus a regression test.

## Why — root cause of the DSv4-Flash SFT NaN (NVIDIA/Megatron-LM#4468)

HF's `partial_rotary_factor` (0.125) is **relative to `head_dim` (512)**. The generic `partial_rotary_factor → rotary_percent` mapping applies it on top of `qk_pos_emb_head_dim` (64) — which **already encodes the rope split** (448 nope + 64 rope). The double-applied factor shrinks the rope cos/sin cache to 64 × 0.125 = **8 dims**, and then:

- the **unfused** path silently rotates only 8 of 64 rope dims (no crash — masked correctness issue on every DSv4 bridge run);
- the **fused** MLA yarn-rope kernel (`fused_mla_rope_inplace`, called with `emb_dim=64`) reads cos/sin **out of bounds**, producing huge garbage activations (q jumps ~5 → ~1e14, captured in-situ) → the iteration-2 `found NaN in local forward loss calculation` in DeepSeek-V4-Flash SFT, reproduced on both H100 and GB300.

Debug trail (single-variable isolations, in-situ capture, and replay) is in [NVIDIA/Megatron-LM#4468](https://github.com/NVIDIA/Megatron-LM/issues/4468). Pretrain never hit it because the mock data path uses fixed-length sequences and the validated pretrain configs didn't exercise the same cache shape.

## Validation (GB300, proxy DSv4-Flash SFT, EP4)

| Run | rotary_percent | apply_rope_fusion | Result |
|---|---|---|---|
| control | 0.125 (status quo) | False | trains, but rotates 8/64 dims |
| repro | 0.125 | True | **NaN at iteration 2** |
| fixed | **1.0** | False | clean, 10/10 iters |
| fixed | **1.0** | True | **clean, matches unfused control to 4 decimals** (12.46808→7.554 vs 12.46882→7.5535) |

Unit test: mocks the parent `provider_bridge` to inject `rotary_percent=0.125` (what the generic mapping produces) and asserts the DSv4 bridge forces it back to 1.0.

## Notes

- Related: #4131 (DSv4-Flash SFT recipes; independent — different hunks of the same file, no stacking needed). Also unblocks re-enabling `apply_rope_fusion` for DSv4 SFT (verl-project/verl#6603 ships it off as a workaround).
- A follow-up suggestion for Megatron-Core (tracked in #4468): `fused_mla_rope_inplace` should assert `cos.shape[-1] == emb_dim` instead of reading out of bounds.
- AI-assisted (Claude); debugging, validation runs, and review by the human author.

### DIFF
--- a/src/megatron/bridge/models/deepseek/deepseek_v4_bridge.py
+++ b/src/megatron/bridge/models/deepseek/deepseek_v4_bridge.py
@@ -346,6 +346,12 @@ class DeepSeekV4Bridge(MegatronModelBridge):
         # head_dim = 512 (nope_dim + rope_dim = 448 + 64)
         provider.v_head_dim = hf_config.head_dim  # 512
         provider.qk_pos_emb_head_dim = hf_config.qk_rope_head_dim  # 64
+        # HF's partial_rotary_factor (0.125) is relative to head_dim (512); the rope split is
+        # already fully encoded by qk_pos_emb_head_dim (64). The generic partial_rotary_factor
+        # -> rotary_percent mapping would shrink the rope cache to 64*0.125 = 8 dims: the
+        # unfused path then silently rotates only 8 of 64 rope dims, and the fused MLA rope
+        # kernel reads cos/sin out of bounds (garbage values -> the SFT loss NaN).
+        provider.rotary_percent = 1.0
         # qk_head_dim and kv_lora_rank derived automatically in DSv4HybridConfig
         provider.q_lora_rank = hf_config.q_lora_rank  # 1024
         provider.o_groups = hf_config.o_groups  # 8
--- a/tests/unit_tests/models/deepseek/test_deepseek_v4_bridge.py
+++ b/tests/unit_tests/models/deepseek/test_deepseek_v4_bridge.py
@@ -340,3 +340,57 @@ class TestMTPEHProjSplit:
             elif isinstance(hf_param, dict):
                 for v in hf_param.values():
                     assert "eh_proj" not in v, f"deprecated eh_proj reference found in hf_param dict value: {v}"
+
+
+class TestDeepSeekV4RotaryPercent:
+    """Regression: HF partial_rotary_factor (relative to head_dim=512) must not shrink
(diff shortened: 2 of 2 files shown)


yaoyu-33 (approved):

src/megatron/bridge/models/deepseek/deepseek_v4_bridge.py
```diff
@@ -346,6 +346,12 @@ def provider_bridge(self, hf_pretrained: PreTrainedCausalLM) -> MLAModelProvider
         # head_dim = 512 (nope_dim + rope_dim = 448 + 64)
         provider.v_head_dim = hf_config.head_dim  # 512
         provider.qk_pos_emb_head_dim = hf_config.qk_rope_head_dim  # 64
+        # HF's partial_rotary_factor (0.125) is relative to head_dim (512); the rope split is
```
yaoyu-33, line 349: is this a different definition issue of HF vs. Megatron, if so, how can we avoid in the future? we can do a auto mapping still. or guard with a check

Meirtz: Yes — definition mismatch: HF's `partial_rotary_factor` is relative to `head_dim` (512×0.125 = the 64 rope dims), while Megatron's `rotary_percent` applies to `qk_pos_emb_head_dim` (64), which already *is* the rope dim — so for MLA it gets applied twice.

For the future, both of your options work: (1) keep the auto-mapping but skip it for MLA providers (`multi_latent_attention=True`), where the factor is always redundant — happy to send that as a follow-up; (2) guards: `fused_mla_rope_inplace` asserting `cos.shape[-1] == emb_dim` (suggested in NVIDIA/Megatron-LM#4468), so the misconfig fails loudly instead of OOB-reading.

Issue #4215, opened 2026-06-08
[feature] LongLiveWan Long-Video Training Recipe


sbhavani:
### User problem

Megatron Bridge currently has WAN video diffusion training support using Megatron Core, Transformer Engine, sequence packing, context parallelism, distributed optimizer, and
  WAN checkpoint conversion. This covers standard WAN flow-matching training over video latents.

  What is missing is LongLive-style long-video generation training, where a single long video is split across GPUs by temporal chunk and each rank owns both:

  - clean history tokens for its chunk
  - noisy target tokens and loss for the same chunk

  This avoids the imbalance of generic sequence sharding, where some ranks may hold mostly clean context while others hold most loss-bearing target tokens.

### Desired outcome

Add a minimal `LongLiveWan` recipe in Megatron Bridge that supports LongLive-2.0 style balanced long-video AR training for WAN.

### Affected area

area:recipe

### Urgency / use case

Nice to have

## Proposed Scope

  - Add a `longlive_wan` recipe/config.
  - Add a `longlive_wan_ar_step` training step.
  - Support paired clean/noisy temporal chunks per CP rank.
  - Construct the teacher-forcing attention mask for clean/noisy AR video training.
  - Compute loss only on noisy target tokens, with correct distributed reduction.
  - Reuse existing Megatron Bridge WAN infrastructure where possible
  
## Initial Implementation Target

  Start with offline WAN latents and text embeddings, matching the existing Bridge WAN data path.

  A later extension can add online raw-video VAE encoding with temporal halo exchange so each rank VAE-encodes only its local video window plus left halo.

  ## Expected Value

  The recipe would enable:

  - training on single long videos that do not fit on one GPU
  - balanced target-loss ownership across ranks
  - clean/noisy AR teacher-forcing semantics
  - a path toward LongLive-2.0 style scalable long-video generation training

suiyoubi:
Thanks for opening this. We have a related implementation proposal in [#4272](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4272). If you have bandwidth, could you please give it a try on your target LongLive WAN workflow and let us know whether it matches what you need?

huvunvidia:
Hi @sbhavani ,
Due to limited bandwidth, we currently not having LongLive style training on our roadmap. 
[...]

PR #4272, opened 2026-06-10
feat(diffusion): add LongLive WAN training path

AndysonYs:
# What does this PR do ?

Adds the initial offline-latents LongLive WAN MVP requested in #4215, covering clean-history plus noisy-target temporal chunks with windowed attention defaults and SP/TP validation.

# Changelog

- Add `longlive_wan_step` registration so WAN recipes can select the LongLive forward step from `scripts/training/run_recipe.py`.
- Add LongLive WAN chunk selection, target-only loss masking, teacher-forcing mask helpers, and CP/SP partition utilities.
- Add `LongLiveWanForwardStep` and `LongLiveWanFlowMatchingPipeline` for clean-history plus noisy-target temporal chunk training.
- Add LongLive WAN 1.3B and 5B SP long-video recipes with explicit sliding-window attention settings to avoid dense `[S, S]` masks for long sequences.
- Update WAN mock/dataset config paths to support LongLive latent shape overrides used by the long-video recipe.
- Keep the Megatron-Core submodule clean by making explicit dense self-attention masks opt-in only when the decoder supports `self_attention_mask`.
- Add focused unit tests for LongLive chunking, noising, masking, recipe wiring, and dense-mask/windowed-attention selection.
- Add `scripts/validation/wan_sp_tp_tiny_parity.py` to verify tiny WAN TP/SP inference parity with exact tensor equality.
- Document the LongLive WAN MVP commands and the 5B SP long-video smoke path in the WAN example README.

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to #4215.
- This PR implements the initial offline WAN latents/text embeddings target from #4215; online raw-video VAE encoding with temporal halo exchange remains a future extension.
- Does not add new required or optional dependencies.
- `pre-commit run --all-files` passed.
- `python -m compileall -q scripts/validation/wan_sp_tp_tiny_parity.py src/megatron/bridge/diffusion/models/wan/longlive_wan_step.py src/megatron/bridge/diffusion/models/wan/longlive_wan_utils.py src/megatron/bridge/diffusion/models/wan/wan_model.py src/megatron/bridge/diffusion/recipes/wan/wan.py tests/unit_tests/diffusion/model/wan/test_longlive_wan_step.py tests/unit_tests/diffusion/recipes/wan/test_wan_recipe.py` passed.
- Slurm unit job `3244385`: 35 passed, 26 warnings in 3.56s on 4x GB200.
- Slurm tiny SP/TP parity job `3244425`: `strict_equal=True`, `max_abs=0.00000000e+00`.
- Slurm long-video smoke job `3244426`: completed 1/1 iteration on 4x GB200 with 0 skipped and 0 NaN iterations.
- Local `uv` was unavailable in the conda environment, so pre-commit was run directly after installing `pre-commit` into the existing `mb-longlive-wan` environment.


yaoyu-33:
Could you please check the default LongLive 1.3B recipe? It looks internally inconsistent: `longlive_wan_1_3b_pretrain_config()` inherits the text-to-video config with `context_parallel_size=4` and the WAN provider default `qkv_format="thd"`, but `LongLiveWanFlowMatchingPipeline.validate_qkv_format()` rejects anything except `qkv_format="sbhd"`.

Please make the default recipe runnable, or mark it as non-runnable and update the README/tests accordingly.

liding-nv:
@yaoyu-33 can you review the latest version?

AndysonYs:
Thanks for catching this. I updated the 1.3B recipe to use the supported CP=1 and qkv_format="sbhd" settings and documented it as a configuration template because the full sequence exceeds the current dense-mask limit. This PR primarily targets the LongLive 2.0 5B TP/SP path, so the README now includes avalidated reduced-resolution four-GPU smoke command using the recipe defaults (TP=4, SP enabled, and CP=1). All four TP ranks initialized successfully and completed one full training iteration.

AndysonYs:
> Could you please check the default LongLive 1.3B recipe? It looks internally inconsistent: `longlive_wan_1_3b_pretrain_config()` inherits the text-to-video config with `context_parallel_size=4` and the WAN provider default `qkv_format="thd"`, but `LongLiveWanFlowMatchingPipeline.validate_qkv_format()` rejects anything except `qkv_format="sbhd"`.
> 
> Please make the default recipe runnable, or mark it as non-runnable and update the README/tests accordingly.

@yaoyu-33  Thanks for catching this. I updated the 1.3B recipe to use the supported CP=1 and qkv_format="sbhd" settings and documented it as a configuration template because the full sequence exceeds the current dense-mask limit. This PR primarily targets the LongLive 2.0 5B TP/SP path, so the README now includes avalidated reduced-resolution four-GPU smoke command using the recipe defaults (TP=4, SP enabled, and CP=1). All four TP ranks initialized successfully and completed one full training iteration.

aroshanghias-nvd:
Thanks for the update. This PR is currently conflicting with `main`. Please rebase and resolve the conflicts so maintainers can continue the review and run the full CI suite on the current branch.

src/megatron/bridge/diffusion/models/wan/longlive_wan_step.py
```diff
@@ -0,0 +1,359 @@
...
+
+def _build_packed_seq_params(
+    qkv_format: str,
+    seq_len_q: torch.Tensor,
+    seq_len_q_padded: torch.Tensor,
+    seq_len_kv: torch.Tensor,
+    seq_len_kv_padded: torch.Tensor,
+) -> dict[str, PackedSeqParams]:
+    zero = torch.zeros(1, dtype=torch.int32, device=seq_len_q.device)
+    cu_seqlens_q = torch.cat((zero, seq_len_q.cumsum(dim=0).to(torch.int32)))
```
yaoyu-33, line 88: Blocking: these cumulative lengths describe every clean/noisy chunk as a separate sequence, but the tensors passed to Transformer Engine remain `sbhd` with `B=1`. For `sbhd`, TE defines `cu_seqlens` as shape `batch_size + 1`. The pinned fused backend truncates this tensor to `B + 1`, which leaves only `[0, first_chunk_end]`; the flash backend instead derives packing indices using `len(cu_seqlens) - 1` as the batch size and can index beyond the actual `B*S` storage. This also makes each chunk a hard packed boundary, contrary to the required noisy-to-previous-clean attention. The cross-attention metadata has the same mismatch. Please keep the self-attention metadata consistent with the real batch layout (for one unpadded sample, `[0, total_S]`) and represent packed-sample/cross-attention isolation separately, with a real pinned-TE forward test containing more than one temporal chunk. Relevant TE contracts: https://github.com/NVIDIA/TransformerEngine/blob/d64bc14dc87eb658ab98839e4b7687595ee53e2d/transformer_engine/pytorch/attention/dot_product_attention/backends.py#L1959-L1967 and https://github.com/NVIDIA/TransformerEngine/blob/d64bc14dc87eb658ab98839e4b7687595ee53e2d/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L1821-L1841

src/megatron/bridge/diffusion/models/wan/longlive_wan_step.py
```diff
@@ -0,0 +1,359 @@
...
+            qkv_format=qkv_format,
+            seq_len_q=seq_len_q,
+            seq_len_q_padded=seq_len_q_padded,
+            seq_len_kv=seq_len_kv,
+            seq_len_kv_padded=seq_len_kv_padded,
+        )
+        global_target_tokens = int(batch["loss_mask"].float().sum().item())
+        use_explicit_mask = self.should_build_explicit_self_attention_mask(paired_latents.size(1), model)
+        if use_explicit_mask:
+            batch["self_attention_mask"] = build_paired_teacher_forcing_self_attention_mask(
```
yaoyu-33, line 231: Blocking: this dense block mask is not enforced by the optimized TE path selected by the WAN layer spec. `get_wan_block_with_transformer_engine_spec()` constructs both attentions with `AttnMaskType.padding`. In pinned TE, FusedAttention forwards only `attn_mask_type` plus cumulative lengths to the kernel, not this arbitrary boolean matrix; FlashAttention uses a padding mask only to derive cumulative lengths when they are absent, but this code always supplies them. A smoke can therefore complete without implementing the clean/noisy visibility rules. Please use an attention implementation with actual block-sparse/arbitrary-mask semantics (as LongLive 2.0 does with FlexAttention), or otherwise demonstrate a supported TE representation, and add an integration test that changes a forbidden key/value and proves the corresponding output is invariant.

src/megatron/bridge/diffusion/recipes/wan/wan.py
```diff
@@ -322,3 +331,96 @@ def wan_1_3b_text2video_pretrain_config() -> ConfigContainer:
...
+
+def longlive_wan_5b_sp_long_video_pretrain_config() -> ConfigContainer:
+    """Return a LongLive-style WAN 5B SP config for long-video development.
+
+    This follows the LongLive AR streaming default latent shape
+    `[B, F, C, H, W] = [1, 320, 48, 44, 80]` with random/mock WAN tensors when
+    `dataset.path` is unset. Sequence parallelism is enabled through TP=4 and
+    CP is disabled so long-video sharding is specifically covered by SP.
+    """
+    cfg = wan_14b_pretrain_config()
```
yaoyu-33, line 367: Blocking: this recipe is named and documented as LongLive/Wan2.2 5B, but it instantiates the existing WAN 14B provider and leaves its 40 layers, 5120 hidden size, 13824 FFN size, and 40 heads unchanged. Updating only the latent channels does not produce the 5B architecture, so the released 5B checkpoint cannot load and the mock smoke exercises a randomly initialized 14B-shaped model. The upstream Wan2.2-TI2V-5B config uses 30 layers, hidden size 3072, FFN size 14336, and 24 heads: https://github.com/NVlabs/LongLive/blob/main/wan_5b/configs/wan_ti2v_5B.py. Please configure the actual 5B dimensions/provider and validate checkpoint compatibility, or rename the recipe and documentation if a 14B model is intended.

scripts/training/run_recipe.py
```diff
@@ -111,6 +111,7 @@
 
 # Diffusion forward steps: use class instances so they can be passed as forward_step_func
 from megatron.bridge.diffusion.models.flux.flux_step import FluxForwardStep
+from megatron.bridge.diffusion.models.wan.longlive_wan_step import LongLiveWanForwardStep
```
yaoyu-33, line 114: This registration is incomplete at the current PR head and also collides with the new launcher on `main`. The advertised command resolves recipes from `megatron.bridge.recipes`, but the two LongLive builders are exported only from `megatron.bridge.diffusion.recipes.wan`; `src/megatron/bridge/recipes/wan/__init__.py` does not expose them, so `load_recipe()` fails before training. During conflict resolution, please port this to the current lazy `scripts/training/recipe_runner.py` step registry, add LongLive default-step metadata in `recipe_metadata.py` (the existing `wan_` prefix does not match `longlive_wan_`), and place/export the recipe through the current WAN recipe package. A launcher-level test should exercise both `load_recipe()` and `load_forward_step()` from current `main`; taking this older `run_recipe.py` wholesale would discard the new launcher.

src/megatron/bridge/diffusion/models/wan/longlive_wan_step.py
```diff
@@ -0,0 +1,359 @@
...
+                chunks=chunks,
+                device=latents.device,
+            )
+        else:
+            batch.pop("self_attention_mask", None)
+
+        context = FlowMatchingContext(
+            noisy_latents=paired_latents,
+            latents=paired_latents,
+            timesteps=timesteps,
```
yaoyu-33, line 242: Blocking: the paired clean and noisy tokens receive the same sampled timestep conditioning here. `sample_timesteps()` returns one value per batch item, and WAN computes one `e/e0` per batch item and expands it across the entire sequence in every AdaLN block and the output head. Consequently, a clean latent copy is modulated as though it were at the noisy target timestep. LongLive teacher forcing instead builds separate conditioning: clean history uses `aug_t=0`, while noisy targets use the sampled `t` (see https://github.com/NVlabs/LongLive/blob/main/wan_5b/distributed/sequence_parallel.py#L318-L336). Please carry timestep metadata aligned with the paired token/chunk layout and update WAN AdaLN/head conditioning to consume it, with a test that observes zero-timestep conditioning on clean tokens and sampled-timestep conditioning on noisy tokens.

yaoyu-33 (changes requested):
Requesting changes because the current path does not implement the stated LongLive training semantics. The focused helper/provider/recipe/mock-data tests pass (43/43), but they stop above Transformer Engine. The pinned-TE source contract and a GPU backend reproducer show that the optimized `padding` path does not enforce the dense clean/noisy block mask, while the chunk-level `cu_seqlens` for an `sbhd`, `B=1` tensor make all but the first chunk padding (or can produce invalid flash packing indices). Separately, clean history receives the noisy target timestep, the named 5B recipe retains the 14B architecture, and the recipe/step is not integrated with the current launcher/export architecture on `main`. Please address the five inline findings and add a real backend-level forward/backward test before treating a one-iteration smoke as validation. I have not triggered CI or attempted a conflict rewrite because the current head is conflicting and the core training semantics need revision first.

src/megatron/bridge/diffusion/models/wan/longlive_wan_step.py
```diff
@@ -0,0 +1,359 @@
...
+        latents = batch["video_latents"].to(device, dtype=dtype)
+        batch_size = latents.shape[0]
+        data_type = batch.get("data_type", "video")
+        task_type = self.determine_task_type(data_type)
+
+        chunks = select_longlive_paired_chunks(
+            grid_sizes=batch["grid_sizes"],
+            seq_len_q=batch["seq_len_q"],
+            seq_len_q_padded=batch["seq_len_q_padded"],
+            target_chunk_frames=self.target_chunk_frames,
```
yaoyu-33, line 171: Follow-up: `chunk_selection_strategy` is now a dead configuration option. The constructor stores it, but this call to `select_longlive_paired_chunks()` does not pass a strategy (and that helper has no strategy parameter), so `"random"`, `"first-valid"`, and even invalid strings all produce the same all-chunks layout. The current test only checks that the string was stored, not behavior. If natural paired teacher forcing intentionally uses every chunk, please remove this option and the obsolete single-target selection helpers/tests; otherwise wire the strategy into selection and add a behavioral test showing the two supported modes differ and invalid values are rejected.

src/megatron/bridge/diffusion/models/wan/longlive_wan_step.py
yaoyu-33: Runtime follow-up: I reproduced this on a GPU with the repository-locked `transformer-engine 2.16.0+d64bc14d`. For an `sbhd` tensor with `B=1`, `S=4`, passing PR-style chunk metadata `cu_seqlens=[0,1,2,3,4]` produced nonzero output rows `[True, False, False, False]`; every row after the first chunk was zero. The focused reproducer exited successfully. This is direct backend evidence for the cumulative-length mismatch, not a claim of full-model validation.

src/megatron/bridge/diffusion/models/wan/longlive_wan_step.py
yaoyu-33: Runtime follow-up: the same focused GPU reproducer completed against the exact repository-locked `transformer-engine 2.16.0+d64bc14d`. With valid full-sequence cumulative lengths, changing the supplied boolean `padding` mask from allow-all to diagonal-only produced identical attention output (`max_abs_diff=0.0`). This directly confirms that this optimized path does not enforce the arbitrary LongLive visibility matrix; it is not a proxy for end-to-end training correctness.

PR #4301, opened 2026-06-11
feat(model): thread weight_dtype through HF export for plain-dtype DeepSeek-V4 output

Meirtz:
## What

Thread `weight_dtype: Optional[torch.dtype] = None` through the HF export path — `export_hf_weights` / `save_hf_pretrained` / `stream_weights_megatron_to_hf` — carried per-task via a new optional `WeightConversionTask.weight_dtype` field. When set, the DeepSeek-V4 bridge emits **plain weights in that dtype** (no `*.scale` companions) instead of re-creating the source repo's quantized layout. Default (`None`) keeps today's behavior. CLI: `--export-weight-dtype` on the export subcommand.

## Why

DSv4 HF export unconditionally re-creates the source repo's quantized weight/scale layout (`maybe_modify_converted_hf_weight` → `requantize_hf_weight_scale_pairs`, from #3969). That's right for checkpoint conversion, but bf16-SFT'd weights get silently post-hoc quantized — a user found `*.scale` tensors in their SFT export and asked about train/inference parity.

**Design (revised after reviewer feedback):** the requantize hook runs on *both* export consumers — online weight streaming to rollout engines (`export_hf_weights`, e.g. verl RL weight sync) and on-disk checkpoints (`save_hf_pretrained`) — so a bridge-level boolean cannot configure them independently. A dtype-typed parameter on each public API lets callers choose per path (e.g. bf16 to rollout for RL parity, quantized to disk for serving-format artifacts, or vice versa). Hook signatures are unchanged (the dtype rides on the task), so the other bridges overriding this hook (dsv3, gemma4, kimi, mimo, flux) are unaffected; DSv3 can adopt the same field later.

## Verified

- the saver streams only yielded tensors (omitting `.scale` keys is safe); exported `config.json` is built fresh (`torch_dtype: bfloat16`, no quantization fields); safetensors index regenerated from written tensors;
- unit tests cover dtype-set (plain weights, non-float tensors untouched) and default (requantize) paths.

## Notes

- AI-assisted (Claude); analysis, validation and review by the human author.

### DIFF
--- a/src/megatron/bridge/models/conversion/model_bridge.py
+++ b/src/megatron/bridge/models/conversion/model_bridge.py
@@ -19,7 +19,7 @@ import itertools
 import logging
 import math
 import re
-from dataclasses import dataclass, replace
+from dataclasses import dataclass
 from typing import (
     Any,
     Callable,
@@ -120,8 +120,8 @@ class WeightConversionTask(Generic[MappingT]):
             sub-module that owns the parameter (required for loads).
         param_weight (Optional[torch.Tensor]): The actual parameter tensor that will
             receive the converted weight (required for loads).
-        weight_dtype (Optional[torch.dtype]): When set, bridges that re-create a quantized
-            source layout on export emit plain weights in this dtype instead.
+        weight_dtype (Optional[torch.dtype]): Export only. Cast float weights to this
+            dtype; bridges that requantize on export skip it (no scale companions).
 
     """
 
@@ -899,6 +899,18 @@ class MegatronModelBridge(
         """
         return converted_weights_dict
 
+    @staticmethod
+    def _cast_export_weight_dtype(
+        weights: Dict[str, torch.Tensor], weight_dtype: Optional[torch.dtype]
+    ) -> Dict[str, torch.Tensor]:
+        """Cast float export weights to ``weight_dtype`` (no-op if None; ints untouched)."""
+        if weight_dtype is None:
+            return weights
+        return {
+            name: (weight.to(weight_dtype) if weight.is_floating_point() else weight)
+            for name, weight in weights.items()
+        }
+
     def _accumulate_grouped_export(
         self,
         task: "WeightConversionTask",
@@ -1254,10 +1266,16 @@ class MegatronModelBridge(
         unwrapped_model_list = unwrap_model(megatron_model)
         # Use provided conversion tasks or build them
         if conversion_tasks is None:
-            conversion_tasks = self.build_conversion_tasks(hf_pretrained, unwrapped_model_list)
-        if weight_dtype is not None:
-            # WeightConversionTask is frozen — rebuild the tasks with the dtype set
-            conversion_tasks = [replace(task, weight_dtype=weight_dtype) for task in conversion_tasks]
+            conversion_tasks = self.build_conversion_tasks(
+                hf_pretrained, unwrapped_model_list, weight_dtype=weight_dtype
+            )
+        elif weight_dtype is not None:
+            # Prebuilt tasks bypass build_conversion_tasks (where weight_dtype is recorded);
+            # fail loudly instead of silently dropping it (this also rejects the fp8-tasks combo).
+            raise ValueError(
+                "weight_dtype is not supported with caller-supplied conversion_tasks; "
+                "omit conversion_tasks so the dtype can be recorded at task-build time."
+            )
 
         # Collect adapter conversion tasks when merge is requested
         adapter_tasks_by_base: Dict[str, List[AdapterWeightConversionTask]] = {}
@@ -1312,6 +1330,7 @@ class MegatronModelBridge(
                     task, converted_weights_dict, model_config, _grouped_buffers, hf_state_dict
                 )
                 if merged_result is not None:
+                    merged_result = self._cast_export_weight_dtype(merged_result, task.weight_dtype)
                     for hf_name, tensor in merged_result.items():
                         yield HFWeightTuple(hf_name, tensor.cpu() if cpu else tensor)
                 continue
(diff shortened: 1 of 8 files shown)


Meirtz:
Reworked per reviewer feedback (offline discussion): the hook serves **two** export consumers — online weight streaming to rollout engines and on-disk checkpoints — so the bridge-level boolean is gone. Now `weight_dtype: Optional[torch.dtype]` on `export_hf_weights` / `save_hf_pretrained`, carried per-task (`WeightConversionTask.weight_dtype`), hook signatures unchanged. `d10a8e7e`.

Meirtz:
**Full-model E2E validation** (DeepSeek-V4-Flash, 43 layers, real weights, TP1/PP4/EP8 on 8×GB300; same imported Megatron checkpoint for both runs):

| export | tensors | `.scale` tensors | dtypes |
|---|---|---|---|
| default (quantized) | 69,187 | 34,167 | F8_E4M3, F8_E8M0, I8 (MXFP4), BF16, F32, I32 |
| `--export-weight-dtype bfloat16` | **35,020** | **0** | **BF16, I32** |

35,020 + 34,167 = 69,187 — the bf16 artifact contains exactly every weight with no scale companions (I32 = `tid2eid` routing buffers, correctly left untouched).

Two notes from the run: (1) the smoke caught a real bug in the first version of this PR — `WeightConversionTask` is a frozen dataclass, so the dtype is now applied via `dataclasses.replace` (`ce66e82`), and the unit tests now use real task instances instead of mocks; (2) with `weight_dtype` set, the saver's completeness check counts the source's `.scale` entries as "not written" (export proceeds with `--not-strict`) — cosmetic, can be polished in a follow-up by excluding scale keys from the expected set.

cuichenx (approved):
LGTM

src/megatron/bridge/models/deepseek/deepseek_v4_bridge.py
```diff
@@ -911,8 +911,16 @@ def maybe_modify_converted_hf_weight(
...
+        """Recreate DSv4 quantized weight/scale pairs expected by the source shard index.
+
+        When ``task.weight_dtype`` is set, plain weights are emitted in that dtype
+        instead (no ``*.scale`` companions).
+        """
+        if task.weight_dtype is not None:
+            return {
+                name: (weight.to(task.weight_dtype) if weight.is_floating_point() else weight)
+                for name, weight in converted_weights_dict.items()
+            }
```
HollowMan6: If we implement this logic them this `weight_dtype` will only be effective for dsv4, better to move the logic under

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/cbc1ab97d326da827820004dbab56c209e3e23b4/src/megatron/bridge/models/conversion/model_bridge.py#L1313-L1317

HollowMan6: Then you can check the task `weight_dtype` in `requantize_hf_weight_scale_pairs`, and return from the method (skip the core logic) if this is set

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -1251,6 +1255,9 @@ def stream_weights_megatron_to_hf(
         # Use provided conversion tasks or build them
         if conversion_tasks is None:
             conversion_tasks = self.build_conversion_tasks(hf_pretrained, unwrapped_model_list)
+        if weight_dtype is not None:
+            # WeightConversionTask is frozen — rebuild the tasks with the dtype set
+            conversion_tasks = [replace(task, weight_dtype=weight_dtype) for task in conversion_tasks]
```
HollowMan6: This looks very hacky, I think you can also add an additional `weight_dtype` to `build_conversion_tasks`.

cuichenx (dismissed):
please check comments above

src/megatron/bridge/models/deepseek/deepseek_v4_bridge.py
Meirtz: Thanks @HollowMan6 — both made it cleaner, addressed in `f7c09879`:

1. **weight_dtype is generic now.** The plain-dtype cast lives in the shared `stream_weights_megatron_to_hf` path (`_cast_export_weight_dtype`, float tensors only) and runs on both the standard and grouped-export branches. The DSv4 hook just returns the converted weights unchanged when `weight_dtype` is set and skips requantization.
2. **Dropped the post-hoc `replace()`** — `build_conversion_tasks` stamps `weight_dtype` at construction; `replace()` stays only for caller-supplied prebuilt tasks.

Re-validated on 32×GB300: bf16 = 35020 tensors / 0 scales, quantized = 69187 / 34167 (unchanged).

Heads-up since your rollout uses this export path: after #4338 this export now needs `flash_mla`/`nvidia-cudnn-frontend[cutedsl]` on Blackwell (config-build dep check, since AutoBridge defaults DSA fusion on) even though export runs no attention — worth a follow-up to force it off for export.

Meirtz: Update after full CI: I had to revert the `build_conversion_tasks` change from point 2 — `NemotronHBridge` and `KimiK25VLBridge` override that method with a fixed signature, so passing `weight_dtype` into it broke their export tests (`TypeError: ... unexpected keyword argument`). Went back to stamping the dtype via `replace()` on the built tasks, which is override-safe. The generic cast — your main point — stays. CI's green now.

HollowMan6: > NemotronHBridge and KimiK25VLBridge override that method with a fixed signature

Will that work if you also change the signature of those overriding methods? I think it's not a good idea to use replace in any case in the long run if later we decide to add new args.

Meirtz: Done in `5a97742` — dropped `replace()` entirely. The main path threads `weight_dtype` through `build_conversion_tasks` (NemotronH/KimiVL overrides forward it to `super()`); the prebuilt-tasks path now raises instead of `replace()`-ing, which also rules out the contradictory fp8-export combo. Also reverted a stray `weight_dtype` left on the dispatch stub. CI green.

HollowMan6 (approved):
Thank you @Meirtz ! Now the code logic looks much better and I don't have more comments

PR #4305, opened 2026-06-11
feat(recipe): enable rope fusion in DeepSeek-V4-Flash SFT recipes

Meirtz:
## What

Flip the two DeepSeek-V4-Flash SFT recipes to `apply_rope_fusion=True` (matching the pretrain recipes), correct the README blocker table, and update tests. Promised fast-follow from #4131.

## Why

The "fused yarn-rope NaNs in SFT" blocker is resolved: the root cause was the bridge-side `partial_rotary_factor → rotary_percent` double-application (fixed in #4271), not the kernel. The README row attributing it to the upstream kernel is corrected here.

## Validation (full model)

DeepSeek-V4-Flash, 43 layers, real HF weights, 8×GB300 (TP1/PP4/EP8), 20 SFT iterations each:

| iter | rope OFF (control) | rope ON (this PR's config) |
|---|---|---|
| 1 | 3.600468 | 3.591948 |
| 5 | 3.488342 | 3.502259 |
| 10 | **2.330603** | **2.331909** |

Zero NaN in both; `mtp_1` loss tracks identically (8.0394 vs 8.0393). Proxy-level isolation previously showed bit-level agreement to 4 decimals.

## Notes

- Also restores varlen-relevant rope-fusion perf for downstream consumers (e.g. verl-project/verl#6603, which currently ships `apply_rope_fusion=False` as a workaround and can now flip it).
- AI-assisted (Claude); analysis, validation runs and review by the human author.


cuichenx (approved):
LGTM Thanks for running the experimets

PR #4306, opened 2026-06-11
fix(ckpt): auto-generate DSv4 pipeline layout on export when run_config lacks it

Meirtz:
## What

PP>1 export falls back to `bridge.generate_pipeline_layout(...)` when `run_config.yaml` cannot provide a usable layout — exactly like the import path already does.

## Why (round-trip bug)

`import` → `export` of a DeepSeek-V4 checkpoint at PP>1 currently fails with `AssertionError: pipeline_model_parallel_layout must be set when using hash MoE layers with pipeline parallelism (PP > 1)`.

Root cause: by save time the provider's layout has been finalized into a `PipelineParallelLayerLayout` **object**, and `run_config.yaml` serializes it as an instantiation stub — the actual layout data is lost:

```yaml
pipeline_model_parallel_layout:
  _call_: true
  _target_: megatron.core.transformer.pipeline_parallel_layer_layout.PipelineParallelLayerLayout
```

The export-side reader only accepts a list, finds the stub, and the provider then asserts. Reproduced on a full DeepSeek-V4-Flash TP1/PP4/EP8 checkpoint (imported with this same script); with the fallback the export proceeds.

A deeper fix would be serializing the layout in its list form in `run_config.yaml` (checkpointing owners' call); this script-level fallback restores the import→export round trip now.

## Notes
- AI-assisted (Claude); analysis, repro and review by the human author.

### DIFF
--- a/examples/conversion/convert_checkpoints_multi_gpu.py
+++ b/examples/conversion/convert_checkpoints_multi_gpu.py
@@ -216,6 +216,18 @@ def export_megatron_to_hf(
                     model_provider.pipeline_model_parallel_layout = saved_layout
                     print_rank_0(f"  Read pipeline layout from checkpoint ({len(saved_layout)} stages)")
                     break
+        # run_config.yaml serializes an already-finalized layout as an object stub
+        # (the layout data is lost), so fall back to auto-generation like import does.
+        if not isinstance(getattr(model_provider, "pipeline_model_parallel_layout", None), list) and hasattr(
+            bridge._model_bridge, "generate_pipeline_layout"
+        ):
+            hf_config = bridge.hf_pretrained.config
+            num_layers = hf_config.num_hidden_layers
+            mtp = getattr(hf_config, "num_nextn_predict_layers", 0) or 0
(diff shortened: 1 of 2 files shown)


Meirtz:
Validated end-to-end on 32x GB300 (TP1/PP4/EP8) with real DeepSeek-V4-Flash weights.

The original script-level fallback turned out to be insufficient on its own: `load_megatron_model` rebuilds the provider from the checkpoint's `run_config.yaml`, where the layout is the serialized object stub, so the PP>1 assert still fired inside `build_and_load_model`. The second commit forwards the resolved layout through `mp_overrides` (and declares the field in `ModelParallelKwargs`).

Test flow: fresh HF->Megatron import (run_config carries the stub — confirmed before export), then export with no manual repair. The fallback log line fires (`Auto-generated pipeline layout for PP=4 (43 layers, 1 MTP)`) and the export completes with 69187 tensors / 34167 scales, identical counts to a reference export from a hand-repaired run_config.

cuichenx (approved):

PR #2656, opened 2026-03-05
[bridge] Fix off-by-one in sliding window size for Gemma2, Gemma3, Mistral, and GPT-OSS

cuichenx:
# What does this PR do?

Fix off-by-one error in sliding window attention size for Gemma2, Gemma3, Mistral, and GPT-OSS bridges.

FlashAttention's `window_size` tuple `(left, right)` uses **inclusive** bounds — `(W, 0)` attends to `W + 1` tokens (W preceding + current). Since HuggingFace `sliding_window` defines the total window size (including the current token), we must subtract 1 when converting to the tuple form. This convention is also observed by Transformer Engine.

# Changelog

- Fix Gemma2 bridge to subtract 1 from `sliding_window` (was passing raw HF value)
- Fix Gemma3 provider to subtract 1 when converting int `window_size` to tuple at the layer level (was passing raw int)
- Fix Mistral bridge to subtract 1 from `sliding_window` (was passing raw HF value)
- Fix GPT-OSS bridge to read `sliding_window` from HF config instead of hardcoding 128, and subtract 1
- Fix GPT-OSS predefined provider default from `(128, 0)` to `(127, 0)`

# GitHub Actions CI

See the CI section in the Contributing doc for how to trigger the CI.
A Nvidia developer will need to approve and trigger the CI for external contributors.

# Before your PR is "Ready for review"

Pre checks:
- [x] Make sure you read and followed Contributor guidelines
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?

# Additional Information

Thanks to @returnL for catching this in https://github.com/NVIDIA/Megatron-LM/pull/2771

PR #4308, opened 2026-06-11
fix+feat(gemma2): fix SWA correctness bugs and add FlexAttention fused softcap+SWA path

nvegesna-netizen:
Fixes four silent correctness bugs in Gemma2DotProductAttention that caused sliding window attention (SWA) to either do nothing or produce wrong masks during pretraining and fine-tuning. On top of the corrected unfused path, adds Gemma2FlexDotProductAttention — a fused attention subclass that uses PyTorch FlexAttention (built-in since PyTorch 2.5) to run softcap and SWA in a single Triton kernel, falling back to the fixed unfused parent when a padding mask is present or dropout is active.

  **Changelog**

_gemma2_provider.py_: Fix SWA gate from if attention_mask is not None and self.window_size is not None: to if self.window_size is not None:. The old gate was never True on MCore's pretrain path because MCore passes attention_mask=None, so SWA was silently skipped for every pretrain step. The mask must always be generated for SWA (even-numbered) layers: these layers use attn_mask_type=arbitrary, and FusedScaleMaskSoftmax with arbitrary + mask=None routes through ScaledSoftmax — plain softmax with no causal masking — so omitting the mask for any sequence length (including short ones where the window covers the full sequence) drops causal masking entirely. get_swa() degenerates to a pure causal mask when the window covers all positions, so always building it is correct.

_gemma2_provider.py_: Fix AttnMaskType.causal → AttnMaskType.arbitrary for SWA (even-numbered) layers. FusedScaleMaskSoftmax with causal routes through ScaledUpperTriangMaskedSoftmax.apply(input, scale), which takes no mask argument and silently discards the externally generated SWA mask. Switching to arbitrary routes through ScaledMaskedSoftmax, which applies it correctly. Odd-numbered layers remain causal and keep the fast fused path.

_gemma2_provider.py_: Fix SWA mask dimensionality — get_swa() returns [sq, sk]; the fused CUDA softmax kernel requires a 4D mask. Now unsqueezed to [1, 1, sq, sk] when there is no padding mask, broadcasting correctly to [b, np, sq, sk].

_gemma2_provider.py_: Fix fine-tuning padding mask combination — SWA mask is now ORed with any incoming padding mask (swa_mask | attention_mask) rather than replacing it. Both masks use True=masked-out; [sq, sk] broadcasts to [b, 1, sq, sk].

_gemma2_provider.py_: Fix Gemma2ModelProvider.window_size default from (4096, 0) to (4095, 0). window_size[0] counts prior tokens exclusively, so a 4096-token window requires 4095, consistent with the gemma2_bridge.py round-trip convention.

_gemma2_provider.py_: Replace bare assert statements (stripped by -O) with raise ValueError for CP > 1 and packed-seq guards.

_gemma2_provider.py_: Add Gemma2FlexDotProductAttention — uses PyTorch FlexAttention (built-in, PyTorch 2.5+) to fuse softcap and SWA into a single Triton kernel via score_mod and block_mask. Module-level _get_softcap_score_mod is lru_cache-decorated so all layers with the same softcap share one function object, avoiding N redundant torch.compile recompilations at startup (torch.compile guards on id(fn)). flex_attention itself is wrapped with torch.compile to trigger the fused Triton path. Falls back to the unfused parent when FlexAttention is unavailable, a padding mask is present, or dropout is active. gemma2_layer_spec updated to use this class as core_attention.

_test_gemma2_provider.py_: Add TestGemma2DotProductAttention with 7 tests covering all correctness fixes: CP ValueError, packed-seq ValueError, SWA gate fires on pretrain path (attention_mask=None), odd layers 
test_gemma2_provider.py: Add TestGemma2DotProductAttention with 7 tests covering all correctness fixes: CP ValueError, packed-seq ValueError, SWA gate fires on pretrain path (attention_mask=None), odd layers have no SWA, even layers use AttnMaskType.arbitrary, SWA combined with padding mask via OR, and window_size default. Add TestGemma2FlexDotProductAttention with 12 tests covering fused path invocation, all fallback conditions (no FlexAttention, dropout, padding mask), output shape, softmax scale, score_mod softcap correctness, lru_cache sharing invariant, block mask caching, SWA layer block mask, and packed-seq ValueError.

**Additional Information**

Related to:
- MCore #2357: arbitrary mask type required for externally generated masks
- MCore #752: confirms MCore passes attention_mask=None on the pretrain path
- MB PR #2656: companion off-by-one fix in the bridge conversion layer


yaoyu-33 (dismissed):

suiyoubi:
/ok to test [d2c59a7](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4308/commits/d2c59a791331f1d5ced4371dbd0a67b4875ffefa)

suiyoubi:
/ok to test [22c6d15](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4308/commits/22c6d1531e97073fda06c161d8ba820a6b4cda7e)

suiyoubi:
/ok to test [8e58592](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4308/commits/8e5859215475ca6538545793b9c7f9db6c30cb19)

suiyoubi:
/ok to test [527fd55](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4308/commits/527fd557e72c5dc3133a68acc9c75bd889003729)

suiyoubi:
/ok to test [b7cf2d3](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4308/commits/b7cf2d3ed01cde9bfa0129a3358f0858aedda2a4)

PR #4310, opened 2026-06-11
[examples] fix: Correct GLM-4.5V fine-tuning scripts

cuichenx:
## What
- Fix GLM-4.5V PEFT and SFT Slurm scripts to use the exported recipe names: `glm_45v_peft_config` and `glm_45v_sft_config`.
- Default `PRETRAINED_CHECKPOINT` to the converted iteration checkpoint `${WORKSPACE}/models/GLM-4.5V/iter_0000000` while preserving env override support.
- Update the GLM-4.5V README to document the actual recipes and the expected converted checkpoint path.

## Why
NVBug 6302885 reports that the example scripts pass nonexistent `glm_45v_finetune_config`. I reproduced that failure in `nvidian/nemo:26.06.rc6`. The second PEFT failure reproduces when the parent checkpoint root is passed; the existing loader accepts the direct `iter_0000000` checkpoint directory, so the examples should point to that path.

## Validation
- Reproduced `glm_45v_finetune_config` `AttributeError` in rc6.
- Verified `glm_45v_sft_config` and `glm_45v_peft_config` load in rc6.
- Reproduced checkpoint-root PEFT guard failure and verified direct `iter_0000000` path passes/resolves in rc6.
- `bash -n examples/models/glm/glm_45v/slurm_peft.sh`
- `bash -n examples/models/glm/glm_45v/slurm_sft.sh`
- `git diff --check`
- `uv run pre-commit run --all-files` in rc6 container

### DIFF
--- a/examples/models/glm/glm_45v/README.md
+++ b/examples/models/glm/glm_45v/README.md
@@ -13,7 +13,7 @@ export WORKSPACE=/your/custom/path
 ```
 
 Directory structure:
-- `${WORKSPACE}/models/` - Converted checkpoints
+- `${WORKSPACE}/models/` - Converted checkpoint roots, with iteration directories such as `iter_0000000/`
 - `${WORKSPACE}/results/` - Training outputs and experiment results
 
 ## Checkpoint Conversion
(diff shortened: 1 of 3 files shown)


cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4310/commits/580d2a78cd71e69fd063fee05f3a35be79e4af9b

yaoyu-33 (approved):

PR #4307, opened 2026-06-11
feat(data): add text chat collate for unified HF datasets

yaoyu-33:
## Summary

Stacked on #4169. Follow-up refactor #4315 is stacked on this PR.

- add a text-only chat collate path under the existing HF conversation/VLM dataset utilities
- reuse the #4169 assistant-mask helpers for text conversations (`build_assistant_loss_mask` and `build_shifted_labels_and_loss_mask`)
- add `text_chat` / `chat` maker aliases and tokenizer fallback for plain LLM checkpoints without an AutoProcessor
- leave existing model-specific VLM collates untouched

## Stack

Current order: #4169 -> #4307 -> #4315.

## Validation

- `ruff check` on changed files
- `python3 -m py_compile` on changed files/tests
- `git diff --check`
- standalone `pre-commit run --all-files`

Full `uv run pre-commit run --all-files` was blocked on this workstation before hooks by a platform wheel availability issue for `nvidia-resiliency-ext==0.6.0`.

## Follow-up validation requested

GPU cluster validation is in progress to compare SQuAD loss curves against #4169 using the same config and seed.

Issue #4041, opened 2026-05-28
PoR: [VLM] Refactor: clean up VLM data pipeline (processing / batching / packing / masking)


yaoyu-33:
## Summary

The current VLM data path in Megatron-Bridge has accumulated several design
issues that make it hard to add new VLM models, share code between the Energon
and Hugging Face data sources, and reason about packing / CP behavior.

This issue tracks a staged refactor to:

1. Unify VLM data **processing** between the Energon `TaskEncoder` and the HF
   processor path.
2. Move **batching / padding / packing** out of the VLM forward step and into
   the processing / collate layer.
3. Replace the current text-search **masking** with a more reliable strategy
   (Tony-style token-search or HF `generation` tag).
4. Investigate and fix the **M-RoPE** issue that surfaces when Qwen2.5-VL is
   wired into the VLM step.

It came out of a design review between @yaoyu-33 and @cuichenx.

---

### 1. Three layers are tangled together

Conceptually the VLM data path has three layers:

| Layer | Responsibility |
|-------|----------------|
| **Loading** | Pure load-from-disk. For Energon: `decode_sample`. For HF: dataset iter. |
| **Processing** | `apply_chat_template` -> tokenize -> mask -> `image_processor` / `processor`. Produces per-sample tensors (input_ids, labels, pixel_values, image_grid_thw, …). Per-model. |
| **Batching / packing** | Per-DP-rank decision. Stack/pad to BSHD or concatenate to THD. Computes `target_seq_len`, etc. |

The code today does not respect this layering:

- Some processing (`apply_chat_template`, masking, tokenization) lives in
  `tasks/vlm/task_encoder.py` for the Energon path, and a near-duplicate
  copy lives in the HF encoder task encoder. These should be a single piece
  of code with one `TaskEncoder`-style API.
- The VLM forward step (`forward_step` / `get_batch`) currently does
  *additional* padding-then-unpadding work on samples that were already
  batched and padded by the Energon `batch` collator. The flow is roughly:
  `pad in collator -> remove pad in step -> repack in step`, which is both
  wasteful and confusing.

### 2. "Packing" in the VLM step is not real packing

What is currently called "packing" in the VLM step is **in-microbatch
packing**: at micro-batch time we concatenate everything in the microbatch
into one THD sequence with no bin size and no packing algorithm. That is
not equivalent to LLM-side offline sequence packing.

Implications:

- We cannot pack across microbatches without changing the data path.
[...]

PR #4315, opened 2026-06-12
refactor(data): move VLM sequence batching to collate

yaoyu-33:
Stacked on #4307. Current order: #4169 -> #4307 -> #4315.

Implements issue #4041 Step 2 by moving generic VLM padding, truncation, and in-batch packing preparation from `vlm_step` into collate-time helpers. The VLM training step now consumes already-collated sequence tensors and packed metadata.

Validation status:
- Targeted static checks passed locally: py_compile, ruff check, ruff format --check, and git diff --check.
- Full pre-commit is currently blocked on this workstation by the known nvidia-resiliency-ext wheel/platform resolver issue.
- Internal loss-parity validation against #4307 head is pending; this PR remains draft until that completes.


src/megatron/bridge/data/hf_datasets/conversation_dataset.py
```diff
@@ -58,24 +61,33 @@ def __init__(
                 )
             selected_impl = COLLATE_FNS[collate_key]
```
yaoyu-33: can we just use `collate_impl` instead of selected_impl

src/megatron/bridge/data/hf_datasets/conversation_dataset.py
```diff
@@ -58,24 +61,33 @@ def __init__(
...
+                collate_kwargs["pad_to_multiple_of"] = pad_to_multiple_of
+
+        if enable_in_batch_packing:
             if "pack_sequences" in sig.parameters:
-                pack_kwargs: dict[str, Any] = {"pack_sequences": True}
+                collate_kwargs["pack_sequences"] = True
                 if "pack_sequences_pad_to_multiple_of" in sig.parameters:
-                    pack_kwargs["pack_sequences_pad_to_multiple_of"] = pack_sequences_pad_to_multiple_of
-                selected_impl = partial(selected_impl, **pack_kwargs)
+                    collate_kwargs["pack_sequences_pad_to_multiple_of"] = in_batch_packing_pad_to_multiple_of
```
yaoyu-33: can you update the signature kwargs name, jus use the same one in_batch_packing_pad_to_multiple_of?

src/megatron/bridge/data/hf_datasets/conversation_dataset.py
```diff
@@ -58,24 +61,33 @@ def __init__(
...
+        # kwargs via functools.partial so the DataLoader just calls f(batch, processor).
+        import inspect
+        from functools import partial
 
-            sig = inspect.signature(selected_impl)
+        sig = inspect.signature(selected_impl)
+        collate_kwargs: dict[str, Any] = {}
+        if sequence_length is not None and "sequence_length" in sig.parameters:
+            collate_kwargs["sequence_length"] = sequence_length
+            if "pad_to_max_length" in sig.parameters:
```
yaoyu-33: can we let all collate can take in these? so no need to check signiture?

src/megatron/bridge/data/sequence_packing.py
```diff
@@ -166,3 +170,64 @@ def pack_padded_sequences_in_batch(
...
+        "attention_mask": padding_mask,
+        "position_ids": position_ids,
+    }
+    try:
+        pack_padded_sequences_in_batch(
+            batch,
+            pad_token_id=pad_token_id,
+            pad_to_multiple_of=pad_to_multiple_of,
+        )
+    except ValueError as exc:
```
yaoyu-33: this should fail, I dont understand the fallback.

src/megatron/bridge/data/hf_datasets/conversation_dataset.py
yaoyu-33: Done in 893129ce: the selected collate function is kept as `collate_impl`, and the ConversationDataset wrapper no longer uses a separate `selected_impl` name.

src/megatron/bridge/data/hf_datasets/conversation_dataset.py
yaoyu-33: Done in 893129ce: removed runtime signature inspection from the ConversationDataset path and made the registered collators accept the common collate kwargs directly.

src/megatron/bridge/data/hf_datasets/conversation_dataset.py
yaoyu-33: Done in 893129ce: standardized the collate kwarg on `in_batch_packing_pad_to_multiple_of` across the provider/dataset/collate call chain.

src/megatron/bridge/data/sequence_packing.py
yaoyu-33: Done in 893129ce: removed the all-padding fallback so inconsistent/empty packed batches now fail with the existing `ValueError`; the unit test was updated to expect the failure.

src/megatron/bridge/data/hf_datasets/text_collate.py
```diff
@@ -170,20 +172,26 @@ def text_chat_collate_fn(
         processor: A HF tokenizer or processor. It must expose
             ``apply_chat_template`` directly or through ``processor.tokenizer``.
         max_length: Optional tokenizer truncation length.
+        sequence_length: Optional tokenizer truncation length used by
+            conversation-dataset providers.
         pad_to_max_length: If set with ``max_length``, pad every row to
             ``max_length`` instead of the longest row in the batch.
+        pad_to_multiple_of: Accepted for parity with VLM collate functions.
```
yaoyu-33: then you should use **kwargs

src/megatron/bridge/models/qwen_vl/qwen3_vl_step.py
```diff
@@ -57,6 +57,11 @@ def get_batch_from_iterator(
         dict[str, torch.Tensor]: A dictionary containing the batch data.
     """
     batch = next(data_iterator)
+    if "cu_seqlens" in batch:
```
yaoyu-33, line 60: This guard conflicts with the updated Qwen3-VL packing flag path. The Qwen3-VL example scripts still run with `--step_func qwen3_vl_step` and set `dataset.enable_in_batch_packing=$pack_config`; `ConversationDataset` now forwards that same flag to the collator as `pack_sequences=True`, and the Qwen Energon encoder does the same through `qwen2_5_collate_fn(..., pack_sequences=self.enable_in_batch_packing)`. When `pack_config=True`, the dataloader emits `cu_seqlens`, so this step raises before its own Qwen3-VL packing branch can run. Please either keep Qwen3-VL packing owned by this step and prevent collate-time packing for these configs, or make this step consume the collate-packed metadata instead of rejecting it.

src/megatron/bridge/data/hf_datasets/conversation_dataset.py
```diff
@@ -43,48 +36,40 @@ def __init__(
...
+            "sequence_length": sequence_length,
+            "pad_to_max_length": pad_to_max_length,
+            "pad_to_multiple_of": pad_to_multiple_of,
+            "pack_sequences": enable_in_batch_packing,
+            "in_batch_packing_pad_to_multiple_of": in_batch_packing_pad_to_multiple_of,
+        }
 
         def _bound_collate(batch: list) -> Dict[str, torch.Tensor]:
-            return selected_impl(batch, self._processor)  # type: ignore[call-arg]
+            return collate_impl(batch, self._processor, **collate_kwargs)
```
yaoyu-33, line 107: This unconditional kwargs call breaks the registered Gemma4 HF path. `COLLATE_FNS["Gemma4Processor"]` points to `gemma4_vl_collate_fn`, but that wrapper still only accepts `(examples, processor)`, unlike the other collators. Any Gemma4 `ConversationDataset` batch will fail with `TypeError: got an unexpected keyword argument 'sequence_length'` before it reaches `ministral3_collate_fn`. Please give the Gemma4 wrapper the same common collate kwargs and forward them.

src/megatron/bridge/models/qwen_audio/data/collate_fn.py
```diff
@@ -120,4 +135,14 @@ def qwen2_audio_collate_fn(examples: list, processor) -> dict[str, torch.Tensor]
     for key in ("input_features", "feature_attention_mask"):
         batch.pop(key, None)
 
+    prepare_vlm_batch_for_training(
+        batch,
+        sequence_length=sequence_length,
+        pad_to_max_length=pad_to_max_length,
+        pad_to_multiple_of=pad_to_multiple_of,
+        pack_sequences=pack_sequences,
```
yaoyu-33: For Qwen2-Audio this creates a double-packing path. The collator now packs and emits `cu_seqlens` when `enable_in_batch_packing` is true, but `audio_lm_step.get_batch()` also keys off the same dataset flag and repacks the already-packed `[1, total_len]` tensor, ignoring the collator's existing boundaries. That collapses multiple examples into one packed segment and loses the per-sample attention boundaries. Please either make `audio_lm_step` consume collate-packed metadata, or keep audio packing in the step and avoid calling `prepare_vlm_batch_for_training(..., pack_sequences=True)` here.

tests/unit_tests/training/test_vlm_step.py
```diff
@@ -15,12 +15,12 @@
 import pytest
 import torch
 
+from megatron.bridge.data.sequence_packing import pack_batch_sequences
```
yaoyu-33: After switching this test module to the new `sequence_packing.pack_batch_sequences`, the all-padding `labels=None` test below is stale. `pack_padded_sequences_in_batch` now raises `ValueError("Cannot pack a batch with no non-padding tokens.")`, but `test_packing_none_labels_empty_batch` still expects the old empty packed tensors, so the core unit job will fail. Please update that test to expect the new failure mode as was done for the labels/loss-mask case.

yaoyu-33 (commented):
Found blocking issues in the current head. Also noting that GitHub currently reports this PR as merge-conflicting, so the final merge assessment is limited until the conflicts are resolved.

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4315/commits/d8adaaf10e61c89a3aec10817b1a385c5fdf3712

cuichenx (approved):

PR #4331, opened 2026-06-12
Reverted back to original config for 70B gb200 nvfp4: cg local, scope…

rsalagame-nvidia:
Reverted back to original config for 70B gb200 nvfp4: cg local, scope=mlp, attn --> cg none, scope=full_iteration


bdubauski (approved):
LGTM

PR #4332, opened 2026-06-12
cp: [training] fix: guard optional nvrx_straggler import against non-import errors (#4124) into r0.5.0

cuichenx:
## Summary

Backport of #4124 to `r0.5.0`.

This cherry-picks `ab3eb055fdbf7fedbef75c5cc33e80d81344e577` as `6954b3c45d8d8a1e2a9b637435424cffa812e70a`, preserving the `-x` provenance footer.

The fix broadens the optional `nvidia-resiliency-ext` straggler import guard so import-time failures from transitive optional dependencies disable NVRx straggler detection instead of breaking unrelated Bridge training/conversion imports.

## Validation

- `git diff --check HEAD~1..HEAD`
- `uv run --no-sync python -m py_compile src/megatron/bridge/training/nvrx_straggler.py`

Note: direct package import in the isolated release worktree was not used as validation because the worktree was not dependency-synced / Megatron-Core was not on the editable path.

### DIFF
--- a/src/megatron/bridge/training/nvrx_straggler.py
+++ b/src/megatron/bridge/training/nvrx_straggler.py
@@ -22,19 +22,30 @@ from megatron.bridge.training.config import NVRxStragglerDetectionConfig
 from megatron.bridge.utils.import_utils import MISSING_NVRX_MSG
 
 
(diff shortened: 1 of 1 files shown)


cuichenx:
needed for deepseek v4 dev branch switch

PR #4347, opened 2026-06-14
fix(model): preserve tp_comm_overlap config during Qwen3VL MTP initialization

### LINKED ISSUE #4227: bug(model): Qwen3VL MTP initialization silently disables TP comm overlap for all decoder layers
## Summary

When `mtp_num_layers > 0` is set for Qwen3.5-VL (dense or MoE), tensor parallel
communication overlap (`tp_comm_overlap`) is silently disabled for all decoder
layers, even when the user explicitly enables it via `CommOverlapConfig`.

## Root Cause

The bug stems from the interaction between MCore's MTP initialization and
`Qwen3VLGPTModel`'s decoder rebuild pattern. The sequence is:

1. `Qwen3VLGPTModel.__init__` calls `super().__init__(..., mtp_block_spec=mtp_block_spec)`
2. Inside MCore's `GPTModel.__init__`, the standard decoder block is initialized first
   — at this point `config.tp_comm_overlap=True` and TP overlap is correctly configured.
3. MCore then initializes the MTP block. The MTP transformer layers have userbuffer
   names (`ub_name`) that are not registered in the `tp_comm_overlap_cfg` preset
   (which only covers standard decoder layer name prefixes). MCore/TE detects this
   mismatch and writes `config.tp_comm_overlap = False` back to the **shared config
   object** as a fallback.
4. `super().__init__()` returns. The shared `config.tp_comm_overlap` is now `False`.
5. `Qwen3VLGPTModel.__init__` rebuilds the decoder as `Qwen3VLTransformerBlock`
   (required for mRoPE support), passing `config=self.config`. Since the config is
   already polluted, `Qwen3VLTransformerBlock` and all its `TELayerNormColumnParallel`
   layers are initialized with `tp_comm_overlap=False`. All overlap-related configs
   (e.g. `ub_name`) are `None`.
6. The final runtime decoder has TP overlap completely disabled, with no warning
   or error surfaced to the user.

### Minimal repro

```shell
1. use any qwen3.5-vl recipe
2. run with tp overlap open
3. get a profiler trace.
```

### Expected behavior

gemm overlap with kuserbuffer push recv.

### Affected area

area:perf

### Regression?

No

### Environment

run on H20 . commit cc3a7d88e54599de691b8f3d56c069fafdd3d554 

### Logs

```shell

```

SophusDavid:
# What does this PR do ?

Fix a bug where Qwen3VL MTP initialization silently disables TP comm overlap for all decoder layers by passing a deepcopy of config to super().__init__(), preventing MTP-induced config pollution from propagating to the decoder rebuild.

# Changelog

- src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/text_model.py

  -  Add from copy import deepcopy import

  - In Qwen3VLGPTModel.__init__, save _original_config = config and pass _config_copy = deepcopy(config) to super().__init__() instead of the original config
  - After super().__init__() returns, restore self.config = _original_config before the Qwen3VLTransformerBlock decoder rebuild, so all downstream layers see the correct tp_comm_overlap=True settings

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Fixes #4227
- Root cause: when mtp_block_spec is provided, MCore's GPTModel.__init__ initializes MTP block layers whose ub_name values are not registered in the tp_comm_overlap_cfg preset. MCore/TE then writes config.tp_comm_overlap = False back to the shared config object as a fallback. Since Qwen3VLGPTModel passes the same config to the subsequent Qwen3VLTransformerBlock rebuild, all decoder layers end up with TP comm overlap silently disabled.
- Scope: only modifies Bridge code, no changes to MCore or any upstream dependency.


src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/text_model.py
```diff
@@ -60,8 +61,28 @@ def __init__(
...
+        # this pollution propagates to the original config, and the
+        # subsequent Qwen3VLTransformerBlock rebuild inherits the broken
+        # settings — silently disabling TP comm overlap for ALL decoder
+        # layers.
+        #
+        # Fix: Give super().__init__ a deepcopy so MTP writes land on the
+        # copy, then restore self.config to the original (unpolluted)
+        # object before rebuilding the decoder.
+        _original_config = config
+        _config_copy = deepcopy(config)
```
yaoyu-33: can we do shallow copy with `copy(config)`, deepcopy sometimes cause other issues.

SophusDavid: ok . shallow copy can solve this issue too  :)

yaoyu-33:
Could you please add a targeted regression test for this path? The shallow `copy(config)` direction looks reasonable, but right now the PR does not prove that after MTP initialization the original config still keeps `tp_comm_overlap=True` / `tp_comm_overlap_cfg`, and that the subsequent Qwen3VL decoder rebuild sees the unpolluted overlap config.

Also, the code now uses shallow `copy(config)`, but the comments still say `deepcopy`. Please update those comments so future readers do not assume nested config isolation.

suiyoubi:
Hi @SophusDavid  could you please try adding the regression test so we could merge this PR ? thanks a lot!

PR #3839, opened 2026-05-15
[training, perf] fix: THD-aware FLOPS via cu_seqlens (Σᵢ sᵢ²)

cuichenx:
## Summary

The FLOPS calculator treated every batch as BSHD with the attention term scaling as `pack_length²` (`seq_length²` after #3529). For THD packed training — both offline-packed LLM SFT (via `PackedSequenceSpecs`) and VLM in-batch packing — the actual attention work is `Σᵢ sᵢ²` over the sub-sequence lengths within each pack, not `pack_length²`. This PR threads `cu_seqlens` from the dataloader into the FLOPS accumulator so the attention term reports the true work, and fixes the per-interval throughput logging to use the same source.

## The fix

`accumulate_flops_metadata()` in `src/megatron/bridge/training/utils/flop_utils.py`, called once per micro-batch from the step functions:

- Accumulates `seqlen_sum` (Σ padded tokens — drives the linear MLP/proj/logit terms) and the THD attention term `seqlen_sq_sum = Σᵢ sᵢ²`, **computed inline from `cu_seqlens`** (preferring `cu_seqlens_unpadded` when `pad_seq_to_mult > 1`; the matching `*_argmin` truncates the padded tail). The squared sub-sequence lengths stay **on-device** (`_scalar_sum_for_accumulator`, no `.item()`), so the per-micro-batch forward path takes no host sync.
- Falls back to the BSHD value `mbs * seq_len²` when `cu_seqlens` is absent or degenerate (dense pretraining / non-packed) — bit-exact with pre-fix behavior.
- Also accumulates `num_vision_patches` (model-agnostic: each step computes the count from its own image/video grid) for the ViT term.

Per step, the main loop resolves the accumulators to a **data-parallel-global** total via `resolve_global_flops_seqlen_stats` — an exact SUM all-reduce over the **pure DP group** (CP ranks share the same `cu_seqlens`, so reducing over DP×CP would double-count) — feeds `num_floating_point_operations`, and folds the result into `floating_point_operations_so_far`. One tiny (3-element) reduce + one host sync per step, at the existing end-of-step sync boundary.

Wired into the step functions:
- `gpt_step.py` — LLM, offline THD via `PackedSequenceSpecs`
- `vlm_step.py` — legacy VLM in-batch packing
- `qwen_vl/qwen3_vl_step.py` — Qwen3-VL family in-batch packing
- `qwen_omni/qwen3_omni_step.py` — Qwen3-Omni (vision-patch tracking + BSHD seqlen²)

[...]

PR #4366, opened 2026-06-15
[training, perf] fix: THD-aware FLOPS via cu_seqlens (Σᵢ sᵢ²) (re-land of #3839)

cuichenx:
Re-lands #3839, which was reverted in #4363 after an accidental merge while CI was red.

This branch is no longer a byte-for-byte reland of #3839. It keeps the THD-aware FLOPs accounting and includes the follow-up fixes needed to make the reland correct and CI-stable.

## What changed

**1. THD-aware FLOPs accounting via `cu_seqlens`.**

For packed/THD batches, attention FLOPs are computed from the real sub-sequence lengths (`Σᵢ sᵢ²`) instead of treating the whole pack as one dense sequence (`seq_len²`). The metadata is accumulated in each forward step and folded into `train_state.floating_point_operations_so_far` once per completed training step.

Throughput logging now uses the cumulative FLOPs delta since the previous log event, so console, WandB, TensorBoard, MLflow, and Comet throughput use the same interval-level FLOPs source as the cumulative counter. This avoids the old current-step-only over/under-reporting behavior for variable-length THD logging intervals.

**2. The DP FLOPs all-reduce is conditional.**

The earlier implementation reduced FLOPs metadata whenever `data_parallel_size > 1`, including dense BSHD training where every DP rank has identical fixed-length sequence stats. The final version adds `_flops_requires_global_reduce`, which is set only when real THD `cu_seqlens` metadata is seen.

- Dense / BSHD path: extrapolates `local * dp_size` for fixed-shape dense batches and avoids the extra collective.
- THD packed path: performs one SUM all-reduce over the pure DP group, because per-rank `Σᵢ sᵢ²` can differ.

**3. CP+packing functional test is stabilized.**

The test now runs a proper pretrain -> SFT flow: it pretrains a tiny CP=2 checkpoint, then finetunes from that checkpoint, satisfying `finetune()`'s checkpoint precondition. The pretrain and SFT model shapes are kept identical so checkpoint load validation is meaningful. The test also uses a tiny in-memory SQuAD-shaped dataset to avoid depending on external `rajpurkar/squad` cache state in CI.

`use_distributed_optimizer=False` remains in this CP+packing test as a test-local stability/memory workaround.

**4. `test_tensor_inspect` memory footprint is reduced.**

The tensor-inspect functional test uses smaller model and batch shapes and disables overlap/distributed-optimizer settings that are unnecessary for the behavior under test. This keeps the shared L0 functional-test job below CI runner memory limits.

## CI failure context

The original CI symptom looked like a hang/NCCL watchdog timeout, but the evidence pointed to CUDA OOM in the shared single-process L0 functional-test session. Once one rank OOMs, the remaining ranks can block in the next collective and eventually hit the watchdog.

The final code avoids the unnecessary dense-path FLOPs all-reduce and trims the two memory-sensitive functional tests. The CP+packing test does not exercise the THD exact-reduce path in this configuration: it uses BSHD fallback FLOPs metadata and does not issue the FLOPs DP all-reduce.

## Validation

- Focused cw-dfw repro reproduced the checkpoint shape mismatch before the final test fix.
- Focused cw-dfw validation passed after the final test fix.
- GitHub CI is green, including `L0_Launch_training` and `Nemo_CICD_Test`.

Reverts the revert (#4363).

### DIFF
--- a/src/megatron/bridge/training/utils/flop_utils.py
+++ b/src/megatron/bridge/training/utils/flop_utils.py
@@ -15,6 +15,8 @@
 import importlib
 from pathlib import Path
 
+import torch
+
 from megatron.bridge.data.datasets.packing_utils import calculate_avg_seqlen
 from megatron.bridge.peft.lora import LoRA
 from megatron.bridge.training.config import ConfigContainer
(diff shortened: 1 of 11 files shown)


src/megatron/bridge/training/utils/flop_utils.py
```diff
@@ -223,6 +224,7 @@ def accumulate_flops_metadata(
     # (which would force a data-dependent-size sync) is needed.
     sub_seq_lens = _real_subseq_lengths(cu_seqlens, cu_seqlens_argmin, cu_seqlens_unpadded, cu_seqlens_unpadded_argmin)
     if sub_seq_lens is not None and sub_seq_lens.numel() > 0:
+        setattr(state, "_flops_requires_global_reduce", True)
```
zhongbozhu, line 227: what does this condition mean and why does it indicate THD layout?

cuichenx: this mean we only do the DP all-reduce when trining with THD. For fixed seqlen training, we don't want to run allreduce

zhongbozhu (approved):
LGTM

PR #4427, opened 2026-06-21
feat(recipe): add Qwen3 30B A3B NVFP4 WBC configs and fix tp_comm_overlap for NVFP4

nvegesna-netizen:
## Summary
- Add NVFP4 WBC presets for Qwen3 30B A3B across all clusters (GB300, GB200,
  VR200, B300, B200) — each aliases the corresponding FP8_CS base config
- Disable `tp_comm_overlap` for NVFP4 precision in all 6 per-GPU builders
  (`tp_comm_overlap=precision != "nvfp4"`), since NVFP4's `fp4_param_gather`
  path is incompatible with TP comm overlap

## Test plan
- [ ] Qwen3 30B A3B NVFP4 runs on Blackwell (GB300/GB200/B300/B200) — confirm
  no TP comm overlap errors and correct precision


yaoyu-33:
Thanks for putting this together. We have recently migrated the performance recipe implementation out of the old `scripts/performance/configs` layout into the new perf recipe structure under `src/megatron/bridge/perf_recipes`.

Because this PR is based on the previous implementation path, I do not think we should merge it as-is. Could you please file a new PR against the new perf recipe layout, carrying over the Qwen3 30B A3B NVFP4 WBC presets and the NVFP4/TP-comm-overlap handling there?

yaoyu-33:
For reference, this is similar to the perf-config style of change in #4434, but the implementation has since moved. Please port the same kind of update to the new perf recipe location instead of the old `scripts/performance/configs/qwen` path.

The new PR should update the corresponding Qwen perf recipe under `src/megatron/bridge/perf_recipes` and adjust any integration coverage to match that new source of truth.

nvegesna-netizen:
Following the same pattern as #4434/#4628: opened a fresh PR based on current `main` that ports the NVFP4 `tp_comm_overlap` fix onto the new `perf_recipes` layout instead of the stale `scripts/performance/configs/qwen/...` path: #4632

A couple of things surfaced while porting that are worth calling out here:
- The H100 `tp_comm_overlap` edit in this PR has no matching `H100_NVFP4_V1` base config (NVFP4 isn't supported on H100), so `qwen3_30b_a3b_pretrain_config_h100(precision="nvfp4")` raises a `ValueError` before that line is ever reached — this PR's own new test case for H100 can't pass as written. #4632 drops the H100 edit for this reason.
- The new `perf_recipes/qwen/` layout already had this same `tp_comm_overlap` bug, unfixed, in the existing Qwen3 235B A22B NVFP4 configs (`tp_comm_overlap=True` left over from their FP8-CS base despite `fp4_param_gather=True`). #4632 fixes those too.

Can close this out in favor of #4632.

PR #4434, opened 2026-06-22
fix(nemotron3-nano): correct moe_router_topk doc, remove dead moe_shared_expert_overlap, fix H100 FP8_CS perf config, and update VL finetune example

nvegesna-netizen:
## Summary

Four correctness and cleanup fixes for the Nemotron 3 Nano model:

- **Doc fix**: `nemotron3-nano.md` stated "5 experts activated per token" but the recipe
  has always used `moe_router_topk=6`. Updated to match.

- **Remove dead constructor arg**: `moe_shared_expert_overlap=True` was passed to
  `MambaModelProvider(...)` in all three builders (pretrain, SFT, PEFT), but was
  overridden to `False` post-construction in every code path: unconditionally in the
  pretrain builder, and via the `if enable_deepep:` branch in SFT and PEFT (where
  `enable_deepep` is hardcoded `True`). Removed the dead constructor arg to eliminate
  the contradiction.

- **H100 FP8_CS perf config**: Aligned `cuda_graph_scope` with the BF16 sibling config
  (`["mamba"]` → `["attn", "mamba"]`; the omission of `"attn"` was an oversight).
  Reduced `recompute_modules` to `["layernorm", "moe_act"]` — dropping `"moe"` and
  `"core_attn"` is safe at MBS=1 and consistent with Super B300/B200 configs.

- **VL finetune example**: Replaced the `config_kwargs` dict with explicit keyword
  arguments to each recipe call, making all parameters visible at the call site.
  `pretrained_checkpoint` is now applied as a post-construction assignment
  (`cfg.checkpoint.pretrained_checkpoint`) rather than a recipe function arg.
  `--hf-model-path` is wired explicitly to both the SFT and PEFT calls.

## Test plan

- [ ] Verify `nemotron_3_nano.py` pretrain/SFT/PEFT builders produce identical configs
  before and after (net-zero change to effective `moe_shared_expert_overlap` value)
- [ ] Run H100 FP8_CS perf config and confirm no regression vs prior throughput baseline
- [ ] Run `finetune_nemotron_nano_v2_vl.py` with `--lora-on-language-model` only and
  confirm LoRA targets only language model layers
- [ ] Confirm doc renders correctly (`moe_router_topk=6` visible in recipe source)


yaoyu-33:
Thanks for the fixes here. One note on the perf-config part: the performance recipe implementation has been migrated from the old `scripts/performance/configs` layout to the new perf recipe source under `src/megatron/bridge/perf_recipes`.

Could you please check the new perf recipe path and port the relevant Nemotron/Nano perf config change there, ideally in a fresh PR if this branch is now based on the old layout? The old `scripts/performance/configs/...` change should not be the source of truth anymore.

nvegesna-netizen:
Thanks for the pointer! Opened a fresh PR based on current `main` that ports all four fixes onto the new `perf_recipes` layout — the perf-config change now lives in `src/megatron/bridge/perf_recipes/nemotronh/h100/nemotronh.py` (`nemotron_3_nano_pretrain_16gpu_h100_fp8cs_config`) instead of the stale `scripts/performance/configs/nemotronh/...` path: #4628

This can be closed in favor of #4628.

PR #4442, opened 2026-06-22
improve Qwen MegatronMIMO data loading

liding-nv:
## Summary
  - Add rank-aware batch projection for Qwen3.5-VL MIMO real-data training.
  - Use metadata-only collate on ranks that do not need visual tensors.
  - Share Qwen VL pixel-bound defaults between visual collate and the MIMO metadata path.
  - Add a unit test covering the shared Qwen pixel defaults.


examples/megatron_mimo/qwen35_vl/finetune_qwen35_vl.py
```diff
@@ -286,13 +397,191 @@ def _normalized_visual_kwargs(batch: dict[str, Any]) -> dict[str, torch.Tensor]:
...
+
+    skipped_tokens = extract_skipped_token_ids(processor)
+    loss_mask = torch.stack(
+        [
+            build_assistant_loss_mask(
+                item,
+                row_input_ids,
+                processor,
+                skipped_tokens,
+                require_matches=False,
```
cuichenx: the function signature changed in this PR: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4169 so this needs to be updated

examples/megatron_mimo/qwen35_vl/finetune_qwen35_vl.py
```diff
@@ -106,6 +112,31 @@ def square_merge_size(self) -> int:
         return self.spatial_merge_size**2
 
 
+@dataclass(frozen=True)
+class MIMOBatchRequirements:
```
cuichenx: Nit, I feel like "requirements" is an unusual name for a container object like this. Is there a precedence somewhere?

examples/megatron_mimo/qwen35_vl/finetune_qwen35_vl.py
liding-nv: updated...

examples/megatron_mimo/qwen35_vl/finetune_qwen35_vl.py
liding-nv: changed to `MIMOBatchSpec`

cuichenx (approved):

PR #4447, opened 2026-06-22
Add Nemotron diffusion DLLM support with CP utilities

linnanwang:
# What does this PR do?

Adds context-parallel (CP) training support for Nemotron diffusion language models (dLLM) using a semi-block-diffusion (sbd_block_diff) attention pattern. The PR introduces zigzag CP sequence sharding, a custom diffusion attention module that manually manages CP collectives, a forward training step for combined DLM+AR loss, and pre-training recipes for 3B/8B/14B model sizes.

# Changelog

- **`diffusion/common/cp_utils.py`** *(new)*: Context-parallelism utilities for zigzag sequence sharding.
  - `zigzag_slice`: Deterministically slices a full-sequence tensor into this rank's two zigzag chunks without any communication; used for RoPE and Llama-4 scaling that are recomputed identically on all ranks.
  - `all_gather_seq_cp` / `_AllGatherSeqCP`: Custom autograd function that all-gathers CP-sharded sequences to full length; backward scatters only this rank's zigzag gradient chunks to avoid double-counting.
  - `scatter_seq_cp` / `_ScatterSeqCP`: Inverse of the above — scatters a full sequence back to this rank's zigzag slice after full-sequence attention.
  - `local_zigzag_mask`: Returns a boolean position mask for loss restriction so standard Megatron CP loss reduction remains valid.

- **`diffusion/common/dllm.py`**: Core masking and sampling primitives for block-diffusion iterative denoising.
  - `forward_process_simple_masking`: Uniform random token masking during training (`t ~ U(eps, 1)` per sequence).
  - `get_num_transfer_tokens` / `get_transfer_index`: Distributes and selects which masked tokens to unmask at each denoising step (supports confidence, random, and entropy-based remasking strategies).
  - `compute_block_mask` / `compute_block_bias`: Builds the sbd_block_diff attention structure over the doubled `[xt | x0]` sequence — block-diagonal for xt, offset-block-causal xt→x0, fully-causal x0. Dense bias variant added for TEDotProductAttention under CP (cuDNN does not support `post_scale_bias + CP`).

- **`diffusion/models/common/nemotron_labs_diffusion_attention.py`**: Custom CP-aware attention for sbd_block_diff training.
  - `NemotronLabsDiffusionAttention`: Under CP, manually all-gathers Q/K/V to full 2L sequence, applies RoPE per-half independently, runs TEDotProductAttention with cp=1 + dense sbd_block_diff bias, then scatters output back to this rank's zigzag slice. Costs `cp_size` × attention FLOPs per rank but keeps cuDNN fused backend functional.
  - `Ministral3RotaryEmbedding`: RoPE with YARN support.
  - Inference path via `_inference_forward` with KV cache, GQA expansion, and causal/bidirectional masking.

- **`diffusion/models/common/dgpt_step.py`**: Training forward step orchestrator for sbd_block_diff models.
  - `DGPTStep.__call__`: Fetches batch → applies uniform masking → doubles sequence to `[xt | x0]` → CP zigzag shards → model forward → gathers CP logits → splits DLM/AR logits → restricts loss to owned positions → computes combined loss.
  - `_masked_loss_sbd_block_diff`: Combined DLM (masked positions, scaled by `1/p_mask`) + AR (clean positions) cross-entropy with NaN/Inf and spiky-loss checks; returns per-component metrics for logging.

- **`diffusion/recipes/nemotron_labs_diffusion/ar_to_dlm.py`**: Pre-training recipes for Nemotron diffusion LM.
  - Size-specific entry points: `nemotron_labs_diffusion_3b/8b/14b_pretrain_config` with sensible TP/MBS/GBS defaults.
  - `_nemotron_labs_diffusion_common`: Builds full `ConfigContainer` from HF checkpoint or model provider — parallelism (TP/PP/CP/SP), recompute, optimizer (distributed fused Adam), scheduler (cosine or WSD annealing), dataset, checkpoint, tokenizer, and logger configs.

# GitHub Actions CI

See the [CI section in the Contributing doc](CONTRIBUTING.md) for how to trigger the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

Pre checks:

- [ ] Make sure you read and followed [Contributor guidelines](CONTRIBUTING.md)

# Additional Information

Related to # (issue)


src/megatron/bridge/diffusion/common/dllm.py
```diff
@@ -211,3 +80,67 @@ def sbd_block_diff_mask(b, h, q_idx, kv_idx):
 
     q_len = max_seq_length * 2
     return create_block_mask(sbd_block_diff_mask, B=None, H=None, Q_LEN=q_len, KV_LEN=q_len)
+
+
+def compute_active_block_bidirectional_mask(block_starts, block_ends, q_len, kv_len, offset=0):
```
sajadn: Is this method used anywhere? i don't find it

src/megatron/bridge/diffusion/models/common/dgpt_step.py
```diff
@@ -165,19 +172,35 @@ def __call__(
 
         timers("batch-generator").stop()
 
+        # Build the doubled [xt | x0] sequence on the FULL (un-sharded) batch.
         (noisy_tokens, labels, loss_mask, attention_mask, position_ids, masked_indices, p_mask, input_ids_len) = (
             self._apply_noise(tokens, labels, loss_mask, attention_mask, position_ids)
         )
+        # position_ids for the doubled sequence: [b, 2L] of the right shape (the
+        # attention recomputes per-half RoPE positions internally).
+        position_ids_2l = torch.cat([position_ids, position_ids], dim=1)
```
sajadn: i think positional embeddings are computed within the model, so why bothering to update the position ids?

src/megatron/bridge/diffusion/models/common/nemotron_labs_diffusion_attention.py
```diff
@@ -209,10 +249,44 @@ def __init__(
...
+        # with qkv_format="thd" AND context parallelism -- unlike the diffusion
+        # dense-bias path. The CP group is supplied per-forward via
+        # packed_seq_params.cp_group, so this single instance serves cp=1 and cp>1.
+        # Built from the original `config` (keeps its context_parallel_size), not
+        # the cp=1 copy used for the bias path.
+        # pg_collection=None matches self.core_attention: TE derives tp/cp from the
+        # global mpu parallel-state at runtime, and the CP group is overridden per
+        # forward by packed_seq_params.cp_group. (Avoids depending on a CP-populated
+        # pg_collection at construction.)
+        self._ar_core_attention = TEDotProductAttention(
```
sajadn: can you put this in a condition that only if _inference_mode is set to true, it's defined?

src/megatron/bridge/diffusion/models/common/nemotron_labs_diffusion_attention.py
```diff
@@ -252,21 +326,40 @@ def forward(
         attention_bias: Tensor = None,
         packed_seq_params: Optional[PackedSeqParams] = None,
     ):
-        assert packed_seq_params is None, "Packed sequence is not supported by NemotronLabsDiffusionAttention."
-
         if self._inference_mode:
-            return self._inference_forward(query, key, value)
+            # AR / causal path: supports packed (thd) sequences + context
+            # parallelism through TE (causal is no_bias, which TE handles for thd+CP).
+            return self._inference_forward(query, key, value, packed_seq_params)
```
sajadn, line 348: i think it might make the PR a bit complicated to do support for CP > 1 inference. My confusion is from the fact that AR with CP>1 uses packed_seq_params while the default training path with CP>1 doesn't use it. So i'd suggest removing inference with CP> 1 to keep things simple.

src/megatron/bridge/diffusion/common/dllm.py
```diff
@@ -60,124 +47,6 @@ def forward_process_simple_masking(input_ids, mask_token_id, eps=1e-3, loss_mask
     return noisy_batch, masked_indices, p_mask
 
 
-def add_gumbel_noise(logits: torch.Tensor, temperature: float) -> torch.Tensor:
```
sajadn, line 63: nit; for the sake of keeping this PR simple i suggest moving functions from dllm.py to dllm_sampling.py as a future work.

src/megatron/bridge/diffusion/common/dllm.py
linnanwang: Good catch — it's not wired up in this PR. `compute_active_block_bidirectional_mask` was added ahead of its consumer (the block-bidirectional attention path). I'll remove it here and reintroduce it together with that consumer to keep this PR tight.

src/megatron/bridge/diffusion/models/common/dgpt_step.py
linnanwang: Agreed — the model computes its positional embeddings internally, so this explicit position-id update looks redundant on the standard path. I'll remove it, and just double-check the CP-sharded path doesn't rely on explicit global positions (if it does I'll re-add with a comment; I expect it's removable).

src/megatron/bridge/diffusion/models/common/nemotron_labs_diffusion_attention.py
linnanwang: Sure — I'll make `_ar_core_attention` construction lazy/conditional so it's only built when inference mode is actually used, instead of unconditionally in `__init__`.

src/megatron/bridge/diffusion/models/common/nemotron_labs_diffusion_attention.py
linnanwang: Good catch on the asymmetry — it's real, but it's forced by a kernel constraint rather than an inconsistency, so let me explain and then propose a path.

**Why AR-CP uses `packed_seq_params` but the diffusion path doesn't:**
- The causal/AR mask is `no_bias`. TE's fused THD (packed) attention supports ring CP for `no_bias`, so the AR path rides TE-native CP via `packed_seq_params` (the THD path).
- The diffusion `sbd_block_diff` mask is a dense `post_scale_bias`. There is no THD attention kernel for a dense bias on any TE version — so packed CP is impossible for it. That path instead does an in-attention all-gather → cp=1 dense-bias core → scatter (unpacked), which is why it carries no `packed_seq_params`.

So the two branches legitimately differ; it's the bias type dictating packed vs unpacked, not a design choice.

On "inference CP>1": worth flagging that this isn't a serving feature — it's the path the RL side runs through. The companion NeMo-RL MR (`cp_nemotron_dllm`) scores/trains with `set_inference_mode(True)`, so "CP>1 inference" here is the diffusion-GRPO recompute + AR-GRPO training path. It's validated bit-identical to cp=1 (AR cp=2; diffusion cp=2 and cp=8, max|Δ logprob| = 0), and it's what enables long-context RL (AR packed-CP trains 128K-token sequences on 8×H100 where cp=1 OOMs around 32K).

**Proposal:** I'm happy to keep this PR simple as you suggest and split the CP>1-inference paths into a follow-up bridge PR (`_packed_causal_forward` / `_packed_position_ids` for cp>1, and the CP gather/scatter branch in `_inference_forward`), leaving this PR with training CP>1 + cp=1 inference. The one caveat: the NeMo-RL MR depends on those paths, so the follow-up would need to land before/with it. Alternatively, if you'd rather keep them here, I'll add a comment documenting the packed-vs-unpacked-bias rationale so the asymmetry is self-explaining. Either works for me — which do you prefer?

src/megatron/bridge/diffusion/common/dllm.py
linnanwang: Already done in the latest commit — the sampling primitives (`add_gumbel_noise`, `get_num_transfer_tokens`, `get_transfer_index`) now live in `common/dllm_sampling.py`, with the importers (`inference_nemotron_labs_diffusion.py`, `inference_llada15.py`) and tests updated. I did it now rather than as future work because removing them from `dllm.py` otherwise broke those `inference_*` imports.

sajadn: sorry i think my comment was a bit ambiguous. What's the motivation for moving these functions to a dedicated file? i think if you avoid doing that this PR becomes much clear/shorter.

linnanwang: Good call — reverted the split. Kept `add_gumbel_noise` / `get_num_transfer_tokens` / `get_transfer_index` in `dllm.py` and dropped `dllm_sampling.py`; the inference modules and tests import them from `dllm` again (`d6af53e6`).

sajadn (dismissed):

src/megatron/bridge/diffusion/common/cp_utils.py
```diff
@@ -0,0 +1,184 @@
...
+    full-sequence tensor (no communication). Used to shard RoPE cos/sin and the
+    Llama-4 scale, which are recomputed identically on every rank.
+  - ``all_gather_seq_cp``: autograd all-gather that reconstructs the full
+    sequence (undoing the zigzag) so the diffusion loss can split ``[xt | x0]``.
+"""
+
+import torch
+
+
+def zigzag_slice(tensor: torch.Tensor, cp_rank: int, cp_size: int, seq_dim: int) -> torch.Tensor:
```
yaoyu-33, line 34: can you check if there any existing cp utils can be used? either MB/MCore. Avoid duplicated implementation

src/megatron/bridge/diffusion/recipes/nemotron_labs_diffusion/ar_to_dlm.py
```diff
@@ -204,7 +204,9 @@ def _nemotron_labs_diffusion_common(
         data_paths, data_args_path, train_data_path, valid_data_path, test_data_path, per_split_data_args_path, mock
     )
     if hf_path is not None:
-        hf_pretrained = PreTrainedCausalLM.from_pretrained(hf_path)
+        # trust_remote_code=True is required: Nemotron Labs diffusion checkpoint uses
+        # custom modeling code (NemotronLabsDiffusionForCausalLM) not shipped with transformers.
+        hf_pretrained = PreTrainedCausalLM.from_pretrained(hf_path, trust_remote_code=True)
```
yaoyu-33: WE DONT allow trust_remote=True for security reasons. You can pass in user args somehow. but cannot hardcode to True.

src/megatron/bridge/diffusion/models/common/nemotron_labs_diffusion_attention.py
```diff
@@ -209,10 +249,49 @@ def __init__(
...
-            block_size=getattr(config, "block_size", 16),
-            max_seq_length=config.seq_length,
+        self.block_size = getattr(config, "block_size", 16)
+
+        # TE core attention run WITHOUT CP (cp=1 config copy): cuDNN supports
+        # post_scale_bias in the non-CP path. We feed it the full gathered 2L
+        # sequence and the dense sbd_block_diff bias; CP comms are done by us.
+        core_cfg = copy.copy(config)
+        core_cfg.context_parallel_size = 1
+        self.core_attention = TEDotProductAttention(
```
sajadn, line 259: can't we still use FlexAttention instead of TEDotProductAttention while doing gather and scatter before and after running the attention?

src/megatron/bridge/diffusion/common/cp_utils.py
linnanwang: Checked MB + MCore. The gather/scatter here (`all_gather_seq_cp`/`scatter_seq_cp`) has no reusable equivalent — MCore's CP sequence gather lives inside TE's ring/a2a attention comm (not a standalone autograd op), and `megatron.core.utils.get_batch_on_this_cp_rank` only *slices* a batch dict forward-only (no gather-back). NeMo-RL independently reimplemented the same thing (`AllGatherCPTensor`), which confirms the gap; Bridge can't depend on NeMo-RL.

`zigzag_slice` does mirror `get_batch_on_this_cp_rank`'s zigzag layout (noted in the docstring), but that helper is dict/`cu_seqlens`-based and forward-only, so it isn't a drop-in for slicing individual in-attention tensors (RoPE cos/sin, Llama-4 scale) or the doubled `[xt|x0]` sequence — which is why dgpt_step moved off it.

Happy to add a comment cross-referencing MCore, or factor the gather into a shared util in a follow-up if you'd prefer a longer-term home.

src/megatron/bridge/diffusion/recipes/nemotron_labs_diffusion/ar_to_dlm.py
linnanwang: Done — removed the hardcoded `trust_remote_code=True`. Added a `trust_remote_code` parameter (default `False`) threaded through the recipe entry points, so the caller must opt in explicitly to load the checkpoint's custom modeling code. (`f99e54dc`)

yaoyu-33 (dismissed):

src/megatron/bridge/diffusion/recipes/nemotron_labs_diffusion/ar_to_dlm.py
sajadn: why do we really need this in the first place?

yaoyu-33 (dismissed):

huvunvidia:
/ok to test [8c711d3](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4447/commits/8c711d3be85398875982078092503f08337d68ba)

huvunvidia:
Hi @linnanwang ,
All CICD tests passed but the code coverage ([codecov/patch](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4447/checks?check_run_id=86287934315) — 58.69% of diff hit (target 80.00%)) is low.
Can you add more tests?

sajadn:
@linnanwang @huvunvidia @yaoyu-33 let's pause merging this PR now. 
@linnanwang do you think we can come up with a better solution rather than all gather/scatter Q/K/V?

PR #4464, opened 2026-06-23
Revert "[training, perf] fix: THD-aware FLOPS via cu_seqlens (Σᵢ sᵢ²) (re-land of #3839)"

malay-nagda:
Reverts NVIDIA-NeMo/Megatron-Bridge#4366


ko3n1g:
FLOP calc now seems to be broken for MoE models too.

PR #4476, opened 2026-06-24
Migrate GPT-OSS to HybridModel

Phlip79:
# What does this PR do ?

Migrates GPT-OSS from using `GPTModel` to `HybridModel`. To accomplish this, I needed to update `HybridModelProvider` to support yarn, which `HybridModel` already supports in MCore.

[Perf comparison results](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/56415800)

Needs https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4508 for correct TFLOPS calculation


yaoyu-33 (dismissed):

src/megatron/bridge/models/gpt_oss/gpt_oss_bridge.py
```diff
@@ -157,78 +179,119 @@ def mapping_registry(self) -> MegatronMappingRegistry:
         Return MegatronMappingRegistry containing parameter mappings from HF to Megatron format.
         Based on the GPT-OSS importer code provided.
         """
+        hf_config = getattr(self, "hf_config", None)
```
yaoyu-33: I dont think it will ever be / should be None here. not use get attr and let it error out?

Phlip79: Good point. Changes made.

yaoyu-33 (approved):

PR #4485, opened 2026-06-24
Enable expert-parallel language modules in MegatronMIMO

liding-nv:
## Summary

  This PR enables MegatronMIMO to describe and build non-colocated language MoE modules with expert parallelism while keeping encoder modules dense.

  Key changes:
  - Adds language-side `expert_model_parallel_size` / `expert_tensor_parallel_size` handling to MegatronMIMO parallelism config.
  - Separates dense rank accounting from expert rank views, so EP/ETP refactor the language rank span instead of multiplying the module world size.
  - Validates heterogeneous module rank ranges tile the distributed world exactly, and keeps encoder modules dense for this MoE path.
  - Rebuilds MegatronMIMO HyperCommGrid setup around a dense base view plus an expert view:
    - dense: `tp, cp, dp, pp`
    - expert: `expt_tp, ep, expt_dp, pp`
  - Wires the resulting expert groups into `ProcessGroupCollection` and MCore compatibility globals.
  - Seeds per-module RNG with real TP/PP/EP/ETP ranks so expert-parallel language ranks do not silently share identical expert seeds.
  - Passes language TP/CP groups into MCore `MimoModel` construction for sequence-parallel partitioning paths.
  - Initializes the MCore global memory buffer for MegatronMIMO custom parallel setup path.
  - Threads `ep` through the MegatronMIMO conversion CLI component parser.
  - Extends unit coverage for config validation, grid construction, provider wiring, RNG seeding, global expert-group bridging, checkpoint/optimizer setup, and conversion CLI parsing.


cuichenx (commented):
LGTM

cuichenx (approved):

PR #4503, opened 2026-06-25
[model, test] fix: handle MCore main bump

yaoyu-33:
## Summary
- Fixes the `main` daily MCore bump failure from #4495.
- Uses the existing MCore ModelOpt hybrid spec layernorm key remap so hybrid quantized checkpoints save/load with TE-compatible layernorm names.

## Original bump
- Bump PR: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4495
- Target: `main`
- MCore range: `a58373f332496f08c6584b3196233275ff69f175...ea967a7a13b1f20145a4aee1b7cfdc78d50b6f5a`

## Classification
MCore broke Bridge.

The NemotronH failures came from checkpoint-load validation expecting `decoder.layers.<n>.norm.weight` while the saved ModelOpt/TE checkpoint state dict contains remapped layernorm keys such as `decoder.layers.<n>.mixer.in_proj.layer_norm_weight`. Bridge was using `remap_te_layernorm=False` for hybrid ModelOpt specs even though the GPT ModelOpt path already enables this remap.

## Fix
- Enable `remap_te_layernorm=True` for hybrid ModelOpt specs in both provider and builder paths.
- Add unit coverage for the call sites and the real MCore sharded state-dict key maps.

## Guards
None added or removed. This uses an existing MCore capability and does not add a compatibility guard.

## Validation
- `git diff --check origin/main...HEAD` -> passed.
- Attempted: `uv run python -m pytest tests/unit_tests/models/hybrid/test_hybrid_builder.py tests/unit_tests/models/hybrid/test_hybrid_provider.py -q` -> blocked before pytest by dependency resolution because `nvidia-resiliency-ext==0.6.0` has no wheel for the local manylinux platform tag.
- Attempted: `uv run pre-commit run --all-files` -> blocked by the same dependency resolution issue.


cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4503/commits/a262f97e9fbfa7aaedeef624191e2313b618fba9

yaoyu-33:
Previous test already passed. so skip ci test this time.

PR #4507, opened 2026-06-25
[data] fix: pack VLM batches directly

yaoyu-33:
## Summary
- build packed THD batches directly from unpadded per-example rows for supported text and VLM collators instead of padding a batch and removing that padding
- emit current MCore / Transformer Engine metadata directly: `cu_seqlens_q`, `cu_seqlens_kv`, padded cu-seqlens when the CP alignment contract requires them, `max_seqlen_q`, and `max_seqlen_kv`
- preserve model-specific packed inputs, including GLM token types, Qwen video timing, and per-segment Qwen/GLM MRoPE positions
- enforce right-padding assumptions in the compatibility adapter and keep non-packed mixed-length loss masks aligned with each row's actual end token
- supervise the complete ChatML assistant stop sequence (`<|im_end|>` and its rendered newline) without exposing user/system/tool turns or right padding
- preserve top-level `tools` and message-level `tool_calls` through Energon, text HF, and registered model collators
- keep valid all-zero masks after truncation instead of falling back to supervision over non-assistant tokens

Refs #4041. Follow-up to #4315.

## MCore reference points
- `megatron/core/packed_seq_params.py` current `PackedSeqParams` fields
- `megatron/core/extensions/transformer_engine.py` TE-version filtering for current packed fields
- `megatron/core/transformer/attention.py` padded cu-seqlens selection
- `megatron/core/models/mimo/model/base.py` current packed metadata construction

## Validation
- targeted collate/packing/unit slice: 160 passed
- targeted offline ChatML SFT mask/truncation slice: 4 passed
- Qwen3-VL-2B-Instruct token-level comparison: VLM HF, text HF, normalized Energon, and full Energon source paths matched for plain multi-turn, reasoning, tool-call, and reasoning-plus-tool-call fixtures
- the same comparison matched the TRL Qwen3-VL training template's effective targets, including the newline after every assistant `<|im_end|>`
- packed CP=2 rows matched their non-packed token and target sequences; current query/KV metadata and padded cu-seqlens matched between HF and Energon paths
- real-image guard: image placeholder remained unsupervised while the effective target was `red<|im_end|>\n`; HF and Energon tensors matched
- `uv run --no-sync pre-commit run --all-files`: passed
- `git diff --check`: passed
- three independent code-review rounds; final round: no actionable findings

## Compatibility notes
- VLM runtime no longer depends on Bridge-only `cu_seqlens_argmin` or `cu_seqlens_unpadded*` fields.
- Existing text/offline packed SFT legacy metadata remains supported through `get_packed_seq_params`.
- Batched processor collators use a dataset-wide shared tool registry; message-level tool calls may vary by row.
- Nemotron Omni HF collate and Nemotron-VL video collate reject in-batch packing explicitly; Nemotron Omni packing remains supported through its Energon task encoder.
- Qwen2-Audio packing remains owned by `audio_lm_step.py`; its collator warns when the collate packing flag is supplied.
- Automatic stop-boundary augmentation recognizes ChatML, `<|turn>model`, and `<start_of_turn>model` formats. Other custom templates must provide correct HF generation masks or an explicit model boundary configuration.
- No files under `3rdparty/Megatron-LM/` are modified.


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4507/commits/593b44e7a1788534610d8d7e1b60eb1b8a962c0a

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4507/commits/38c5a6f57deb5548a3e45f392f09510ccfac6e07

src/megatron/bridge/data/sequence_batching.py
```diff
@@ -99,7 +98,158 @@ def _pad_or_truncate_attention_mask(attention_mask: torch.Tensor | None, target_
...
+        if labels is not None
+        else None
+    )
+    packed_loss_mask = (
+        torch.zeros((1, total_len), dtype=loss_mask.dtype, device=loss_mask.device) if loss_mask is not None else None
+    )
+
+    offset = 0
+    for batch_idx, length, padded_length in zip(valid_indices, unpadded_lengths, padded_lengths):
+        packed_tokens[0, offset : offset + length] = tokens[batch_idx, :length]
```
yaoyu-33: This packs the first `length` columns after deriving `length` from `attention_mask.sum()`. That is only correct for right-padded rows. If a tokenizer/processor left-pads, this will copy pad tokens into the packed THD stream and drop real tokens at the end. Some collators force right padding before this helper, but the shared text/Qwen paths do not. Can we either force right-padding before all callers that enable direct packing, or make this helper gather true-token indices from the 2D attention mask and reset packed position ids accordingly?

src/megatron/bridge/data/energon/nemotron_omni_task_encoder.py
```diff
@@ -535,15 +535,10 @@ def batch(self, samples: List[NemotronOmniTaskSample]) -> NemotronOmniTaskBatch:
...
-            cu_seqlens_t = torch.tensor(cu_seqlens, dtype=torch.int32)
-            cu_seqlens_unpadded_t = cu_seqlens_t.clone()
-            # get_packed_seq_params truncates cu_seqlens_padded[: argmin.item()]; the
-            # trick in the fixed-size-batched case is sentinel=-1 padding with argmin
-            # pointing at the first sentinel. Here we emit an unpadded cu_seqlens and
-            # set argmin = len(cu_seqlens) so the slice is a no-op (keeps every entry).
-            cu_seqlens_argmin_t = torch.tensor(len(cu_seqlens), dtype=torch.int32)
-            cu_seqlens_unpadded_argmin_t = torch.tensor(len(cu_seqlens), dtype=torch.int32)
-            max_seqlen_t = torch.tensor(max(lengths), dtype=torch.int32)
+            cu_seqlens_q_t = torch.tensor(cu_seqlens, dtype=torch.int32)
```
yaoyu-33, line 553: This custom packing path should honor `self.in_batch_packing_pad_to_multiple_of` like `prepare_padded_or_packed_sequence_batch` does. Config validation fills that multiple for CP/SP constraints, but this branch concatenates raw lengths and never emits `cu_seqlens_q_padded` / `cu_seqlens_kv_padded`. For packed CP runs with lengths that are not aligned to `2 * context_parallel_size`, TE partitioning can become inconsistent with the actual token layout.

src/megatron/bridge/data/sequence_batching.py
```diff
@@ -140,27 +288,13 @@ def pad_or_pack_sequence(
         raise ValueError("Sequence batch preparation expects a 2D token tensor.")
 
     if enable_in_batch_packing:
```
yaoyu-33, line 439: [P2] The packing branch returns before validating or applying `sequence_length`. An overlength processor row therefore produces metadata/position IDs beyond the configured model cap. Apply the per-row cap (or reject overlength rows) before building packed tensors, with an overlength regression test.

src/megatron/bridge/models/qwen_vl/qwen3_vl_step.py
```diff
@@ -57,7 +57,7 @@ def get_batch_from_iterator(
         dict[str, torch.Tensor]: A dictionary containing the batch data.
     """
     batch = next(data_iterator)
-    if "cu_seqlens" in batch:
+    if "cu_seqlens_q" in batch or "cu_seqlens" in batch:
```
yaoyu-33: [P1] This rejects metadata by key presence, but `QwenVLTaskEncoder.encode_batch()` includes `cu_seqlens_q=None` on ordinary unpacked Energon batches. Those batches are rejected even though no packed metadata exists. Please check for a non-`None` value (for current and legacy keys) and add the deferred-batch regression case.

src/megatron/bridge/models/glm_vl/data/collate_fn.py
```diff
@@ -94,7 +94,7 @@ def glm4v_collate_fn(
         if key in batch:
             visual_kwargs[key] = batch.pop(key)
     batch["visual_inputs"] = GenericVisualInputs(**visual_kwargs) if visual_kwargs else None
-    pad_or_pack_sequence(
+    prepare_padded_or_packed_sequence_batch(
```
yaoyu-33, line 167: [P1] `mm_token_type_ids` is removed into `visual_inputs` before the sequence packer runs. For a multi-row packed batch, `input_ids` becomes `[1, packed_len]` but this sequence-aligned tensor remains `[B, padded_len]`; GLM MRoPE then sees incompatible batch/sequence dimensions. Please pack it with the same row offsets/padding and cover a two-row GLM batch.

src/megatron/bridge/data/sequence_batching.py
```diff
@@ -99,7 +98,157 @@ def _pad_or_truncate_attention_mask(attention_mask: torch.Tensor | None, target_
...
+        batch["labels"] = packed_labels
+    if packed_loss_mask is not None:
+        batch["loss_mask"] = packed_loss_mask
+    batch["position_ids"] = packed_position_ids
+    batch["attention_mask"] = None
+
+    cu_seqlens_t = torch.tensor(cu_seqlens, dtype=torch.int32, device=device)
+    batch["cu_seqlens_q"] = cu_seqlens_t
+    batch["cu_seqlens_kv"] = cu_seqlens_t
+    if cu_seqlens_padded != cu_seqlens:
```
yaoyu-33: [P1] When a CP/SP padding multiple is requested, aligned lengths still need explicit padded cu-seqlens. MCore's LLaVA THD CP branch dereferences `cu_seqlens_q_padded` unconditionally; omitting it when values equal the unpadded offsets passes `None` into `thd_get_partitioned_indices`. Emit padded fields whenever `pad_to_multiple_of > 1`, including the aligned case.

src/megatron/bridge/data/sequence_batching.py
```diff
@@ -99,7 +98,157 @@ def _pad_or_truncate_attention_mask(attention_mask: torch.Tensor | None, target_
...
+    else:
+        active_mask = tokens != pad_token_id
+
+    lengths_t = active_mask.sum(dim=1)
+    positions = torch.arange(tokens.size(1), device=tokens.device).unsqueeze(0)
+    expected_mask = positions < lengths_t.unsqueeze(1)
+    if not torch.equal(active_mask, expected_mask):
+        raise ValueError("Direct sequence packing requires right-padded input rows.")
+
+    return [int(length.item()) for length in lengths_t]
```
yaoyu-33, line 135: [P2] Rejecting non-right-padded input is safe only if every supported collator forces and restores right padding. Qwen/text do, but GLM, Ministral, and Nemotron-VL currently do not, so a user-configured left-padding tokenizer now fails in packing mode. Please centralize the right-padding context or apply it consistently to those processor paths, with restoration coverage.

src/megatron/bridge/models/qwen_vl/data/collate_fn.py
```diff
@@ -105,38 +105,46 @@ def qwen2_5_collate_fn(
...
+    saved_padding_side = getattr(tokenizer_for_padding, "padding_side", None)
+    if tokenizer_for_padding is not None and saved_padding_side is not None:
+        tokenizer_for_padding.padding_side = "right"
+    try:
+        if idx_with:
+            texts_with = [texts[i] for i in idx_with]
+            images_with = [per_example_images[i] for i in idx_with]
+            videos_with = [per_example_videos[i] for i in idx_with]
+            processor_kwargs = {
+                "text": texts_with,
```
yaoyu-33, line 207: [P2] The packing branch still requests processor batch padding and later copies valid prefixes into a second packed tensor. That remains pad-then-unpad/repack, which is the behavior this follow-up is meant to remove. In packing mode, process rows without batch padding and construct THD tensors directly from those per-example rows; assert `padding=False` in the packed-collate test.

src/megatron/bridge/models/qwen_vl/data/collate_fn.py
```diff
@@ -221,7 +229,7 @@ def pad_to(x, tgt_len, value):
     for key in THW_GRID_VISUAL_KEYS:
         batch.pop(key, None)
     batch["visual_inputs"] = visual_inputs
-    pad_or_pack_sequence(
+    prepare_padded_or_packed_sequence_batch(
```
yaoyu-33, line 318: [P1] Packed Qwen2.5 rows are later passed to `Qwen2_5_VLModel.get_rope_index` as one flattened sample, so MRoPE does not reset at sequence boundaries and CP alignment gaps participate in position construction. The model path needs to reconstruct segment rows from current cu-seqlens for MRoPE, then scatter the 3-axis positions back to THD layout. Please compare packed segment positions with standalone rows.

src/megatron/bridge/models/glm_vl/data/collate_fn.py
yaoyu-33, line 167: [P1] Even after packing `mm_token_type_ids`, GLM calls `get_rope_index` once over the flattened row. That preserves tensor shape but not per-example MRoPE resets. Reconstruct segments from `cu_seqlens_q` / padded offsets for position generation, then repack positions; cover the downstream MRoPE values, not only collate shapes.

src/megatron/bridge/data/sequence_batching.py
yaoyu-33, line 439: [P1] The shared packing entry still starts from a processor-padded `[B, S]` tensor. Gemma, GLM, Ministral, Nemotron-VL, Kimi, and Nemotron-Omni all reach this branch after batch padding or manual padding, so the required direct collate behavior is still incomplete outside Qwen. Supported packing collators should build per-example rows and call a direct THD row builder.

src/megatron/bridge/models/nemotron_omni/data/collate_fn.py
```diff
@@ -400,7 +398,7 @@ def nemotron_omni_collate_fn(
     batch["labels"] = batch["labels"].masked_fill(loss_mask_t == 0, IGNORE_INDEX)
     batch["loss_mask"] = loss_mask_t
 
-    pad_or_pack_sequence(
+    prepare_padded_or_packed_sequence_batch(
```
yaoyu-33, line 429: [P2] Nemotron-Omni processor/tokenizer calls still use ambient `padding_side`. A left-padding tokenizer produces rows that the right-padded packing adapter rejects. Apply the temporary right-padding context consistently and restore it on both success and exceptions.

src/megatron/bridge/data/sequence_batching.py
```diff
@@ -99,7 +98,157 @@ def _pad_or_truncate_attention_mask(attention_mask: torch.Tensor | None, target_
...
+        if (
+            not isinstance(attention_mask, torch.Tensor)
+            or attention_mask.dim() != 2
+            or attention_mask.shape != tokens.shape
+        ):
+            raise ValueError("'attention_mask' must match token shape for direct sequence packing.")
+        padding_mask = attention_mask.to(device=tokens.device)
+
+    lengths = _right_padded_sequence_lengths(tokens, pad_token_id=pad_token_id, padding_mask=padding_mask)
+    valid_indices = [idx for idx, length in enumerate(lengths) if length > 0]
```
yaoyu-33: [P2] Filtering zero-length rows here changes text batch indexing while visual/audio tensors retain the original example order. Reject any empty row (or synchronously filter every batch-aligned modality) instead of silently dropping only its sequence tensors.

src/megatron/bridge/models/nemotron_vl/data/collate_fn.py
```diff
@@ -143,7 +143,7 @@ def nemotron_nano_v2_vl_collate_fn(
     loss_mask_t = torch.cat([loss_mask_t[:, 1:], torch.zeros_like(loss_mask_t[:, :1])], dim=1)
     batch["labels"] = batch["labels"].masked_fill(loss_mask_t == 0, IGNORE_INDEX)
     batch["loss_mask"] = loss_mask_t
-    pad_or_pack_sequence(
+    prepare_padded_or_packed_sequence_batch(
```
yaoyu-33, line 225: [P1] Nemotron-VL video collation requires `len(examples) == 1`, while in-batch packing validation requires micro-batch size greater than 1. The video branch therefore cannot support this flag and otherwise reaches the padded adapter for a single row. Please reject video packing explicitly before processor work; image packing can remain supported.

src/megatron/bridge/data/sequence_batching.py
```diff
@@ -99,7 +98,158 @@ def _pad_or_truncate_attention_mask(attention_mask: torch.Tensor | None, target_
...
+        else:
+            non_pad_mask = tokens[idx] != pad_token_id
+            if non_pad_mask.all():
+                length = seq_len
+            elif non_pad_mask.any():
+                length = int(non_pad_mask.nonzero(as_tuple=True)[0][-1].item()) + 1
+            else:
+                length = 0
+        lengths.append(length)
+    return lengths
```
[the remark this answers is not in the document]
yaoyu-33: Fixed in ea912ca9d. The compatibility adapter now validates contiguous right padding, supported processor paths temporarily force and restore right padding, and regression tests cover invalid left-padded input.

src/megatron/bridge/data/sequence_batching.py
yaoyu-33: Fixed in ea912ca9d. Direct packing consumes unpadded per-example rows; the compatibility adapter validates right padding, and supported collators force and restore right padding around processor calls.

src/megatron/bridge/data/energon/nemotron_omni_task_encoder.py
yaoyu-33: Fixed in ea912ca9d. The Energon path now emits padded query/key-value cu-seqlens whenever the configured alignment multiple is greater than one, for both aligned and unaligned lengths.

src/megatron/bridge/data/sequence_batching.py
yaoyu-33: Fixed in ea912ca9d. Direct row packing validates every row against sequence_length and rejects overlength input; targeted tests cover the failure.

src/megatron/bridge/models/qwen_vl/qwen3_vl_step.py
yaoyu-33: Fixed in ea912ca9d. Qwen3 now rejects packed metadata only when a current or legacy field has a non-None value, with coverage for deferred ordinary batches.

src/megatron/bridge/models/glm_vl/data/collate_fn.py
yaoyu-33: Fixed in ea912ca9d. GLM mm_token_type_ids now use the same direct row layout and alignment padding as tokens, with a two-row packed regression test.

src/megatron/bridge/data/sequence_batching.py
yaoyu-33: Fixed in ea912ca9d. Padded cu-seqlens are emitted whenever pad_to_multiple_of is greater than one, including already-aligned batches required by the MCore THD CP path.

src/megatron/bridge/data/sequence_batching.py
yaoyu-33: Fixed in ea912ca9d. A shared processor right-padding context is applied consistently and restores the original setting on both normal and exceptional exits; tests cover restoration.

src/megatron/bridge/models/qwen_vl/data/collate_fn.py
yaoyu-33: Fixed in ea912ca9d. Supported packing collators process each example without batch padding and build THD tensors directly from rows; tests assert the no-padding processor calls.

src/megatron/bridge/models/qwen_vl/data/collate_fn.py
yaoyu-33: Fixed in ea912ca9d. Qwen reconstructs logical rows from current cu-seqlens only for MRoPE generation, resets each segment independently, excludes alignment gaps, and scatters positions back to THD layout.

src/megatron/bridge/models/glm_vl/data/collate_fn.py
yaoyu-33: Fixed in ea912ca9d. GLM now generates MRoPE independently per logical segment and repacks the resulting positions; tests compare packed segments with standalone rows.

src/megatron/bridge/data/sequence_batching.py
yaoyu-33: Fixed in ea912ca9d. Text, Gemma, GLM, Ministral, Nemotron-VL image, Kimi, and Qwen supported paths now build direct per-example rows. Unsupported HF packing combinations fail explicitly.

src/megatron/bridge/models/nemotron_omni/data/collate_fn.py
yaoyu-33: Fixed in ea912ca9d. Nemotron-Omni HF collate now rejects packing before processor work; its supported Energon packed path constructs current metadata directly without depending on tokenizer padding side.

src/megatron/bridge/data/sequence_batching.py
yaoyu-33: Fixed in ea912ca9d. The compatibility adapter rejects empty rows instead of filtering them, preserving alignment between sequence and modality tensors; tests cover this case.

src/megatron/bridge/models/nemotron_vl/data/collate_fn.py
yaoyu-33: Fixed in ea912ca9d. Nemotron-VL video plus in-batch packing is rejected before processor work, while the image path retains direct packing support; a targeted regression test covers the guard.

yaoyu-33:
Follow-up mask/tool audit is included in `f8a6cf902`.

Key results:
- Assistant targets now include the rendered ChatML stop sequence `<|im_end|>\n`; user, system, developer, tool, padding, and image-placeholder tokens remain masked.
- Nonzero HF generation masks are augmented with boundary tokens only. All-zero masks use boundary fallback when a complete assistant span exists, but stay zero after truncation when no assistant target remains.
- Full Energon source normalization preserves top-level `tools` and message-level `tool_calls`; registered chat-template collators forward the shared tool registry.
- Offline ChatML SFT uses the same stop-boundary augmentation and no longer changes an empty post-truncation mask to all ones.

Targeted validation:
- 160 collate/packing/unit tests passed.
- 4 offline ChatML SFT mask/truncation tests passed.
- Qwen3-VL-2B-Instruct token-level comparison matched across VLM HF, text HF, normalized Energon, and full Energon source paths for plain multi-turn, reasoning, tool-call, and reasoning-plus-tool-call fixtures.
- Effective targets matched the TRL Qwen3-VL training template, including the newline after each assistant end token.
- Packed CP=2 rows matched non-packed targets and current query/KV packed metadata.
- Real-image check kept the image placeholder unsupervised; the effective target was `red<|im_end|>\n`, with HF and Energon tensors identical.
- Full pre-commit passed. Three independent review rounds were run; the final round reported no actionable findings.

Remaining compatibility boundary: automatic stop augmentation recognizes ChatML, `<|turn>model`, and `<start_of_turn>model`; custom templates outside those formats still need correct HF generation masks or an explicit boundary configuration.

yaoyu-33:
CI follow-up for `e728ac76d522e5fb6ee39d8903b40bd0c4de15fd`:

- Merged latest `origin/main` (`6d5e92457`).
- The prior `L0_Launch_training_finetune` failure was the VLM masking assertion: the test accepted assistant text and older terminal variants, but this PR intentionally supervises the complete ChatML terminator `<|im_end|>\n`.
- Updated that functional assertion to accept the complete supervised terminator.

Targeted validation:
- 87 focused VLM boundary/collator unit tests passed.
- The previously failing VLM masking functional test passed under the same 2-process distributed launch shape used by CI.

New CI run: https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/28327723320 (currently waiting for the shared test queue).

yaoyu-33:
## Cross-path check: samples and results

This is the detailed tokenizer/collator comparison behind the earlier validation summary. The capture was produced with `Qwen/Qwen3-VL-2B-Instruct` after the mask/tool fixes that landed in `f8a6cf902`; current head `e728ac76d` additionally merges latest `main` and fixes the CI assertion for the complete `<|im_end|>\n` target. The post-merge focused unit and 2-process functional checks also pass.

This is a preprocessing comparison: it loads processor/tokenizer assets, not model weights. No internal paths, hosts, or credentials are needed to interpret the results.

### Executed paths

1. Bridge Qwen VLM HF collate (`qwen2_5_collate_fn`)
2. Bridge generic text HF collate (`text_chat_collate_fn`)
3. Bridge normalized Energon wrapper (`QwenVLTaskSample -> QwenVLTaskEncoder.batch`)
4. Bridge full Energon source (`ChatMLSample -> encode_sample -> batch`)
5. HF tokenizer with TRL's Qwen3-VL training template as the target-policy reference
6. LLaMA-Factory's Qwen3-VL supervised preprocessing as an additional external reference
7. Read-only contract comparison with Megatron-LM GitLab commit `50ffcd26a99d9e4368115078a318761f2418dd33`, `examples/multimodal/data_loading/task_encoder.py` and `examples/multimodal/train.py` (not `examples/mimo`)

### Chat samples

<details>
<summary>Four fixtures used for every HF/Energon/external-template path</summary>

```json
{
  "plain_multiturn": [
    {"role": "system", "content": "You are a concise assistant."},
    {"role": "user", "content": "What is 2 + 3?"},
    {"role": "assistant", "content": "5"},
    {"role": "user", "content": "Spell that number."},
    {"role": "assistant", "content": "five"}
  ],
  "reasoning_multiturn": [
    {"role": "user", "content": "What is 7 minus 4?"},
    {"role": "assistant", "content": "<think>Subtract four from seven.</think>\n\n3"},
    {"role": "user", "content": "Why?"},
    {"role": "assistant", "content": "<think>Check the inverse operation.</think>\n\nBecause 3 + 4 = 7."}
  ],
  "tool_call_multiturn": [
    {"role": "user", "content": "What is the weather in Paris?"},
    {"role": "assistant", "content": "", "tool_calls": [{"type": "function", "function": {"name": "get_weather", "arguments": {"city": "Paris"}}}]},
    {"role": "tool", "content": "Sunny, 21 C."},
    {"role": "assistant", "content": "It is sunny and 21 C in Paris."}
  ],
  "reasoning_tool_call": [
    {"role": "user", "content": "Should I carry an umbrella in Paris?"},
    {"role": "assistant", "content": "<think>I need current weather data.</think>", "tool_calls": [{"type": "function", "function": {"name": "get_weather", "arguments": {"city": "Paris"}}}]},
    {"role": "tool", "content": "Light rain, 14 C."},
    {"role": "assistant", "content": "<think>Rain means an umbrella is useful.</think>\n\nYes, carry one."}
  ]
}
```

Both tool fixtures used the same top-level `get_weather(city: string)` tool definition.
</details>

### Token/label/mask result

All four Bridge paths were exactly equal for `input_ids`, `attention_mask`, effective target IDs, and effective target positions on all four fixtures. "Targets" below means tokens that actually enter causal cross entropy after normalizing Bridge's prediction-slot-shifted labels and the external token-aligned labels.

| Fixture | Sequence length | Targets | Supervised `<|im_end|>` | Supervised post-end newline | 4 Bridge paths equal | HF + TRL template equal |
|---|---:|---:|---:|---:|---:|---:|
| Plain multi-turn | 45 | 6 | 2 | 2 | yes | yes |
| Reasoning multi-turn | 68 | 32 | 2 | 2 | yes | yes |
| Tool call + final answer | 220 | 34 | 2 | 2 | yes | yes |
| Reasoning + tool call | 238 | 47 | 2 | 2 | yes | yes |

Effective target text:

```text
plain:     5<|im_end|>\nfive<|im_end|>\n
reasoning: <think>Subtract four from seven.</think>\n\n3<|im_end|>\n
           <think>Check the inverse operation.</think>\n\nBecause 3 + 4 = 7.<|im_end|>\n
tool:      <tool_call>...get_weather("Paris")...</tool_call><|im_end|>\n
           It is sunny and 21 C in Paris.<|im_end|>\n
combined:  <think>I need current weather data.</think>\n<tool_call>...</tool_call><|im_end|>\n
           <think>Rain means an umbrella is useful.</think>\n\nYes, carry one.<|im_end|>\n
```

Concrete raw slice from the first assistant turn of the plain fixture:

```text
input_ids[23:33] = [198, 151644, 77091, 198, 20, 151645, 198, 151644, 872, 198]
labels[23:33]    = [-100, -100, -100, 20, 151645, 198, -100, -100, -100, -100]
loss_mask[23:33] = [0, 0, 0, 1, 1, 1, 0, 0, 0, 0]

effective target IDs       = [20, 151645, 198, 52670, 151645, 198]
effective target positions = [27, 28, 29, 42, 43, 44]
```

Thus user/system/tool context remains masked while assistant content, `<|im_end|>`, and its newline are supervised. Stock Qwen3-VL HF templates return an all-zero assistant mask because they lack `{% generation %}` blocks, so the stock mask is not a valid target reference. The TRL training template is the valid HF reference and matched exactly. LLaMA-Factory matched the first three fixtures; its legacy converter intentionally/structurally replaces assistant content when a message combines reasoning text and `tool_calls`, so the combined fixture is not byte-identical there.

### Direct packed + CP=2 result

The four fixtures above were collated directly into one THD batch with `pad_to_multiple_of=4`:

```text
input shape:                 [1, 576]
cu_seqlens_q / kv:           [0, 45, 113, 333, 571]
cu_seqlens_q_padded / kv:    [0, 48, 116, 336, 576]
max_seqlen_q / kv:           240 / 240
attention_mask:              None (THD metadata is the segmentation contract)
```

For every logical row:

- packed `input_ids`, effective targets, and target positions matched its non-packed row;
- physical alignment gaps had `labels=-100` and `loss_mask=0`;
- HF and normalized Energon batches were equal for `input_ids`, `position_ids`, `labels`, `loss_mask`, q/kv cu-seqlens, padded q/kv cu-seqlens, and max q/kv seqlens.

### Real image guard

A generated 56x56 red image plus "What color is this image?" was run through the HF and full Energon image paths:

```text
input shape:                    [1, 28]
<|image_pad|> input count:      1
<|image_pad|> supervised count: 0
effective target:               red<|im_end|>\n
```

HF and Energon were equal for `input_ids`, attention, positions, labels, loss mask, `pixel_values`, and `image_grid_thw`. The visual placeholder remains masked.

### GitLab `examples/multimodal` comparison

The Bridge collator was also run on three small SQuAD-derived records to compare current metadata with the inspected GitLab implementation:

| Key | Prompt | Answer | Image |
|---|---|---|---:|
| squad-0 | To whom did the Virgin Mary allegedly appear in 1858 in Lourdes France? | Saint Bernadette Soubirous | yes |
| squad-1 | What is in front of the Notre Dame Main Building? | a copper statue of Christ | yes |
| squad-2 | The Basilica of the Sacred Heart at Notre Dame is beside which structure? | the Main Building | no |

Observed Bridge results:

```text
logical lengths:                    [60, 48, 45]
pad64 shape:                        [1, 192]
pad64 cu_seqlens_q:                 [0, 60, 108, 153]
pad64 cu_seqlens_q_padded:          [0, 64, 128, 192]
CP2/pad4 shape:                     [1, 156]
CP2/pad4 cu_seqlens_q:              [0, 60, 108, 153]
CP2/pad4 cu_seqlens_q_padded:       [0, 60, 108, 156]
non-ignore labels / loss-mask sum:  16 / 16
```

HF-provider and Bridge Energon outputs were exactly equal for both cases, including all current packed metadata fields.

The GitLab CP=2 `hack=True` rules produce shape `[1, 156]`, padded boundaries `[0, 60, 108, 156]`, and `max_seqlen=60`, matching Bridge's physical layout. Two representation differences are expected:

1. GitLab places padded per-sample boundaries in both cu-seqlens arrays; Bridge preserves logical boundaries in `cu_seqlens_q/kv` and emits physical boundaries separately in `cu_seqlens_q_padded/kv_padded`, matching the current MCore/TE API.
2. GitLab emits labels with length `tokens + 1` and slices them in `train.py`; Bridge emits already-shifted labels with the token shape. After that runtime shift, label shapes and prediction positions agree.

So the Bridge paths are tensor-identical to each other, while the GitLab path is contract-equivalent after normalizing its older padding and label-shift representation; it is not claimed to be raw byte-identical.

cuichenx (approved):

PR #4517, opened 2026-06-26
[test] fix: avoid false skip in Nemotron Omni conversion

cuichenx:
## Summary

Fix the Nemotron Omni conversion functional test so it does not false-green before exercising the toy checkpoint.

PR #4287's intended toy-checkpoint conversion coverage did not actually run in GHA. Both the H100 `L0_Launch_models_nemotron_omni` job (`80716633167`) and GB200 `gb200_L0_Launch_models_nemotron_omni` job (`80716633028`) collected one test, skipped it in fixture setup with `No module named 'transformer_engine_torch'`, and still passed with no coverage data.

This PR:

- Changes the fixture preflight from `pytest.importorskip("transformer_engine_torch")` to a normal import of `transformer_engine.pytorch`, the public TE entry point that loads the native extension for source builds. If TE is unavailable or broken, the test now fails instead of silently skipping.
- Keeps the shrunken toy LLM config internally consistent by shortening `layer_types` and `layers_block_type` to match `num_hidden_layers=6` and `hybrid_override_pattern="MEMEM*"`. Transformers 5.8 validates these lengths when the conversion subprocess reloads the saved toy config.
- Adds output assertions for the shrunken LLM layer count/type lists so this does not regress quietly.

## Validation

- `python -m py_compile tests/functional_tests/test_groups/models/nemotron_omni/test_nemotron_omni_conversion.py`
- `git diff --check`
- Local two-GPU container run on NGC PyTorch 26.04 confirms the test no longer skips, builds the toy checkpoint, reloads the six-layer `MEMEM*` config, and reaches MCore `HybridModel` construction. The trimmed local container then stops on missing `mamba-ssm`; the CI image is built with all extras, including the SSM extra.

`uv run pre-commit run --all-files` without `--no-sync` was attempted first and failed during local dependency resolution because `nvidia-resiliency-ext==0.6.0` has no wheel for this host's `manylinux_2_31_x86_64` platform. The hook set itself passes with `--no-sync`.

GHA CICD NeMo is triggered for the current head and pending.

## Blast Radius

One functional-test fixture only: `tests/functional_tests/test_groups/models/nemotron_omni/test_nemotron_omni_conversion.py`.

Recommended CI scope: L0 is sufficient, specifically H100 and GB200 `L0_Launch_models_nemotron_omni`. No L1/L2 labels requested.


cuichenx:
(done) ~need radio model cache~

PR #4521, opened 2026-06-26
[ckpt, data] feat: complete Energon dataloader checkpointing save/res…

### LINKED ISSUE #3993: [feature] Complete Energon dataloader checkpointing: expose `dataloader_save` and implement restore

zsombor-kaiko:
Closes NVIDIA-NeMo/Megatron-Bridge#3993

# What does this PR do ?

Completes end-to-end Megatron Energon dataloader checkpointing: makes the save path a discoverable config field and implements the previously-missing restore, so a resumed run continues over the **same data stream** instead of restarting from an arbitrary position. Reworks the approach from [#4235](<https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/4235>) (@rob-luke) and folds in the review feedback left there, opened fresh with @rob-luke's agreement.

# Changelog

* Declare `dataloader_save` / `dataloader_load` on `DataloaderConfig`, replacing the undocumented `getattr(cfg.dataset, "dataloader_save", None)` convention so the path is discoverable in the schema. Both default to an `energon/` subdirectory of the checkpoint dir when unset, so saving model checkpoints implies saving/restoring dataloader state with no extra config.
* Return the train split un-wrapped from `EnergonProvider.build_datasets` (drop the `iter(...)`) so the downstream `RerunDataIterator.iterable` retains `save_state` / `restore_state`.
* Add `EnergonDataloader.restore_state`, symmetric to `save_state`.
* Add `maybe_load_dataloader_state`, called from `setup()` on resume, plus default save-path resolution in `save_checkpoint`.
* Document the feature in `docs/training/checkpointing.md`, including the per-sample determinism requirement. Add unit + functional tests.

# Design notes (differences from [#4235](<https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/4235>))

These address the review on [#4235](<https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/4235>):

1. **Restore is not symmetric to save.** Save writes one file per DP replica (gated to a single model-parallel writer, since the per-DP-rank state is identical across model-parallel ranks). Restore runs on **every** rank, keyed by the **pure DP rank** — each tensor/pipeline/context rank pulls from its own data iterator (e.g. `qwen3_vl` `get_batch`), so all must be rewound. Keying by pure DP rank (excluding CP) means CP ranks in a replica restore from the same file, which is the context-parallel case @yaoyu-33 flagged.
2. **Default-on.** `dataloader_save` / `dataloader_load` default to `{checkpoint.save|load}/energon`, so resume reproduces the data stream without the user having to set a path. Save is gated on the dataloader actually supporting `save_state`, so non-Energon runs are unaffected.
3. **Fail loud on a real mismatch.** If the dataloader state dir is absent, the checkpoint predates this feature → start fresh (logged). If the dir exists but the current rank's per-DP-rank file is missing, restore **raises** — this almost always means the data-parallel size changed since the checkpoint was saved, and silently resuming would change the data order.
4. **Determinism.** Documented that restore replays the in-flight samples by re-running the pipeline, so encoding must be deterministic per sample (`restore_seeds=True`).

# Testing

## Unit tests

`maybe_load_dataloader_state` branch matrix (no path / no iterator / non-Energon iterable / **restores on a non-zero pp/tp rank** / dir-absent skip / dir-present-but-file-missing **raises** / happy path); `build_datasets` train-split preservation.

## End-to-end training

Bit-for-bit resume validated across parallelism configs on an internal Qwen-VL training stack (sequence packing enabled). In every case **step 101 — the first step after resuming from step 100 — matches the fresh run's** `lm loss` **to 16 digits**, the packed-sample `file_names` are identical, and the per-step packed `full_sequence` MD5s match. The \~1e-4 drift at later steps is ordinary CUDA nondeterminism. In each pair, **A** is a fresh run and **B** resumes from A at step 100.

### Qwen3.5-VL 2B - (ws4-dp4-tp1-pp1-cp1)

| Check | Step 101 | Step 200 | Step 299 |
| -- | -- | -- | -- |
| `lm loss` (A / B) | **2.0937845706939697 / 2.0937845706939697** | 1.9516996145248413 / 1.9516639709472656 (Δ=3.6e-5) | 1.9557422399520872 / 1.9552744626998900 (Δ=4.7e-4) |
| `file_names` | ✓ identical | ✓ identical | ✓ identical |
| `full_sequence` MD5 | ✓ `9566d8a7c5dbcde904e3cbdcb18e5715` | ✓ `58049a738b9749ffdd119dbdd571df68` | ✓ `58cae7ae1cf510bf10d5b3ec246d2db3` |

### Qwen3.5-VL 2B - (ws4-dp1-tp2-pp2-cp1)

| Check | Step 101 | Step 200 | Step 299 |
| -- | -- | -- | -- |
| `lm loss` (A / B) | **2.2256877422332764 / 2.2256877422332764** | 2.0771758556365967 / 2.0770578384399414 (Δ=1.2e-4) | 2.2031960487365723 / 2.2033071517944336 (Δ=1.1e-4) |
| `file_names` | ✓ identical | ✓ identical | ✓ identical |
| `full_sequence` MD5 | ✓ `456b80e0fc6b7f599e0f60ce62b3c7da` | ✓ `c20be30a119f78d21d37781c595f6928` | ✓ `0c776aa3787d6c694990fdba3cdc5bd7` |

### Qwen3-VL 8B - (ws8-dp4-tp1-pp1-cp2)

| Check | Step 101 | Step 200 | Step 299 |
| -- | -- | -- | -- |
| `lm loss` (A / B) | **1.9300434589385984 / 1.9300434589385984** | 1.7472652196884155 / 1.7441542148590088 (Δ=3.1e-3) | 1.9940789937973025 / 1.9956401586532593 (Δ=1.6e-3) |
| `file_names` | ✓ identical | ✓ identical | ✓ identical |
| `full_sequence` MD5 | ✓ `0d87afa8224bc95cc2f1d944b999d00b` | ✓ `b154e41d18d7463630c3348300557092` | ✓ `bc0d7df88a49dbb29046fba7dc9eb884` |

# GitHub Actions CI

See the [CI section](<https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci>) in the Contributing doc for how to trigger the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

**Pre checks**:

- [X] Make sure you read and followed [Contributor guidelines](<https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md>)
- [X] Did you write any new necessary tests? (unit + functional)
- [X] Did you add or update any necessary documentation? (`docs/training/checkpointing.md`)
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc) — No optional-install components are affected.

# Additional Information

* Fork's PR [https://github.com/kaiko-ai/Megatron-Bridge/pull/3](<https://github.com/kaiko-ai/Megatron-Bridge/pull/3>)
* Supersedes [#4235](<https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/4235>) (@rob-luke), opened fresh with his agreement.


src/megatron/bridge/training/checkpointing.py
```diff
@@ -1613,9 +1626,14 @@ def maybe_save_dataloader_state(
...
 
-    # Resolve process groups and save dataloader state for each DP rank only once.
+    # Resolve process groups and write the per-DP-rank state from a single writer. Tensor-,
+    # pipeline-, and context-parallel ranks of a DP replica all hold the identical per-DP-rank
+    # state, so only the tp/pp/cp leader writes avoiding racing to write the same
+    # train_dataloader_dprank{dp}.pt file.
     pg_collection = pg_collection or get_pg_collection(model)
-    is_first_rank = (pg_collection.pp.rank() == 0) and (pg_collection.tp.rank() == 0)
+    is_first_rank = (
+        (pg_collection.pp.rank() == 0) and (pg_collection.tp.rank() == 0) and (pg_collection.cp.rank() == 0)
```
yaoyu-33: Could we use `megatron.core.utils.get_pg_rank` for this gate (and the rank lookups below)? `ProcessGroupCollection.cp` may be `None` for a valid singleton/partially populated collection—Energon’s `_build_worker_config` already handles that case—so direct `.rank()` makes checkpoint save fail. MCore’s helper defines a missing/uninitialized group as rank 0 and is the established pattern in `megatron/training/training.py`. Please add a `cp=None` test as well.

src/megatron/bridge/training/setup.py
```diff
@@ -365,6 +368,22 @@ def modelopt_pre_wrap_hook(model):
...
     barrier_and_log("after dataloaders are built")
 
+    # Resume the dataloader stream position from the loaded checkpoint so a resumed run continues
+    # over the same data (currently only Megatron Energon). Runs after the iterator is built and the
+    # model checkpoint load restored state.train_state.step. Defaults the source to an `energon`
+    # subdir of checkpoint.load; an explicit dataset.dataloader_load overrides. No-op for
+    # non-Energon dataloaders or checkpoints saved without dataloader state.
+    if should_load_checkpoint:
+        dataloader_load_path = getattr(cfg.dataset, "dataloader_load", None)
+        if dataloader_load_path is None and cfg.checkpoint.load:
```
yaoyu-33: The checkpoint just loaded is not necessarily rooted at `checkpoint.load`. `should_load_checkpoint` can be true solely because a local non-persistent checkpoint exists; with `checkpoint.load=None`, this leaves the path unset and silently skips restore even though save wrote the state under `checkpoint.save/energon`. If `load` and `save` differ, recovery can also look in the old run. A direct iteration load has the inverse layout problem (`.../iter_N/energon/iter_N` instead of the saved `.../energon/iter_N`). Please resolve the default from the checkpoint source/layout actually selected and cover local non-persistent and direct-iteration resume.

tests/unit_tests/training/test_checkpointing.py
```diff
@@ -4089,3 +4091,139 @@ def test_load_global_checkpoint_uses_standard_load_state_dict(
         # Standard load_state_dict must be called; per-rank file loader must NOT be called.
         mock_layer_wise_optim.load_state_dict.assert_called_once_with(mock_state_dict["optimizer"])
         mock_layer_wise_optim.load_state_dict_from_file.assert_not_called()
+class TestMaybeLoadDataloaderState:
```
yaoyu-33, line 4542: Please run the repository formatter before merging. With the pinned Ruff 0.9.9, `ruff format --check tests/unit_tests/training/test_checkpointing.py` reports that this file would be reformatted (starting with the missing blank lines before this class).

src/megatron/bridge/training/checkpointing.py
zsombor-kaiko: Makes sense, fixed in [0ba8d13](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4521/commits/0ba8d131bdb871ae675de43f6da27136abcaac20)

src/megatron/bridge/training/setup.py
zsombor-kaiko: Good catch, thanks for raising! I derived the default from the model checkpoint to cover all cases in [d2c040e](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4521/commits/d2c040e685d8de0054e1f362fb415b3ce0a6cc36)

tests/unit_tests/training/test_checkpointing.py
zsombor-kaiko: Changes in [e7f0923](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4521/commits/e7f09238bc10e81f146e0d1fb99b19f9f267ed55)

zsombor-kaiko:
Thanks for the review @yaoyu-33! I've pushed the requested changes. Let me know what else I can do to get this merged as soon as possible.

liding-nv:
@zsombor-kaiko thanks for the update, can you fix the DCO check so we can proceed with CI testing

zsombor-kaiko:
Oh right, I forgot to sign off the last two commits... I added them now and also fixed some conflicts with main. Should be good to run the CI @liding-nv

psycherun-kaiko:
Hi, just commenting here in order to get notifications about updates/comments on this PR

zsombor-kaiko:
Thanks @liding-nv. I see some tests already passed but some still need approval. I will be off after today for two weeks but my colleague @psycherun-kaiko can assist if something else is needed from our side.

huvunvidia:
Hi @psycherun-kaiko,
I see that all the tests have passed but there is a merge conflict. Can you resolve it and I can help trigger the tests again?

psycherun-kaiko:
Hi @huvunvidia, I resolved the merge conflict, could you please trigger the tests again? Thanks!

huvunvidia:
/ok to test [979a43d](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4521/commits/979a43db3f4b7e4040d974e808202273167cc8ef)

huvunvidia:
Hi @psycherun-kaiko ,
The tests ran and passed all important tests, but now there is a merge conflict. Can you have a quick look?
Thanks.

psycherun-kaiko:
@huvunvidia done, in short: PR adds a couple fields to DataloaderConfig, which upstream moved from `src/megatron/bridge/training/config.py` to `src/megatron/bridge/data/base.py`. So the merge commit [a6d2680](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4521/commits/a6d26808c59bad04ab4f20a877a394fb9283f186) just moves these fields over. There is no functional change.

huvunvidia:
/ok to test [a6d2680](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4521/commits/a6d26808c59bad04ab4f20a877a394fb9283f186)

psycherun-kaiko:
Hi @huvunvidia, is there anything I can do to help you with the merging of this PR?

huvunvidia:
/ok to test [b2fa411](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4521/commits/b2fa4110483e1c6a1dc004dc05bc33ea62001b32)

huvunvidia:
Hi @psycherun-kaiko ,
Currently no. We are having reviewers to review your PRs. Thank you.

huvunvidia:
Hi @psycherun-kaiko ,
It seems like the curerrnt PR has conflicts with the latest main. Can you have another look?
I can trigger the test again when you're finished.

psycherun-kaiko:
@huvunvidia done. The merge conflict that needed resolution was that #4774 deleted the energon_provider.py, moving the logic over to `src/megatron/bridge/data/builders/energon.py`. For dataloader resume, we remove the iter() on `train_dataloader()`. So I moved this logic over as well.

yaoyu-33 (approved):

zsombor-kaiko:
Can someone take a look at the failing code coverage test? It seems that it's independent from this PR's changes and possibly merging main and triggering the test would solve it?

PR #4532, opened 2026-06-26
[training] fix Qwen3-VL packed vlm_step MRoPE

cuichenx:
## Summary

This PR makes the generic `vlm_step` path work for Qwen3.5-VL in-batch packed-sequence training, so the specialized `qwen3_vl_step` path can be deprecated later without losing Qwen MRoPE behavior.

The original packed-path bug had two parts:

1. Qwen3-VL needs 3-axis MRoPE positions computed from the full multimodal token stream.
2. The model then repacked/retargeted tensors that the collator had already emitted in MCore THD layout, which could make positions, hidden states, supervision tensors, and packed metadata describe different token streams.

The generic path now treats collate-time packing as authoritative. Qwen computes or consumes full packed MRoPE positions, applies one Transformer Engine partition index when CP is enabled, and passes the original packed metadata to the language model.

## Changes

- Keep token IDs and position IDs available on every pipeline rank.
- Keep visual metadata on later pipeline ranks, while dropping heavy image/video payload tensors after the first rank.
- Reject `defer_in_batch_packing_to_step=True` in `vlm_step`; generic VLM packing must happen in the collator.
- Remove Qwen/model-shape detection and deferred Qwen packing logic from the generic step.
- Treat collate-owned `[1,T]` THD batches as authoritative with no second model-side compaction or `cu_seqlens` retargeting. Retain the original `[B,S]` conversion only for the deprecated `qwen3_vl_step` compatibility path.
- For packed CP, derive one TE THD partition index from padded `cu_seqlens` and apply it consistently to LM input IDs, embeddings, vision/deepstack masks, 3-axis MRoPE positions, labels, and loss masks.
- Give packed MTP an isolated metadata copy whose logical boundaries use the physical padded offsets, preventing future-token rolls from crossing alignment gaps while leaving attention metadata unchanged.
- Share the packed-CP partition-index helper with `slice_batch_for_context_parallel()`.
- Keep the non-packed CP protection that avoids double-slicing tensors which are already rank-local.
- Handle a model-returned sliced loss mask in both `vlm_step` and the deprecated `qwen3_vl_step`.
- Fix Qwen3-VL PP + MTP tied-embedding checkpoint state-dict generation by dropping the redundant output-layer checkpoint entry on non-preprocess MTP ranks.

### Pipeline input design note

We considered a `PipelineBatchSpec`-style, per-model contract so later PP ranks could receive only model-specific MRoPE metadata. That would require every model to inherit a default specification and models such as Qwen to override it, adding a new configuration surface solely to avoid broadcasting small integer tensors. Model detection inside generic `vlm_step` had the same coupling problem. We therefore did not add that abstraction: token IDs and position IDs are sent to every rank, while the much larger image/video payloads remain first-stage-only.

## Experiment labels

- **A**: pre-fix generic `vlm_step` with collate-time in-batch packing. This path completed but produced the wrong loss trajectory.
- **B**: fixed generic `vlm_step` with the same collate-time in-batch packing setup.
- **C**: existing specialized `qwen3_vl_step` baseline, which defers packing to the step.

The expected result is that B matches C wherever C provides a valid baseline, while A differs.

## Validation

Latest review-driven implementation:

- `uv run --no-sync --with pre-commit pre-commit run --all-files`
- 26.06 GPU container: 69 Qwen model/rope, legacy-step compatibility, generic `vlm_step`, and packed MTP boundary tests passed, with 2 expected skips.
- 26.06 GPU container: 72 collator, packed metadata, and CP slicing tests passed.
- The Qwen collator test enables both `pad_to_max_length` and in-batch packing and verifies the output is packed `[1,T]`, confirming the pack-or-pad branch selects packing rather than max-length padding.

Direct packed A/B/C comparison (global batch size 32, sequence length 2048):

- A (pre-fix `vlm_step`, MBS=4) final LM/MTP: `0.0846988` / `0.159827`
- B (fixed `vlm_step`, MBS=4) final LM/MTP: `0.0877338` / `0.162998`
- C (`qwen3_vl_step`, MBS=2) final LM/MTP: `0.0877426` / `0.169921`
- B follows the C baseline trajectory, while A follows the incorrect pre-fix trajectory.

Final H100 end-to-end validation on commit `f41f3dc8c` in the 26.06 RC8 container:

- All six jobs completed 20 iterations with zero skipped iterations, zero NaN iterations, and no fatal exceptions.
- TP=4 + sequence parallel:
  - B mean LM loss: `0.860150`; final LM/MTP: `0.598935` / `0.853095`
  - C mean LM loss: `0.866036`; final LM/MTP: `0.601689` / `0.856164`
- PP=4:
  - B mean LM loss: `0.859580`; final LM/MTP: `0.597122` / `0.851291`
  - C mean LM loss: `0.861437`; final LM/MTP: `0.599258` / `0.853385`
- CP=4 generic `vlm_step`:
  - Packed mean LM/MTP: `0.862322` / `0.977947`; final LM: `0.597955`
  - Padded mean LM/MTP: `0.864237` / `0.961900`; final LM: `0.598279`
  - The specialized `qwen3_vl_step` CP path remains unsupported and was not used as a C baseline.
- The main packed comparison also confirms that pre-fix A follows the wrong loss trajectory while fixed B matches C.


src/megatron/bridge/training/vlm_step.py
```diff
@@ -82,13 +85,81 @@ def _filter_visual_kwargs_for_model(model: Any, visual_kwargs: Mapping[str, torc
     return {key: value for key, value in visual_kwargs.items() if key in supported_kwargs}
 
 
+def _defer_in_batch_packing_to_step(cfg: ConfigContainer) -> bool:
```
yaoyu-33: why we need this? we should not defer for vlm step in general?

src/megatron/bridge/training/vlm_step.py
```diff
@@ -82,13 +85,81 @@ def _filter_visual_kwargs_for_model(model: Any, visual_kwargs: Mapping[str, torc
...
+    """Return whether the current dataset expects the training step to pack the batch."""
+    dataset_cfg = getattr(cfg, "dataset", None)
+    return bool(
+        dataset_cfg is not None
+        and getattr(dataset_cfg, "enable_in_batch_packing", False)
+        and getattr(dataset_cfg, "defer_in_batch_packing_to_step", False)
+    )
+
+
+def _has_qwen_mrope_position_ids(batch: Mapping[str, Any]) -> bool:
```
yaoyu-33: is this a bit hacky? can we check model type?

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
```diff
@@ -52,6 +52,187 @@
...
+        total_tokens=None,
+    )
+
+
+def _is_qwen_mrope_position_ids(position_ids: torch.Tensor | None) -> bool:
+    """Return whether ``position_ids`` is explicit Qwen MRoPE metadata."""
+    return isinstance(position_ids, torch.Tensor) and position_ids.dim() == 3 and position_ids.size(0) == 3
+
+
+def _get_packed_cp_index(
```
yaoyu-33: Could we extract this packed-CP index calculation into a shared helper, perhaps in `training/utils/packed_seq_utils.py`? `slice_batch_for_context_parallel()` already performs the same padded `cu_seqlens_q` selection and `thd_get_partitioned_indices()` call. The existing function cannot be reused directly because Qwen has 3-axis MRoPE and additional sequence-aligned tensors, but sharing the lower-level index calculation would keep CP partition semantics in one place.

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
```diff
@@ -52,6 +52,187 @@
...
+    if (
+        use_padded_cu_seqlens
+        and total_sequence_length is not None
+        and packed_seq_params is not None
+        and packed_seq_params.cu_seqlens_q_padded is not None
+        and packed_seq_params.cu_seqlens_kv_padded is not None
+    ):
+        cu_seqlens_q = packed_seq_params.cu_seqlens_q_padded
+        cu_seqlens_kv = packed_seq_params.cu_seqlens_kv_padded
+        cu_seqlens_q, max_seqlen_q = _retarget_cu_seqlens(
```
yaoyu-33: why need re-target?

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
```diff
@@ -52,6 +52,187 @@
...
+    if total_sequence_length is None or int(cu_seqlens[-1].item()) == total_sequence_length:
+        return cu_seqlens, max_seqlen
+
+    cu_seqlens = cu_seqlens.clone()
+    cu_seqlens[-1] = total_sequence_length
+    max_seqlen = int((cu_seqlens[1:] - cu_seqlens[:-1]).max().item())
+    return cu_seqlens, max_seqlen
+
+
+def _language_model_packed_seq_params(
```
yaoyu-33: this looks complicated, why in general needs this

src/megatron/bridge/training/vlm_step.py
cuichenx: qwen 3 vl model still packs inside the forward pass so this was added. But I think you're right, a better approach is to make deferred configuration fail when vlm_step is used

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
cuichenx: will refactor this part after the pack or pad change on main.

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
```diff
@@ -644,4 +943,6 @@ def forward(
         if self.use_dist_train:
             if not is_pp_last_stage(self.pg_collection.pp):
                 return {"language_module": output}
+        if return_compacted_loss_mask:
+            return output, loss_mask
         return output
```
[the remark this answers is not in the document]
cuichenx: Handled. `qwen3_vl_step` now unpacks `(output, loss_mask)` and uses the model-returned mask when constructing the loss function. `test_forward_step_uses_compacted_loss_mask_from_model` covers the compatibility path.

src/megatron/bridge/training/vlm_step.py
cuichenx: Agreed. Deferred packing support has been removed from generic `vlm_step`. It now requires collate-time packing and raises early when `enable_in_batch_packing=True` is combined with `defer_in_batch_packing_to_step=True`.

src/megatron/bridge/training/vlm_step.py
cuichenx: Removed this detection entirely. The generic step now keeps token IDs and position IDs on every PP rank for every VLM; only the heavy pixel/video payloads are dropped after the first stage. This avoids both model-type checks and Qwen metadata heuristics.

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
cuichenx: Done in `fc59880dd`. The padded-boundary selection and TE `thd_get_partitioned_indices` call now live in `get_packed_seq_cp_partition_indices()` under `training/utils/packed_seq_utils.py`, and both Qwen and `slice_batch_for_context_parallel()` use it.

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
cuichenx: The retargeting is removed in `fc59880dd`. With current collate-time pack-or-pad behavior, the collator owns the THD stream and its packed metadata; rewriting only the final `cu_seqlens` entry in the model could make that metadata inconsistent with the actual layout.

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
cuichenx: Simplified in `fc59880dd`. Qwen no longer runs a second packed compaction or builds replacement packed metadata in `forward()`. It consumes the collator's THD layout directly and only applies one shared partition index when packed CP is enabled.

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
```diff
@@ -50,6 +50,43 @@
     split_deepstack_embs,
 )
 from megatron.bridge.models.qwen_vl.modelling_qwen3_vl.vision_model import Qwen3VLVisionModel
+from megatron.bridge.training.utils.packed_seq_utils import get_packed_seq_cp_partition_indices
+
+
+def _is_qwen_mrope_position_ids(position_ids: torch.Tensor | None) -> bool:
```
cuichenx: rename to _is_mrope_position_ids

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
```diff
@@ -50,6 +50,43 @@
...
+    """Return whether ``position_ids`` is explicit Qwen MRoPE metadata."""
+    return isinstance(position_ids, torch.Tensor) and position_ids.dim() == 3 and position_ids.size(0) == 3
+
+
+def _is_legacy_packed_bshd(input_ids: torch.Tensor | None, packed_seq_params: PackedSeqParams | None) -> bool:
```
cuichenx: this doesn't need a helper function, it's only used once

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/rope.py
```diff
@@ -31,6 +31,79 @@
 from megatron.bridge.models.qwen_vl.modelling_qwen3_vl.transformer_config import Qwen3VLTransformerConfig
 
 
+def _get_packed_cu_seqlens(
```
cuichenx: remove this helper function

yaoyu-33 (approved):

Issue #4294, opened 2026-06-11
Improve Megatron Bridge UX based on prototype MBridge comparison


yaoyu-33:
## Context

We received feedback that Megatron Bridge can feel worse from a user-experience perspective than the earlier prototype repo:

- Prototype: https://github.com/ISEEKYAN/mbridge/tree/main/mbridge
- Megatron Bridge: https://github.com/NVIDIA-NeMo/Megatron-Bridge

The prototype README now points users to Megatron Bridge for production support, but the prototype still has a compact UX that downstream users liked:

```python
from mbridge import AutoBridge

bridge = AutoBridge.from_pretrained(HF_MODEL_PATH)
model = bridge.get_model(weight_path=HF_MODEL_PATH)

for key, weight in bridge.export_weights(model):
    ...

bridge.save_weights(model, "path/to/save/model", memory_efficient=False)
```

Megatron Bridge has broader production functionality: training loop, recipes, PEFT, FP8/BF16/FP4, conversion verification, many model families, and more robust checkpoint/export flows. However, the first-use path can feel heavier because users often need to understand providers, `to_megatron_provider()`, provider parallelism fields, `finalize()`, distributed model construction, and separate conversion scripts before they can complete a simple import/export task.

## Goal

Create a concrete comparison between the prototype MBridge UX and current Megatron Bridge UX, then identify improvements that preserve Megatron Bridge's production architecture while reducing friction for common users.

## Proposed investigation

Please compare these areas side by side:

1. **Quickstart API ergonomics**
   - Prototype path: `AutoBridge.from_pretrained()`, `get_model()`, `export_weights()`, `save_weights()`.
   - Current path: `AutoBridge.from_hf_pretrained()`, `to_megatron_provider()`, provider config/finalize, `provide_distributed_model()`, `export_hf_weights()`, `save_hf_pretrained()`.
   - Identify whether docs alone are enough, or whether we should add a narrow convenience wrapper for the simple path.

2. **Common workflows**
   - HF -> Megatron in-memory load.
   - Megatron -> HF export.
   - Round-trip verification.
   - Single-GPU smoke test.
   - Multi-GPU TP/PP import/export.
   - RL/downstream integration where users want a Megatron model and streamed HF weights without learning the full training stack.

3. **Discoverability and naming**
   - Are method names intuitive for users coming from the prototype?
   - Are aliases or docs callouts useful, e.g. mapping old prototype names to current Megatron Bridge names?
[...]

PR #2810, opened 2026-03-15
Integrate ModelConfig & ModelBuilder interface into conversion

maanug-nv:
# What does this PR do ?
A possible implementation of integrating ModelConfig and ModelBuilder into the Bridge APIs. This PR is incomplete, mainly just for reference on what changes are needed.

# Key Changes

- Add `to_megatron_config()` to `AutoBridge`. Eventually deprecate `to_megatron_provider()`. This requires a `config_bridge()` API in `ModelBridge` which may need to be overriden for specific models' bridges.
- When importing HF checkpoint to Megatron format, initializing parallel state (if not already initialized) is moved to `AutoBridge` from the mixin, since `ModelBuilder` should not perform this logic
- Since `build_model()` now only passes `TransformerConfig` to the MCore model, (old `provide()` interface passed the entire provider to MCore model), need to keep track of the full ModelConfig when saving a megatron checkpoint through the bridge

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci)in the Contributing doc for how to trigger the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

# Before your PR is "Ready for review"

**Pre checks**:

- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?

If you haven't finished some of the above items you can still open "Draft" PR.

# Additional Information

- Related to # (issue)

PR #4543, opened 2026-06-28
feat(conversion): add direct ModelConfig mapping

yaoyu-33:
## What does this PR do?

Add the first direct `ModelConfig` / `ModelBuilder` foundation to the conversion API for RL and external-trainer integrations.

- add `AutoBridge.to_megatron_model_config()`
- add `MegatronModelBridge.hf_config_to_model_config()` as the canonical HF-config-to-`ModelConfig` path
- reuse the existing `CONFIG_MAPPING` without routing the new path through `provider_bridge()`
- preserve Bridge's established construction defaults where upstream `ModelConfig` defaults differ
- cover stock GPT via Llama and Hybrid via Nemotron-H
- keep the provider path compatible while its callers migrate

Related to #4294. This also incorporates the useful direction from the incomplete reference draft #2810, rebuilt against current `main` with focused parity tests.

## Design

The dependency direction is intentionally:

```text
HF config -> CONFIG_MAPPING -> hf_config_to_model_config() -> ModelConfig
                         \-> hf_config_to_provider_kwargs() -> provider (compatibility path)
```

`to_megatron_model_config()` is configuration-only: it does not load weights, initialize distributed state, finalize the config, or build a model.

Model families opt in explicitly with `MODEL_CONFIG_CLASS` and may specialize `hf_config_to_model_config()` for architecture defaults. Llama maps directly to `GPTModelConfig`; Nemotron-H maps directly to `HybridModelConfig`. Tests patch `provider_bridge()` and the provider-era mapping alias to fail if the new path accidentally depends on either one.

The shared `CONFIG_MAPPING` traversal is kept private. `hf_config_to_provider_kwargs()` remains as the provider compatibility hook, and `provider_bridge()` continues calling it so downstream subclasses that override the hook are not broken.

The direct path explicitly carries Bridge construction defaults such as loss/gradient fusions and pipeline output deallocation. Nemotron-H additionally preserves its Hybrid-specific RMSNorm, flash-attention, fusion, and MTP defaults. This avoids silently changing training behavior merely because upstream builder dataclasses use more conservative defaults.

This PR does **not** emit a runtime deprecation warning for `provider_bridge()` / `to_megatron_provider()` yet: current model construction and checkpoint import still depend on them. Formal deprecation should follow the builder-backed `to_megatron_model()` migration, when callers have a complete replacement.

## Tests

Added focused coverage for:

- Llama/GPT provider-to-`ModelConfig` parity across all overlapping transformer and model fields
- a non-default Llama rotary base, Llama 2 RoPE defaults, and Llama 3 scaling preservation
- Nemotron-H/Hybrid provider-to-`ModelConfig` parity across all overlapping fields
- direct-path independence from `provider_bridge()` and `hf_config_to_provider_kwargs()`
- `AutoBridge.to_megatron_model_config()` delegation

Validation:

- nightly NeMo container focused pytest: **4 passed**
- `uvx pre-commit run --all-files`: passed
- `git diff --check`: passed
- changed-source compile check: passed

The focused pytest command covered the two Llama config tests, the Nemotron-H parity test, and AutoBridge delegation. The local macOS ARM environment cannot resolve the full project because `nvidia-resiliency-ext` is distributed only as Linux wheels; no dependency or lockfile changes were made.

## Intentional limitations / follow-up

- only explicitly opted-in GPT and Hybrid families are supported in this first PR
- no overrides argument, aliases, or generic `TypeVar` helper are added
- no provider mutation or duplicate `provider_bridge` abstraction is introduced
- no MLA, VLM, metadata serialization, weight-loading, distributed initialization, or builder-backed construction changes
- the next construction PR can migrate `to_megatron_model()` and then formally deprecate the public provider path

## Checklist

- [x] Public APIs include type hints and docstrings
- [x] Focused parity coverage added and run in a Linux nightly container
- [x] DCO sign-off included
- [x] No dependency, CI workflow, or `3rdparty/Megatron-LM` changes

### DIFF
--- a/src/megatron/bridge/models/common/__init__.py
+++ b/src/megatron/bridge/models/common/__init__.py
@@ -18,7 +18,6 @@ from megatron.bridge.models.common.base import (
     ModelConfig,
     ModelT,
     Serializable,
-    apply_model_config_overrides,
     compose_hooks,
 )
 from megatron.bridge.models.common.unimodal import build_virtual_pipeline_stages, unimodal_build_distributed_models
@@ -30,7 +29,6 @@ __all__ = [
     "ModelConfig",
     "ModelT",
     "Serializable",
-    "apply_model_config_overrides",
     "compose_hooks",
     "build_virtual_pipeline_stages",
     "unimodal_build_distributed_models",
--- a/src/megatron/bridge/models/common/__init__.py
+++ b/src/megatron/bridge/models/common/__init__.py
@@ -18,6 +18,7 @@ from megatron.bridge.models.common.base import (
     ModelConfig,
     ModelT,
     Serializable,
+    apply_model_config_overrides,
     compose_hooks,
 )
 from megatron.bridge.models.common.unimodal import build_virtual_pipeline_stages, unimodal_build_distributed_models
@@ -29,6 +30,7 @@ __all__ = [
     "ModelConfig",
     "ModelT",
     "Serializable",
+    "apply_model_config_overrides",
     "compose_hooks",
     "build_virtual_pipeline_stages",
     "unimodal_build_distributed_models",
--- a/src/megatron/bridge/models/common/base.py
+++ b/src/megatron/bridge/models/common/base.py
@@ -12,10 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections.abc import Mapping
 from dataclasses import fields as dataclass_fields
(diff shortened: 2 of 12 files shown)


yaoyu-33:
Closing as superseded by #4545, which includes this direct ModelConfig/ModelBuilder foundation and carries the broader AutoBridge migration. Keeping both open would duplicate implementation and review effort.

PR #4545, opened 2026-06-28
[DO NOT MERGE] refactor(models): migrate AutoBridge to model builders

yaoyu-33:
## Summary

- migrate all 52 non-diffusion registered AutoBridge families to serializable ModelConfig plus stable standalone ModelBuilder contracts
- remove provider fallbacks from primary AutoBridge build and adapter-export paths while retaining deprecated provider compatibility entry points
- migrate official text and VLM recipe call sites plus inference paths off provider build APIs
- move every in-scope registered config target onto a provider-neutral import path, including transitive family-local dependencies for legacy bridge-local configs
- keep family-only state outside exact Megatron-Core TransformerConfig dataclasses, including Qwen multimodal, YaRN, and Nemotron-H MTP construction
- preserve flat and nested Hydra model overrides while retaining the outer serializable ModelConfig across construction, checkpoint save, config reload, and reconstruction
- add registry-complete manifests, provider-free primary-consumer/import-graph guards, config fidelity/round-trip coverage, and direct builder contract tests

## Stacked follow-ups

- #4556 migrates diffusion bridges and recipes
- #4557 migrates MegatronMIMO construction, conversion, checkpointing, and data loading

Both follow-ups are based on this branch so the generic ModelConfig and ModelBuilder infrastructure remains centralized here.

## Validation

- `uvx pre-commit run --all-files`
- EOS changed-surface suite: 839 passed, 22 deselected before the domain split
  - the deselected TestExportAdapterScript tests are unchanged and shadowed by the EOS container preloaded `examples` namespace
- EOS provider-neutral config/bridge suite: 322 passed
- EOS recipe migration validation: 1054 passed in the broad slice; the 11 initially exposed compatibility failures were fixed and all affected paths passed in a 98-test focused rerun
- EOS CI-regression suite: 67 passed across Falcon H1, MiMo v2 Flash, Qwen Omni/VL, Stepfun, Kimi VL, and provider-neutral import coverage
- EOS concrete smoke: tiny Llama completed AutoBridge build -> distributed checkpoint save -> ModelConfig reload -> model reload on CPU; both constructed models used exact Megatron-Core TransformerConfig
- EOS functional round trip: Nemotron3 Nano AutoBridge build -> save -> load -> HF export
- the original combined tree completed all 90 required PR checks successfully; split-specific EOS and CI validation is being rerun on each branch
- independent implementation and review passes covered text, multimodal/Qwen, recipes/inference, registry completeness, checkpoint round trips, serialization/import neutrality, provider-path removal, and split-boundary correctness

## DeepSeek-V4 compatibility

DeepSeek-V4 has the same serializable config/builder contract as every other in-scope registration, with no provider fallback. The pinned Megatron-Core only exposes the `gated_delta_net` and `dsa` experimental attention variants and does not yet contain DSv4 CSA/mHC/hash-layer fields. Its builder therefore raises an actionable NotImplementedError instead of mutating an exact MCore config with phantom attributes or routing through a provider that cannot construct DSv4 on this pin either.

## Notes

- Deprecated provider classes and explicit compatibility entry points remain available, but repository-owned text/VLM recipes, inference, conversion, checkpoint, and adapter-export primary paths no longer use them.
- No dependency, lockfile, CI workflow, or Megatron-LM submodule changes are included.
- The H100 local-checkpoint functional launcher was moved from active to flaky after three identical iteration-8 NaN failures; its GB200 launcher was already in the flaky tier.

## API migration: old → new

The provider-based and one-call conversion APIs remain available for backward compatibility, but the public compatibility entry points now emit `DeprecationWarning`. New integrations should use the explicit builder-backed lifecycle below.

| Use case | Deprecated compatibility API | Canonical API |
|---|---|---|
| Obtain a build configuration | `bridge.to_megatron_model_config()` | `model_config = bridge.get_model_config()` |
| Build through a provider | `bridge.to_megatron_provider(...)` followed by `provider.provide_distributed_model(...)` | Finalize a `ModelConfig`, then call `bridge.get_megatron_model(model_config, load_weights=..., ...)` |
| One-call HF → Megatron construction | `bridge.to_megatron_model(load_weights=..., ...)` | `get_model_config()` → apply overrides → `finalize()` → `get_megatron_model(...)` |
| Implement model-family config conversion | Override `provider_bridge()` / use provider kwargs | Override `hf_config_to_model_config_kwargs()` or `model_config_bridge()` and provide a serializable `ModelConfig` + standalone `ModelBuilder` |
| Choose whether HF weights are loaded | Pass `load_weights` while creating a provider/config | `get_model_config()` is config-only; pass `load_weights=True` or `False` only to `get_megatron_model()` |

Canonical construction example:

```python
bridge = AutoBridge.from_hf_pretrained(hf_model_id)
model_config = bridge.get_model_config()
# Apply model and parallelism overrides here.
model_config.finalize()
model = bridge.get_megatron_model(
    model_config,
    load_weights=True,
    wrap_with_ddp=False,
)
```

Compatibility is intentionally retained in this PR: old calls continue to work while directing users to the builder-backed API.


src/megatron/bridge/inference/vlm/base.py
```diff
@@ -73,13 +73,13 @@ def setup_model_and_tokenizer(
...
 
     # Initialize model parallel before loading
-    model_provider = bridge.to_megatron_provider(load_weights=False)
-    model_provider.tensor_model_parallel_size = tp
-    model_provider.pipeline_model_parallel_size = pp
-    model_provider.pipeline_dtype = torch.bfloat16
-    model_provider.parallel_output = False
-    model_provider.finalize()
-    model_provider.initialize_model_parallel(seed=0)
+    model_config = bridge.to_megatron_model_config(load_weights=False)
```
yaoyu-33: can we rename api "to_megatron_model_config" to "get_model_config", and when get_config, load_weights=False should already be default

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
```diff
@@ -88,8 +91,11 @@ def __init__(
         pg_collection: ProcessGroupCollection = None,
         mtp_block_spec: Optional[ModuleSpec] = None,
         vp_stage: Optional[int] = None,
+        model_config: "Qwen3VLModelConfig | None" = None,
     ) -> None:
         super().__init__(config=language_transformer_config)
+        build_config = model_config if model_config is not None else language_transformer_config
```
yaoyu-33: I dont understand the fallback here, it seems would cause issues.

skills/adding-model-support/vlm-patterns.md
```diff
@@ -65,29 +50,27 @@ Read these from the correct config level:
 @MegatronModelBridge.register_bridge(
     source="MyModelForConditionalGeneration",   # HF class name (string if not importable)
     target=MyVLModel,                            # Megatron model class
-    provider=MyVLModelProvider,                  # Provider class
     model_type="my_model",                       # HF model_type for export
 )
 class MyVLBridge(MegatronModelBridge):
-    def provider_bridge(self, hf_pretrained: PreTrainedVLM) -> MyVLModelProvider:
+    MODEL_CONFIG_CLASS = MyVLModelConfig
+    CUSTOM_PROVIDER_MODEL_CONFIG_SUPPORTED = True  # compatibility registration, if present
```
yaoyu-33: please delete this, looks not necessary.

src/megatron/bridge/inference/text_generation.py
```diff
@@ -258,11 +268,16 @@ def load_bridge_model(
...
         )
-        provider = bridge.to_megatron_provider(load_weights=True)
-        _apply_provider_parallelism(provider, cache_mla_latents=cache_mla_latents, **parallelism)
-        provider.finalize()
-        provider.initialize_model_parallel(seed=seed)
-        model_list = provider.provide_distributed_model(wrap_with_ddp=False)
+        model_config = bridge.to_megatron_model_config(load_weights=True)
+        _apply_model_config_parallelism(model_config, cache_mla_latents=cache_mla_latents, **parallelism)
+        model_config.finalize()
+        pg_collection = bridge._get_or_initialize_pg_collection(model_config.transformer, seed=seed)
```
yaoyu-33, line 276: is it okay we pass in model_config directly?

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -398,6 +404,12 @@ def mapping_registry(self) -> MegatronMappingRegistry:
     # For MLA models, use DeepSeekModelProvider or similar; for standard GPT, use GPTModelProvider
     PROVIDER_CLASS = None  # Set by @register_bridge(provider=...) or defaults to GPTModelProvider
 
+    # ModelConfig classes used by the non-provider bridge path. A model family that
+    # needs a different config or builder should override ``model_config_bridge``.
+    MODEL_CONFIG_CLASS: ClassVar[type[ModelConfig]] = BridgeGPTModelConfig
+    TRANSFORMER_CONFIG_CLASS: ClassVar[type[TransformerConfig]] = TransformerConfig
```
yaoyu-33, line 445: TRANSFORMER_CONFIG_CLASS is this needed?

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -398,6 +404,12 @@ def mapping_registry(self) -> MegatronMappingRegistry:
...
+    CUSTOM_PROVIDER_MODEL_CONFIG_SUPPORTED: ClassVar[bool] = False
```
yaoyu-33: remove CUSTOM_PROVIDER_MODEL_CONFIG_SUPPORTED

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -607,6 +619,186 @@ def hf_config_to_provider_kwargs(self, hf_config) -> dict:
...
+
+        model_kwargs = {name: value for name, value in config_kwargs.items() if name in model_fields}
+        transformer_kwargs = {
+            name: value
+            for name, value in config_kwargs.items()
+            if name not in model_fields and name in transformer_fields
+        }
+        return model_kwargs, transformer_kwargs
+
+    def model_config_bridge(self, hf_pretrained: HFPreTrained) -> ModelConfig:
```
yaoyu-33: some of these are class method?

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1570,10 +1602,13 @@ def _load_and_export_adapter(model):
             nullcontext() if torch.distributed.is_initialized() else temporary_distributed_context(backend="gloo")
         )
         with model_context:
-            model = provider.provide_distributed_model(
+            model_config.finalize()
```
yaoyu-33, line 1765: model_config's finalize should call when model config is finalized, not when creating a model

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1583,15 +1618,189 @@ def to_megatron_model(
...
+            if torch.cuda.is_available():
+                from megatron.core.tensor_parallel import model_parallel_cuda_manual_seed
+
+                model_parallel_cuda_manual_seed(seed)
+
+        return ProcessGroupCollection.use_mpu_process_groups()
+
+    def to_megatron_model_config(
+        self,
+        load_weights: bool = True,
```
yaoyu-33, line 2002: default to False.

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1583,15 +1618,189 @@ def to_megatron_model(
...
+                expert_tensor_parallel_size=transformer_config.expert_tensor_parallel_size,
+            )
+            if torch.cuda.is_available():
+                from megatron.core.tensor_parallel import model_parallel_cuda_manual_seed
+
+                model_parallel_cuda_manual_seed(seed)
+
+        return ProcessGroupCollection.use_mpu_process_groups()
+
+    def to_megatron_model_config(
```
yaoyu-33: rename to "get_model_config"

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1755,23 +1964,23 @@ def get_conversion_tasks(
 
     @property
     def transformer_config(self) -> TransformerConfig:
```
yaoyu-33, line 2279: is this still needed / used? shouldn't need just tranformer config, but the whole model config

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1755,23 +1964,23 @@ def get_conversion_tasks(
...
+        model_config = self.to_megatron_model_config(load_weights=False)
+        if hasattr(model_config, "finalize"):
+            model_config.finalize()
+        transformer_config = getattr(model_config, "transformer", None)
+        if not isinstance(transformer_config, TransformerConfig):
+            raise TypeError(f"Expected TransformerConfig, got {type(transformer_config).__name__}.")
+        return transformer_config
 
     @property
     def mla_transformer_config(self) -> MLATransformerConfig:
```
yaoyu-33, line 2287: similar as above.

src/megatron/bridge/models/conversion/auto_bridge.py
yaoyu-33: maybe need a model config property instead

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1755,23 +1964,23 @@ def get_conversion_tasks(
...
-        return self._create_config_from_provider(_model_provider, MLATransformerConfig)
+        model_config = self.to_megatron_model_config(load_weights=False)
+        if hasattr(model_config, "finalize"):
+            model_config.finalize()
+        transformer_config = getattr(model_config, "transformer", None)
+        if not isinstance(transformer_config, MLATransformerConfig):
+            raise TypeError(f"Expected MLATransformerConfig, got {type(transformer_config).__name__}.")
+        return transformer_config
 
     @property
```
yaoyu-33, line 2294: please help to review again, which should be property / static method / class method from ux point of view.

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -1057,8 +1249,9 @@ def load_weights_hf_to_megatron(
 
         _hf_import_cache: Dict[str, torch.Tensor] = {}
         for task in self._with_progress_tracking(hf_to_megatron_tasks, description):
-            # None means megatron module not on current rank, skip if this task is not going to happen
-            if task.megatron_module is None:
+            # A missing mapping leaves an empty task slot; a missing module means
+            # the mapped parameter lives on another pipeline rank.
+            if task is None or task.megatron_module is None:
```
yaoyu-33: task should NEVER be none, unless something goes wrong. Please add in comment as well. Tasks are synced across all gpus, so should never be none.

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -1487,10 +1683,17 @@ def _get_provider_from_model(self, model: MegatronModule) -> ModelProviderTarget
 
     def _share_embeddings_and_output_weights(
         self,
-        model_config: TransformerConfig,
+        model_or_config: object,
```
yaoyu-33: why need to change from config only to both?

src/megatron/bridge/models/deepseek/deepseek_v2_bridge.py
```diff
@@ -83,6 +80,38 @@ def provider_bridge(self, hf_pretrained: PreTrainedCausalLM) -> MLAModelProvider
...
+            make_vocab_size_divisible_by=3200,
+            seq_length=4096,
+            moe_layer_freq=[0] * hf_config.first_k_dense_replace
+            + [1] * (hf_config.num_hidden_layers - hf_config.first_k_dense_replace),
+            moe_shared_expert_intermediate_size=hf_config.moe_intermediate_size * hf_config.n_shared_experts,
+        )
+        return config_kwargs
+
     @classmethod
     def megatron_to_hf_config(cls, provider) -> dict:
```
yaoyu-33: this still takes in provider, might need to check the whole thing.

src/megatron/bridge/models/conversion/auto_bridge.py
yaoyu-33: rename to "get_megatron_model"; but you can keep an alias for the old name

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1583,15 +1618,189 @@ def to_megatron_model(
...
 
-        # Finalize the provider before creating models
-        if hasattr(provider, "finalize"):
-            provider.finalize()
+        if hf_identifier:
+            metadata = dict(model_config.extra_checkpoint_metadata or {})
+            metadata["hf_model_id"] = hf_identifier
+            model_config.extra_checkpoint_metadata = metadata
 
-        return provider.provide_distributed_model(**kwargs)
```
yaoyu-33, line 1887: is provide_distributed_model no longer relevant? can we keep it backward compatible but use config + builder instead inside `provide_distributed_model` and all old model building code.

src/megatron/bridge/inference/vlm/base.py
yaoyu-33: Done. The canonical API is now `get_model_config()`, and its `load_weights` default is `False`. The old name remains as a compatibility alias with its legacy default.

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
yaoyu-33: Removed the fallback. `Qwen3VLModel` now requires an explicit outer `model_config`; new builders pass the `Qwen3VLModelConfig`, while deprecated providers pass their compatibility config explicitly.

skills/adding-model-support/vlm-patterns.md
yaoyu-33: Deleted this compatibility-registration line and removed the flag from every bridge and from the base implementation.

src/megatron/bridge/inference/text_generation.py
yaoyu-33: Yes—updated this path to pass the finalized `model_config` directly to `get_megatron_model(model_config, ...)`, rather than constructing the builder here.

src/megatron/bridge/models/conversion/model_bridge.py
yaoyu-33: It is needed only to select the exact nested MCore config type for MLA/heterogeneous families; that cannot be inferred from inherited dataclass annotations. I documented that reason and removed every redundant `= TransformerConfig` override, leaving only MLA/heterogeneous specializations.

src/megatron/bridge/models/conversion/model_bridge.py
yaoyu-33: Removed `CUSTOM_PROVIDER_MODEL_CONFIG_SUPPORTED` completely. A custom legacy provider must now implement an explicit config mapping or `model_config_bridge()`.

src/megatron/bridge/models/conversion/model_bridge.py
yaoyu-33: Audited the method UX. Stateless helpers are now static methods; HF config export remains a classmethod for subclass mapping metadata; `model_config_bridge()` remains an instance method because family overrides use instance mapping hooks/runtime bridge state.

src/megatron/bridge/models/conversion/auto_bridge.py
yaoyu-33: Changed the canonical lifecycle: `get_megatron_model(model_config, ...)` never calls `finalize()`. Callers modify and explicitly finalize the config first. Only the old one-call alias finalizes internally to preserve backward compatibility.

src/megatron/bridge/models/conversion/auto_bridge.py
yaoyu-33: Done: `get_model_config(load_weights=False)` now defaults to `False` and call sites use the no-argument form where appropriate.

src/megatron/bridge/models/conversion/auto_bridge.py
yaoyu-33: Renamed to `get_model_config`; `to_megatron_model_config` remains a backward-compatible alias.

src/megatron/bridge/models/conversion/auto_bridge.py
yaoyu-33: Removed the `transformer_config` convenience property. Callers now retrieve the whole `ModelConfig` with `get_model_config()` and access `.transformer` explicitly when needed.

src/megatron/bridge/models/conversion/auto_bridge.py
yaoyu-33: Removed `mla_transformer_config` for the same reason; the exact MLA config is available as `bridge.get_model_config().transformer`.

src/megatron/bridge/models/conversion/auto_bridge.py
yaoyu-33: Completed a method-kind audit: stateful public conversion/build operations remain instance methods, subclass-metadata export remains classmethod, and pure dtype/vocab/name/config lookup helpers are static methods. Added typed builder-only kwargs for the canonical model-build API.

src/megatron/bridge/models/conversion/model_bridge.py
yaoyu-33: Enforced this invariant. Standard and FP8 task builders now return only real synchronized tasks and raise a clear `RuntimeError` if any slot remains `None`; remote PP parameters use real placeholder tasks with `megatron_module=None`.

src/megatron/bridge/models/conversion/model_bridge.py
yaoyu-33: Changed the export path to recover the whole outer builder `ModelConfig` (with a legacy-provider fallback), then read `share_embeddings_and_output_weights` from that single config. It no longer probes both model and nested config.

src/megatron/bridge/models/deepseek/deepseek_v2_bridge.py
yaoyu-33: Updated DeepSeek V2/V3/V4 HF config export to accept and read the builder-backed `model_config`, while retaining legacy provider compatibility. Added builder-config export regression tests for all three.

src/megatron/bridge/models/conversion/auto_bridge.py
yaoyu-33: Added canonical `get_megatron_model(model_config, ...)`; the existing `to_megatron_model(...)` remains as the compatibility alias and delegates through ModelConfig + ModelBuilder.

src/megatron/bridge/models/conversion/auto_bridge.py
yaoyu-33: Kept the old one-call model-building behavior through `to_megatron_model()`: it translates legacy dtype/device/hook options into the ModelConfig, finalizes it, and delegates to the builder-backed canonical API. Training setup likewise supports old providers while routing ModelConfig instances through ModelBuilder.

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4545/commits/1d02b3cdc5c33392fbd917777bace50eb407b277

src/megatron/bridge/models/bailing/model_config.py
maanug-nv, line 1: Don't understand the need for this entire file. 
Firstly, I thought we wanted to avoid introducing model-specific configs/providers just for the sake of overriding defaults? Seems this one is just used for Bailing model bridge.
Secondly, I don't understand the need for this Bailing specific layer spec (on line 35). I think the default layer spec here will do the same thing (https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/models/gpt.py#L81-L88)? As long as `transformer_impl` is set to use TE, also assuming `num_moe_experts` is set.

src/megatron/bridge/models/common/base.py
maanug-nv, line 1: didn't realize we needed our own version of the ModelConfig base class. Should we upstream any of this? is there any plan?

src/megatron/bridge/models/config_proxy.py
maanug-nv, line 1: It seems this mixin is meant to support model configs that have duplicated fields that are already on transformer config? Is that correct? i.e. , `MaanuModelConfig.hidden_size = 1024` will update it on `MaanuModelConfig` and the nested `TransformerConfig`? Why is that necessary? Can't we just remove the duplicate fields from the child model configs? The getattr and setattr proxies on GPTModelConfig will already handle retrieving `hidden_size` from `MaanuModelConfig.transformer.hidden_size` in the case of `MaanuModelConfig.hidden_size`.

src/megatron/bridge/models/metadata.py
```diff
@@ -0,0 +1,46 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+
+
+def get_hf_model_id_from_model_config(model_config: object) -> str | None:
```
maanug-nv, line 20: do we need a corresponding set hf model id?

src/megatron/bridge/models/gpt/model_config.py
```diff
@@ -0,0 +1,113 @@
...
+
+    restored = dict(data)
+    restored_transformer = dict(transformer)
+    restored_transformer["activation_func"] = str_to_callable(activation_name)
+    restored["transformer"] = restored_transformer
+    return restored
+
+
+@dataclass(kw_only=True)
+class BridgeGPTModelConfig(FlatTransformerConfigMixin, GPTModelConfig):
```
maanug-nv, line 128: having trouble understanding why this is needed, could you explain?

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1570,28 +1660,298 @@ def _load_and_export_adapter(model):
...
+                "AutoBridge.from_hf_config() does not include weights. "
+                "Pass load_weights=False or provide hf_path to load weights."
+            )
+
+        original_perform_initialization: list[tuple[TransformerConfig, bool]] = []
+        original_metadata = model_config.extra_checkpoint_metadata
+        original_pre_wrap_hooks = model_config.pre_wrap_hooks
+        original_pre_wrap_hook_values = list(original_pre_wrap_hooks)
+        succeeded = False
+        try:
```
maanug-nv, line 1906: some comments would be good for this block explaining what overrides need to be done and why... afaict, this is handling load weights and skipping initialization, which means setting perform_init false everywhere and adding the hf load weights hook

maanug-nv (commented):
overall makes sense. my main requirement was that we can convert HF config to GPTModelConfig , and customize that conversion as needed. and then build the MCore model from that config. so makes sense at a high-level, but i have a few concerns/questions for my understanding below.

src/megatron/bridge/models/config_proxy.py
maanug-nv: Let me know if there's a good reason that we need to support duplicated fields like this

maanug-nv:
### Code review

Reviewed the ModelConfig+ModelBuilder migration with a focus on paradigm adherence and design (not minor bugs). The core design is sound: Bridge configs subclass MLM's `GPTModelConfig`/`HybridModelConfig`, reuse MLM's builders, keep family-only state off the exact `TransformerConfig`, and the strict `FlatTransformerConfigMixin` correctly rejects phantom fields. All imported `megatron.training.models.*` symbols resolve at the pinned MCore commit. Four concerns, most severe first — the first two are confirmed defects, the last two are design concerns.

**1. VLM inference is broken for migrated (builder-backed) models.** `setup_model_and_tokenizer` was migrated to `get_model_config()` / `load_megatron_model()`, so the built `Qwen25VLModel`'s `.config` is now a plain `TransformerConfig` (`Qwen25VLModel.__init__` does `super().__init__(config=language_transformer_config)`), not a `*ModelProvider`. The unchanged `isinstance(config, Qwen25VLModelProvider)` dispatch in `setup_inference_wrapper` therefore always falls through to `raise ValueError("Unknown model config: ...")`. This is the `isinstance(*, *ModelProvider)` update that has to accompany a provider→config migration — dispatch on the model type or on `model.model_config` (the new `ModelConfig`) instead.

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/9a2afd89b4306a678af1e8f2eb2499e2ad2c8ef6/src/megatron/bridge/inference/vlm/base.py#L73-L85
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/9a2afd89b4306a678af1e8f2eb2499e2ad2c8ef6/src/megatron/bridge/inference/vlm/base.py#L147-L156

**2. Step3.7's custom layer spec is silently dropped → wrong architecture.** `transformer_layer_spec = build_step35_layer_spec` is a bare class attribute with **no type annotation**, so `@dataclass` never registers it as the field default (contrast the correct `field(default_factory=...)` used by `Gemma2ModelConfig`, `SarvamMoEModelConfig`, etc.). Instances inherit the parent `GPTModelConfig.transformer_layer_spec` default of `None`, and the new bridge path (`step37_bridge.hf_config_to_model_config_kwargs`) never sets it — only the deprecated `provider_bridge` does. So `Step37ModelBuilder.build_model → super().build_model()` sees `None` and falls back to `default_layer_spec`, building a generic GPT stack instead of Step3.7's spec (head-wise attn gate, sliding attention, per-layer rotary, swiglu limits). `test_step37_model_config.py` doesn't assert the spec, so CI won't catch it. Same bare-assignment pattern (currently benign since MoE routes to the same spec) at `sarvam/model_config.py:42`.

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/9a2afd89b4306a678af1e8f2eb2499e2ad2c8ef6/src/megatron/bridge/models/stepfun/step37_model_config.py#L29-L42
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/9a2afd89b4306a678af1e8f2eb2499e2ad2c8ef6/src/megatron/bridge/models/sarvam/model_config.py#L40-L43

**3. GPT (non-MLA) YaRN isn't representable on the stock builder path, and the compat fallback is dead.** `MistralModelConfig` declares outer `yarn_*` fields but points `builder` at the stock `GPTModelBuilder`, which never reads them — only `build_gpt_with_yarn` (used by Ministral3's dedicated builder) does. A Mistral YaRN checkpoint therefore builds with YaRN silently disabled. Separately, the base `hf_config_to_model_config_kwargs` raises a bare `NotImplementedError` for non-MLA YaRN, but the deprecated `to_megatron_model` compat path only catches its subclass `ModelConfigNotSupportedError` — so an `except` on the child never catches the parent and it hard-crashes instead of falling back to the still-present provider. Either raise `ModelConfigNotSupportedError` there, or route GPT-YaRN through a dedicated builder like Ministral3 does.

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/9a2afd89b4306a678af1e8f2eb2499e2ad2c8ef6/src/megatron/bridge/models/mistral/model_config.py#L24-L35
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/9a2afd89b4306a678af1e8f2eb2499e2ad2c8ef6/src/megatron/bridge/models/conversion/model_bridge.py#L682-L688
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/9a2afd89b4306a678af1e8f2eb2499e2ad2c8ef6/src/megatron/bridge/models/conversion/auto_bridge.py#L1806-L1812

**4. `finalize()` is skipped on the explicit-config build path (lifecycle footgun).** `get_megatron_model` only finalizes when it derives the config itself (`model_config is None`); when the caller passes an explicit config — the documented `get_model_config()` → override → `get_megatron_model()` flow — it doesn't, and `get_model_config()` doesn't finalize either. Forgetting the manual `finalize()` builds from an unvalidated config with no error (silent divergence). Consider finalizing idempotently inside `get_megatron_model` (or asserting a finalized flag) so the canonical two-step flow can't skip validation.

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/9a2afd89b4306a678af1e8f2eb2499e2ad2c8ef6/src/megatron/bridge/models/conversion/auto_bridge.py#L1704-L1713

Minor: `AutoBridge.transformer_config` was removed rather than deprecated like the other methods (breaking; AGENTS.md flags `models/conversion/` signature changes as ASK-FIRST — no live references remain, but it warrants the `breaking-change` label). And `docs/.../bridge-tech-details.mdx` still teaches the deprecated `provider_bridge` as the canonical way to add a bridge, contradicting the migrated `adding-new-models.mdx`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/megatron/bridge/models/common/base.py
yaoyu-33: I dont think so, we can keep this on our end. Only used in model specific case.

yaoyu-33:
Conflict resolution and the generic text-model `ModelConfig` cleanup are now pushed in `0fe9cb1dd9b1db0fb9ca59f943c8f6553200e4bb`.

Validation:
- affected model/config suite: 465 passed, 1 CUDA-only test deselected
- provider-free consumer/recipe/instantiate suite: 122 passed
- focused DeepSeek/Qwen/Stepfun/config regression suite: 158 passed
- `uv run pre-commit run --all-files`: passed
- three independent review rounds completed; the final two reviews found no actionable issues
- no `3rdparty/Megatron-LM`, `uv.lock`, or CI workflow changes

The resolved implementation preserves the Config + ModelBuilder architecture, removes redundant thin text-family config wrappers, and includes the identified MTP/DeepSeek-V4/layer-spec regression fixes.

PR #4566, opened 2026-06-29
support modelopt W4A16 & W4A4 NVFP4 export for grouped MoE weights

HollowMan6:
# What does this PR do ?

support modelopt W4A16 & W4A4 NVFP4 export for grouped MoE weights

# Changelog

- Extend AutoBridge.export_hf_weights_modelopt to support quant_mode="w4a16_nvfp4". https://github.com/NVIDIA/Model-Optimizer/commit/a451a2baf85b5e5274f64772a3e9689ac28aaebe
- Collect ModelOpt quant metadata from the actual quantized weight/module, including NVFP4 weight_scale_2.
- Map Megatron ModelOpt metadata onto exported HF names before quantization.
- Support grouped MoE expert exports by syncing EP metadata, stacking per-expert metadata, and emitting vLLM ModelOpt fused-MoE tensor names:
    - *.experts.w13_weight
    - *.experts.w13_weight_scale
    - *.experts.w13_weight_scale_2
    - *.experts.w2_weight
    - *.experts.w2_weight_scale
    - *.experts.w2_weight_scale_2
- Keep ignored weights unquantized and skip quantizer tensors during export.

# GitHub Actions CI

**Pre checks**:

- [X] Did you write any new necessary tests?

# Additional Information

- Related to # (issue)


src/megatron/bridge/models/conversion/modelopt_utils.py
```diff
@@ -66,56 +77,288 @@ def matches_quant_ignore_pattern(name: str, patterns: list[str]) -> bool:
...
+    if leading_dim % 2 != 0:
+        return meta
+
+    midpoint = leading_dim // 2
+    split = slice(0, midpoint) if hf_key == "gate" else slice(midpoint, leading_dim)
+    return _with_quant_meta_tensors(
+        meta,
+        weight_amax=_slice_optional_quant_tensor(meta.weight_amax, split, leading_dim),
+        weight_scale_2=meta.weight_scale_2,
+    )
```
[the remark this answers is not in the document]
HollowMan6: weight_scale_2 is ModelOpt’s global per-tensor NVFP4 scale and is intentionally shared by gate and up. ModelOpt’s exporter also writes the same scalar to both projections. For grouped export, we stack this value across experts, so its leading dimension represents experts rather than gate/up rows. Only the row-aligned weight_amax should be sliced here.

src/megatron/bridge/models/conversion/modelopt_utils.py
```diff
@@ -66,56 +77,288 @@ def matches_quant_ignore_pattern(name: str, patterns: list[str]) -> bool:
...
+            for expert_number, expert_meta in _iter_grouped_quant_meta(task, metadata):
+                for hf_key, hf_name in hf_items:
+                    grouped_metadata.setdefault(hf_name, {})[expert_number] = _slice_gated_quant_meta(
+                        expert_meta,
+                        hf_key,
+                    )
+            continue
+
+        for _, hf_name in hf_items:
+            hf_metadata[hf_name] = meta
```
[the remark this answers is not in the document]
HollowMan6: Good catch. The non-grouped branch was dropping hf_key, so a GatedMLPMapping assigned full fused metadata to both split tensors. I’ll preserve the key and apply _slice_gated_quant_meta, matching the grouped path

zhangyimi:
LGTM

tests/unit_tests/models/test_modelopt_utils.py
```diff
@@ -350,8 +616,50 @@ def fake_export_weight(name, tensor, meta):
...
+    )
+    monkeypatch.setattr(
+        modelopt_utils,
+        "get_modelopt_quant_exporter",
+        lambda quant_mode: (QUANTIZATION_NVFP4, fail_export_weight),
+    )
+    monkeypatch.setattr(
+        "megatron.bridge.models.conversion.auto_bridge.model_bridge._get_pp_group",
+        lambda _model: None,
+    )
```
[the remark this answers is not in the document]
HollowMan6: removed

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -700,16 +714,24 @@ def export_hf_weights_modelopt(
...
-                if megatron_name is not None:
-                    meta = metadata.get(megatron_name)
+            if is_modelopt_quantizable_weight_name(
+                hf_name,
+            ) and not matches_quant_ignore_pattern(
+                hf_name,
+                ignore_patterns,
+            ):
+                meta = hf_metadata.get(hf_name)
+                if meta is None and metadata:
```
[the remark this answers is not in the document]
HollowMan6: added

yaoyu-33 (dismissed):

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -671,56 +823,238 @@ def export_hf_weights_modelopt(
...
+            ep_group = model_bridge._get_ep_group(model)
+        else:
+            pp_group = None
+            ep_group = None
+
+        # Every PP stage needs the metadata for its EP-local experts because
+        # the normal conversion broadcasts those source weights across PP.
+        # Keep this view EP-local: pre-EP packing must use E/EP metadata
+        # entries, not a full global-E metadata copy.
+        pp_world_size = dist.get_world_size(group=pp_group) if dist.is_initialized() and pp_group is not None else 1
```
yaoyu-33: utils, get_pg_size.

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -1634,6 +1659,7 @@ def build_conversion_tasks(
                 if not mapping:
                     logger.warning(f"WARNING: No mapping found for megatron_param: {global_name}")
                     continue
+                mapping.set_process_groups_from_pg_collection(pg_collection)
```
yaoyu-33: Could we centralize process-group installation in `MegatronMappingRegistry` rather than repeating it after each lookup? This PR now performs the same installation at multiple task-building and placeholder/FP8 paths, making it easy for a future lookup to omit it. The registry could retain the `pg_collection` and apply it to every exact or resolved mapping returned by both `megatron_to_hf_lookup()` and `hf_to_megatron_lookup()`. Explicit clones created through `resolve()` should likewise preserve the configured collection. Process-group installation is an invariant of a usable mapping, so callers should not need to remember this step.

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -671,56 +823,238 @@ def export_hf_weights_modelopt(
...
 
-        hf_to_megatron_name = build_hf_to_megatron_name_map(conversion_tasks)
-        metadata = collect_modelopt_quant_metadata(conversion_tasks)
+        if ignore_patterns is None:
+            ignore_patterns = []
 
-        pp_group = model_bridge._get_pp_group(model)
-        if pp_group is not None and dist.is_initialized() and dist.get_world_size(group=pp_group) > 1:
-            sync_modelopt_quant_metadata(metadata, pp_group)
+        local_metadata = collect_modelopt_quant_metadata(conversion_tasks)
```
yaoyu-33: This method now owns the full ModelOpt export implementation: metadata synchronization, expert-family eligibility, mapping replacement, metadata partitioning, quantization, and byte-level EP gathering. Could we extract that into a dedicated `modelopt_export.py`, for example as `build_modelopt_export_plan(...)`, and leave `AutoBridge` responsible only for building the initial tasks and invoking `export_hf_weights()` with the prepared tasks?

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -80,6 +79,158 @@
 _MISSING = object()
 
 
+@dataclasses.dataclass(frozen=True)
```
yaoyu-33: Could we avoid introducing a ModelOpt-specific task subclass here? It only adds export callbacks while manually copying every `WeightConversionTask` field, so future fields could be silently dropped. A format-agnostic optional `export_hook` on `WeightConversionTask`, installed with `dataclasses.replace()`, would be simpler. For pre-EP tasks, quantization and EP gathering can be composed into that single hook because both happen before final CPU placement. This would also remove the private `getattr(task, "_export_hf_weight", ...)` protocol from the stream path.

src/megatron/bridge/models/conversion/modelopt_utils.py
```diff
@@ -25,11 +26,27 @@
...
+    re.compile(r"(local_experts\.)(\d+)(\.)"),
+    re.compile(r"((?:weight|bias))(\d+)(?=$|\.)"),
+    re.compile(r"(experts\.)(\d+)(\.)"),
+)
+_FUSED_MOE_NVFP4_NAME_MAP = {
+    ".experts.gate_up_proj": ".experts.w13_weight",
+    # Keep Nemotron-H's single (non-gated) projection distinguishable from a
+    # gate+up w13 tensor. The receiver expands this fused expert batch without
+    # splitting its projection dimension.
+    ".experts.up_proj": ".experts.up_proj",
```
yaoyu-33: This non-gated fused transport name does not match the current vLLM ModelOpt loader contract. Nemotron-H still stores its first expert projection in `w13_weight`; when `is_act_and_mul=False`, that tensor has one shard. The fallback checkpoint mapping otherwise expects per-expert `experts.<id>.up_proj.*` names, not index-free `experts.up_proj*` tensors. Could we emit the one-shard `w13_weight`, `w13_weight_scale`, `w13_weight_scale_2`, and `w13_input_scale` family here, or keep this family on the regular per-expert path? An end-to-end vLLM load test would catch this contract mismatch.

yaoyu-33 (dismissed):

yaoyu-33 (approved):

PR #4592, opened 2026-06-30
Per-layer expert count + MoE aux-loss tracker sizing (the NCCL-deadlo…

chochowski:
Small fixes that allow support for heterogeneous MoE layers

### DIFF
--- a/src/megatron/bridge/models/conversion/model_bridge.py
+++ b/src/megatron/bridge/models/conversion/model_bridge.py
@@ -310,8 +310,7 @@ def _megatron_local_name_to_global(
     is_expert_param = (is_grouped_expert_param or is_local_expert_param) and ".adapter." not in param_name
     ep_group = _get_ep_group(models) if is_expert_param else None
     if is_expert_param and ep_group is not None and get_pg_size(ep_group) > 1:
-        num_experts = config.num_moe_experts
-        num_experts_per_rank = num_experts // ep_group.size()
+        num_experts_per_rank = layer_module.mlp.num_local_experts   # per-layer, heterogeneous-safe
 
         def _update_grouped_expert_number(param_name: str, param_type: str) -> str:
             """Update expert number from local to global for weight or bias parameters."""
--- a/src/megatron/bridge/models/conversion/model_bridge.py
+++ b/src/megatron/bridge/models/conversion/model_bridge.py
@@ -286,6 +286,11 @@ def _megatron_local_name_to_global(
     vp_stage: Optional[int] = None,
 ) -> str:
     """Adjust layer number and expert number from local to global numbering."""
+    # ``layer_module`` is resolved lazily and shared between the PP and EP
+    # branches. It stays ``None`` on the EP-only path (PP=1, EP>1) until the EP
+    # branch resolves it, so it must never be read before being assigned.
+    layer_module = None
+
     # PP
     pp_group = _get_pp_group(models)
     if "layers." in param_name and get_pg_size(pp_group) > 1:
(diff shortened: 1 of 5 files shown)


src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -290,8 +290,7 @@ def _megatron_local_name_to_global(
     is_expert_param = (is_grouped_expert_param or is_local_expert_param) and ".adapter." not in param_name
     ep_group = _get_ep_group(models) if is_expert_param else None
     if is_expert_param and ep_group is not None and get_pg_size(ep_group) > 1:
-        num_experts = config.num_moe_experts
-        num_experts_per_rank = num_experts // ep_group.size()
+        num_experts_per_rank = layer_module.mlp.num_local_experts   # per-layer, heterogeneous-safe
```
yaoyu-33: This regresses the EP-only conversion path. `layer_module` is only assigned in the PP renumbering block above (`if "layers." in param_name and get_pg_size(pp_group) > 1`). When `expert_model_parallel_size > 1` and PP is 1, we still enter this EP block for expert params, but `layer_module` is undefined, so import/export now raises `UnboundLocalError` before it can renumber experts. Existing tests exercise this PP=1 + EP>1 path in `tests/unit_tests/models/test_quant_mapping.py` via `_call_local_to_global`. Please resolve the layer module independently of the PP branch, or keep a config-based fallback, before reading `layer_module.mlp.num_local_experts`.

src/megatron/bridge/training/utils/train_utils.py
```diff
@@ -1041,10 +1041,11 @@ def training_log(
...
             track_names.append("z_loss")
 
-        if getattr(config.model, "is_hybrid_model", False):
-            layers = getattr(config.model, "hybrid_layer_pattern", "").count("E")
-        else:
-            layers = getattr(config.model, "num_layers", None)
+        num_layers = getattr(config.model, "num_layers", None)
+        moe_layer_freq = getattr(config.model, "moe_layer_freq", None)
+        pattern = getattr(config.model, "hybrid_layer_pattern", "") or ""
+        if moe_layer_freq is None and pattern:                 # per-layer averaging denominator
```
yaoyu-33: This still ignores `hybrid_layer_pattern` for common hybrid-MoE configs because `moe_layer_freq` defaults to `1` on the inherited TransformerConfig/MambaModelProvider path, so the `is None` guard is false. Recipes such as `nemotron_3_nano` set a pattern with `E` layers and do not override `moe_layer_freq`, so this now passes `(num_layers=52, moe_layer_freq=1)` into `_count_moe_layers` and averages aux/z-loss over every layer instead of only the MoE layers. The storage size should stay at full `num_layers` for the deadlock fix, but the averaging denominator still needs to be derived from the hybrid pattern when the frequency is just the default.

yaoyu-33 (changes requested):
I found two regression risks that should be fixed before merge. I could not run the targeted pytest locally because `uv` cannot install `nvidia-resiliency-ext==0.6.0` on macOS arm64, but the EP-only failure below is also covered by existing unit tests that exercise PP=1 + EP>1 renumbering.

src/megatron/bridge/models/conversion/model_bridge.py
chochowski: added solution

src/megatron/bridge/training/utils/train_utils.py
chochowski: added solution

kevalmorabia97:
/ok to test [cacd727](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4592/commits/cacd7278aabfd6de269a9b5ad6dff43aedee462c)

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -290,8 +295,25 @@ def _megatron_local_name_to_global(
...
+            if models is not None and layer_match is not None:
+                try:
+                    _, layer_module = get_module_and_param_from_name(
+                        models=models, param_name=layer_match.group(1), vp_stage=vp_stage
+                    )
+                except (ValueError, AttributeError):
+                    layer_module = None
+
+        if isinstance(layer_module, MegatronModule):
+            num_experts_per_rank = layer_module.mlp.num_local_experts  # per-layer, heterogeneous-safe
```
yaoyu-33: Thanks for addressing the earlier PP=1 + EP>1 regression. I am still concerned about relying on `layer_module` for the expert count here.

Conversion/name-mapping infrastructure should not need to traverse an instantiated model and assume a concrete structure such as `layer_module.mlp.num_local_experts`. This couples conversion to module wrappers and runtime layout (PP/VPP, MTP, custom MoE implementations, and whether a live model is available). The broad exception plus scalar fallback can also silently produce incorrect expert numbering for the heterogeneous case when traversal fails.

Can we make the expert layout an explicit config contract instead? For example, expose a validated sequence or mapping of global expert counts by global layer. Conversion can select the count for the resolved layer and compute `num_experts_per_rank = expert_counts[layer_idx] // ep_size`, while homogeneous models continue using `config.num_moe_experts`.

The existing scalar field alone cannot represent heterogeneous counts, but I would prefer extending the config metadata rather than adding more dependency on `LayerModule`. Is there a reason the heterogeneous expert layout cannot be represented in the model config?

chochowski:
/ok to test [4bbddb1](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4592/changes/4bbddb10cf32d584ea988c5948dc36893d46b891)

kevalmorabia97:
/ok to test [4bbddb1](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4592/changes/4bbddb10cf32d584ea988c5948dc36893d46b891)

kevalmorabia97:
/ok to test [a34d52d](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4592/commits/a34d52d10803b28a0eb71a414af00ca8f9411711)

Issue #4490, opened 2026-06-25
PoR: [model] tracking: GLM-5.2 support


HollowMan6:
### Hugging Face repository

https://huggingface.co/zai-org/GLM-5.2

### Architecture family

MoE decoder-only (e.g. DeepSeek V2/V3, OLMoE, Qwen3-MoE, MiniMax-M2)

### Required deliverables

- [x] Model providers
- [x] HF conversion bridge
- [x] Unit tests (config and bridge)
- [x] Model conversion functional tests
- [x] Optimal pretraining recipe
- [x] Optimal finetuning recipe
- [x] Recipe unit tests
- [ ] Recipe functional tests
- [ ] End-to-end CI coverage

### Owner if known

@HollowMan6

### Status: core dependencies and bridge enablement merged - TileLang kernel pending

**GLM-5.2 support is implemented across the dependency stack, but Bridge support is still pending merge.** The core DSA CP/THD dependency chain in Megatron-Core is now merged, and the Bridge-side MTP / LoRA helper PRs are merged. The remaining open items are the fused DSA kernel PRs and the Bridge GLM-5.2 enablement / recipe PR. This issue tracks landing those remaining PRs and updating verification evidence.

GLM-5.2 is a small delta over the GLM-5 / GLM-5.1 support that already lives in this repo. The only genuinely new model-side piece is **indexer share** (HF "IndexShare"): the DSA lightning indexer is computed once and **shared/reused across a block of layers** instead of being recomputed every layer, cutting indexer FLOPs at long context. Everything else that differs from DeepSeek V3.2 (e.g. the indexer RoPE handling) has been made **configurable in Megatron-Core** and is set from the Megatron-Bridge provider — so enabling GLM-5.2 is config + weight mapping only, no new module logic.

### Implemented in Megatron-Bridge

| Area | Location |
|------|----------|
| Conversion bridge (MLA + DSA indexer + MoE) | `src/megatron/bridge/models/glm_moe_dsa/glm5_bridge.py` — `GLM5Bridge`, `model_type="glm_moe_dsa"`, `GlmMoeDsaForCausalLM` → `GPTModel`, provider `MLAModelProvider` |
| Provider | `glm5_provider.py` (alias of `MLAModelProvider`) |
[...]

PR #4056, opened 2026-05-29
[data] fix: pad loss_mask and handle tensors in packed-sequence chat …

JRD971000:
…path

tokenize_dataset() crashed with chat=True and pad_seq_to_mult > 1 (issue #2610):

1. pre_pad_dataset concatenated `tensor + [pad_id]*n`, but the chat path (_chat_preprocess) returns torch tensors, raising TypeError.
2. loss_mask was never padded, so samples that round to the same padded input_ids length but had different original lengths kept mismatched loss_mask lengths, causing np.array() in fill_packing_strategy to fail with an inhomogeneous-shape error.

Extract the padding logic to a testable module-level helper `_pre_pad_data_point`, normalize tensor/array values to lists before concatenating, and pad loss_mask (with 0) to the same length as input_ids. Add unit tests covering the chat-tensor, equal-length, non-chat, and truncation paths.

Fixes #2610

# What does this PR do ?

Add a one line overview of what this PR aims to accomplish.

# Changelog

- Add specific line by line info of high level changes in this PR.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci) in the Contributing doc for how to trigger the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

# Before your PR is "Ready for review"

**Pre checks**:

- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?

If you haven't finished some of the above items you can still open "Draft" PR.

# Additional Information

- Related to # (issue)

PR #4598, opened 2026-06-30
Fix packed-sequence SFT prep edge cases for long-context THD+CP (#4593 §2.3-2.5)

### LINKED ISSUE #4593: PoR: GLM-5.2 packed SFT Bridge data/config blockers
## Summary

This is a Bridge-side companion issue for NVIDIA-NeMo/Megatron-Bridge#4490 tracking the remaining Megatron-Bridge data/config blockers we hit while bringing up full SFT for GLM-5.2 (`GlmMoeDsaForCausalLM`: MLA + DeepSeek Sparse Attention indexer + MoE) on 64x GB200 with TP2/PP4/CP8/EP64, `seq_len=150000`, packed THD, container `nvcr.io/nvidia/nemo:26.06.00`.

This issue intentionally excludes the fused DSA kernel work and container/JIT deployment robustness. It covers the Bridge-side priorities that should be upstreamed or documented:

1. Keep packed metadata available on middle pipeline stages.
2. Fix packed-sequence preparation edge cases for long-context CP/THD SFT.
3. Document or automate required packed/CP/DSA recipe settings.

## 1. Middle PP stages drop `packed_seq_params` for packed THD batches

With PP > 2, `gpt_step._get_batch` short-circuits middle pipeline stages (`not is_first and not is_last`) by returning an all-`None` batch. That means `cu_seqlens` is not loaded and `packed_seq_params` is never constructed on middle stages.

For GLM-5.2 DSA, the index-share holder expects `packed_seq_params` to exist on every stage. The run crashes in Megatron-LM DSA when the middle stage reaches the index-share path.

Source pointers:

* Bridge: `src/megatron/bridge/training/gpt_step.py` `_get_batch`
* Core: `megatron/core/transformer/experimental_attention_variant/dsa.py` `_get_index_share_topk_holder`

Suggested fix:

* Load the batch on every pipeline stage when packed metadata is present.
* Only use the middle-stage short-circuit when the batch truly carries no packed metadata, e.g. `is_middle and batch.get("cu_seqlens") is None`.
* `get_batch_from_iterator` already loads the `cu_seqlens` family on all stages, and `_partition_packed_batch_for_cp` is already `None`-safe.

### 2.1 `pre_pad_dataset` breaks on tensor `input_ids`

The `pad_seq_to_mult > 1` pre-pad path does list concatenation directly against `input_ids`:

```python
val = val + [pad_value] * (...)
```

When `input_ids` is a tensor, this raises `TypeError`.

Source pointer:

* Bridge: `src/megatron/bridge/data/datasets/packed_sequence.py` `pre_pad_dataset`

Suggested fix:

* Normalize per-token values to lists before padding/truncation.

### 2.2 Only `input_ids` / `context_ids` are padded, leaving per-token fields ragged

The current pre-pad path pads only `input_ids` and `context_ids`. It does not pad `loss_mask` and `labels` to the same length. This can produce ragged arrays inside a length bucket and later fail during packing with inhomogeneous-shape / dtype errors.

Suggested fix:

* Pad all per-token fields consistently:
  * `input_ids`
  * `context_ids`
  * `loss_mask` as bool
  * `labels`

### 2.3 Near-pack-size documents can overflow the pack after pre-padding

For `seq_len=150000`, a near-pack-size document such as length 149999 can be padded to 150001, then truncated back to 150000. The runtime segment length becomes `stored_doc_len - 1 = 149999`, which is not divisible by `2 * CP` and later crashes THD+CP RoPE partitioning.

Suggested fix:

* Cap the pad target at `((max_seq_length - 1) // pad_seq_to_mult) * pad_seq_to_mult`.
* Store documents with length `target + 1`, so the runtime segment length `stored_len - 1` remains divisible by `2 * CP` and never exceeds the pack size.

For CP8, this keeps runtime segment lengths divisible by 16.

### 2.4 `create_hist` can IndexError on sequences longer than the pack size

`create_hist` indexes `counts[seq_len]` without a bounds check. A sequence longer than `truncate_seq_len` goes out of range, especially when the tokenizer/truncation path is not exercised before histogram creation.

Source pointer:

* Bridge: `src/megatron/bridge/data/datasets/packing_utils.py` `create_hist`

Suggested fix:

* Add an explicit oversize guard: count/log skipped sequences and continue instead of indexing past the histogram.

### 2.5 High-worker packing can fail with multiprocessing FD / `ancdata` errors

Packed-sequence tokenization uses `multiprocessing.Pool`. At high `num_tokenizer_workers`, tensor sharing via the default `file_descriptor` strategy can exhaust file descriptors and fail with:

```text
RuntimeError: received 0 items of ancdata
```

Source pointers:

* Bridge: `src/megatron/bridge/data/datasets/packed_sequence.py` tokenization Pool / `imap`
* Script: `scripts/training/pack_sft_data.py`

Suggested fix:

* Set `torch.multiprocessing.set_sharing_strategy("file_system")` before constructing the Pool, or raise the soft `RLIMIT_NOFILE`.
* Ensure `pack_sft_data.py` can run with high worker counts without falling back to effectively single-process throughput.

## 3. Required packed/CP/DSA recipe settings should be documented or automated

For GLM-5.2 DSA SFT with CP and packed THD, these settings are required in practice:

* `cp_comm_type="all_gather"` because DSAttention CP currently only supports all-gather.
* packed sequences are mandatory for DSA cross-layer index sharing because the holder is attached to `packed_seq_params`.
* pipeline layout must start each stage on a DSA compute layer; for 78 layers / PP4 this requires `[18, 20, 20, 20]` rather than an even `[20, 20, 20, 18]`-style split.
* CP > 1 finetuning requires `calculate_per_token_loss=True`.
* `ddp.average_in_collective=False` should be paired with CP finetuning to avoid the wrong loss normalization.
* packed SFT `pad_seq_to_mult` should be set to `2 * context_parallel_size` for CP, or `lcm(2 * CP, CP * TP)` when sequence parallelism also requires it.

Suggested improvements:

* Document these requirements in the packed-sequence / GLM-5.2 recipe docs.
* Auto-set `calculate_per_token_loss=True` and `ddp.average_in_collective=False` for finetune configs with CP > 1 when safe, or raise a targeted validation error with the exact required settings.
* Auto-derive a valid DSA pipeline layout from `(num_layers, pp_size, topk_freq, skip_offset)` when GLM-5.2 DSA index sharing is enabled, or emit a clear error explaining the valid stage starts.

## 4. Tokenizer chat template compatibility gap

Megatron's `HuggingFaceTokenizer.apply_chat_template` currently differs from the HF tokenizer signature and the Bridge SFT helper can encounter nested tokenizer wrappers. In practice, callers expecting the HF signature can hit:

```text
TypeError: ... missing 1 required positional argument: 'chat_template'
```

Source pointers:

* Core: `tokenizers/text/libraries/huggingface_tokenizer.py`
* Bridge: `src/megatron/bridge/data/datasets/utils.py` `_assistant_mask_via_generation_prompt`

Suggested Bridge-side fix:

* Bounded recursive unwrap through `.tokenizer` / `._tokenizer` to reach the underlying `transformers` tokenizer before calling `apply_chat_template`.
* Keep the unwrap bounded, e.g. <= 8 hops, to avoid cycles.

## Validation target

A fix should make the following class of setup work without local patches:

* GLM-5.2 full SFT
* 64x GB200
* TP2 / PP4 / CP8 / EP64
* `seq_len=150000`
* offline packed THD SFT
* PP layout `[18, 20, 20, 20]`
* `cp_comm_type="all_gather"`
* `calculate_per_token_loss=True`
* `ddp.average_in_collective=False`
* `pad_seq_to_mult=16` for CP8, or documented equivalent when additional parallelism constraints apply

sen-ppl:
**Draft** for NVIDIA-NeMo/Megatron-Bridge#4593 (Bridge-side packed-SFT data blockers; companion to the GLM-5.2 tracker #4490). Filed as a starting point — please take it from here.

These are the genuinely-unaddressed packed-sequence-prep items from #4593 §2, exercised on a 64×GB200 GLM-5.2 (`GlmMoeDsa`) full SFT at `seq_len=150000`, TP2/PP4/CP8/EP64, offline packed THD. They were required to reach training, and with them the run reproduced the no-fix baseline's iter-1 loss bit-exactly.

### §2.3 — near-pack-size pre-pad overflow (`packed_sequence.py`)
`tokenize_dataset` set the pad target to `min(max_seq_length, ceil_to_nearest(len, mult))`, so a near-pack-size document (e.g. len 149999 at `max_seq_length` 150000) padded to `max_seq_length + 1`, overflowing the pack; and a document in the `(max_length_to_pad, max_seq_length]` gap was left at a length whose runtime segment (`stored - 1`) is not divisible by `pad_seq_to_mult` (= 2·CP), crashing THD+CP RoPE partitioning.
**Fix:** cap the pad target at `((max_seq_length - 1) // mult) * mult`; in `_pre_pad_data_point`, trim a gap-length doc to that capped target and add the single trailing pad, so the stored length is exactly `target + 1` (runtime `target`, divisible by `pad_seq_to_mult`, ≤ pack). Adds a regression test; `_pre_pad_data_point`'s signature and the existing overlong-truncation test are unchanged.

### §2.4 — `create_hist` IndexError on oversize sequences (`packing_utils.py`)
`counts[seq_len]` was indexed without a bounds check → IndexError for `seq_len > truncate_seq_len` when the truncation path isn't exercised first. Count/log/skip oversize sequences; in-range behavior unchanged.

### §2.5 — high-worker packing FD exhaustion (`packed_sequence.py`, `pack_sft_data.py`)
The tokenization `Pool` used the default `file_descriptor` tensor sharing → `RuntimeError: received 0 items of ancdata` at high worker counts. Set `file_system` sharing + raise `RLIMIT_NOFILE` before the `Pool` (multi-worker path only; the `num_workers <= 1` path is untouched). `pack_sft_data.py` honors `PACK_NUM_WORKERS`.

## Already handled upstream (intentionally not in this PR)
- **§2.1** (tensor `val + [pad]*n` TypeError) and **§2.2** (loss_mask padding): already fixed by merged #4056 — `_pre_pad_data_point` coerces via `.tolist()` and pads `loss_mask`.
- **Item 1** (middle-PP `packed_seq_params`): handled by `_uses_packed_sequence_metadata` / `_middle_pp_stage_needs_batch` (returns True for `enable_offline_packing`), so middle stages load `cu_seqlens`.
- **Item 4** (`apply_chat_template` / chat mask): superseded by the `_chat_preprocess` + `build_assistant_loss_mask` rework.

## Known related gap (flagging, not fixed here)
The pre-existing overlong branch (`len > max_seq_length` → `val[:max_seq_length]`) is left as upstream. For THD+CP it produces a runtime length (`max_seq_length - 1`) not divisible by `pad_seq_to_mult`, and `create_hist` keeps it (`≤ truncate_seq_len`). A divisibility-preserving variant would truncate to `((max_seq_length - 1) // mult) * mult` instead. Left out because it changes upstream's general (non-CP) truncation semantics and the corresponding unit test — flagging for a maintainer decision.

### DIFF
--- a/scripts/training/pack_sft_data.py
+++ b/scripts/training/pack_sft_data.py
@@ -57,6 +57,12 @@ def main() -> None:
         "--packed-val-data-path", default=None, help="Optional output path for packed validation data."
     )
     parser.add_argument("--packed-metadata-path", default=None, help="Optional output path for packing metadata.")
+    parser.add_argument(
+        "--num-tokenizer-workers",
+        type=int,
+        default=1,
+        help="Tokenizer worker processes. Values less than or equal to 1 run serially (default: 1).",
+    )
     args = parser.parse_args()
 
     import megatron.bridge.recipes as all_recipes
@@ -94,15 +100,11 @@ def main() -> None:
     if offline_packing_specs is None:
         sys.exit(f"Error: recipe '{args.recipe}' has no offline packing specs.")
 
-    # Cap tokenizer workers to avoid /dev/shm OOM from multiprocessing shared memory.
-    # Default is -1 (all CPUs) which exhausts /dev/shm even on CPU nodes.
-    # Use 1 worker so serial materialization avoids multiprocessing shared memory
-    # (see packed_sequence._materialize_dataset_items).
-    offline_packing_specs.num_tokenizer_workers = 1
+    offline_packing_specs.num_tokenizer_workers = args.num_tokenizer_workers
 
     print(f"Recipe:   {args.recipe}")
     print(f"Seq len:  {offline_packing_specs.packed_sequence_size}")
-    print(f"Workers:  {offline_packing_specs.num_tokenizer_workers} (single-threaded, no /dev/shm)")
+    print(f"Workers:  {offline_packing_specs.num_tokenizer_workers}")
     print()
 
     print("Building tokenizer...")
--- a/src/megatron/bridge/data/datasets/packed_sequence.py
+++ b/src/megatron/bridge/data/datasets/packed_sequence.py
@@ -13,12 +13,14 @@
 # limitations under the License.
 import json
 import logging
+import resource
 import warnings
 from dataclasses import dataclass
 from multiprocessing import Pool
 from pathlib import Path
 
 import numpy as np
+import torch
 from megatron.core.msc_utils import MultiStorageClientFeature
 from tqdm import tqdm
 
@@ -48,8 +50,31 @@ def _init_shared_dataset_worker(dataset):
 def _materialize_dataset_items(dataset, num_workers):
     if num_workers <= 1:
         return np.array([dataset[i] for i in tqdm(range(len(dataset)))])
-    with Pool(num_workers, initializer=_init_shared_dataset_worker, initargs=(dataset,)) as pool:
-        return np.array(list(tqdm(pool.imap(_get_shared_dataset_item, range(len(dataset))), total=len(dataset))))
+
+    # File-backed tensor sharing avoids one descriptor per returned tensor; the pool still needs descriptors.
+    previous_sharing_strategy = torch.multiprocessing.get_sharing_strategy()
+    torch.multiprocessing.set_sharing_strategy("file_system")
+
+    previous_nofile_limit = None
+    try:
+        soft_limit, hard_limit = resource.getrlimit(resource.RLIMIT_NOFILE)
+        if soft_limit != hard_limit:
+            resource.setrlimit(resource.RLIMIT_NOFILE, (hard_limit, hard_limit))
+            previous_nofile_limit = (soft_limit, hard_limit)
+    except (ValueError, OSError) as error:
+        logger.warning("Unable to raise the file-descriptor limit for tokenizer workers: %s", error)
+
+    try:
+        with Pool(num_workers, initializer=_init_shared_dataset_worker, initargs=(dataset,)) as pool:
+            items = tqdm(pool.imap(_get_shared_dataset_item, range(len(dataset))), total=len(dataset))
+            return np.array(list(items))
+    finally:
+        if previous_nofile_limit is not None:
+            try:
+                resource.setrlimit(resource.RLIMIT_NOFILE, previous_nofile_limit)
+            except (ValueError, OSError) as error:
+                logger.warning("Unable to restore the file-descriptor limit after tokenization: %s", error)
+        torch.multiprocessing.set_sharing_strategy(previous_sharing_strategy)
 
 
 def _pre_pad_data_point(data: dict, max_seq_length: int, max_length_to_pad: int, pad_id: int) -> None:
@@ -77,16 +102,19 @@ def _pre_pad_data_point(data: dict, max_seq_length: int, max_length_to_pad: int,
         val = data[key]
         # _chat_preprocess returns torch tensors / numpy arrays; normalize to a plain list.
         val = val.tolist() if hasattr(val, "tolist") else list(val)
-        if len(val) <= max_length_to_pad:
+        sequence_length = len(val)
+        if sequence_length <= max_length_to_pad:
             # input_ids are truncated by 1 for labels; add 1 extra pad token
-            val = val + [pad_value] * (max_length_to_pad - len(val) + 1)
-        elif len(val) > max_seq_length:
-            logger.info(
-                "Sequence length %d is larger than max_seq_length %d; truncating for packing.",
-                len(val),
-                max_seq_length,
-            )
-            val = val[:max_seq_length]
+            val = val + [pad_value] * (max_length_to_pad - sequence_length + 1)
+        else:
+            if sequence_length > max_seq_length:
+                logger.info(
+                    "Sequence length %d exceeds max_seq_length %d; truncating for packing.",
+                    sequence_length,
+                    max_seq_length,
+                )
+            # Keep the runtime segment length (stored length minus one) at the divisible pad target.
+            val = val[:max_length_to_pad] + [pad_value]
         data[key] = val
     return
 
@@ -113,6 +141,8 @@ def tokenize_dataset(
             Can include 'chat', 'use_hf_tokenizer_chat_template', 'tool_schemas', etc.
         pad_seq_to_mult (int | None): Optional multiple to pad each sequence to during packing
             preparation (e.g., set to 2 * context_parallel_size for THD CP).
+        num_tokenizer_workers: Number of worker processes used to materialize tokenized samples.
+            Values less than or equal to 1 run serially.
 
     Returns:
         np.ndarray: A NumPy array containing the tokenized data.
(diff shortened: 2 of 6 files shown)


scripts/training/pack_sft_data.py
```diff
@@ -94,15 +95,19 @@ def main() -> None:
     if offline_packing_specs is None:
         sys.exit(f"Error: recipe '{args.recipe}' has no offline packing specs.")
 
-    # Cap tokenizer workers to avoid /dev/shm OOM from multiprocessing shared memory.
-    # Default is -1 (all CPUs) which exhausts /dev/shm even on CPU nodes.
-    # Use 1 worker so serial materialization avoids multiprocessing shared memory
-    # (see packed_sequence._materialize_dataset_items).
-    offline_packing_specs.num_tokenizer_workers = 1
```
yaoyu-33, line 101: should probably be an argument, do not import from env.

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4598/commits/d9ce0cc8c5425abe2490af512f6fb9f149f3a810

yaoyu-33:
I think the current padding fix is still mixing stored length and runtime length in a couple of places.

For packed SFT, the packed file stores `input_ids`, and runtime `tokens`/`labels` are produced later by shifting:

```python
tokens = input_ids[start : end - 1]
labels = input_ids[start + 1 : end]
```

So the divisibility/cap should apply to runtime length `len(input_ids) - 1`, while the stored sample length should be `runtime_target + 1`.

Concretely, the current PR code still has these issues:

1. `((max_seq_length - 1) // mult) * mult` prevents reaching the runtime cap. For example, `max_seq_length=150000, mult=16` caps runtime length at `149984`, not `150000`.
2. `ceil_to_nearest(len(data["input_ids"]), mult)` rounds the stored length. A stored length of `17` already has runtime length `16`, but this rounds to `24` and pads the stored sample to `25`.
3. The tokenizer path still needs a stored-token budget of `max_seq_length + 1` when this padding path is active. Otherwise a real SFT dataset can truncate the final shifted label token before `_pre_pad_data_point()` sees it, and padding can only add a pad token back.

I pushed a tested reference patch here:

https://github.com/NVIDIA-NeMo/Megatron-Bridge/compare/d9ce0cc8c5425abe2490af512f6fb9f149f3a810...yuya/pr4598-runtime-pad-fix

The key change is:

```python
runtime_max_seq_length = max_seq_length
max_runtime_pad_cap = (runtime_max_seq_length // pad_seq_length_to_mult) * pad_seq_length_to_mult
stored_max_seq_length = runtime_max_seq_length + 1

# create_sft_dataset(..., seq_length=stored_max_seq_length, ...)

runtime_len = max(len(data["input_ids"]) - 1, 0)
runtime_length_to_pad = min(max_runtime_pad_cap, ceil_to_nearest(runtime_len, pad_seq_length_to_mult))
max_stored_length_to_pad = runtime_length_to_pad + 1
```

EOS validation on that patch:

```text
uv run python -m pytest tests/unit_tests/data/datasets/test_packed_sequence.py -q
# 12 passed

uv run pre-commit run --files src/megatron/bridge/data/datasets/packed_sequence.py tests/unit_tests/data/datasets/test_packed_sequence.py
# passed
```

I also ran an inspection script through `tokenize_dataset -> create_hist -> fill_packing_strategy -> GPTSFTPackedDataset.collate_fn`; the relevant printed values were:

```text
factory_seq_length 17
stored_lengths [17, 17, 17, 17]
runtime_lengths [16, 16, 16, 16]
histogram_nonzero {16: 4}
tokens_shape (1, 16)
labels_shape (1, 16)
cu_seqlens [[0, 16, -1]]
labels_tail [14, 15, 16]
rounding_factory_seq_length 41
rounding_stored_lengths [17, 25, 41, 41, 41]
rounding_runtime_lengths [16, 24, 40, 40, 40]
inspection_ok
```

yaoyu-33 (approved):

PR #4604, opened 2026-06-30
fix(recipe): correct Qwen3.5-VL 122B H100 VPP

cuichenx:
## Summary

- change the canonical flat Qwen3.5-VL 122B H100 BF16/FP8 recipes from VPP4 to VPP2
- keep the legacy workload-base configurations synchronized during their deprecation window
- add a regression test covering canonical and legacy BF16/FP8 configurations

## Root cause

The H100 presets use 48 transformer layers with PP8, which assigns six layers to each pipeline rank. MCore requires the layers per pipeline rank to be divisible by the virtual pipeline size. VPP4 violates that requirement because `6 % 4 != 0`, so the model fails during construction.

VPP2 produces two three-layer virtual chunks while preserving the existing 128-GPU TP2/PP8/EP16 topology and keeping VPP enabled for expert-parallel communication overlap.

## Impact

Both the canonical `megatron.bridge.perf_recipes` path used by `run_script.py` and the legacy `scripts/performance/configs` compatibility path now produce a valid layout. Other hardware presets and model families are unchanged.

## Validation

- canonical/legacy BF16 and FP8 parity and divisibility test in `nvcr.io/nvidia/nemo:26.06`: 2 passed
- canonical `run_script.py --dryrun` for H100 BF16: passed, emitted VPP2
- canonical `run_script.py --dryrun` for H100 FP8 current scaling: passed, emitted VPP2
- `uv run pre-commit run --all-files`: passed

## Test scope

Blast radius is limited to the Qwen3.5-VL 122B H100 performance recipes and their legacy compatibility configs. L0/unit validation is sufficient; no CI tier label is requested.

### DIFF
--- a/scripts/performance/configs/qwen_vl/qwen35_vl_workload_base_configs.py
+++ b/scripts/performance/configs/qwen_vl/qwen35_vl_workload_base_configs.py
@@ -267,7 +267,7 @@ QWEN35_VL_122B_A10B_PRETRAIN_CONFIG_H100_BF16 = replace(
     num_gpus=128,
     tensor_model_parallel_size=2,
     pipeline_model_parallel_size=8,
-    virtual_pipeline_model_parallel_size=4,
+    virtual_pipeline_model_parallel_size=2,
     expert_model_parallel_size=16,
     moe_a2a_overlap=True,
 )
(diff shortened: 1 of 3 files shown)


yaoyu-33 (approved):

tomlifu (approved):
LGTM

malay-nagda (approved):

PR #4605, opened 2026-06-30
fix(data): handle packed SFT preparation edge cases

yaoyu-33:
## Summary

This is an org-owned migration and cleanup of fork PR #4598 for the still-unaddressed packed SFT preparation items in #4593 §2.3–§2.5.

- Keep near-pack-size and overlong stored samples within the pack while making their runtime lengths divisible by the effective THD/CP padding multiple.
- Reject padding configurations that cannot produce a positive runtime segment.
- Skip and report oversize samples in `create_hist` instead of indexing beyond the histogram.
- Use file-backed tensor sharing and a temporary file-descriptor limit increase around multi-worker tokenization, restoring both settings afterward.
- Replace the `PACK_NUM_WORKERS` environment behavior from #4598 with an explicit `--num-tokenizer-workers` argument and a conservative default of 1, addressing the maintainer inline comment.

Areas already handled on `main` remain out of scope, as described in #4598.

## Validation

- `uvx ruff check` on all six changed files — passed
- `uvx ruff format --check` on all six changed files — passed
- `uv run --active --no-sync pre-commit run --all-files` — passed
- `uv run --active --no-sync python -m pytest --confcutdir=tests/unit_tests/scripts tests/unit_tests/scripts/test_pack_sft_data.py -q` — 5 passed
- The focused packed-sequence and packing-utils pytest files were attempted with the standard project command, but this host cannot install `nvidia-resiliency-ext==0.6.0`: available wheels require `manylinux_2_39`, while the host is `manylinux_2_31_x86_64`. Those tests are included for CI.

No training E2E verification was run for this migrated branch.

## Residual risk

The multiprocessing mitigation is intended for the supported Linux default-fork path. A non-default `spawn` context would not inherit the parent's torch sharing strategy and was not validated here.

### DIFF
--- a/scripts/training/pack_sft_data.py
+++ b/scripts/training/pack_sft_data.py
@@ -57,6 +57,12 @@ def main() -> None:
         "--packed-val-data-path", default=None, help="Optional output path for packed validation data."
     )
     parser.add_argument("--packed-metadata-path", default=None, help="Optional output path for packing metadata.")
+    parser.add_argument(
+        "--num-tokenizer-workers",
+        type=int,
+        default=1,
+        help="Tokenizer worker processes. Values less than or equal to 1 run serially (default: 1).",
+    )
     args = parser.parse_args()
 
     import megatron.bridge.recipes as all_recipes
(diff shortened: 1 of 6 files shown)


yaoyu-33:
Closing this replacement PR because the same cleaned-up changes were pushed directly to the original fork PR #4598.

PR #4608, opened 2026-07-01
feat(data): intra-microbatch reordering for MegatronMIMO (+ sequence packing, scalable DP)

### LINKED ISSUE #4609: [feature] intra-microbatch reordering for MegatronMIMO (+ sequence packing, scalable DP)
### User problem

Non-colocated MegatronMIMO training (e.g. Qwen3.5-VL: separate vision encoder + language model wired by `BridgeCommunicator`) has a per-step **data-parallel straggler**: per-micro-batch multimodal load
(vision patch count) is uneven across samples, so a DP rank that draws a heavy-image shard stalls the whole group every step. Throughput is bounded by the slowest rank.

Two supporting gaps make this worse / block the fix:

1. **Sequence packing is unused.** MegatronMIMO on Megatron-Core supports packed sequences, but in Megatron Bridge, there is no logic to actually feed a packed language shard, so padding is computed.
2. **Data parallelism is not scalable (vision DP = 1 only).** Today every DP worker reads the *full* global micro-batch and slices its own part locally in the forward step → CPU/host-IO overhead that scales with DP, and vision DP > 1 is effectively unsupported. When the vision data parallel degress scale linearly, the total i/o volume across the cluster will scale quadratically.

### Desired outcome

**Main — Intra-microbatch Reordering** (cf. DistTrain §5.2, https://arxiv.org/abs/2408.04275):
rebalance each micro-batch's vision load across the module DP group, so per-rank vision
load is even and the straggler tail is removed. Works with heterogeneous DP (`vision_dp != language_dp`)
via canonical `n_groups` pairing (vision replica r ↔ language replica r) and a variable number of
images per sample (0 / 1 / N).

**Auxiliary 1 — Sequence Packing:** add the logic to pack a language DP shard's real tokens into a
single `[1, T]` THD sequence so the LM skips padding compute (MCore already accepts packed input).

**Auxiliary 2 — Scalable Data Parallelism:** let each rank's sampler read only its disjoint `1/dp`
shard instead of every rank reading the full batch and slicing locally — reducing the per-rank CPU/IO
overhead and enabling vision DP > 1. This is the read-sharding base the reorder rebalances on top of.

All three are off by default and gated by `MegatronMIMOFeatureConfig`. Expected effect (16-GPU
validation): median throughput dominated by the read-sharding win, plus a large straggler-tail
reduction; same total work / same images processed.

> The name/implementation of the configuration class can be a subject of discussion.

### Alternatives considered

- **Status quo** — full-batch read + forward-time local slice (`slice_batch_for_megatron_mimo`):
  wastes host IO, supports vision DP = 1 only, cannot rebalance per-rank cost.

### Affected area

`area:data` (primary), `area:training`

### Urgency / use case

Important but not blocking. 

Use case: non-colocated MegatronMIMO VLM finetuning (Qwen3.5-VL + receipt/doc datasets) where vision load is the per-step straggler.

### Extra context

- Implementation: PR #4608 .
- Out of scope: Inter-microbatch (DistTrain S5.3) reordering is a future follow-up — the
  transport is built inter-ready but the balancer stays intra-only.

sailor1493:
# What does this PR do ?

> close #4609

**Main Feature**
- Intra-microbatch Reordering ([DistTrain](https://arxiv.org/abs/2408.04275) §5.2): rebalance each
  micro-batch's vision load across the module DP group by a per-sample cost all-gather + ragged
  `all_to_all` (GPU-resident, overlapped with compute) so the per-step DP straggler is removed.

image

image

**Auxiliary Features (needed to implement the main feature)**
- Sequence Packing — Megatron Core accepts packed sequences, but there was no logic to actually
  build and feed a packed sequence; added it (`[1, T]` THD).
- Scalable Data Parallelism — previously only vision DP=1 was supported (every DP worker reads the
  full global micro-batch and slices its own part locally → CPU/IO overhead). Now each rank reads
  only its disjoint `1/dp` shard, and vision DP>1 is supported.

All three are off by default, gated by the new `MegatronMIMOFeatureConfig`. **No breaking change**
(purely additive: new config + example CLI flags; no existing key/flag/symbol changed).

# Changelog

### Config (`src/megatron/bridge/training/config.py`)
- Add `MegatronMIMOFeatureConfig`: `scalable_dp`, `intra_microbatch_reorder`,
  `overlap_intra_microbatch_reorder`, `reorder_window_size`, `pack_sequences_in_batch`,
  cost coefficients (`cost_linear_vit`, `cost_linear_lm`), `pad_token_id`.
- `finalize()` rejects negative coefficients/pad id, `reorder_window_size < 1`, and an all-zero cost
  when reorder is enabled. Wired into `ConfigContainer` via a new `mimo` field.

### Reorder engine (`src/megatron/bridge/data/megatron_mimo/`)
- `reorder_buffer.py` (new): per-sample cost all-gather (Gloo) + ragged `all_to_all` (NCCL) on a
  dedicated CUDA stream; ragged serialize/deserialize; `balanced_assignment` (contiguous-block,
  het-DP canonical `n_groups`); `split_microbatch`/`merge_samples`; variable-images-per-sample
  (`cu_img`, `empty_like_vision`); `W`-micro-batch reorder window with cross-window prefetch overlap.
- The per-sample cost is the **module-independent image-placeholder token count** in `input_ids`
  (`cost = count(image_token) · spatial_merge_size²`), identical on the vision and language modules and
  on every PP stage — so both modules derive the same assignment with no cross-module communication.
- `intra_microbatch_pack.py` (new): pack each language shard's real tokens into a single `[1, T]`
  THD sequence (`pack_language_shard` / `assemble_packed_sequence`); packs `position_ids`/`labels`/
  `loss_mask` to the same `[1, T]` on every PP stage so the THD rotary is sized to `T`.
- `dp_utils.py`: scalable-DP sampling info (each rank reads its module-local shard); image-boundary
  vision handling; non-scalable `vision_dp > 1` explicitly raises (out of scope); colocate
  single-consumer helpers.

### Training integration (`src/megatron/bridge/training/`)
- `megatron_mimo_step.py`: thread the config through the forward step — `scalable_dp` skips the
  forward-time local slice (sampler already delivered the shard); optional in-batch sequence packing.
  Keeps `input_ids` on every language PP stage when reorder or packing is active so per-sample lengths
  are derivable on stages > 0. DP loss reduction is unchanged from non-scalable runs.
- `train_megatron_mimo.py`: build/route the scalable-DP sampler and the reorder exchange from config.

### Data / packing (`src/megatron/bridge/data/`)
- `datasets/packing_utils.py`: shared THD packing helper (placement plan across DP workers).

### Example (`examples/megatron_mimo/qwen35_vl/finetune_qwen35_vl.py`)
- Wire the config + CLI flags (`--scalable-dp`, `--intra-microbatch-reorder`/`--no-…`,
  `--pack-sequences-in-batch`, `--reorder-window-size`, cost coefficients).
- Guard the one remaining unsupported config: `NotImplementedError` for `dataloader_type != "single"`.
  `PP > 1` is supported on untied checkpoints (no packing/PP guard); see Known limitations.

### Docs (`docs/`)
- `training/mimo-intra-microbatch-reorder.md`: feature, config, validation matrix, single-node throughput,
  gaps. Index entries: `index.md`, `training/README.md`.

### Tests (`tests/`)
- Unit: `data/megatron_mimo/test_reorder_buffer.py` (44), `data/megatron_mimo/test_intra_microbatch_pack.py`
  (14), `training/test_mimo_feature_config.py` (14), `data/datasets/test_packing_utils.py`, plus updates to
  `training/megatron_mimo/test_megatron_mimo_step.py` — all green.
- Functional (2-GPU): `test_groups/megatron_mimo/test_reorder_exchange.py` — on-device exchange smoke.

# Performance

Throughput, single **8×A100-80GB** node, vision dp4 / language dp4, PP=1, TP=1, Qwen3.5-0.8B (VL) +
CORD-v2, seq 2048, **sequence packing on**, MBS/GBS 32 (= 8 examples/rank), patch-only cost.
**500 iters**, stats over `elapsed time per iteration` with the first 10 iters excluded (compile + the
one-time side-NCCL `new_group` warmup). `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`,
`CUDA_DEVICE_MAX_CONNECTIONS=8`.

> Note: Due to PP bug regarding tied embedding, we turn of the tie_embedding option

| config | p10 | p25 | p50 | p75 | p90 | p99 | mean | max |
|---|---|---|---|---|---|---|---|---|
| base + pack (full-batch read, no scalable-dp) | 826 | 885 | 968 | 1089 | 1523 | 2342 | 1075 | 2659 |
| scalable read only + pack (`--no-intra-microbatch-reorder`) | 726 | 761 | 793 | 832 | 864 | 910 | 798 | 1493 |
| scalable + reorder + pack (default, overlap) | 703 | 736 | 752 | 771 | 786 | 863 | **755** | 1410 |
| scalable + reorder + pack, `--no-overlap-intra-microbatch-reorder` | 747 | 787 | 811 | 831 | 852 | 891 | 811 | 1496 |

(ms/iter.)

1. **Scalable read (base → read-only) is the dominant win**, as expected — each rank reads only its `1/dp`
   shard instead of every rank redundantly reading and decoding all 32 samples. mean **1,075 → 798 (1.35×)**,
   and the full-read tail collapses: **p99 2,342 → 910**, p90 1,523 → 864.
2. **Reorder pays off only with overlap, as a single-digit-% gain here.** With packing, each language rank
   packs its `mbs/dp` samples into one `[1, T]` THD sequence whose `T = Σ` image-placeholder tokens, so an
   uneven per-rank image load skews `T`. The all-to-all evens per-rank patch cost (balance probe: spread
   ≈1.27× → ≈1.07×, ~18–24 of 32 samples exchanged):
   - **read-only → reorder + overlap:** mean **798 → 755 (~5%)**, p90 864 → 786 (~9%), p99 910 → 863 (~5%);
     inter-percentile spread (p10–p99) tightens 184 → 160 ms.
   - **Overlap is what makes it net-positive.** Without it, the synchronous all-to-all sits on the critical
     path: `--no-overlap` mean **811 ≈ read-only 798** and behind overlapped **755**. So at this config the balancing gain is real only because the exchange is hidden behind compute.
   - We believe the gain can be larger with more extreme settings - more images, larger image size etc., yet the CORD-v2 dataset was enough to show the proposed gain.
3. The per-rank imbalance here is mild (~1.3×, CORD-v2 natural image-size variance at dp4/dp4), so balancing
   buys single-digit %, not multiples. **The reorder win grows with imbalance** — larger DP (more ranks →
   higher chance one rank draws a heavy shard), larger image-size variance, and larger per-rank batch.

# Known limitations / out of scope

- **`PP>1` is supported with untied checkpoints.** Reorder + `PP>1` and packing + `PP>1` are fixed and
  verified at `dp2/dp2/pp2` (`lm loss < 2`, tracking the no-reorder PP=2 baseline). **Tied-embedding +
  `PP>1`** is still not working at the upstream, independent of this feature — use an untied checkpoint
  (`tie_word_embeddings=false`, LM head = copy of the input embedding) for `PP>1`.
- **Intra-microbatch reordering** does not support a non-`single` sampler (guarded with an explicit error),
  and `TP>1` is untested. `CP>1` is blocked upstream (`bridge_communicator` asserts language-grid CP size 1).
- Distributed exchange needs ≥4 GPUs → CI coverage is the unit-emulated all-to-all + a 2-GPU smoke; the
  full on-device path was validated manually (16-GPU multi-node + 8×A100 single-node).
- Inter-microbatch reordering ([DistTrain](https://arxiv.org/abs/2408.04275) §5.3): a future follow-up; the transport is inter-ready in code, but the balancer stays intra-only. We leave this as out of scope for this PR.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci)
in the Contributing doc for how to trigger the CI.

**Pre checks**:

- [x] Did you write any new necessary tests? (3 new unit suites + updates + 1 functional)
- [x] Does the PR affect components that are optional to install? — No (no new optional deps).

Signed-off-by: Kayeon Song <kayeon.song@navercorp.com>


sailor1493:
@liding-nv 
Please be noted with this comment(https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/4609#issuecomment-4933142521)
Wish you good luck!

examples/megatron_mimo/qwen35_vl/finetune_qwen35_vl.py
```diff
@@ -862,12 +941,84 @@ def _build_data_iterators(cfg, _megatron_mimo_infra, *, train_state=None):
...
+
+            # Precondition for the per-sample exchange. exchange_window infers each sample's global
+            # index from CONTIGUOUS sharding (rank r holds [r*local, (r+1)*local)), which only holds
+            # for the "single" sampler. A cyclic/batch sampler would silently mispair the
+            # vision/language fan-out (each rank ends up with the wrong samples), so fail loud here
+            # instead of training on corrupted pairings. PP is orthogonal: it does not participate in
+            # DP sharding, so every PP stage of a given dp_rank reads the same contiguous shard and
+            # derives the identical deterministic reordering within its own per-stage DP group —
+            # reorder + PP>1 is therefore supported.
+            if cfg.dataset.dataloader_type != "single":
```
yaoyu-33, line 966: Blocking: this guard is nested under `intra_microbatch_reorder`, but scalable DP already gives each module its own DP sampler. With vision DP 2, language DP 4, and logical MBS 8, `single` pairs vision-r0 `[0,1,2,3]` with language-r0+r1 `[0,1,2,3]`; `cyclic` with sharding produces `[12,10,9,6]` versus `[4,0,12,8]` (unsharded cyclic also differs). `BridgeCommunicator` then attaches vision activations to different text examples even when reorder is disabled. Please require the paired/contiguous sampler whenever `scalable_dp` is enabled, or route actual global sample IDs, and add a heterogeneous-DP cyclic regression test.

src/megatron/bridge/data/megatron_mimo/dp_utils.py
```diff
@@ -278,3 +305,51 @@ def slice_batch_for_megatron_mimo(
...
+    Returns:
+        An ``int64`` tensor of shape ``[B]`` with each sample's real length.
+    """
+    if (
+        isinstance(attention_mask, torch.Tensor)
+        and attention_mask.dim() == 2
+        and attention_mask.shape == input_ids.shape
+    ):
+        return attention_mask.to(torch.bool).sum(dim=1).to(torch.long)
+    return (input_ids != pad_token_id).sum(dim=1).to(torch.long)
```
yaoyu-33, line 355: Blocking: this fallback treats every occurrence of `pad_token_id` as padding, while the Qwen adapter builds a true mask and then projects `attention_mask=None`. For EOS-as-pad (or any valid interior pad-ID token), `[10,0,11,0-pad]` is measured as length 2 and `assemble_packed_sequence` copies `[10,0]`, silently dropping `11`; the authoritative mask gives length 3. Current main's merged #5131 intentionally requires `attention_mask` for MIMO packing. Please use that source of truth and fail closed when it is unavailable rather than infer packed boundaries from token values.

yaoyu-33 (changes requested):
This needs to remain split rather than be conflict-resolved as one PR. In #4609 the maintainers and contributors explicitly agreed on three PRs: (1) MIMO sequence packing, (2) scalable data loading, and (3) intra-microbatch balancing. The first split, #5131, is already merged, and current main now has the authoritative `sequence_pack.py` path plus subsequent packed-PP/ragged-shape and padding-mask fixes. Please drop the duplicate packing implementation here, submit scalable loading independently against current main, then layer reorder/exchange/overlap on that.

There are also two correctness blockers inline: scalable loading can silently mismatch vision and language samples with the cyclic sampler even when reorder is disabled, and this branch's packing length fallback can truncate real tokens when the pad ID appears inside a sequence.

I reviewed the full 19-file diff, current checks/threads, process-group ownership/ordering, ragged serialization, stream handoff, sampler semantics, PP fields, loss handling, and zero/multi-image behavior. Focused CPU validation passed: 68 DP/reorder tests and 30 packing tests. The canonical assignment and stream synchronization look coherent under the documented all-modules-DP>1 + single-sampler constraints, but the bundled, conflicting head should not be merged in its current form.

liding-nv:
i think we should close this PR because we have aligned to split this one into several smaller PRs
https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5131 was the first one and has been merged

PR #4619, opened 2026-07-01
feat(model): add lazy transformers compatibility guards

### LINKED ISSUE #4314: Add lazy per-model Transformers version guards for Bridge
## Summary

Megatron Bridge currently has a tight global `transformers` dependency. We want to loosen the package-level version requirement so downstream libraries do not have to track the newest HF release unless they use models that actually require it.

The key requirement: `megatron.bridge` and model registration should import successfully under older viable `transformers` versions. If a user selects a model whose HF support requires a newer version, Bridge should fail lazily with a clear model-specific error.

## Plan

Add a Bridge-local compatibility system:

- Add a compatibility helper module with:
  - `get_transformers_version()`
  - `is_transformers_min_version(version)`
  - `TransformersVersionError`
  - `require_transformers_version(model_name, min_version, *, symbols=(), action=None)`

- Extend `MegatronModelBridge.register_bridge(...)` with:
  - `min_transformers_version: str | None`
  - `required_transformers_symbols: tuple[str, ...] = ()`

- Store this metadata on the bridge class and validate only after a bridge is selected, before provider construction, HF loading, export, or conversion.

- Convert import-time risky bridge registrations to string-based `source="ArchitectureName"` registrations so missing HF classes do not break import.

- Move HF model/config imports that require newer `transformers` into guarded runtime paths.

- Wrap `AutoConfig.from_pretrained()` failures for known guarded models so users get a Bridge compatibility error instead of a generic HF “architecture not recognized” error.

## Model Metadata Policy

- Every registered bridge gets a manually defined `min_transformers_version`.
- Unknown model minima default to `5.3.0`.
- Models requiring newer HF internals get higher explicit values and required-symbol checks.
- First PR should add the guard system and metadata only; dependency range loosening can follow after validation.

## Test Plan

- Unit tests for version helper behavior.
- Registration/import tests proving `import megatron.bridge` succeeds when newer HF symbols are unavailable.
- AutoBridge tests for string dispatch, lazy errors, and config-load error wrapping.
- Full-suite compatibility validation under:
  - current pinned `transformers`
  - intended future lower-bound version
  - `transformers==5.3.0`
- Older-version skips/xfails must be explicit and tied to model-level metadata.
- Any import-time failure from a registered but unused model is a bug.

## Success Criteria

- Downstream users can import Bridge with an older supported `transformers`.
- Unsupported models fail only when selected/used.
- Error messages clearly state model, installed HF version, required version, missing symbol/module when applicable, and upgrade guidance.

cuichenx:
## Summary

- add a Bridge-local, PEP 440-aware Transformers compatibility helper with actionable version and missing-symbol errors
- attach optional minimum-version and required-symbol metadata to registered bridge classes, enforcing it lazily after model selection and before provider/model loading or conversion
- narrowly translate known guarded `AutoConfig` architecture-recognition failures without masking network, authentication, malformed-config, or unrelated errors
- migrate GLM-5, Qwen3.5 text/VL, and DeepSeek V4 registrations and provider imports to the new mechanism
- add focused coverage for version parsing, dotted symbols, registration metadata, guard timing, config error handling, and fresh-process registration
- update the adding-model-support skill with compatibility discovery, registration, lazy-import, and boundary-test requirements

## Motivation

This is the guard and metadata foundation proposed in #4314. It allows package and registry discovery to remain independent of newer model-specific Transformers APIs while preserving the current global dependency range. This PR does not loosen dependency bounds.

`min_transformers_version=None` intentionally means that compatibility research is incomplete. It is not treated as an implicit minimum; compatibility metadata is added only where the version boundary is supported by direct evidence.

## Compatibility evidence

- GLM-5 and Qwen3.5 required symbols are absent in Transformers 5.1.0 and present in 5.2.0+, so those bridges declare `>=5.2.0` with exact symbols.
- DeepSeek V4's native config and model symbols are available and its bridge constructs with Transformers 5.8.0, while 5.7.0 produces the exact lazy guard. It therefore declares `>=5.8.0`.
- The 26.06 container baseline uses Transformers 5.8.1. Disposable wheel overlays exercised exact Transformers 5.7.0 and 5.6.0 without changing project dependency bounds.
- Every registered model family reached Bridge's loading progress path in all eligible environments. DeepSeek V4 is the only family with a newly observed version-specific pre-load incompatibility.

## Validation

- `uv run pre-commit run --all-files` in `nvcr.io/nvidia/nemo:26.06`
- focused impacted unit tests: 87 passed on Transformers 5.8.1; 58 passed on 5.7.0; 58 passed on 5.6.0
- scoped `tests/unit_tests/models`: 2783 passed on 5.8.1; 2753 passed on 5.7.0; 2753 passed on 5.6.0
- the same three Qwen3-VL processor setup errors occurred in each broad run because of process-global auto-doc registration order; the full Qwen3-VL test file passed 10/10 in a fresh process on every version
- public tiny-model plus repository toy/synthetic import probes covered every registered model-family directory and reached loading progress on 5.8.1, 5.7.0, and 5.6.0 where compatible
- exact DeepSeek V4 boundary probes passed at 5.8.0 and produced the expected lazy version error at 5.7.0
- DeepSeek V4 on the 5.8.1 baseline proceeds past Transformers compatibility, then encounters an independent pinned-MCore `dsv4_hybrid` layer-type mismatch
- `quick_validate.py skills/adding-model-support`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/skills/adding-model-support/SKILL.md
+++ b/skills/adding-model-support/SKILL.md
@@ -1,7 +1,6 @@
 ---
 name: adding-model-support
-description: Guide for adding support for new LLM or VLM models in Megatron-Bridge. Covers bridge, provider, recipe, tests, docs, and examples.
-when_to_use: User asks to add, onboard, or integrate a new model family; 'add Qwen4 support', 'onboard Llama 5', 'create a bridge for X', 'write a recipe for Y'.
+description: Guide for adding, onboarding, or integrating new LLM or VLM model families in Megatron-Bridge, including bridges, providers, recipes, tests, docs, and examples.
 ---
 
 # Adding New Model Support in Megatron-Bridge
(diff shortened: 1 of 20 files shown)


cuichenx:
no longer needed

PR #4622, opened 2026-07-02
refactor(recipes): add h100 recipe namespace

yaoyu-33:
## Summary
- Move library recipe implementations under hardware-specific h100 modules while keeping legacy recipe names as compatibility wrappers.
- Normalize canonical library PEFT recipe names to use `_peft_` instead of `_lora_`.
- Flatten Qwen3/Qwen3.5-VL recipe bodies and remove parameterized recipe helper patterns.
- Update recipe docs, Fern nightly pages, and recipe tests for the new naming.

## Split
This PR is now recipes-only. The legacy performance config removal has been split into #4623.

## Validation
- `python3 -m compileall -q src/megatron/bridge/recipes tests/unit_tests/recipes tests/functional_tests/test_groups/recipes/test_qwen35_vl_recipes_pretrain.py scripts/translate_mlm_to_bridge.py`
- `uvx pre-commit run --all-files`

Note: `uv run pre-commit run --all-files` is blocked on macOS by `nvidia-resiliency-ext==0.6.0` Linux-only wheels.

PR #4623, opened 2026-07-02
refactor(performance): remove legacy perf configs

yaoyu-33:
## Summary
- Remove the legacy scripts/performance/configs preset tree and dump_perf_configs.py.
- Resolve performance launcher workload defaults from flat megatron.bridge.perf_recipes recipes with v2 as the canonical suffix-less variant.
- Update perf env plugin/default config variant handling, training-mode inference for legacy lora recipe names, and perf integration/unit tests.

## Split
This is the legacy performance config cleanup split out of #4622. #4622 is now recipes-only.

## Validation
- python3 -m compileall -q scripts/performance scripts/training/run_recipe.py tests/functional_tests/test_groups/recipes/test_perf_config_integration.py tests/unit_tests/scripts/performance tests/unit_tests/recipes/test_override_precedence.py
- git diff --check
- uvx pre-commit run --all-files

Note: uv run pre-commit run --all-files is blocked on macOS by nvidia-resiliency-ext==0.6.0 Linux-only wheels.


malay-nagda:
@yaoyu-33 with all the changes, do we now need `megatron.core` and `src` (other than perf recipes) on login node to launch jobs?

Before the refactor, we only needed nemo_run on login node to launch jobs which avoided dependency issues on login node.

cc: @gautham-kollu

yaoyu-33:
@malay-nagda  thanks for reminding me, I will double check, this pr not fully ready yet. I will do some proper guarding.

scripts/performance/utils/utils.py
```diff
@@ -12,31 +12,109 @@
...
+    ("qwen3_30b_a3b", "pretrain", "h100", "bf16", "v2"): 16,
+    ("qwen3_30b_a3b", "pretrain", "h100", "fp8cs", "v2"): 16,
+}
 
 
 @dataclass
 class WorkloadBaseConfig:
-    """Container for workload base configs. This object exists because we cannot import MBridge on the headnode but need a place to store recipe overrides."""
-
-    # NOTE: `num_gpus` is for representation purposes only. It is only meant to
```
yaoyu-33, line 33: do NOT  remove the comments in this part

scripts/performance/setup_experiment.py
```diff
@@ -53,12 +51,6 @@
 except (ImportError, ModuleNotFoundError):
     from .perf_plugins import NsysPlugin, PerfEnvPlugin, PyTorchProfilerPlugin
 
-try:
```
yaoyu-33, line 56: why remove?

scripts/performance/run_script.py
```diff
@@ -91,7 +91,9 @@ def _find_perf_recipe(name: str) -> Callable[[], ConfigContainer] | None:
 
 def _flat_recipe_variant_suffix(config_variant: str | None) -> str:
     """Return the suffix used in flat perf recipe function names."""
-    return f"_{config_variant}" if config_variant and config_variant not in {"v1", "v2", "v3"} else ""
+    if config_variant is None or config_variant.lower() == "v2":
```
yaoyu-33: dont need to change this? we want to cover potential v#

scripts/performance/utils/utils.py
```diff
@@ -193,53 +413,44 @@ def get_exp_name_config(
...
+
+    precision_name = _normalize_precision_name(compute_dtype)
+    prefix = f"{model_recipe_name}_{task}_"
+    suffix = f"gpu_{gpu}_{precision_name}"
+    variants: set[str] = set()
+    for name in flat_perf_recipe_names():
+        if not name.startswith(prefix) or not name.endswith("_config") or suffix not in name:
+            continue
+        middle = name.removesuffix("_config").split(suffix, maxsplit=1)[1]
+        variants.add(middle.removeprefix("_") or "v2")
```
yaoyu-33: if only v2 exists, do we need to keep? should do something like more general for vx number

cuichenx (dismissed):

scripts/performance/perf_plugins.py
```diff
@@ -237,7 +237,8 @@ class PerfEnvPlugin(Plugin):
     gpu: str
     compute_dtype: str
     train_task: str
-    config_variant: str = "v1"
+    num_gpus: int | None = None
```
malay-nagda: @yaoyu-33 how is this used in PerfEnvPlugin? seems a placeholder?

yaoyu-33: You are right: `num_gpus` was unused inside `PerfEnvPlugin`. I removed the field and its constructor argument. GPU count remains handled by `setup_experiment` for recipe selection, node count, and experiment naming; it does not belong in this plugin. The stack was then rebased and synced, with the fix now at `ff99429a56b7d978a4e33c1c577d60a7ff7485b9`.

yaoyu-33:
Checked the nemo-ci impact and opened internal companion MR !2645.

- The 14 standard Llama 70B LoRA perf cases map one-to-one to the new `peft` exports; the recipe bodies are unchanged.
- nemo-ci also passes legacy `v1`/`v2`/`v3` config variants. With this PR those become literal suffixes and fail lookup, so the companion omits them at runtime for the suffix-less canonical recipes.
- The shim keeps the old `lora` and version tokens in golden artifact/W&B names, preserving existing baselines and performance history. Named variants such as `fsdp` and `large_scale` remain unchanged.
- It detects removal of the legacy perf-config tree, so the nemo-ci MR is safe to merge before this PR.

Separate upstream coverage gap: nemo-ci has 16-GPU LLMB LoRA cases, while the flat Llama PEFT exports here are 8-GPU only and rank-local lookup requires an exact GPU-count name. That gap already exists with the current 8-GPU-only flat LoRA exports, but it should be handled independently by adding 16-GPU recipes or defining supported scaling fallback behavior.

ko3n1g (dismissed):

yaoyu-33:
Paired nemo-ci L0/L1 validation completed against this PR at `eb2c1ff94be9` and internal nemo-ci MR !2645 (`4846a1d8d88a`). Internal pipeline: **57165478**.

Matrix: six targeted perf jobs across the requested H100, GB200, and GB300 environments (five PEFT/legacy-LoRA finetune cases and one pretrain case). **Result: 0/6 reached Slurm submission.**

What is verified:
- nemo-ci renders runtime `--task peft` for the renamed Llama recipes;
- legacy `v1` is omitted from runtime arguments for the suffix-less canonical recipe;
- the golden/W&B stem stays `...finetune_config_lora...v1_perf`, preserving history.

The first run exposed an unnecessary recipe lookup used only for auto-generated experiment names. Commit `eb2c1ff94be9` now skips that lookup when CI supplies `wandb_experiment_name`, and the launcher proceeds further.

The remaining blocker is `PerfEnvPlugin.setup()`: it calls `get_workload_base_config()` -> `perf_recipe_family_modules()` and imports `megatron.bridge.perf_recipes` on the login node. The dependency-light launcher venv intentionally does not install Megatron-Bridge, so all six cases fail before submission with `ModuleNotFoundError: No module named 'megatron'`.

This should not be worked around by disabling/defaulting `PerfEnvPlugin`, because its recipe-derived fields control performance-critical environment settings. The refactor needs either dependency-free recipe metadata for the launcher or to defer recipe-dependent environment derivation into the training container.

Conclusion: the nemo-ci name mapping works, but the pair is **not runtime-ready** until the remaining login-node recipe import is removed. The separate 16-GPU LLMB recipe-coverage gap also remains.

yaoyu-33:
Superseding validation update after fixing both login-node recipe imports. Current head: `b5a4de80e1a9`. Internal nemo-ci pipelines: **57168244** and **57174001** (nemo-ci MR !2645 at `4846a1d8d88a`).

Bridge fixes added during validation:
- `eb2c1ff94be9`: skip recipe lookup when CI supplies the experiment name.
- `b5a4de80e1a9`: keep launcher-only actions in `PerfEnvPlugin`; apply recipe-derived performance env inside the training container, then `exec` `run_script.py` so env vars precede training imports.

Results:

| Case | Level | Training | Current step | Baseline | Delta | CI verdict |
|---|---:|---|---:|---:|---:|---|
| H100 pretrain FP8 CS | L0 config | passed | — | — | — | passed |
| H100 PEFT BF16 | L1 | passed | 11972.7 ms | 12461.1 ms | -3.9% | post-check failed |
| GB300 PEFT BF16 | L1 | passed | 3266.0 ms | 3290.0 ms | -0.7% | post-check failed |
| GB300 PEFT FP8 MX | L0 | passed | 2204.4 ms | 2266.0 ms | -2.7% | post-check failed |
| GB200 PEFT BF16 | L1 | passed | 4171.4 ms | 3979.5 ms | +4.8% | post-check failed |
| GB200 PEFT FP8 MX | L0 | passed | 2782.0 ms | 2728.2 ms | +2.0% | post-check failed |

All five renamed PEFT/legacy-LoRA recipes completed 50 training steps. Step-time deltas are within the existing 5% threshold: **no perf slowdown and no OOM were observed**.

The paired nemo-ci mapping is verified live: runtime uses `--task peft`, legacy version variants are omitted for suffix-less recipes, and golden/W&B stems remain `...finetune_config_lora...v1_perf`.

The five red finetune jobs are post-validation only. The field named `GPU utilization` changed scale on the library-recipe path (for example, H100 ~11,443 current vs ~366 baseline), so the evaluator flags an impossible “unexpected improvement” even though step time is stable. The remaining CI follow-up is to reconcile/regenerate that throughput baseline or update evaluation for the current MODEL_TFLOP/s calculation.

The separate missing 16-GPU LLMB flat recipes remain a follow-up; this matrix validated the existing 8-GPU mappings.

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4623/commits/b5a4de80e1a978ec9a51ff2f04f910179fcc8b37

PR #4631, opened 2026-07-02
fix(misc): preserve nested Hugging Face config mappings

### LINKED ISSUE #4616: bug: instantiate_utils recursive instantiation breaks HF config classes with __post_init__ re-instantiation
## Description

Megatron-LM's `instantiate_utils.instantiate_node()` recursively instantiates all nested config entries that have a `_target_` key in `run_config.yaml`. This causes a crash when loading certain HF model configs (e.g., `Qwen3OmniMoeTalkerConfig`) because:

1. `instantiate_utils` pre-instantiates nested configs like `code_predictor_config` and `text_config` into Python objects
2. The parent config's `__post_init__` then tries to re-instantiate them via `**self.code_predictor_config` (dict unpacking), which fails because the value is already an object, not a dict

```
TypeError: Qwen3OmniMoeTalkerCodePredictorConfig() argument after ** must be a mapping,
not Qwen3OmniMoeTalkerCodePredictorConfig
```

## Root Cause

In `instantiate_utils.py` L308-314, `instantiate_node()` recursively processes all child values:

```python
for key in node.keys():
    if key not in exclude_keys:
        value = node[key]
        value = instantiate_node(value, mode=mode)  # recursively instantiates nested _target_ nodes
        kwargs[key] = _convert_node(value)
```

When `run_config.yaml` contains nested configs with `_target_`:

```yaml
code_predictor_config:
  _target_: transformers.Qwen3OmniMoeTalkerCodePredictorConfig
  num_codebooks: 2
  ...
```

`instantiate_node` first instantiates `code_predictor_config` into a `Qwen3OmniMoeTalkerCodePredictorConfig` object. This object is then passed as a kwarg to the parent `Qwen3OmniMoeTalkerConfig(code_predictor_config=<already instantiated>)`.

The HF config's `__post_init__` (transformers 5.8.1, `configuration_qwen3_omni_moe.py` L489-495) then crashes:

```python
def __post_init__(self, **kwargs):
    if self.code_predictor_config is None:
        self.code_predictor_config = Qwen3OmniMoeTalkerCodePredictorConfig()
    else:
        # CRASH: self.code_predictor_config is already an object, not a dict
        self.code_predictor_config = Qwen3OmniMoeTalkerCodePredictorConfig(**self.code_predictor_config)
```

## Affected Models

- `Qwen3OmniMoeTalkerConfig` (confirmed)
- Potentially any HF config class whose `__post_init__` re-instantiates nested configs without an `isinstance` guard

## Current Workaround

NeMo-RL applies a monkey-patch in `setup.py` that adds `isinstance` guards to `__post_init__` before `ConfigContainer.from_yaml()` is called (see [NVIDIA-NeMo/RL#3021](https://github.com/NVIDIA-NeMo/RL/pull/3021)).

## Suggested Fix

One or both of the following:

1. **In `instantiate_utils`**: skip recursive instantiation for kwargs that will be passed to an HF `PreTrainedConfig` subclass, since those classes handle their own nested config instantiation in `__post_init__`
2. **In Megatron-Bridge**: when generating `run_config.yaml`, avoid emitting `_target_` on nested HF sub-configs that the parent config's `__post_init__` already knows how to construct

## Reproducibility

- **transformers**: 5.8.1
- **Model**: `Qwen/Qwen3-Omni-30B-A3B-Instruct`
- **Trigger**: delete the mcore checkpoint cache and re-run — the newly generated `run_config.yaml` includes `_target_` on nested configs, causing the crash on next load

cuichenx:
# What does this PR do?

Fixes recursive instantiation of nested Hugging Face configs when Megatron Bridge serializes a checkpoint `run_config.yaml`.

The parent `PreTrainedConfig` remains instantiable through `_target_`, while nested HF configs are emitted as plain mappings so the parent can construct its own children. Existing model-specific `to_cfg_dict()` hooks remain authoritative, and ordinary Bridge/MCore dataclasses keep recursive `_target_` behavior.

Fixes [#4616](https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/4616).

## Root cause

MCore recursively instantiated every nested `_target_` before invoking the parent Qwen3-Omni config. The parent then attempted to unpack an already-instantiated child as keyword arguments and failed:

```text
TypeError: Qwen3OmniMoeTalkerCodePredictorConfig() argument after ** must be a mapping,
not Qwen3OmniMoeTalkerCodePredictorConfig
```

## Config shape

Before, both the parent and children were independently targeted:

```yaml
talker_config:
  _target_: transformers.models.qwen3_omni_moe.configuration_qwen3_omni_moe.Qwen3OmniMoeTalkerConfig

  code_predictor_config:
    _target_: transformers.models.qwen3_omni_moe.configuration_qwen3_omni_moe.Qwen3OmniMoeTalkerCodePredictorConfig
    hidden_size: 1024
    num_hidden_layers: 5

  text_config:
    _target_: transformers.models.qwen3_omni_moe.configuration_qwen3_omni_moe.Qwen3OmniMoeTalkerTextConfig
    hidden_size: 1024
    num_hidden_layers: 20
```

After, only the parent is targeted. Nested mappings retain their HF `model_type` discriminator:

```yaml
talker_config:
  _target_: transformers.models.qwen3_omni_moe.configuration_qwen3_omni_moe.Qwen3OmniMoeTalkerConfig

  code_predictor_config:
    model_type: qwen3_omni_moe_talker_code_predictor
    hidden_size: 1024
    num_hidden_layers: 5

  text_config:
    model_type: qwen3_omni_moe_talker_text
    hidden_size: 1024
    num_hidden_layers: 20
```

# Changelog

- Add HF-aware Bridge config serialization without changing generic MCore instantiation.
- Preserve explicit `to_cfg_dict()` serializers such as Qwen3-ASR.
- Preserve dynamic nested HF child classes through canonical `model_type` mappings.
- Add Qwen3-Omni object/dictionary YAML round trips, a non-default Llava SigLIP/Mistral round trip, and a non-HF recursive control.

# Validation

- Real config-only round trip using `Qwen/Qwen3-Omni-30B-A3B-Instruct` revision `26291f793822fb6be9555850f06dfe95f2d7e695`.
- All five Qwen3-Omni nested configs retained exact `to_dict()` equality after load.
- 199 affected unit tests passed, including Qwen3-Omni and Qwen3-ASR.
- Pre-commit passed.
- Newly emitted YAML loaded successfully with the pre-fix Bridge loader.

# GitHub Actions CI

Not triggered locally.

- [x] Added regression and control tests.
- [x] Documentation changes are not required.
- [x] No optional-install component is affected.

# Additional Information

Already-generated Qwen3-Omni YAML containing recursive child `_target_` entries must be regenerated or repaired; this PR does not migrate invalid existing YAML. Checkpoint tensor files are unaffected.

### DIFF
--- a/src/megatron/bridge/training/utils/config_utils.py
+++ b/src/megatron/bridge/training/utils/config_utils.py
@@ -13,9 +13,11 @@
 # limitations under the License.
 
 import logging
+from dataclasses import fields as dataclass_fields
+from dataclasses import is_dataclass
 from typing import Any, Mapping
 
-from megatron.training.config.container import ConfigContainerBase as _ConfigContainerBase  # noqa: F401
+from megatron.training.config.container import ConfigContainerBase as _MCoreConfigContainerBase
 from megatron.training.config.utils import (
     _get_init_false_fields,  # noqa: F401
     _resolve_target_class,  # noqa: F401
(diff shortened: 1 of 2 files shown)


yuekaizhang (approved):
Many thanks!

PR #4414, opened 2026-06-18
[model] fix: use hf_config to check whether model is dense

pavelgein:
# What does this PR do ?

Using hf config to check whether model is dense or not

# Changelog

- Add specific line by line info of high level changes in this PR.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci) in the Contributing doc for how to trigger the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

# Before your PR is "Ready for review"

**Pre checks**:

- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?

If you haven't finished some of the above items you can still open "Draft" PR.

# Additional Information

- Related to #4413

PR #4634, opened 2026-07-02
fix(ckpt): select gemma4 dense mapping registry by hf_config and guard None conversion tasks

pruprakash:
## Summary

Fixes NVBug 6367442: gemma4 dense (text-only) HF↔Megatron conversion selected the MoE parameter-mapping registry for dense models, producing unmapped params and crashing during weight loading.

## Changes

- src/megatron/bridge/models/gemma/gemma4_bridge.py:
  `mapping_registry()` falls back to `hf_config.enable_moe_block` when the stateful `_is_dense` flag is unset (MoE path unchanged).
- src/megatron/bridge/models/conversion/model_bridge.py:
  guard None conversion tasks in `_load_weights_hf_to_megatron` and `stream_weights_hf_to_megatron`.
- tests/unit_tests/models/test_model_bridge_load_weights.py:

  new red-green regression test.

## Verification
- Red-green: FAIL pre-fix, PASS after (nemo:26.06.rc9).
- Real roundtrip reproduced on origin/r0.5.0
  (RED: shape-mismatch crash) → GREEN on this branch:
  all tensors verified, HF ckpt written, 0 unmapped params.

## Reference
- NVBug 6367442

### DIFF
--- a/src/megatron/bridge/models/conversion/model_bridge.py
+++ b/src/megatron/bridge/models/conversion/model_bridge.py
@@ -1187,8 +1187,10 @@ class MegatronModelBridge(
             conversion_tasks = self.build_conversion_tasks(hf_pretrained, megatron_model)
 
         for task in conversion_tasks:
-            # None means megatron module not on current rank, skip if this task is not going to happen
-            if task.megatron_module is None:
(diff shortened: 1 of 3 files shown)


pruprakash:
This is already fixed, via #4414

PR #4651, opened 2026-07-05
fix(model): adapt MegatronMIMO grid to MCore expert-view API

ko3n1g:
## Background

Current Megatron-Core uses a dedicated `expert` view on `HyperCommGrid` for expert-parallel process groups. Megatron Bridge maintains its own MegatronMIMO grid builder, and that implementation had not been updated with the corresponding process-group contract.

The mismatch causes MegatronMIMO training to fail when MCore requests an expert-view group:

```text
KeyError: "View 'expert' is not registered. Registered views: ['base']"
```

This PR contains only the Bridge compatibility fix and its tests. The original `.main.commit`, MCore submodule, and `uv.lock` bump were removed.

## What changed

- Register the dense base view and the dedicated expert view in the MegatronMIMO grid builder.
- Preserve distinct expert-model, expert-tensor, and expert-data process-group semantics.
- Populate the MCore process-group collection with the required expert and distributed-optimizer groups.
- Update MegatronMIMO training and conversion bridges to use the current MCore field and parallel-state contracts.
- Add focused structural and contract tests for the expert-view topology, including `etp > 1`.

## Validation

- Focused and adjacent MegatronMIMO tests: 24 passed.
- Applicable provider tests: 29 passed; one unrelated CUDA-device test was excluded because the local environment could not provide that device path.
- Python compilation: passed.
- `git diff --check`: passed.
- GitHub review thread: resolved.

Full CI has been restarted for the dependency-cleanup SHA.


src/megatron/bridge/models/megatron_mimo/megatron_mimo_provider.py
```diff
@@ -296,10 +297,10 @@ def _get_pg_collections_from_grids(
                     dp=grid.get_pg(["dp"]),
                     pp=pp_group,
                     cp=grid.get_pg(["cp"]),
-                    ep=grid.get_pg(["ep"]),
+                    ep=grid.get_pg(["expt_tp"], view=_EXPERT_VIEW),
```
yaoyu-33: This doesn't look right.

yaoyu-33: Confirmed and fixed in eedec1bdff7ec81e4a00e2c50151472b65bb22ff. `ProcessGroupCollection.ep` now comes from `["ep"]` on the expert view, while `["expt_tp"]` is assigned separately to `expt_tp`; `tp_ep`, `tp_ep_pp`, `expt_dp`, and the matching pinned-MCore parallel-state globals are also wired explicitly. Focused contract and adjacent MIMO tests pass (24/24), and the applicable provider suite passes (29/29).

liding-nv (dismissed):

liding-nv (approved):

PR #4656, opened 2026-07-05
fix(training): pass training_log optional args by keyword to avoid misbinding

khazic:
## What does this PR do?

`train.py` passed `log_max_attention_logit` as the 15th positional argument of `training_log`, but the 15th parameter of `training_log` is `pg_collection`. The value therefore bound to the wrong slot:

- With `qk_clip` enabled, `clip_qk()` returns a float, which landed in `pg_collection`. The `pg_collection or get_pg_collection(model)` guard keeps a truthy float, so the first logging step crashes with `AttributeError: 'float' object has no attribute 'mp'` at `reduce_max_stat_across_model_parallel_group(..., mp_group=pg_collection.mp)`.
- Without `qk_clip`, the misbinding is masked (`None or get_pg_collection(model)` recovers), but the real `log_max_attention_logit` parameter always stays `None`, so the `max-attention-logit` metric is silently never written to TensorBoard/W&B/MLflow.

The sibling caller in `train_megatron_mimo.py` already passes everything by keyword and is unaffected.

## Changes

- Pass `pg_collection` and `log_max_attention_logit` by keyword at the `train.py` call site. Passing the in-scope `pg_collection` also removes a per-iteration `get_pg_collection(model)` recomputation inside `training_log`.
- Make the optional parameters of `training_log` keyword-only (`*` separator) so any future positional call fails immediately with a `TypeError` instead of misbinding.
- Add unit tests: the `max-attention-logit` metric reaches the writers when a value is provided, and a signature guard asserting the optional parameters stay keyword-only.

## Before submitting

- [x] Did you read the contributor guideline?

### DIFF
--- a/src/megatron/bridge/training/train.py
+++ b/src/megatron/bridge/training/train.py
@@ -610,7 +610,8 @@ def train(
                 global_state,
                 history_wct,
                 model,
-                log_max_attention_logit,
+                pg_collection=pg_collection,
+                log_max_attention_logit=log_max_attention_logit,
                 loaded_iteration=start_iteration,
                 # training_log recomputes seqlen/FLOPS from the (log-step) accumulators
                 # itself; pass None so it uses that path (the per-step seqlen_sum is no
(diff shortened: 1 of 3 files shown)


yaoyu-33 (approved):

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4656/commits/d8efe8918690fa1000c39eb4e71b2474d7611a22

PR #4657, opened 2026-07-05
fix(model): map router expert_bias in DeepSeek-family common mapping list

### LINKED ISSUE #4199: [bug]  expert_bias is lack in deepseek mode convert between hf and megatron format
### Problem

The parameter expert_bias exists not only in the MTP layer but also in other layers. However, in models/deepseek/common.py, this parameter is not added to the param_mappings in the get_common_mapping_list function.

### Minimal repro

```shell
param_mappings = {
        ......

        # MoE
        "decoder.layers.*.mlp.router.weight": "model.layers.*.mlp.gate.weight",
        "decoder.layers.*.mlp.experts.linear_fc2.weight*": "model.layers.*.mlp.experts.*.down_proj.weight",
        "decoder.layers.*.mlp.shared_experts.linear_fc2.weight": "model.layers.*.mlp.shared_experts.down_proj.weight",
        # expert bias
        "decoder.layers.*.mlp.router.expert_bias": "model.layers.*.mlp.gate.e_score_correction_bias",     # this line is needed.
       ......
    }
```

### Expected behavior

The model can be correctly converted between Hugging Face and Megatron formats.

### Affected area

area:model

### Regression?

No

### Environment

_No response_

### Logs

```shell

```

khazic:
## What does this PR do?

Fixes #4199

The DeepSeek-family shared mapping list (`get_common_mapping_list`) mapped `router.expert_bias <-> gate.e_score_correction_bias` only for MTP layers, not for the main decoder layers. Any bridge that consumes the common list and enables `moe_router_enable_expert_bias` therefore converted HF checkpoints with the router bias left at its zero init.

Kimi K2 was affected directly: `KimiK2Bridge` sets `moe_router_enable_expert_bias = True` but never appended the mapping itself, so after HF-to-Megatron conversion its noaux-tc top-k routing diverged from the HF checkpoint. DeepSeekV3, GLM-4.7-Flash, and Kimi K2.5 VL each carried a private copy of the same `AutoMapping` append to work around the gap.

## Changes

- Add `decoder.layers.*.mlp.router.expert_bias -> model.layers.*.mlp.gate.e_score_correction_bias` to the `param_mappings` dict in `models/deepseek/common.py`. The MTP generation loop derives the MTP variant from the same entry, so the previously hand-written MTP `AutoMapping` is removed (it generated an identical mapping).
- Drop the now-redundant per-bridge appends in `deepseek_v3_bridge.py`, `glm47_flash_bridge.py`, and `kimi_k25_vl_bridge.py`.
- Families without the bias (for example DeepSeek-V2) simply never match the wildcard, the same established pattern as the coexisting MLA and non-MLA query mappings in the same dict.

## Tests

- New: common-list lookup resolves decoder `expert_bias` in both directions for Kimi K2, and the MTP variant is still generated.
- Existing assertions in `test_glm47_flash_bridge.py` and `test_kimi_k25_vl_bridge.py` keep passing since the common list now supplies identical wildcard entries.

## Before submitting

- [x] Did you read the contributor guideline?

### DIFF
--- a/src/megatron/bridge/models/deepseek/common.py
+++ b/src/megatron/bridge/models/deepseek/common.py
@@ -44,6 +44,7 @@ def get_common_mapping_list(hf_config=None) -> list:
         "decoder.layers.*.mlp.linear_fc2.weight": "model.layers.*.mlp.down_proj.weight",
         # MoE
         "decoder.layers.*.mlp.router.weight": "model.layers.*.mlp.gate.weight",
+        "decoder.layers.*.mlp.router.expert_bias": "model.layers.*.mlp.gate.e_score_correction_bias",
         "decoder.layers.*.mlp.experts.linear_fc2.weight*": "model.layers.*.mlp.experts.*.down_proj.weight",
         "decoder.layers.*.mlp.experts.local_experts.*.linear_fc2.weight": "model.layers.*.mlp.experts.*.down_proj.weight",
         "decoder.layers.*.mlp.shared_experts.linear_fc2.weight": "model.layers.*.mlp.shared_experts.down_proj.weight",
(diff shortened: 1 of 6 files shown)


yaoyu-33 (approved):

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4657/commits/8a3cf0ca74eab9e9bc250ca481a83b0d1c46386b

PR #4658, opened 2026-07-06
fix(ckpt): remove all non-persistent checkpoints when most_recent_k is 0

khazic:
## What

`cleanup_old_non_persistent_checkpoint` computes the checkpoints to delete with:

```python
rm_iter_ckpts = sorted_iter_ckpts[:-leave_ckpt_num]
```

The caller passes `leave_ckpt_num=ckpt_cfg.most_recent_k` and only guards `most_recent_k > -1`, so `0` is an accepted value meaning "keep zero most-recent checkpoints" (delete all). But with `leave_ckpt_num == 0`, Python evaluates `sorted_iter_ckpts[:-0]` as `sorted_iter_ckpts[:0]`, i.e. an empty list, so **nothing is ever removed** and non-persistent checkpoints accumulate unboundedly, the exact opposite of the intended behavior. The "to be kept" log line has the same `[-0:]` (whole list) pitfall.

## Fix

Guard the zero case explicitly: when `leave_ckpt_num <= 0`, schedule all checkpoints for removal and keep none.

## Test

Added `test_cleanup_old_non_persistent_checkpoint_leave_zero` asserting all checkpoints are removed when `leave_ckpt_num=0`.

## Notes

Local unit-test execution isn't possible on this machine (megatron.core / CUDA deps); ruff check and format pass. Happy to trigger CI on request.

### DIFF
--- a/src/megatron/bridge/training/checkpointing.py
+++ b/src/megatron/bridge/training/checkpointing.py
@@ -1574,9 +1574,16 @@ def cleanup_old_non_persistent_checkpoint(
     sorted_iter_ckpts = sorted(iter_ckpts, key=lambda ckpt_name: int(ckpt_name.name[len(iter_prefix) :]))
     if not sorted_iter_ckpts:
         return
-    rm_iter_ckpts = sorted_iter_ckpts[:-leave_ckpt_num]
+    # `leave_ckpt_num == 0` means keep none: slice with `[:-0]` would be `[:0]` (empty),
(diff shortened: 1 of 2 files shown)


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4658/commits/5f6bebee57ad4f21305c7e28b82e7307d53da8e2

PR #4659, opened 2026-07-06
fix(data): return None for Energon test split instead of aliasing validation

khazic:
## What

`EnergonProvider.build_datasets` returned the validation dataloader for **both** the validation and test slots:

```python
return (
    iter(dataset.train_dataloader()),
    iter(dataset.val_dataloader()),   # <- test slot reuses validation
)
```

But `EnergonMultiModalDataModule.test_dataloader()` intentionally returns `None` and logs `"Multimodal dataloader test dataset split does not exist"` — there is no distinct test split.

Because a non-None test dataloader was returned, downstream:

```python
do_test = test_dataloader is not None and cfg.validation.eval_iters > 0
```

evaluates to `True`, so end-of-training test evaluation runs on the validation set and **reports validation metrics as test metrics**, hiding whether the model generalizes to a genuinely held-out split.

## Fix

Return `iter(dataset.test_dataloader())` if present, else `None` — honoring the datamodule's own contract. With `None`, `do_test` stays `False` and test evaluation is correctly skipped (the consumer in `data/loaders.py` already handles a `None` test dataloader).

## Test

Updated `test_init_and_build_datasets` to assert the test iterator is `None` and `test_dataloader()` is consulted; updated the shared mock helper to mirror the real datamodule (test split returns `None`).

## Notes

### DIFF
--- a/src/megatron/bridge/data/energon/energon_provider.py
+++ b/src/megatron/bridge/data/energon/energon_provider.py
@@ -69,8 +69,12 @@ class EnergonProvider(DatasetProvider):
             num_workers=self.num_workers,
             pg_collection=context.pg_collection,
         )
+        # EnergonMultiModalDataModule.test_dataloader() returns None (no distinct test split);
+        # honor that instead of aliasing the validation loader as a fake test set, which would
+        # otherwise report validation metrics as test metrics whenever eval_iters > 0.
+        test_dataloader = dataset.test_dataloader()
         return (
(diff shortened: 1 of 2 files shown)


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4659/commits/570fcb2b12e3a0fab3005c639c2b8a4b592eb22f

PR #4660, opened 2026-07-06
fix(data): mask MegatronMIMO padding by position, not pad_token_id value

khazic:
## What

`MegatronMIMODataset.__getitem__` built its loss mask like this:

```python
pad_token_id = self.tokenizer.pad_token_id or 0
...
loss_mask[labels == pad_token_id] = 0.0
```

Both providers set `tokenizer.pad_token = tokenizer.eos_token` when the tokenizer has no dedicated pad token:

```python
# hf_provider.py / mock_provider.py
if tokenizer.pad_token is None:
    tokenizer.pad_token = tokenizer.eos_token
```

This is the common case for Llama/GPT-2-style tokenizers, so `pad_token_id == eos_token_id`. Masking loss on `labels == pad_token_id` then zeroes the loss on **every genuine EOS token in the corpus**, not just the padding tail, so the model never learns to emit EOS / stop generating.

## Fix

`_tokenize_with_placeholders` now also returns the number of real (non-padding) tokens, and `__getitem__` masks padding **by position** (from the last real token onward, since `labels[i] = input_ids[i+1]`) instead of by `pad_token_id` value. Genuine EOS tokens inside the real region keep their loss.

## Test

Added `test_loss_mask_does_not_drop_genuine_eos`: a tokenizer with `pad_token_id == eos_token_id == 0` and a real EOS (id 0) mid-sequence. Asserts the EOS target position keeps `loss_mask == 1.0` and only padding targets are masked. (Under the old code that EOS position was incorrectly masked to 0.)

## Notes

Local unit-test execution isn't possible on this machine (megatron.core / CUDA deps); ruff check and format pass. The loss-mask arithmetic was verified independently.


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4660/commits/75225fc5422b5e571e3baef3cdd360eedda3a1ee

yashaswikarnati (approved):

huvunvidia:
All tests passed. Merging.

PR #4661, opened 2026-07-06
fix(model): use tanh-approximate GELU for Gemma-1 bridge

khazic:
## What

`GemmaBridge.provider_bridge` set the activation function to exact GELU:

```python
provider.activation_func = self.hf_to_megatron_activation("gelu")  # -> F.gelu (exact)
```

But Gemma runs **tanh-approximate** GELU (HF `gelu_pytorch_tanh`). The gemma-1 HF config carries the legacy `hidden_act="gelu"` field, yet:

- `GemmaModelProvider` **defaults** to `fast_gelu` (tanh approximation)
- `Gemma2Bridge` hardcodes `fast_gelu` for exactly this reason
- `Gemma3`/`Gemma4` providers all default to `fast_gelu`

Only the gemma-1 bridge overrode the provider default back to exact GELU. Running exact GELU on weights trained with the tanh approximation produces slightly wrong activations and logits.

## Fix

Map the gemma-1 activation to `fast_gelu`, matching `Gemma2Bridge`, the provider default, and the actual model semantics.

## Test

Extended `test_provider_bridge_basic_2b` to assert the bridged provider's `activation_func is fast_gelu`.

## Notes

### DIFF
--- a/src/megatron/bridge/models/gemma/gemma_bridge.py
+++ b/src/megatron/bridge/models/gemma/gemma_bridge.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from megatron.core.activations import fast_gelu
 from megatron.core.models.gpt.gpt_model import GPTModel
 from megatron.core.transformer.enums import AttnBackend
 from transformers import GemmaForCausalLM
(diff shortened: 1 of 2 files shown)


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4661/commits/740dc973e5790d708a53db99f68d9d37421546f8

PR #4662, opened 2026-07-06
fix(inference): reject pipeline parallelism in QwenVLInferenceWrapper

khazic:
## What

`setup_model_and_tokenizer` for VLM inference accepts a `pp` (pipeline) argument, but `QwenVLInferenceWrapper` only implements the tensor-parallel / no-pipeline forward path (`forward_pass_without_pipeline_parallel`), whose inputs are keyed on `"input_ids"` and carry the multimodal tensors (`pixel_values` / `image_grid_thw`).

When `pipeline_model_parallel_size > 1`, the base `run_one_forward_step` takes the pipeline branch:

```python
# AbstractModelInferenceWrapper.run_one_forward_step (PP branch)
tokens = inference_input["tokens"]   # <- this wrapper only provides "input_ids"
```

so decoding dies with a cryptic `KeyError: 'tokens'`. That path also reads only `tokens`/`position_ids`/`attention_mask` and calls the model positionally, dropping the multimodal inputs entirely — so it cannot run the VLM correctly even if the key existed.

## Fix

Override `run_one_forward_step` to fail fast with a clear `NotImplementedError` when `pipeline_model_parallel_size > 1`, and delegate to the base implementation otherwise. This turns a confusing mid-generation `KeyError` into an explicit, actionable message. (Full PP support for the VLM path would additionally require threading the multimodal inputs through the pipeline forward, which is out of scope here.)

## Test

Added:
- `test_run_one_forward_step_rejects_pipeline_parallel` — asserts `NotImplementedError` when `pipeline_model_parallel_size = 2`.
- `test_run_one_forward_step_delegates_without_pipeline_parallel` — asserts delegation to the base method when `= 1`.

## Notes

### DIFF
--- a/src/megatron/bridge/inference/vlm/qwenvl_inference_wrapper.py
+++ b/src/megatron/bridge/inference/vlm/qwenvl_inference_wrapper.py
@@ -40,6 +40,32 @@ class QwenVLInferenceWrapper(AbstractModelInferenceWrapper):
     def __init__(self, model, inference_context=None):
         super().__init__(model, inference_context=inference_context)
 
+    def run_one_forward_step(self, inference_input: Dict[str, Any], recv_buffer_seq_len: Optional[int] = None):
+        """Run a single forward pass, rejecting pipeline parallelism.
+
+        This wrapper only implements the tensor-parallel / no-pipeline path
+        (``forward_pass_without_pipeline_parallel``), whose inputs are keyed on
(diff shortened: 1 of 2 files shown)


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4662/commits/eeec8a23ceafda5e12873e8bbaaf24c054799b5f

PR #4682, opened 2026-07-06
feat(recipes): support recipe environment variables

yaoyu-33:
## What does this PR do?

Adds recipe-owned process environment defaults to `ConfigContainer` and makes each benchmark's effective process environment visible in the recipe that owns it.

- gives all 396 canonical flat performance recipe builders across 57 hardware modules an explicit `cfg.env_vars` mapping immediately before `return cfg`
- gives all 247 supported H100 model recipe builders across 41 modules the same explicit mapping structure; the remaining canonical Qwen3-Next PEFT stub still intentionally raises `NotImplementedError`
- keeps `COMMON_PERF_ENV_VARS` limited to the value shared by every flat performance workload and `COMMON_RECIPE_ENV_VARS` limited to the seven allocator/NCCL/cuDNN defaults shared by every supported model recipe
- writes model-, precision-, CUDA graph-, NCCL-, Transformer Engine-, and topology-specific values directly in the owning recipe with comments
- removes the environment decorators, generic runtime composition helpers, and `PerfEnvPlugin`
- preserves a narrow target-finalization adapter for the three HybridEP topology values because some H100 model recipe builders are reused on other hardware targets; explicit recipe/CLI values retain precedence
- resolves environment state in three stages: recipe base, argparse/Hydra overrides, then target finalization
- launches `bootstrap.py` on every rank; it resolves the recipe environment before importing the training stack, applies defaults without replacing explicit process values, and `exec`s exactly one training entrypoint (`run_script.py` or `run_recipe.py`)
- keeps deployment-owned orchestration, cache/offline, credential, scheduler, and fabric values in the executor
- keeps DeepSeek LayerNorm SM margins at `20` to preserve the established performance environment
- carries environment mappings through performance workload discovery and MLM-to-Bridge translation
- documents serialization, precedence, coupled overrides, and ownership boundaries

Cluster fabric settings remain deployment-owned. Recipe configs do not contain credentials, cache policy, scheduler metadata, or cluster-specific network tuning.

## Testing

- exhaustive AST coverage: 396 flat performance builders and 247 supported H100 model recipe builders, with no missing or malformed inline mappings
- legacy parity audit: all 396 flat builders preserve their effective process environment; model recipe common defaults, HybridEP topology, explicit launcher precedence, manual GC, and DeepSeek margin `20` remain equivalent
- invariant audits cover first-position common mappings, direct-return placement, paired settings, complete HybridEP topology, NVFP4, DeepSeek, and the two newly added MiniMax-M3 builders
- focused bootstrap, executor/topology, recipe, and mock-loader checks passed in independent review; Ruff, Python AST parsing, shell syntax, and `git diff --check` pass on the rebased head
- end-to-end CI validation has passed representative model-recipe and flat-performance entrypoints on three accelerator targets; the latest full PR CI rerun is in progress after rebasing onto current `main`

Recipe instantiation tests run in the standard PR container because the lightweight local environment does not contain a usable Transformer Engine binary.


yaoyu-33:
Validation update for `ed7eb781e98cb5bd9d7c394d6ef6c6fb0021b620`:

- Exhaustive flat-recipe parity: all 396 builders instantiated with real parser defaults; zero build errors, zero default `env_vars` identity changes, zero effective-env differences from the merge-base launcher plus `PerfEnvPlugin`, and zero manual-GC differences. Final manual GC is `true` with interval `100` for every flat recipe.
- Explicit legacy CLI coupling was compared against the merge base for TP, CP, MoE A2A, HybridEP EP, PP, NCCL-UB, deterministic mode, and explicit Hydra env precedence. Effective env matched in every case.
- Static library coverage is complete for all 245 builders. GB200 library dry run without an explicit EP override correctly produced NVL72 topology, ranks-per-domain from the final recipe EP, `USE_MNNVL=1`, and DeepSeek margins of 20.
- Real nemo-ci argument-builder and launcher dry runs were compared for representative H100, GB200, B300, VR200, and library paths. Normalized rank-local commands, node/task layout, segment, srun options, and mounts are unchanged. Legacy nemo-ci `v1` and `v2` labels now resolve to the canonical unsuffixed flat recipes; named variants remain suffixed.
- Actual worker dry runs passed for representative v2 H100, GB200, and B300 recipes and for the GB200 library path. Generated configs are semantically identical to the merge-base controls outside the newly explicit `env_vars` mapping and normalized output paths; flex/HybridEP selection and DeepSeek margin 20 are preserved.
- The expected launcher-text difference is only env placement: legacy sbatch exported recipe perf vars through the executor, while this PR applies those same values rank-locally before training-framework imports. The effective training-process env is unchanged.
- Validation passed: 87 expanded targeted unit tests, 40 final bootstrap/environment tests, full-repository pre-commit, and three independent review rounds. No PR-local blocker remains.

Non-blocking paired nemo-ci follow-up: GCP `KUBEFLOW_ENV_LIST_JSON` values still have launcher precedence. Current overlapping library values match the recipes, but that list should be narrowed before enabling flat perf recipes that intentionally need different values.

scripts/performance/run_recipe.py
```diff
@@ -167,8 +161,6 @@ def set_user_overrides(config, args):
         config.model.context_parallel_size = args.context_parallel_size
     if args.virtual_pipeline_model_parallel_size != -1:
         config.model.virtual_pipeline_model_parallel_size = args.virtual_pipeline_model_parallel_size
-    if args.expert_model_parallel_size:
```
yaoyu-33, line 170: I dont see why this need to change.

scripts/performance/utils/executors.py
```diff
@@ -37,16 +37,9 @@
 """
 
 PERF_ENV_VARS = {
```
yaoyu-33: rename to offline benchmark vars.

scripts/performance/setup_experiment.py
```diff
@@ -60,27 +61,85 @@
 logger.setLevel(logging.DEBUG)  # pin level so nemo_run's WARNING root doesn't suppress INFO
 
 
+def _configure_slurm_gpu_tuning(executor, *, enable_vboost: bool, lock_gpu_freq: int | None) -> None:
```
yaoyu-33: put in utils

scripts/performance/run_recipe.py
yaoyu-33: set_user_overrides now calls apply_library_argparse_overrides at the start, and that shared helper applies expert_model_parallel_size. The bootstrap pass uses the same helper before self-exec. Removing the second assignment avoids duplicate logic, keeps both passes identical, and preserves Hydra final precedence.

scripts/performance/utils/executors.py
yaoyu-33: Renamed the executor-only defaults to OFFLINE_BENCHMARK_ENV_VARS and updated the implementation comments and tests.

scripts/performance/setup_experiment.py
yaoyu-33: Moved the Slurm GPU tuning helper into scripts/performance/utils/utils.py, updated the launcher import and call, and moved the direct test to the utility.

yaoyu-33:
Launched a nemo-ci nightly validation pipeline against exact PR head `5f863b497261c438f4185212e23ea349d27e4b04`.

- Root pipeline: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/57880164
- MBridge module pipeline: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/57880245
- Suites: `L0,L1` (the selected flat perf template is currently classified as L1)
- EOS: `llama3_8b_8gpu_h100` — recipe / `run_recipe`
- Pretyche: `llama3_8b_8gpu_gb200` — GB200 topology/env / `run_recipe`
- Lyris: `llama3_8b_8gpu_gb300_fp8_mx_perf` — flat perf env / `run_script`

Lyris and Pretyche setup completed successfully. EOS setup is still waiting for a runner, so the pretrain child pipeline has not expanded the target training jobs yet. There is no launch, authentication, or configuration failure so far.

yaoyu-33:
Validation update for the recipe-owned environment migration:

- Core unit CI passed on the current head (`6de1da8e`).
- The three-target L0/script-system validation passed 3/3, covering `setup_experiment` through bootstrap into both `run_recipe` and `run_script`.
- The paired nemo-ci golden/known-issue refresh is in [nemo-ci !2681](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/merge_requests/2681).
- Final-branch DeepSeek V3 BF16 GB200 validation passed: [job 363934110](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/363934110), Slurm succeeded, 50/50 steps, no OOM, correlation 1.0000, final-loss delta 0, performance and memory checks passed. The full loss/grad-norm trajectory exactly matched the promoted run.
- Runtime artifacts confirm the intended effective environment: EP ranks/domain 64, NVLink domain 72, MNNVL enabled, and DeepSeek forward/backward LayerNorm SM margins preserved at 20. This also verifies parity with the legacy effective env rather than only source-level parity.
- Final-branch GB300 verification is still waiting for scheduler allocation: [job 363890636](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/363890636). It remains healthy/PENDING with no error or OOM marker.

I will post the terminal GB300 result here when it completes.

yaoyu-33:
Final validation update:

- Core unit CI passed on the current head (`6de1da8e`).
- The three-target L0/script-system validation passed 3/3, covering `setup_experiment` through bootstrap into both `run_recipe` and `run_script`.
- Paired nemo-ci refresh: [!2681](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/merge_requests/2681).
- DeepSeek V3 BF16 GB200 passed: [job 363934110](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/363934110), 50/50, no OOM, correlation 1.0000, exact final-loss parity, performance and memory checks green.
- DeepSeek V3 BF16 GB300 passed on the controlled retry: [job 364806360](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/364806360), 50/50, no OOM, exact final loss `7.155369`, convergence and memory green. Its evaluator window was `16,198.66 ms` / `1051.88 TFLOP/s/GPU`, `+0.72%` versus the refreshed golden and within the ±5% gate.
- Runtime artifacts on both targets confirm the intended effective environment: EP ranks/domain `64`, NVLink domain `72`, MNNVL enabled, and DeepSeek forward/backward LayerNorm SM margins preserved at `20`. This verifies runtime parity with the legacy effective env, not only source-level parity.

The first GB300 allocation completed training but showed two recoverable mid-run slowdown waves and failed the performance gate at `-15.57%`. A fresh-allocation retry with the exact same MB SHA, MCore SHA, image, config, env, loss trajectory, evaluator code, and window passed; neither wave recurred. We therefore classify the first result as a transient allocation/runtime performance anomaly, not a deterministic recipe-env, evaluator-formula, MB, or MCore regression.

No bootstrap, architecture, OOM, convergence, or environment-propagation issue remains in the final correct-architecture validation.

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4682/commits/b251a9fb8a70516c5157246a7f1dc6130307de4a

malay-nagda (approved):

dingqingy-nv (approved):

PR #4688, opened 2026-07-07
fix(checkpoint): preserve quantization recipe config

cuichenx:
## Summary

NVBug: 6419767

- preserve the complete MCore `RecipeConfig` state in `run_config.yaml`, including ordered matchers and quantization configs
- support serializable custom matchers and defer to a future upstream serializer
- load lossy legacy recipe stubs as `None` with a warning instead of failing checkpoint config reconstruction

## Root cause

`RecipeConfig` is a stateful, non-dataclass object without `to_cfg_dict()`. Existing serialization recorded only its class target and discarded the required `matchers` and `config_dict`, so loading tried to call `RecipeConfig()` without its required arguments.

## Compatibility

Existing affected checkpoints cannot recover the recipe state that was already discarded; they load with `quant_recipe=None`. Checkpoints saved with this change retain the full state and instantiate `RecipeConfig` and its matchers normally when loaded.

### DIFF
--- a/src/megatron/bridge/training/utils/checkpoint_utils.py
+++ b/src/megatron/bridge/training/utils/checkpoint_utils.py
@@ -505,7 +505,11 @@ def _sanitize_run_config_object(obj: Any) -> Any:
         target = obj.get("_target_")
         if isinstance(target, str) and target in _RUNTIME_ONLY_TARGETS:
             return None
-        if target == _RECIPE_CONFIG_TARGET and set(obj).issubset({"_target_", "_call_"}):
+        if (
+            target == _RECIPE_CONFIG_TARGET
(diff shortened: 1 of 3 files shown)


src/megatron/bridge/training/utils/checkpoint_utils.py
```diff
@@ -35,6 +35,7 @@
 
 logger = logging.getLogger(__name__)
 _RUNTIME_ONLY_TARGETS = frozenset({"megatron.core.timers.Timers"})
+_RECIPE_CONFIG_TARGET = "megatron.core.quantization.quant_config.RecipeConfig"
```
yaoyu-33, line 38: need frozen set?

PR #4686, opened 2026-07-07
fix(model): enable attention output-gate fallback for step35 head-wis…

pruprakash:
## Summary

Step-3.5-Flash HF→Megatron conversion failed with a linear_qkv shape mismatch (Expected [10240, 4096], Got [10304, 4096]) at layer 0. The published config sets `use_head_wise_attn_gate` but no `attention_output_gate`, and the gate-mode block only ever disabled the output gate — the documented fallback for MCore without native head-wise gate support was never enabled, so the model was built without gate rows while QKVGMapping merged g_proj in.

## Changes

- `src/megatron/bridge/models/stepfun/step35_bridge.py`: the gate-mode block now sets `attention_output_gate = not _mcore_supports_head_wise_attn_gate()` whenever head-wise gates are requested (native support → off, as before; no native support → fallback on). Note this intentionally normalizes `attention_output_gate` to a bool whenever head-wise gates are requested.
- `tests/unit_tests/models/stepfun/test_step35_bridge.py`: two new tests covering both MCore-support branches with the published config shape (`attention_output_gate` absent). These pin the new behavior; the pre-existing gate tests remain the old-behavior guards.

## Verification

- Red-green (nvcr.io/nvidian/nemo:26.06.rc9): RED on unpatched source — exactly the new distributed test fails (2 failed, 46 passed); GREEN on the branch — 48/48.
- E2E: `examples/models/stepfun/step35/conversion.sh` completes the full stepfun-ai/Step-3.5-Flash conversion on the branch (checkpoint saved), past the previous crash point.

## Reference

- NVBug 6270246

PR #4690, opened 2026-07-07
fix(model): make step35 layer-spec builder public so converted checkpoints load

pruprakash:
## Summary

Step-3.5-Flash inference from a converted checkpoint fails: with #4686 applied, `examples/models/stepfun/step35/conversion.sh` succeeds, but `examples/models/stepfun/step35/inference.sh` (with `MEGATRON_MODEL_PATH` set) crashes on all ranks before generation:

```text
InstantiationException: Instantiation of
'megatron.bridge.models.stepfun.step35_bridge._build_step35_layer_spec'
is not allowed because private target path segments are not supported.
full_key: transformer_layer_spec
```

Root cause: the bridge assigns a private function as `transformer_layer_spec`; checkpoint save serializes its dotted path into `run_config`, and the checkpoint loader's target validation rejects private path segments. This blocks anything that loads a converted Step-3.5 checkpoint — inference, finetuning resume, and `test_step35_inference_sh`. Every other bridge uses public spec builders.

## Changes

- `src/megatron/bridge/models/stepfun/step35_bridge.py`: rename `_build_step35_layer_spec` → `build_step35_layer_spec` (public).
- `src/megatron/bridge/models/stepfun/step37_bridge.py`, `modelling_step37/transformer_block.py`: same rename at import/usage sites (Step-3.7 shares the builder).
- `tests/unit_tests/models/stepfun/test_step35_bridge.py`: new test asserting the serialized spec target (module.qualname, as checkpoint save produces it) passes the checkpoint loader's target validation.

No backward-compat shim: Step-3.5 conversion could not complete before #4686, so no existing checkpoints carry the private target.

## Verification

- Red-green (nvcr.io/nvidian/nemo:26.06.rc9): RED — on unpatched code the real validator raises `InstantiationException` on the serialized target; GREEN — full `test_step35_bridge.py` passes on the branch.
- E2E: converted `stepfun-ai/Step-3.5-Flash` (with #4686), then `inference.sh` with `MEGATRON_MODEL_PATH` loads the Megatron checkpoint and generates coherent text.

## Reference

- NVBug 6270246. Companion: #4686 (conversion produces the checkpoint; this PR makes it loadable for inference).

### DIFF
--- a/src/megatron/bridge/models/stepfun/modelling_step37/transformer_block.py
+++ b/src/megatron/bridge/models/stepfun/modelling_step37/transformer_block.py
@@ -17,16 +17,16 @@
 Mirrors ``qwen_vl/modelling_qwen3_vl/transformer_block.py``: returns the
 per-layer ``TransformerLayerSubmodules`` spec consumed by Megatron's
 ``GPTModel``. Step3.7 reuses Step-3.5's hybrid full/sliding decoder layer
-(``_build_step35_layer_spec``) verbatim — the function lives in
+(``build_step35_layer_spec``) verbatim — the function lives in
 ``step35_bridge`` to keep all Step-3.5 spec-construction logic in one place.
 """
 
-from megatron.bridge.models.stepfun.step35_bridge import _build_step35_layer_spec
+from megatron.bridge.models.stepfun.step35_bridge import build_step35_layer_spec
 
 
 def get_step37_text_layer_spec(*args, **kwargs):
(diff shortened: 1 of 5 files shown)


yaoyu-33 (dismissed):

huvunvidia (approved):
LGTM!

qiaochuz-nv:
Retrospective QA-fix tracking issue: #6012. It records the fix authored by @pruprakash and delivered by this merged PR.

PR #4691, opened 2026-07-07
fix(misc): pass trust_remote_code when loading tokenizer from checkpoint in compare_text_generation

pruprakash:
## Summary

`examples/conversion/compare_text_generation.py` fails in `megatron_generate_from_checkpoint()` when loading the tokenizer from the just-imported checkpoint:

```text
ValueError: Checkpoint tokenizer config requested trust_remote_code=True.
Pass trust_remote_code=True to load_tokenizer() only if you trust the
checkpoint tokenizer code.
```

Root cause: #4502 added a trust gate to `load_tokenizer()`. This example reuses the checkpoint it imports with `trust_remote_code=True` (as all five of its other HF/bridge/tokenizer call sites do), so the checkpoint's tokenizer config always records the flag — and the one bare `load_tokenizer(checkpoint_path)` call then trips the gate unconditionally. This is the only bare `load_tokenizer` caller in the repo; it blocks the script's phase-3/4 comparison (generate from converted checkpoint, HF export round-trip).

## Changes

- `examples/conversion/compare_text_generation.py`: forward `trust_remote_code=True` to `load_tokenizer()`, matching the script's other call sites. The checkpoint being loaded is created moments earlier by the same script, so the trust decision was already made at import time.
- `tests/unit_tests/doc_consistency/test_compare_text_generation_consistency.py`: consistency test asserting the checkpoint tokenizer load opts into `trust_remote_code`.

## Verification

- Red-green (local, stdlib test): RED — fails on the unpatched call with the trust-gate message; GREEN — passes on the branch.
- E2E (nvcr.io/nvidian/nemo:26.06.01.rc0, EOS, meta-llama/Llama-3.2-1B): the full comparison pipeline completes; HF → Megatron-from-HF, Megatron-from-checkpoint, and HF-from-export all produce exactly matching ids/text (previously crashed at the tokenizer load).

## Reference

- NVBug 5991025 (the bug's earlier layers — eos_id AttributeError and BOS mismatch — were fixed by #2853 / #2889 in March; this addresses the regression from #4502 found while re-verifying on latest main).
```

### DIFF
--- a/examples/conversion/compare_text_generation.py
+++ b/examples/conversion/compare_text_generation.py
@@ -97,7 +97,9 @@ def validate_path(path: str, must_exist: bool = False) -> Path:
     return path_obj
 
 
-def transformers_generate(hf_model: str, prompt: str, max_new_tokens: int = 20) -> GeneratedData:
+def transformers_generate(
+    hf_model: str, prompt: str, max_new_tokens: int = 20, trust_remote_code: bool = False
+) -> GeneratedData:
     """
     Generate text from a HuggingFace model using transformers.
 
@@ -112,7 +114,7 @@ def transformers_generate(hf_model: str, prompt: str, max_new_tokens: int = 20) 
     print_rank_0(f"Loading model and tokenizer: {hf_model}")
 
     tokenizer = AutoTokenizer.from_pretrained(hf_model)
-    model = AutoModelForCausalLM.from_pretrained(hf_model, trust_remote_code=True).cuda()
+    model = AutoModelForCausalLM.from_pretrained(hf_model, trust_remote_code=trust_remote_code).cuda()
 
     print_rank_0("Generating text using HF weights...")
     in_tokens = tokenizer(prompt, return_tensors="pt").to(model.device)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: examples/conversion/compare_text_generation.py [hunk 1]
```diff
@@ -445,7 +445,9 @@ def get_iter_number(folder_name):
 
     # Initialize and load the model and tokenizer
     megatron_model = build_and_load_model(str(checkpoint_path), model_provider)
-    tokenizer = load_tokenizer(str(checkpoint_path))
+    # The checkpoint was imported with trust_remote_code=True (as all HF loads in
+    # this example are), so the tokenizer load must opt in as well.
+    tokenizer = load_tokenizer(str(checkpoint_path), trust_remote_code=True)
```
yaoyu-33: we cannot always pass True here, it's a security issue. We can pass in trust_remote_code as args if needed.

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4691/changes/1f440b7530d58f9877817c690dd81133d477a87e

examples/conversion/compare_text_generation.py
pruprakash: Done - replaced the hardcoded trust_remote_code=True with a --trust-remote-code flag (default off), threaded through all HF loads in the example. Remote code now runs only when explicitly opted in. E2E comparison test still passes on Llama-3.2-1B with the default.

liding-nv (approved):

qiaochuz-nv:
Retrospective QA-fix tracking issue: #6013. It records the fix authored by @pruprakash and delivered by this merged PR.

PR #4706, opened 2026-07-07
fix(ckpt): load fused MoE experts into sequential (non-grouped) layout

kevalmorabia97:
## What

Follow-up to https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4499

Qwen3.5 / Qwen3-VL routed experts are stored **fused** (`experts.gate_up_proj` / `down_proj`) on disk, but the sequential-expert mappings (`local_experts.*.linear_fc{1,2}`, used when `moe_grouped_gemm=False`, e.g. ModelOpt pruning) read **per-expert** HF keys that don't exist in a fused checkpoint. `build_conversion_tasks` then warns and skips them, leaving the pruned model's experts **randomly initialized**.

## Changes

- `qwen35_bridge.py`, `qwen3_vl_bridge.py`: switch the `local_experts.*.linear_fc{1,2}` mappings to `FusedGatedExpertMapping` / `FusedExpertMapping` (read the fused HF tensors).
- `model_bridge.py`: assemble the fused layout from per-expert weights in `maybe_modify_loaded_hf_weight`, covering checkpoints re-saved in the per-expert format.

## Testing

Validated in ModelOpt CI on `nemo:26.06`; needs CI on `main`.

### DIFF
--- a/src/megatron/bridge/models/qwen/qwen35_bridge.py
+++ b/src/megatron/bridge/models/qwen/qwen35_bridge.py
@@ -33,6 +33,7 @@ from megatron.bridge.models.conversion.param_mapping import (  # noqa: F401
     RMSNorm2ZeroCenteredRMSNormMapping,
 )
 from megatron.bridge.models.conversion.transformers_compat import full_attention_interval_from_hf
+from megatron.bridge.models.conversion.utils import moe_experts_stored_packed
 from megatron.bridge.models.gpt_provider import GPTModelProvider
 
 
@@ -101,6 +102,39 @@ def _apply_qwen35_moe_config(provider: GPTModelProvider, text_config) -> None:
     provider.moe_permute_fusion = True
 
 
+def _moe_routed_expert_mappings(hf_prefix, megatron_prefix, experts_packed, transpose_on_export=False):
+    """Routed-expert mappings for a MoE decoder, for both the grouped-GEMM and SequentialMLP layouts.
+
+    ``experts_packed`` selects fused HF tensors (``experts.gate_up_proj`` / ``experts.down_proj``)
+    vs per-expert (``experts.<i>.gate_proj`` / ``up_proj`` / ``down_proj``), matching the checkpoint so
+    HF->Megatron->HF round-trips. ``transpose_on_export`` applies to the fused mappings (e.g. Qwen3-VL).
+    """
+    grouped_fc1 = f"{megatron_prefix}decoder.layers.*.mlp.experts.linear_fc1.weight*"
+    grouped_fc2 = f"{megatron_prefix}decoder.layers.*.mlp.experts.linear_fc2.weight*"
+    seq_fc1 = f"{megatron_prefix}decoder.layers.*.mlp.experts.local_experts.*.linear_fc1.weight"
+    seq_fc2 = f"{megatron_prefix}decoder.layers.*.mlp.experts.local_experts.*.linear_fc2.weight"
+    if experts_packed:
+        gate_up = f"{hf_prefix}layers.*.mlp.experts.gate_up_proj"
+        down = f"{hf_prefix}layers.*.mlp.experts.down_proj"
+        return [
+            FusedGatedExpertMapping(
+                megatron_param=grouped_fc1, hf_param=gate_up, transpose_on_export=transpose_on_export
+            ),
+            FusedExpertMapping(megatron_param=grouped_fc2, hf_param=down, transpose_on_export=transpose_on_export),
+            FusedGatedExpertMapping(megatron_param=seq_fc1, hf_param=gate_up, transpose_on_export=transpose_on_export),
+            FusedExpertMapping(megatron_param=seq_fc2, hf_param=down, transpose_on_export=transpose_on_export),
+        ]
+    gate = f"{hf_prefix}layers.*.mlp.experts.*.gate_proj.weight"
+    up = f"{hf_prefix}layers.*.mlp.experts.*.up_proj.weight"
+    down = f"{hf_prefix}layers.*.mlp.experts.*.down_proj.weight"
+    return [
+        GatedMLPMapping(megatron_param=grouped_fc1, gate=gate, up=up),
+        AutoMapping(megatron_param=grouped_fc2, hf_param=down),
+        GatedMLPMapping(megatron_param=seq_fc1, gate=gate, up=up),
+        AutoMapping(megatron_param=seq_fc2, hf_param=down),
+    ]
+
+
 @MegatronModelBridge.register_bridge(source=Qwen3_5MoeForCausalLM, target=GPTModel, model_type="qwen3_5_moe_text")
 class Qwen35MoEBridge(MegatronModelBridge):
     """
@@ -132,7 +166,7 @@ class Qwen35MoEBridge(MegatronModelBridge):
     """
 
     @staticmethod
-    def _get_moe_lm_mappings(hf_prefix="model.", megatron_prefix=""):
+    def _get_moe_lm_mappings(hf_prefix="model.", megatron_prefix="", experts_packed=False):
         """Get language model parameter mappings for MoE Qwen3.5.
 
         Args:
@@ -140,6 +174,9 @@ class Qwen35MoEBridge(MegatronModelBridge):
                 for LM and "model.language_model.layers.*" for VL models.
             megatron_prefix: Prefix for Megatron param names. Use "" for LM
                 (default) and "language_model." for VL models.
+            experts_packed: Whether the routed experts are stored fused (``experts.gate_up_proj`` /
+                ``down_proj``) rather than per-expert (``experts.<i>.gate_proj`` / ``up_proj`` /
+                ``down_proj``). Selected from the checkpoint layout by ``mapping_registry``.
 
         Returns:
             List of mapping objects for the MoE LM portion.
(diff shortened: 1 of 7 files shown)


kevalmorabia97:
Thanks for the review! Addressed the flagged gap in 306b995:

- **VLM test guard** (the silent-regression risk): added `TestQwen35VLMoEBridgeMappingRegistry::test_mapping_registry_sequential_experts_are_fused`, mirroring the LLM guard — it asserts the VLM `local_experts.*.linear_fc1/2` mappings are `FusedGatedExpertMapping`/`FusedExpertMapping` reading `gate_up_proj`/`down_proj`. Verified against the real `Qwen35VLMoEBridge` registry.

On the suggested `test_model_bridge.py::maybe_modify_loaded_hf_weight` case: the assembly itself (numeric ordering, gate-above-up, empty-state KeyError) is covered by `test_fuse_per_expert_hf_weight`; the remaining bit is the thin try/except dispatch (suffix match + re-raise). Happy to add a dedicated test for that wrapper if you'd like it as well.

kevalmorabia97:
/ok to test [306b995](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4706/commits/306b995a558ebb69e880654a8c0c7e2a47a11f8c)

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -874,7 +903,13 @@ def maybe_modify_loaded_hf_weight(
             The loaded weights.
         """
         if isinstance(hf_param, str):
-            hf_weights = hf_state_dict[hf_param]
+            try:
```
yaoyu-33: This looks too hacky could cause unexpected fallback. There is no way to know if it's fused or not?

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -323,6 +323,35 @@ def _update_grouped_expert_number(param_name: str, param_type: str) -> str:
...
+
+    Routed experts are stored either fused (``experts.gate_up_proj`` / ``experts.down_proj`` of shape
+    ``[num_experts, ...]``) or per-expert (``experts.{i}.gate_proj`` / ``up_proj`` / ``down_proj``),
+    depending on the transformers version that saved the checkpoint. The expert mappings expect the
+    fused layout, so for a per-expert checkpoint we stack the per-expert weights into the fused tensor.
+    Raises ``KeyError`` if per-expert weights are also absent.
+    """
+    base, proj = hf_param.rsplit(".", 1)  # base = "...experts", proj = "gate_up_proj" | "down_proj"
+    per_expert_re = re.compile(rf"^{re.escape(base)}\.(\d+)\.")
+    expert_ids = sorted({int(m.group(1)) for k in hf_state_dict.keys() if (m := per_expert_re.match(k))})  # noqa: SIM118
```
yaoyu-33: can we make a util to get id from hf param?

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -323,6 +323,35 @@ def _update_grouped_expert_number(param_name: str, param_type: str) -> str:
...
+        raise KeyError(hf_param)
+    if proj == "gate_up_proj":
+        # Fused gate_up: [num_experts, 2 * ffn, hidden], gate stacked above up per expert.
+        return torch.stack(
+            [
+                torch.cat(
+                    [hf_state_dict[f"{base}.{i}.gate_proj.weight"], hf_state_dict[f"{base}.{i}.up_proj.weight"]],
+                    dim=0,
+                )
+                for i in expert_ids
```
yaoyu-33: [High] Preserve expert IDs instead of compacting gaps

`expert_ids` becomes the stack order, but `FusedGatedExpertMapping`/`FusedExpertMapping` later index this tensor with the numeric ID extracted from the Megatron parameter. For IDs `{0, 2}`, slot 1 contains expert 2; a two-expert target therefore silently loads expert 2 into expert 1. Please validate a zero-based contiguous set (and the required projection keys) before stacking, or build a representation that preserves the IDs. The added test only exercises a complete `0..11` set.

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -874,7 +903,13 @@ def maybe_modify_loaded_hf_weight(
             The loaded weights.
         """
         if isinstance(hf_param, str):
-            hf_weights = hf_state_dict[hf_param]
+            try:
+                hf_weights = hf_state_dict[hf_param]
+            except KeyError:
+                # MoE experts stored per-expert (some transformers versions) rather than fused; assemble.
+                if hf_param.endswith((".experts.gate_up_proj", ".experts.down_proj")):
+                    return _fuse_per_expert_hf_weight(hf_param, hf_state_dict)
```
yaoyu-33: [High] Apply this fallback to streaming imports too

This hook is used by `load_weights_hf_to_megatron`, but `stream_weights_hf_to_megatron` still directly evaluates `hf_state_dict[task.mapping.hf_param]` (lines 1226-1230). With these new fused mappings and a per-expert checkpoint, the public streaming path requests the absent `gate_up_proj`/`down_proj` key and raises `KeyError`. Please route streaming loads through `maybe_modify_loaded_hf_weight` as well (with the same grouped cache if needed), or resolve the checkpoint layout before creating the mappings.

src/megatron/bridge/models/qwen/qwen35_bridge.py
```diff
@@ -242,17 +242,17 @@ def _get_moe_lm_mappings(hf_prefix="model.", megatron_prefix=""):
...
-                GatedMLPMapping(
+                # Sequential (non-grouped) experts, used when moe_grouped_gemm is disabled
+                # (e.g. ModelOpt pruning). The decoder stores experts fused (gate_up_proj /
+                # down_proj); per-expert-stored checkpoints are assembled into the fused layout
+                # by maybe_modify_loaded_hf_weight before slicing.
+                FusedGatedExpertMapping(
                     megatron_param=f"{megatron_prefix}decoder.layers.*.mlp.experts.local_experts.*.linear_fc1.weight",
-                    gate=f"{hf_prefix}layers.*.mlp.experts.*.gate_proj.weight",
-                    up=f"{hf_prefix}layers.*.mlp.experts.*.up_proj.weight",
+                    hf_param=f"{hf_prefix}layers.*.mlp.experts.gate_up_proj",
```
yaoyu-33: [High] Preserve the source layout on export

For a per-expert source, this imports only because of the new fallback, but `FusedGatedExpertMapping`/`FusedExpertMapping` always group export under these fused `hf_param` keys. `SafeTensorsStateSource.save_generator(strict=True)` validates against the original sharding map (`experts.{i}.gate_proj.weight`, etc.), so HF→Megatron→HF raises on the unexpected fused key; `strict=False` skips it and omits the experts. The same happens to whichever projection is per-expert in a mixed-layout checkpoint. Please select mappings from the detected source layout or expand grouped output back to the expected per-expert keys.

src/megatron/bridge/models/conversion/model_bridge.py
kevalmorabia97: Agreed — removed this fallback entirely. `mapping_registry()` now inspects the checkpoint keys up front via the new `moe_experts_stored_packed()` helper and selects fused vs per-expert mappings accordingly, so there's no guessing (or hacky fallback) at load time.

src/megatron/bridge/models/conversion/model_bridge.py
kevalmorabia97: Moot in the reworked approach — the assembly helper and its ID extraction are gone. Per-expert checkpoints are now read directly by per-expert mappings (`experts.*.gate_proj.weight`), so no ID needs to be parsed out of the fused param.

src/megatron/bridge/models/conversion/model_bridge.py
kevalmorabia97: Fixed by dropping the assembly path. A per-expert checkpoint now uses `GatedMLPMapping`/`AutoMapping` that index HF params by the actual expert ID (`experts.*.gate_proj/up_proj/down_proj`), so there is no stacking and no gap-compaction — non-contiguous ID sets can't be mislabeled.

src/megatron/bridge/models/conversion/model_bridge.py
kevalmorabia97: Resolved — there is no fallback now, so `maybe_modify_loaded_hf_weight` is back to a plain lookup. Because mappings are selected to match the on-disk layout before loading, the streaming path (`stream_weights_hf_to_megatron`, which reads `hf_state_dict[hf_param]` directly) requests the keys that actually exist; a per-expert checkpoint never asks for `gate_up_proj`/`down_proj`.

src/megatron/bridge/models/qwen/qwen35_bridge.py
kevalmorabia97: Resolved — mappings now mirror the source layout. A per-expert source uses per-expert `hf_param` keys, so `SafeTensorsStateSource.save_generator(strict=True)` validates on export and HF→Megatron→HF round-trips for both layouts (and the mixed-layout case raises early in `moe_experts_stored_packed`). Verified the detection on a real fused Qwen3.5-MoE checkpoint in-container.

kevalmorabia97:
Reworked to address the review (force-pushed, rebased on latest `main`).

**What changed:** dropped the per-expert→fused assembly fallback in `maybe_modify_loaded_hf_weight` and instead detect the checkpoint's expert layout up front. New `moe_experts_stored_packed()` (in `conversion/utils.py`) inspects the HF keys; `mapping_registry()` then emits **fused** (`FusedGatedExpertMapping`/`FusedExpertMapping`) or **per-expert** (`GatedMLPMapping`/`AutoMapping`) mappings for both the grouped-GEMM and SequentialMLP layouts, via a shared `_moe_routed_expert_mappings()` helper reused by the Qwen3.5 / Qwen3.5-VL / Qwen3-VL MoE bridges.

**Why this resolves the 5 comments:** mappings now match the on-disk layout, so (1) no hacky fallback / no guessing, (2) no ID extraction or gap-compaction, (3) streaming works (correct keys requested directly), and (4) export round-trips under `strict=True`. Mixed-layout checkpoints raise early.

Verified detection + mapping selection on a real fused Qwen3.5-MoE checkpoint in-container (this transformers version stores experts fused). Lint/format clean.

kevalmorabia97:
/ok to test [dfe5e18](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4706/commits/dfe5e18574cfe4f5b645f113612963f574759356)

kevalmorabia97:
/ok to test [afcb6ae](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4706/commits/afcb6aeb90a71627d1e0ef2aaf088a016888c2b7)

yaoyu-33 (approved):

PR #4707, opened 2026-07-07
feat(distill): defer KD conversion to a pre-wrap hook and support submodule distillation for VLMs

kevalmorabia97:
## What

Two related changes to `DistillationProvider`:

1. **Defer the KD conversion.** `mtd.convert` moves out of `provide()` into a `_convert_hook` pre-wrap hook that runs *after* the student's weights are loaded. This lets a caller restore a quantized student (QAD) via an earlier prepended pre-wrap hook, and lets a weight-loaded submodule be extracted before it is wrapped with the teacher.

2. **`distill_submodule` option.** VLMs can distill only the `language_model` submodule while the rest of the model (vision encoder, projector) is trained and exported unchanged. The full built model is retained on `full_model` for export.

Replaces the `student_pre_conversion_hook` parameter, which is superseded by the `register_pre_wrap_hook(fn, prepend=True)` pattern (no external callers).

## Testing

Unit test updated for the deferred flow. Validated end-to-end in ModelOpt CI on `nemo:26.06`.


src/megatron/bridge/models/distillation_provider.py
```diff
@@ -163,7 +189,15 @@ def convert_to_distillation_provider(
...
+    student_provider.distill_submodule = distill_submodule
     student_provider.__post_init__()
 
+    # Convert after the bridge's weight-load hook (appended => runs last), so the student is fully
+    # weight-loaded before it is wrapped with the teacher. Set _pre_wrap_hooks via object.__setattr__
+    # (not register_pre_wrap_hook) to bypass __setattr__'s teacher-mirroring: when the student starts
+    # with no hooks (e.g. QAD builds it with load_weights=False), the mirror would share the hook list
+    # with the teacher, so building the teacher inside _convert_hook would recurse into _convert_hook.
+    hooks = [*getattr(student_provider, "_pre_wrap_hooks", []), student_provider._convert_hook]
+    object.__setattr__(student_provider, "_pre_wrap_hooks", hooks)
```
AAnoosheh, line 206: This is a crazy hack 😅

Not opposed, just never seen anything like it.

src/megatron/bridge/models/distillation_provider.py
```diff
@@ -72,37 +75,55 @@ def __post_init__(self):
...
             "criterion": kd_cfg.criterion,
             "loss_balancer": kd_cfg.loss_balancer,
         }
+        # ``mtd.convert`` mutates in place, so for the submodule case ``full_model`` already holds the
+        # distilled submodule for export.
         kd_model = mtd.convert(student_model, mode=[("kd_loss", modelopt_cfg)])
         mtd_mcore.adjust_distillation_model_for_mcore(kd_model, kd_cfg)
-
-        return kd_model
+        return [kd_model]
```
AAnoosheh, line 130: Should we keep it non-list (i.e. `return kd_model`) and make `_convert_hook()` accept a non-list too for simplicity? (unless something else in between expects a list)

kevalmorabia97: The pre-wrap-hook contract is `list[MegatronModule] -> list[MegatronModule]`: `register_pre_wrap_hook`'s composed hook does `model = hook(model)` over the model-chunks list, so `_convert_hook` has to accept and return a list to fit that interface. The single-chunk assumption is enforced by the `len(model_chunks) == 1` assert at the top.

src/megatron/bridge/models/distillation_provider.py
kevalmorabia97: Yeah 😄 — the `object.__setattr__` is deliberate. `DistillationProvider.__setattr__` mirrors every attribute onto the teacher, so a plain `self._pre_wrap_hooks = [...]` would share the hook list with the teacher; then building the teacher inside `_convert_hook` re-runs the hook and recurses. `object.__setattr__` sets it on the student only, breaking that loop.

src/megatron/bridge/models/distillation_provider.py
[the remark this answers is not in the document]
kevalmorabia97: Addressed in bf08c92: added `assert getattr(self.full_model, self.distill_submodule) is kd_model` to make the in-place `mtd.convert` contract explicit, reworded the docstring to "exported unchanged (only the submodule is trained)", and added unit tests for the `distill_submodule` branch and the QAD prepend-hook ordering.

kevalmorabia97:
Thanks for the reviews! Addressed in bf08c92:

- **In-place mutation contract** (@claude): added `assert getattr(self.full_model, self.distill_submodule) is kd_model` so a regression in `mtd.convert`'s in-place behavior fails loudly instead of silently exporting the un-converted submodule.
- **Docstring wording** (@claude): the rest of the model is now described as "exported unchanged (only the submodule is trained)".
- **Test coverage** (@claude): added unit tests for (a) the `distill_submodule` branch — `full_model` retained, submodule KD model returned — and (b) QAD prepend-hook ordering (`_convert_hook` is appended last, so a `register_pre_wrap_hook(fn, prepend=True)` runs before it).
- Replied inline to @AAnoosheh's list-return and `object.__setattr__` questions.

Note: the new unit tests can't run on the nemo:26.06 container (its Megatron-Bridge predates `distill_submodule`), but I validated both against this branch's code by prepending its `src/` on `PYTHONPATH` — both pass; they'll run in CI on `main`.

AAnoosheh (dismissed):

kevalmorabia97:
/ok to test [bf08c92](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4707/commits/bf08c922be3d23f45c2ac5627c100faf38870f29)

kevalmorabia97:
/ok to test [2487ea2](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4707/commits/2487ea2d42a13a258a5130bb401fc0561e246c68)

kevalmorabia97:
/ok to test [94d8dc4](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4707/commits/94d8dc4850c6552369fbb44c6df2059a5b85aaf6)

AAnoosheh (approved):

PR #4721, opened 2026-07-07
[peft] fix: support TE 2.17 grouped expert LoRA

cuichenx:
## What

- Make `GroupedExpertLinearAdapter` compatible with the Transformer Engine 2.14 through 2.17 private `_GroupedLinear` call layouts.
- Detect explicit split arguments and workspace outputs independently, then pass the matching non-tensor tuple and return only the output tensor.
- Add a behavioral CUDA regression that exercises normal adapter training, gradients, and no-grad inference while creating the real TE grouped-linear helper as a byproduct.

## Why

Transformer Engine 2.17 moved `m_splits` out of `non_tensor_args` into a dedicated `_GroupedLinear.forward` parameter. Megatron-Bridge called that private function using the older layout, which shifted every subsequent argument. The same unexercised path also omitted a TE 2.15/2.16 tuple field and propagated `(output, workspaces)` instead of the output tensor.

An independent review also caught that TE 2.14 uses a helper-module field and returns a bare tensor. The feature detection preserves that legacy contract while handling the workspace-based TE 2.15/2.16 contract and explicit-splits TE 2.17 contract.

This was discovered while validating the TE bump in #4696, but the fix is independent and can land before that bump.

## Validation

- `uv run python -m pytest tests/unit_tests/peft/test_utils.py` — 71 passed
- Real TE 2.16 CUDA grouped-linear adapter forward, backward, and no-grad inference — passed
- Exact TE 2.14 and 2.17 source contracts — inspected

The test runs against the TE version installed in CI rather than duplicating private signatures in test code, so it will naturally cover TE 2.17 when the dependency bump is applied.

### DIFF
--- a/src/megatron/bridge/peft/utils.py
+++ b/src/megatron/bridge/peft/utils.py
@@ -12,10 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import inspect
 import logging
 import math
 import re
 from dataclasses import dataclass, fields
+from functools import cache
 from importlib import import_module
 from importlib.metadata import version
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


cuichenx:
Superseded by #4772. We are migrating the TE 2.16/2.17 compatibility dispatch there only for FP8, while keeping BF16 on the public grouped-MM path and documenting removal once Transformer Engine provides a public functional grouped-linear API.

PR #4727, opened 2026-07-08
build(deps): constrain tokenizers below 0.23 so CLIP tokenizers construct

pruprakash:
## Summary

FLUX inference crashes on every rank before generation, in
`FluxInferencePipeline.load_text_encoders` when constructing the CLIP tokenizer:

    TypeError: RobertaProcessing.__new__() got an unexpected keyword argument 'cls'

Root cause: `[tool.uv]` sets `prerelease = "allow"` and `tokenizers` had no upper
bound, so the lock resolved **`tokenizers 0.23.0rc0`** — a release candidate that
renames `RobertaProcessing`'s `cls` keyword to `cls_token`. transformers (including
current main) still passes `cls=`, so any `CLIPTokenizer` construction raises at
`__init__`. This blocks FLUX inference and any other path that builds a CLIP tokenizer.

## Changes

- `pyproject.toml`: add `tokenizers>=0.22.0,<0.23` to `[tool.uv] override-dependencies`
  (alongside the existing onnx / langchain pins), resolving tokenizers to **0.22.2**.
- `uv.lock`: regenerated — `tokenizers 0.23.0rc0 → 0.22.2`.
- `tests/unit_tests/utils/test_tokenizers_compat.py`: regression guard asserting the
  installed `RobertaProcessing` accepts the exact kwargs transformers' `CLIPTokenizer`
  passes (`sep`, `cls`, `add_prefix_space`, `trim_offsets`) — fails fast if the lock ever
  drifts back to a renaming build.

## Why this layer (not the prerelease policy or transformers)

- `prerelease = "allow"` is global and intentional (other deps here depend on prerelease /
  internal wheels); disabling it would churn unrelated resolutions. Scoping the fix to the
  one offending package via `override-dependencies` is the surgical, repo-idiomatic move.
- The cap is `<0.23`, not `!=0.23.0rc0`: the `cls → cls_token` rename lives in the whole
  0.23 line, so excluding only the rc would let a future stable `0.23.0` reintroduce the
  crash. Lift the bound once transformers adopts the renamed kwarg.
- transformers is the component behind the API change, but it can't be fixed from this repo;
  pinning tokenizers to the compatible line is the correct lock-side fix.

## Verification

- **Dep-compat (red-green):** `RobertaProcessing(sep=…, cls=…, add_prefix_space=…, trim_offsets=…)`
  — RED on `0.23.0rc0` (`TypeError: unexpected keyword argument 'cls'`); GREEN on `0.22.2`
  (constructs). This is exactly what the new unit test asserts.
- **E2E (`nvcr.io/nvidian/nemo:26.06.01.rc0`, EOS):** FLUX inference — the failing artifact —
  runs the full HF→Megatron convert → inference → denoise pipeline to completion and writes an
  image with tokenizers `0.22.2` (the version this constraint resolves to); it crashed at the
  CLIP tokenizer load before. (The shipped container venv still bakes `0.23.0rc0`; the lock fix
  takes effect once a container rebuilds from the updated lock — the e2e confirms `0.22.2` is
  the correct, working target.)

## Reference

- NVBug 6317367 (prior conversion-step layer: 6169776).


cuichenx (dismissed):

cuichenx:
Upstream compatibility follow-up:

Transformers support for tokenizers 0.23.1 is **not released and is not on current main yet**. The adaptation is in open [huggingface/transformers#46381](https://github.com/huggingface/transformers/pull/46381); the specific CLIP/Roberta positional-argument change is [commit 876dc367](https://github.com/huggingface/transformers/commit/876dc3674556f744f2ba63aac2a8dd08dd51511e). The PR currently proposes `tokenizers>=0.23.1,<0.24.0`.

Verified with isolated construction probes:

| Transformers | tokenizers | Result |
|---|---:|---|
| 5.8.1 (current Bridge pin) | 0.22.0 | Incompatible resolver constraints through `huggingface-hub` |
| 5.8.1 | 0.22.1 | `CLIPTokenizer()` passes |
| 5.8.1 | 0.22.2 | `CLIPTokenizer()` passes |
| 5.8.1 | 0.23.0rc0 | Import passes; `CLIPTokenizer()` fails on unexpected `cls` |
| 5.8.1 | 0.23.1 | Rejected by Transformers metadata and API-incompatible |
| 5.13.0 (latest release checked) | 0.23.0rc0 | Same `cls` construction failure |
| 5.13.0 | 0.23.1 | Unsatisfiable from package metadata |
| #46381 head | 0.23.1 | `CLIPTokenizer()` passes |

This supports the temporary global `tokenizers<0.23` override for Bridge's current `transformers>=5.8.1,<5.9.0` range. It should be lifted atomically with a future Transformers bump that actually contains the upstream adaptation.

qiaochuz-nv:
Retrospective QA-fix tracking issue: #6014. It records the fix authored by @pruprakash and delivered by this merged PR.

PR #4736, opened 2026-07-08
[training] fix: Run eval callbacks in evaluation mode

yaoyu-33:
## What does this PR do?

Runs validation and test lifecycle callbacks while model chunks are actually in evaluation mode.

### Supported trigger and impact

Any `pretrain()` run that registers `on_eval_start`, `on_eval_end`, `on_test_start`, or `on_test_end` reaches this path during interval validation or final validation/test. The public callback contract says start hooks run after `model.eval()` and end hooks run before `model.train()`.

Previously, start hooks fired before `evaluate()` switched modes, while end hooks fired after `evaluate()` restored training mode. Callback inference or dropout-sensitive work therefore ran with training semantics, and `module.training` contradicted the documented lifecycle contract.

### Root cause and fix

Lifecycle hooks were owned by `evaluate_and_print_results()`, but the model-mode transitions are owned by `evaluate()`. This moves only the four lifecycle hook calls next to those transitions:

- start hooks fire immediately after all model chunks enter eval mode;
- end hooks fire after loss aggregation and before model chunks return to train mode.

Eval/test event selection, per-step hooks, reduced `total_loss_dict`, and the existing time-limit return behavior are preserved.

### Regression evidence

A deterministic CPU contract reproducer loaded and executed the real `evaluate_and_print_results()` and `evaluate()` functions with zero eval iterations:

```text
MB_WORKTREE=<checkout> uv run --no-project python <contract-reproducer>
```

- Base: exit 1 — `eval lifecycle modes were [('start', True), ('end', True)]`
- Fixed: exit 0 — `eval/test lifecycle callbacks observed evaluation mode`

The existing functional callback test now asserts evaluation mode for all validation/test lifecycle hooks and training mode for training hooks as a negative control.

Validation:

```text
uv run --no-project --with pytest python -m pytest --confcutdir=tests/unit_tests/training -p callback_loader tests/unit_tests/training/test_callbacks.py -q
# 42 passed

git diff --check
# passed

uv run --no-project --with pre-commit pre-commit run --all-files
# all hooks passed
```

### Scope

No checkpoint, training-step accounting, model conversion, dependency, workflow, public signature, or MCore code changes are included. The GPU functional test was not run locally; the ordering defect was exercised with the deterministic CPU harness. No full-suite, convergence, performance, or model-quality validation is claimed.

### DIFF
--- a/src/megatron/bridge/training/eval.py
+++ b/src/megatron/bridge/training/eval.py
@@ -93,6 +93,8 @@ def evaluate(
             - timelimit_hit: Boolean indicating if the time limit was reached.
     """
     # Determine callback event names based on whether this is test or eval
+    start_event = "on_test_start" if is_test else "on_eval_start"
+    end_event = "on_test_end" if is_test else "on_eval_end"
     step_start_event = "on_test_step_start" if is_test else "on_eval_step_start"
     step_end_event = "on_test_step_end" if is_test else "on_eval_step_end"
     # Prepare forward_step_func (check signature and inject state if needed)
@@ -105,6 +107,16 @@ def evaluate(
     # Turn on evaluation mode which disables dropout.
     for model_module in model:
         model_module.eval()
+
+    if should_fire(callback_manager, start_event):
+        callback_manager.fire(
+            start_event,
+            CallbackContext(
+                state=state,
+                model=model,
+                user_state=callback_manager.user_state,
+            ),
+        )
 
     # Retrieve process group collection and model config from the model
     # Use injected pg_collection if provided, otherwise extract from model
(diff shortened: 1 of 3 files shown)


yaoyu-33:
Diagnosis: the two core-unit failures were stale wrapper-level assertions, not a removed time-limit contract. Callback ownership moved into `evaluate()`, while these tests mocked `evaluate()` and therefore bypassed the real `should_fire()` calls. The real path still fires the start event after `model.eval()`; a time-limited incomplete evaluation returns without firing the end event, as before.

Updated coverage now verifies callback-manager delegation from `evaluate_and_print_results()`, eval/test lifecycle callbacks observing evaluation mode, train-mode restoration after completed evaluation, and start-without-end behavior on the time-limit path.

Verification at `3606dc14a51e90cef257e325dd430e86325b235d`:
- `uv run python -m pytest tests/unit_tests/training/test_eval.py tests/unit_tests/training/test_callbacks.py -q` — 47 passed
- `uv run pre-commit run --all-files` — all hooks passed

cuichenx (approved):

PR #4750, opened 2026-07-08
fix(training): prevent VPP FLOPS metadata overcounting

cuichenx:
## Summary

- accumulate FLOPs sequence/vision metadata only on the primary virtual model chunk, so each logical data microbatch contributes exactly once
- preserve #4511's resolver behavior: do not blindly divide already-normalized metadata by VPP
- apply the same invariant to GPT, VLM, Qwen3-VL, and Qwen3-Omni step functions
- add VPP=1/2/10 regression coverage for dense token, THD attention, and vision metadata totals

## Regression

NeMo CI job [356425840](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/356425840) ran packed Llama 3 70B SFT with PP=8, VPP=10, CP=1, and DP=4. Stable iterations reported about 14,946 MODEL_TFLOP/s/GPU versus a 1,502.74 golden average (+894.59%, or 9.9459x) at unchanged ~1.14 s iteration time.

This regressed in #4511 (`fb039fd1`), which removed resolver-side division by VPP on the premise that the accumulators already represented one complete training step. That premise is path-dependent:

- dense GPT metadata can exist on only one boundary virtual chunk, so unconditional division undercounts it
- packed SFT exposes the same tokens and `cu_seqlens` on every virtual chunk, so the forward-step accumulator runs VPP times

MCore's interleaved schedule executes the forward step once for every `(logical microbatch, model chunk)` pair. `resolve_global_flops_seqlen_stats` runs once after the completed training step, and `training_log` consumes the cumulative per-step FLOPs delta once at each log boundary. Therefore neither resolver nor logging adds an independent VPP multiplier; they expose the per-chunk over-accumulation.

## Invariant and fix

FLOPs metadata describes the logical data microbatch, while VPP only partitions the full model into scheduled chunks. Each physical PP rank must accumulate each logical microbatch exactly once, independent of VPP size.

The accumulator now accepts the current model chunk's explicit `vp_stage`; stages other than `None`/`0` are no-ops. The stage is read through mixed-precision/DDP wrappers. This fixes packed VPP without assuming every path accumulates exactly VPP times or introducing integer division/truncation in the resolver.

Controlled schedule-table reproduction with four logical microbatches:

| VPP | forward-step calls | resolved token total before | resolved token total after |
| ---: | ---: | ---: | ---: |
| 1 | 4 | 32 (1x) | 32 (1x) |
| 2 | 8 | 64 (2x) | 32 (1x) |
| 10 | 40 | 320 (10x) | 32 (1x) |

## Validation

- `426 passed`: FLOPs, GPT step, training/logging/state, VLM, Qwen3-VL, Qwen3-Omni, and Llama performance recipe unit tests
- VPP=1/2/10 controlled reproduction in the NeMo 26.06 RC container
- `uv run --no-project -- pre-commit run --all-files`

The `--no-project` flag uses the RC container's pre-synced environment; a fresh host sync cannot install the pinned NRX wheel on the host's older manylinux baseline.

No user-facing configuration or checkpoint format changes.

### DIFF
--- a/src/megatron/bridge/models/qwen_omni/qwen3_omni_step.py
+++ b/src/megatron/bridge/models/qwen_omni/qwen3_omni_step.py
@@ -28,7 +28,7 @@ from megatron.core.utils import get_batch_on_this_cp_rank, get_model_config
 from megatron.bridge.training.losses import (
     create_masked_next_token_loss_function as _create_loss_function,
 )
-from megatron.bridge.training.utils.flop_utils import accumulate_flops_metadata
+from megatron.bridge.training.utils.flop_utils import accumulate_flops_metadata, get_model_chunk_vp_stage
 from megatron.bridge.training.utils.padding_utils import (
     pad_or_truncate_2d_to_len,
     pad_or_truncate_attn_to_len,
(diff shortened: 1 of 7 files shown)


yaoyu-33 (approved):

yaoyu-33:
not waiting for GB runs

PR #4751, opened 2026-07-09
feat(conversion): preserve weights in toy model generator

yaoyu-33:
## What does this PR do?

Replaces the existing random-weight `examples/conversion/create_hf_toy_model.py` implementation with a streaming CLI that creates shallow Hugging Face checkpoints while preserving pretrained tensor bytes.

- accepts a local checkpoint directory or Hugging Face model ID
- keeps the first `N` transformer layers and updates layer-related config metadata
- supports unsharded and sharded safetensors checkpoints
- rewrites shard indexes and preserves index metadata
- copies tokenizer and other non-weight model metadata
- does not load checkpoint tensors into CPU or GPU memory
- rejects overlapping source/output directories before `--overwrite` can remove data

The module docstring and conversion README include Qwen3 usage and a Bridge round-trip command.

## Why?

Randomly initialized toy models do not exercise conversion with real pretrained weights. The replacement produces smaller checkpoints whose retained tensors are byte-for-byte identical to the source, making them suitable for checkpoint conversion, parity, and functional checks.

## Testing

- `uvx pre-commit run --all-files`
- `uvx --with huggingface-hub mypy --strict examples/conversion/create_hf_toy_model.py`
- manual source/output overlap guard checks for same, parent, and child paths
- EOS: truncated cached `Qwen/Qwen3-0.6B` from 28 to 4 layers (47 tensors, 714 MiB)
- all 47 retained tensors matched the source exactly
- Hugging Face CUDA forward passed with finite logits
- Megatron Bridge HF -> MCore -> HF round-trip matched all 47 weights and exported successfully
- final CLI output SHA256 matched the EOS-validated checkpoint: `e817676a6990bb8a25fc8d62fcf01ddb8c5b21b4e8c3948ab5d47d134d5a961f`


examples/conversion/truncate_hf_checkpoint.py
```diff
@@ -0,0 +1,303 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Create a shallow Hugging Face checkpoint by retaining its first transformer layers.
+
+Unlike ``create_hf_toy_model.py``, this tool preserves weights from a pretrained
```
yaoyu-33: use this to replace create_hf_toy_model, give how to use in docstring.

examples/conversion/truncate_hf_checkpoint.py
```diff
@@ -0,0 +1,303 @@
+#!/usr/bin/env python3
```
yaoyu-33: no need unit test for example.

PR #4753, opened 2026-07-09
test(functional): migrate legacy perf recipe coverage

yaoyu-33:
Stacked on #4752 (and therefore #4742 / #4679).

## Summary
- replace the manually copied DeepSeek V3 FSDP performance configuration with a compact wrapper around the canonical flat GB300 FSDP recipe
- make performance config integration tests import `megatron.bridge.perf_recipes` directly instead of legacy `scripts/performance/configs` adapters
- separate no-checkpoint performance proxies from focused training feature tests in the shared test helpers
- reclassify the two Llama CUDA Graph cases as training feature coverage rather than performance recipe coverage
- preserve model-specific proxy modules and launchers

## DeepSeek FSDP compatibility proxy
Functional CI currently has a four-GPU GB200 runner for this test, while the canonical FSDP recipe targets GB300. The test now states that relationship explicitly: it uses the production GB300 recipe as the source of truth and only reduces layers, pipeline layout, and EP size for the GB200 topology.

## Validation in progress
- targeted Ruff lint and format checks pass
- changed launcher shell syntax checks pass
- canonical recipe symbols verified from their defining modules
- EOS H100 runtime validation is pending
- GB200 DeepSeek FSDP runtime and golden validation are delegated to PR CI

## Dependencies
Recipe-owned environment variables are intentionally not folded into this PR because #4682 is not merged. The existing launchers continue to provide the required environment until that dependency lands.


yaoyu-33:
H100 validation completed for follow-up commit `3745cf204`:

- Canonical performance recipe construction: **12 passed**.
- Qwen3 30B-A3B H100 FP8 current-scaling proxy: **8 GPUs, 10/10 training steps passed** on the indexed CI corpus; no OOM or NaN. The first step included kernel compilation (~90.7 s), then steady steps were ~0.1–0.6 s; final LM loss was ~0.696.
- Llama CUDA Graph feature tests: **2 passed** on 2 GPUs (local full-iteration graph and Transformer Engine attention graph), 10/10 steps each; no OOM or NaN.

Validation-driven fixes in this commit:

- broadcast rank 0's downloaded corpus path so every torchrun/pytest rank uses the same dataset index cache;
- use a train-only split for the no-eval real-data proxy;
- make the no-checkpoint helper actually disable checkpoint load/save/pretrained restore;
- align the FP8 assertion with the canonical `hybrid` + `tensorwise` current-scaling config;
- set `NCCL_GRAPH_REGISTER=0` in the CUDA Graph launcher when expandable segments are enabled.

PR #4776, opened 2026-07-10
[data] fix: isolate and unify Nemotron Omni collation

cuichenx:
# Summary

This PR has five goals:

1. **Fix cross-sample contamination and multimodal metadata handling.** `adjust_image_tokens` now adjusts each row and image independently using the correct `num_patches`, without flattening the batch or misaligning labels and loss masks. Nemotron Omni also distinguishes `num_frames` temporal grouping from `num_image_tiles` language-model replacement counts and preserves the tokenizer's real pad ID.
2. **Consolidate the Direct-HF and Energon data paths.** Both now use the same model-aware collator for tokenization, masking, image/video/audio processing, padding, and packing; the Energon task encoder only adapts source samples. Omni processors remain on this model-owned collator instead of selecting the generic text collator.
3. **Make the public model contract explicit.** Nemotron Omni supports only dynamic resolution, uses squared-ReLU projection and square-then-crop C-RADIO position embeddings, validates image/sound token and configuration IDs, and accepts only its model-owned `pixel_values` visual input.
4. **Support correct multimodal online in-batch packing.** Image-text and video/audio/text batches use THD boundaries computed after modality-token expansion, with row-major modality ownership, aggregate model-length guards, CP/SP alignment, fixed-width pipeline tails, and masked physical gaps that prevent cross-sample attention or supervision. Generic text packing retains its established emergent-width behavior.
5. **Harden pipeline-parallel training and inference.** Labels, expanded loss masks, exact vision replacement counts, and packed metadata reach the stages that consume them; model-returned loss masks are honored, receive buffers use the correct fixed width, and generation samples from the last real merged token rather than a padded position.

# Full-data HF/Energon parity

Slurm job `13975195` trained on the complete MedPix splits (17,420 train / 3,080 validation) for 2,000 packed optimizer steps per path on the same 8xH100 node and source revision. The experiment matched sample and DP order to isolate data-pipeline behavior.

- All 312 traced Direct-HF/Energon runtime records were tensor-identical across tokens, labels, loss masks, positions, visual inputs, tile counts, and THD metadata, including the epoch boundary and train/validation transitions.
- LM-loss correlation was `0.99977`; the worst validation-loss gap was `0.065%`. Sequence load-balancing correlation was `0.99383`, with zero skipped or NaN iterations.

Direct-HF and Energon matched-order training curves

# Compatibility and scope

- Recipe/CLI usage is unchanged: `--dataset vlm-hf`, `--dataset vlm-energon`, and `dataset.path=...` remain valid.
- The `dynamic_resolution` provider field is retained for serialized/config compatibility, but Nemotron Omni validates that it is `True`; legacy fixed-tile processors fail explicitly.
- Omni's `visual_keys` setting is retained for configuration compatibility and must be `("pixel_values",)`. Previously ignored alternatives now fail explicitly.
- Non-temporal images retain public variable-resolution processing. Temporal mode uses an antialiased 512×512 compatibility canvas for every visual item because pinned MCore cannot pixel-shuffle ragged non-square tubelets.
- Low-level `NemotronOmniTaskSample` constructors must now provide the shared pre-collation `example`; batch `visual_tensors` and `tokens` compatibility adapters remain available.
- No dependency, CI workflow, or third-party submodule change.


cuichenx:
more refactor coming to unify the hf and energon paths for nemotron omni. i'll absorb this fix to that refactor

yaoyu-33 (approved):

PR #4796, opened 2026-07-10
fix(training): account for rampup_batch_size in scheduler step calculation (#2609)

### LINKED ISSUE #2609: Using `rampup_batch_size` breaks the learning rate schedule
In `ConfigContainer` there is [`_calculate_scheduler_steps`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/383b610dca1fe553346b37fc0172f54917b02069/src/megatron/bridge/training/config.py#L1694) which computes samples/steps for warmup/training/decay. Problem is that this doesn't account for batch size warmup. Steps are computed as `iters x global_batch_size`. Because of this warmup lasts longer in terms of iters and decay starts and ends later. For warmup and cosine scheduler it's not a big issue because you can just adjust those to last less or stop sooner. For WSD scheduler though, it's a big problem because start of the decay is computed from the end. Here is an example training setup:
- `train_iters: 100`
- `global_batch_size: 16`
- `rampup_batch_size: [8, 8, 160]`
- `lr_decay_style: WSD`
- `lr_wsd_decay_style: linear`
- `lr_warmup_iters: 10`
- `lr_wsd_decay_iters: 20`

The resulting lr schedule looks like this:
Image

As you can see warmup lasts 20 iters instead of 10 because during those 20 iters the batch size is half of `global_batch_size` and decay starts 10 iters later and doesn't get a chance to decay completely and stops half way. Increasing `train_iters` to 110 would just shift the decay to the right again because for WSD start of the decay is computed from the end.
Here is a quick fix that copies Megatron LM batch size warmup logic in order to compute correct steps/samples.

``` python
    def _get_consumed_samples_for_iterations(self, iterations: int) -> int:
        """Calculates consumed samples after a given number of iterations."""
        if iterations <= 0:
            return 0

        global_batch_size = self.train.global_batch_size

        rampup_batch_size = self.train.rampup_batch_size
        if rampup_batch_size is None:
            return iterations * global_batch_size

        # Same behavior as in Megatron LM
        # megatron/core/num_microbatches_calculator.py::RampupBatchsizeNumMicroBatchesCalculator
        assert len(rampup_batch_size) == 3, (
            'expected the following '
            'format: --rampup-batch-size <start batch size> '
            '<batch size incerement> <ramp-up samples>'
        )
        start_batch_size = int(rampup_batch_size[0])
        batch_size_increment = int(rampup_batch_size[1])
        rampup_samples = int(rampup_batch_size[2])

        diff_batch_size = global_batch_size - start_batch_size
        assert diff_batch_size >= 0, (
            'expected global batch size to be greater than or equal to start batch size, '
            f'got {global_batch_size} and {start_batch_size}'
        )
        assert diff_batch_size % batch_size_increment == 0, (
            'expected '
            f'global batch size interval ({diff_batch_size}) to be divisible by global batch '
            f'size increment ({batch_size_increment})'
        )

        num_increments = diff_batch_size // batch_size_increment
        rampup_samples_per_increment = rampup_samples / num_increments

        consumed_samples = 0
        for _ in range(iterations):
            if consumed_samples > rampup_samples:
                current_batch_size = global_batch_size
            else:
                steps = int(consumed_samples / rampup_samples_per_increment)
                current_batch_size = start_batch_size + steps * batch_size_increment
                assert current_batch_size <= global_batch_size

            consumed_samples += current_batch_size

        return consumed_samples

    def _calculate_scheduler_steps(self) -> None:
        """Calculate scheduler steps for both iteration-based and sample-based training."""
        is_sample_based = self.train.train_samples is not None

        if is_sample_based:
            if self.scheduler.lr_decay_samples is None:
                self.scheduler.lr_decay_samples = self.train.train_samples
            self.scheduler.lr_decay_steps = self.scheduler.lr_decay_samples
            self.scheduler.wd_incr_steps = self.train.train_samples

            if self.scheduler.lr_wsd_decay_samples is not None:
                self.scheduler.wsd_decay_steps = self.scheduler.lr_wsd_decay_samples

            # Warmup calculation for sample-based training
            if self.scheduler.lr_warmup_fraction is not None:
                self.scheduler.lr_warmup_steps = self.scheduler.lr_warmup_fraction * self.scheduler.lr_decay_steps
            else:
                self.scheduler.lr_warmup_steps = self.scheduler.lr_warmup_samples
        else:
            # Iteration-based training
            if self.scheduler.lr_decay_iters is None:
                self.scheduler.lr_decay_iters = self.train.train_iters
            self.scheduler.lr_decay_steps = self._get_consumed_samples_for_iterations(self.scheduler.lr_decay_iters)
            self.scheduler.wd_incr_steps = self._get_consumed_samples_for_iterations(self.train.train_iters)

            if self.scheduler.lr_wsd_decay_iters is not None:
                wsd_start_iters = max(self.scheduler.lr_decay_iters - self.scheduler.lr_wsd_decay_iters, 0)
                self.scheduler.wsd_decay_steps = (
                    self.scheduler.lr_decay_steps - self._get_consumed_samples_for_iterations(wsd_start_iters)
                )

            if self.scheduler.lr_warmup_fraction is not None:
                self.scheduler.lr_warmup_steps = self.scheduler.lr_warmup_fraction * self.scheduler.lr_decay_steps
            else:
                self.scheduler.lr_warmup_steps = self._get_consumed_samples_for_iterations(self.scheduler.lr_warmup_iters)

```

lonexreb:
## What does this PR do?

Fixes #2609.

`ConfigContainer._calculate_scheduler_steps()` converts iteration-based scheduler settings to sample counts as `iters * global_batch_size`. With `rampup_batch_size` enabled the early iterations consume fewer samples, so:

- LR warmup lasts longer (in iterations) than `lr_warmup_iters`
- decay starts and ends later than configured
- for `lr_decay_style=WSD` (decay anchored to the end) the decay never completes — it stops partway, exactly as shown in the issue's plot

## Fix

Add `_get_consumed_samples_for_iterations(iterations)`, mirroring `megatron.core.num_microbatches_calculator.RampupBatchsizeNumMicroBatchesCalculator` (same assertions and batch-size stepping), and use it in the iteration-based branch for:

- `lr_decay_steps = consumed(lr_decay_iters)`
- `wd_incr_steps = consumed(train_iters)`
- `lr_warmup_steps = consumed(lr_warmup_iters)`
- `wsd_decay_steps = consumed(lr_decay_iters) - consumed(lr_decay_iters - lr_wsd_decay_iters)` — the WSD decay window covers the *last* `lr_wsd_decay_iters` iterations, so it must be measured as the sample span between those two iteration marks

Without `rampup_batch_size` every value is unchanged (`consumed(n) == n * global_batch_size`).

With the issue's example (`train_iters=100`, `gbs=16`, `rampup=[8,8,160]`, `lr_warmup_iters=10`, `lr_wsd_decay_iters=20`): warmup now ends exactly at iteration 10 (80 samples), and WSD decay spans iterations 80→100 and reaches its end value at the final iteration.

## Tests

Added `TestSchedulerStepsWithRampupBatchSize` in `tests/unit_tests/training/test_config.py`: consumed-samples math (no ramp / ramp / start==global), the full WSD example from the issue, and a no-ramp regression check.

## Changelog

- fix(training): scheduler warmup/decay/WSD sample counts now honor `rampup_batch_size`

### DIFF
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -1459,6 +1459,54 @@ class ConfigContainer(Container):
         )
         self._calculate_scheduler_steps()
 
+    def _get_consumed_samples_for_iterations(self, iterations: int) -> int:
+        """Return the number of samples consumed after ``iterations``, honoring batch-size ramp-up.
+
+        Mirrors the ramp-up behavior of
+        ``megatron.core.num_microbatches_calculator.RampupBatchsizeNumMicroBatchesCalculator`` so
+        scheduler sample counts line up with the iterations actually trained (#2609).
+        """
+        if iterations <= 0:
+            return 0
+
+        global_batch_size = self.train.global_batch_size
+        rampup_batch_size = self.train.rampup_batch_size
+        if rampup_batch_size is None:
+            return iterations * global_batch_size
+
+        assert len(rampup_batch_size) == 3, (
+            "expected rampup_batch_size format: [<start batch size>, <batch size increment>, <ramp-up samples>]"
+        )
+        start_batch_size = int(rampup_batch_size[0])
+        batch_size_increment = int(rampup_batch_size[1])
+        rampup_samples = int(rampup_batch_size[2])
+
+        diff_batch_size = global_batch_size - start_batch_size
+        assert diff_batch_size >= 0, (
+            "expected global batch size to be greater than or equal to start batch size, "
+            f"got {global_batch_size} and {start_batch_size}"
+        )
+        assert batch_size_increment > 0, f"expected a positive batch size increment, got {batch_size_increment}"
+        assert diff_batch_size % batch_size_increment == 0, (
+            f"expected global batch size interval ({diff_batch_size}) to be divisible by global batch "
+            f"size increment ({batch_size_increment})"
+        )
+
+        num_increments = diff_batch_size // batch_size_increment
+        if num_increments == 0:
+            return iterations * global_batch_size
+        rampup_samples_per_increment = rampup_samples / num_increments
+
+        consumed_samples = 0
+        iterations_done = 0
+        while iterations_done < iterations and consumed_samples < rampup_samples:
+            steps = int(consumed_samples / rampup_samples_per_increment)
+            current_batch_size = min(start_batch_size + steps * batch_size_increment, global_batch_size)
+            consumed_samples += current_batch_size
+            iterations_done += 1
+
+        return consumed_samples + (iterations - iterations_done) * global_batch_size
+
     def _calculate_scheduler_steps(self) -> None:
         """Calculate scheduler steps for both iteration-based and sample-based training."""
         is_sample_based = self.train.train_samples is not None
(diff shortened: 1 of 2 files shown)


yaoyu-33:

- For this change specifically: a before-and-after scheduler trace showing the warmup and WSD boundaries at the intended iterations with rampup_batch_size enabled.

PR #4797, opened 2026-07-10
fix(ckpt): honor CheckpointConfig.async_strategy on the sync save path (#4639)

### LINKED ISSUE #4639: [bug] CheckpointConfig.async_strategy is silently ignored on the sync save path
## Summary

`CheckpointConfig.async_strategy` (default `"nvrx"`, defined at [`src/megatron/bridge/training/config.py:586`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/training/config.py#L586)) is plumbed into `dist_checkpointing.save(...)` at [`src/megatron/bridge/training/checkpointing.py:1320-1322`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/training/checkpointing.py#L1320-L1322), but it has **no effect** when `async_save=False`. In that case the code path in Megatron-LM is:

```python
# megatron/core/dist_checkpointing/serialization.py
if not async_sharded_save:
    sharded_strategy.save(sharded_state_dict, checkpoint_dir)   # <- async_strategy dropped
    ...
    return None
```

and `TorchDistSaveShardedStrategy.save()` internally hardcodes `async_strategy="mcore"`:

```python
# megatron/core/dist_checkpointing/strategies/torch.py
def save(self, sharded_state_dict, checkpoint_dir):
    """Sync save always uses the built-in implementation."""
    async_request = self.async_save(
        sharded_state_dict, checkpoint_dir, async_strategy="mcore"
    )
    async_request.execute_sync()
```

Since `save_megatron_model` sets `async_save=False` unconditionally ([`model_load_save.py:561`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/training/model_load_save.py#L561)), every HF->Megatron conversion via `AutoBridge.save_megatron_model` runs through the deprecated `"mcore"` async-finalize path — the same path whose deprecation warning tells users to switch to `"nvrx"`.

## Impact

The `"mcore"` async-finalize path can crash with `TypeError: cannot pickle code objects` in `gather_object(write_results)` for certain models, which forced a downstream monkey-patch workaround in NeMo-RL: [NVIDIA-NeMo/RL#2989](https://github.com/NVIDIA-NeMo/RL/pull/2989) wraps `bridge.save_megatron_model(...)` in a context manager that temporarily replaces `TorchDistSaveShardedStrategy.save` so the HF->Megatron sync save uses `async_strategy="nvrx"`. That shim should not be required — the config knob is right there.

## Ask

Prefer **(a)**:

**(a) Route the sync save through the same code path as async, honoring `CheckpointConfig.async_strategy`.** E.g. at the MB layer, when `async_save=False`, call `sharded_strategy.async_save(..., async_strategy=ckpt_cfg.async_strategy).execute_sync()` instead of `dist_checkpointing.save(..., async_sharded_save=False)`. This makes the already-exposed config actually work for both sync and async saves and removes the need for the NeMo-RL shim.

**(c) Alternatively**, flip the Megatron-LM sync default in `TorchDistSaveShardedStrategy.save` from `"mcore"` to `"nvrx"` to match the `async_save` default. Simpler patch but changes upstream behavior.

## Reference

- Workaround PR in NeMo-RL: [NVIDIA-NeMo/RL#2989](https://github.com/NVIDIA-NeMo/RL/pull/2989)
- Failure symptom in the mcore async finalize path: `TypeError: cannot pickle code objects` during `gather_object(write_results)`

lonexreb:
## What does this PR do?

Fixes #4639.

`CheckpointConfig.async_strategy` (default `"nvrx"`) is plumbed into `dist_checkpointing.save(...)`, but has no effect when `async_save=False`: MCore's sync path calls `TorchDistSaveShardedStrategy.save()`, which internally hardcodes `async_strategy="mcore"` — the deprecated finalize path that can crash with `TypeError: cannot pickle code objects` in `gather_object(write_results)`. Since `save_megatron_model` sets `async_save=False` unconditionally, every HF→Megatron conversion runs through that deprecated path, which forced the monkey-patch workaround in NVIDIA-NeMo/RL#2989.

## Fix

Implements option **(a)** from the issue at the MB layer: a `SyncSaveWithConfiguredAsyncStrategy` subclass of `TorchDistSaveShardedStrategy` whose `save()` routes through `async_save(..., async_strategy=<configured>).execute_sync()`. The torch_dist construction site in `save_checkpoint` now instantiates it with `sync_async_strategy=ckpt_cfg.async_strategy`.

Fallbacks to the built-in `save()` (unchanged behavior) when:
- `async_strategy="mcore"` is explicitly configured,
- the installed MCore's `async_save` no longer accepts the `async_strategy` kwarg (dev branch),
- the configured strategy's modules are missing (`ImportError`, e.g. no `nvidia-resiliency-ext`).

Async saves (`async_save=True`) are untouched — `dist_checkpointing.save` never calls `strategy.save()` on that path.

## Tests

Added `TestSyncSaveWithConfiguredAsyncStrategy` in `tests/unit_tests/training/test_checkpointing.py` covering: nvrx routing + `execute_sync`, explicit mcore passthrough, ImportError fallback, and missing-kwarg fallback.

## Changelog

- fix(ckpt): sync torch_dist saves now honor `CheckpointConfig.async_strategy` instead of the deprecated hardcoded `mcore` path

### DIFF
--- a/src/megatron/bridge/training/checkpointing.py
+++ b/src/megatron/bridge/training/checkpointing.py
@@ -140,6 +140,38 @@ _DIRECT_ITERATION_DIR_SENTINEL = -2
 HF_WEIGHTS_SUBDIR = "hf"
 
 
+class SyncSaveWithConfiguredAsyncStrategy(TorchDistSaveShardedStrategy):
+    """torch_dist save strategy whose synchronous ``save`` honors ``CheckpointConfig.async_strategy``.
+
+    MCore's ``TorchDistSaveShardedStrategy.save`` hardcodes the deprecated ``mcore``
+    async-finalize path for synchronous saves, silently ignoring the configured
+    ``async_strategy``. Route the sync save through ``async_save`` with the configured
+    strategy followed by ``execute_sync`` so ``async_save=False`` saves (e.g.
+    ``AutoBridge.save_megatron_model``) use the same strategy as async ones (#4639).
+    """
+
+    def __init__(self, *args: Any, sync_async_strategy: str = "nvrx", **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.sync_async_strategy = sync_async_strategy
+
+    def save(self, sharded_state_dict: ShardedStateDict, checkpoint_dir: Path) -> None:
+        """Synchronously save using the configured async strategy's writer/finalize path."""
+        if (
+            self.sync_async_strategy == "mcore"
+            or "async_strategy" not in inspect.signature(self.async_save).parameters
+        ):
+            # "mcore" matches the built-in behavior; newer MCore branches removed the kwarg.
+            return super().save(sharded_state_dict, checkpoint_dir)
+        try:
+            async_request = self.async_save(
+                sharded_state_dict, checkpoint_dir, async_strategy=self.sync_async_strategy
+            )
+        except (ImportError, ModuleNotFoundError):
+            # Environment lacks the configured strategy's modules (e.g. no nvidia-resiliency-ext).
+            return super().save(sharded_state_dict, checkpoint_dir)
+        async_request.execute_sync()
+
+
 # ============================================================================
 # Checkpoint version and utilities
 # ============================================================================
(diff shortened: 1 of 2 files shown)


yaoyu-33:

- For this change specifically: a real synchronous torch_dist checkpoint or conversion save showing the original failure before the change and a successful save and restore with the configured strategy after it.

PR #4798, opened 2026-07-10
fix(data): retry packed-Parquet resolution with metadata refresh after prep barrier (#4207)

### LINKED ISSUE #4207: [bug] Multi-node finetuning intermittently crashes with 'NoneType' object is not an iterator — packed-Parquet not visible to non-producer nodes after the dataset-prep barrier
### Problem

Multi-node PEFT/SFT finetuning (packed-sequence dataset, `dataset.dataloader_type="batch"`, `pipeline_model_parallel_size=1`) intermittently crashes at the very start of the training loop (iteration 0) with:

```
TypeError: 'NoneType' object is not an iterator
```

at `prepare_finetuning_batch` -> `global_batch = next(data_iterator)`. The train `data_iterator` is `None` on an entire non-rank-0 node (e.g. all 4 ranks of the second node), while the first node's ranks proceed. With `pipeline_model_parallel_size == 1` every rank is both first and last pipeline stage and must hold a valid finetuning iterator, so a `None` iterator on a whole node is an invalid state.

### Root cause

`FinetuningDatasetBuilder.build()` prepares the packed Parquet on rank 0 only, then all ranks call `torch.distributed.barrier()`, then every rank calls `_create_dataset()`, which resolves the packed-Parquet path via `resolve_packed_parquet_paths()` / `Path.exists()`. On a shared filesystem (e.g. with NFS-style attribtue caching), a `torch.distributed.barrier()` guarantees the producer *finished writing* but does NOT guarantee that other nodes' filesystem clients have refreshed their directory metadata. A non-producer node can therefore still observe a stale negative directory entry and resolve zero files, so `_create_dataset` returns `None` -> `train_ds` / `train_dataloader` / `train_data_iterator` all become `None` on that whole node -> crash at `next()`. The pretraining path tolerates a `None` iterator (it forwards it straight to `forward_backward_func`); only the finetuning branch eagerly calls `next()`, so only finetuning crashes this way. This is the same shared-filesystem race family as the already-fixed mkdir race (PR #3888), but the read-visibility-after-write variant.

### Minimal repro

```shell
# 2 nodes x 4 GPUs, data-parallel across nodes, pipeline_model_parallel_size=1,
# packed-sequence SQuAD finetuning of an 8B model on a shared filesystem (Lustre).
# Intermittent (~1 in 6 observed) at iteration 0, right after the dataset cache
# is written by rank 0 and all ranks pass the dataset-prep barrier.
#   - model: 8B causal LM
#   - finetuning: peft (lora) or sft
#   - dataset: packed-sequence SQuAD, dataloader_type="batch"
#   - parallelism: TP intra-node, DP inter-node, PP=1
```

### Expected behavior

With `pipeline_model_parallel_size == 1`, every rank should receive a valid finetuning data iterator before the training loop begins; the iterator should never be `None`. If a `None` iterator can legitimately occur, `prepare_finetuning_batch` should fail fast with a clear, actionable error rather than raising a bare `TypeError` from `next(None)`.

### Proposed fix

1. In `FinetuningDatasetBuilder._create_dataset`, when resolving a packed-Parquet spec that is expected to exist (we are past the `build()` barrier, so rank 0 has finished writing), retry resolution with bounded backoff and a parent-directory metadata refresh (e.g. `os.scandir(parent)`) before concluding the path is absent. This tolerates shared-filesystem read-visibility latency on non-producer nodes instead of silently returning `None`.
2. Defensively, make `prepare_finetuning_batch` fail fast with a clear, rank-attributed error (or handle `None` the way the pretraining path does) instead of calling `next(None)`.

### Affected area

area:data (finetuning packed-Parquet dataset construction / shared-FS visibility), area:training

### Regression?

Not sure. Intermittent and reproduces on byte-identical library versions; appears to be a pre-existing shared-filesystem race exposed by multi-node finetuning timing rather than a recent version bump.

### Validation

A bounded-retry resolution in `_create_dataset` was validated experimentally on the reproducing topology: it recovered the visibility lag (resolved the packed Parquet after ~7 retries on the second node) and eliminated the crash — 0 None-iterator crashes in 20 multi-node attempts with the fix, versus 3 in 20 without it.

lonexreb:
## What does this PR do?

Fixes #4207.

Multi-node finetuning with packed-sequence datasets intermittently crashes at iteration 0 with `TypeError: 'NoneType' object is not an iterator` on an entire non-rank-0 node. Root cause (as analyzed in the issue): `FinetuningDatasetBuilder.build()` prepares the packed Parquet on rank 0, all ranks pass a `torch.distributed.barrier()`, then every rank resolves the packed-Parquet path. On a shared filesystem (e.g. Lustre) the barrier guarantees the producer *finished writing* but not that other nodes' filesystem clients refreshed their directory metadata — a non-producer node can still observe a stale negative directory entry, resolve zero files, and silently build a `None` dataset → `None` iterator → crash.

## Fix

- New `resolve_packed_parquet_paths_with_refresh()` in `data/packing/paths.py`: retries resolution up to 5 times, forcing a parent-directory listing (`os.listdir`) between attempts with linear backoff, and logs each retry. Raises the last resolution error if all attempts fail (unchanged failure semantics).
- `build_gpt_sft_split()` (the post-barrier consumer path) now uses the retrying resolver for packed-Parquet specs.
- The rank-0 pre-preparation check (`_packed_path_exists`) intentionally keeps the single-attempt resolver — a missing artifact there is the expected signal to produce it, and retrying would delay preparation.

No behavior change when files resolve on the first attempt (no sleeps, same results).

## Tests

New `tests/unit_tests/data/packing/test_paths.py`: immediate resolution without retry, delayed visibility resolved on a later attempt (backoff sequence asserted), exhaustion raising `ValueError`, empty-directory retry, and metadata-refresh no-op on missing directories.

## Changelog

- fix(data): packed-Parquet resolution retries with directory-metadata refresh to tolerate shared-filesystem visibility lag after the dataset-prep barrier

### DIFF
--- a/src/megatron/bridge/data/builders/gpt_sft.py
+++ b/src/megatron/bridge/data/builders/gpt_sft.py
@@ -34,6 +34,7 @@ from megatron.bridge.data.packing import PackedSequenceSpecs
 from megatron.bridge.data.packing.paths import (
     is_packed_parquet_spec,
     resolve_packed_parquet_paths,
+    resolve_packed_parquet_paths_with_refresh,
 )
 from megatron.bridge.data.sft_processing import (
     ChatSFTPreprocessingConfig,
@@ -436,7 +437,9 @@ def build_gpt_sft_split(
     path_str = str(path)
     if is_packed_parquet_spec(path_str):
         try:
-            path_exists = bool(resolve_packed_parquet_paths(path_str))
+            # Retry with directory-metadata refresh: on shared filesystems a non-producer
+            # node can transiently see zero files right after rank 0 wrote them (#4207).
+            path_exists = bool(resolve_packed_parquet_paths_with_refresh(path_str))
         except ValueError:
             path_exists = False
     elif MultiStorageClientFeature.is_enabled():
--- a/src/megatron/bridge/data/packing/paths.py
+++ b/src/megatron/bridge/data/packing/paths.py
@@ -15,10 +15,15 @@
 """Path detection and resolution for packed Parquet artifacts."""
 
 import glob
+import logging
 import os
+import time
 from pathlib import Path
 
 from megatron.core.msc_utils import MultiStorageClientFeature
+
+
+logger = logging.getLogger(__name__)
 
 
 def is_packed_parquet_file(path) -> bool:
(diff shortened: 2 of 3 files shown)


yaoyu-33:

- For this change specifically: a real multi-node shared-filesystem reproduction showing stale packed-dataset visibility before the change and successful dataset construction after it; mocked retry tests alone are not sufficient for this root-cause claim.

PR #4801, opened 2026-07-10
feat(training): add ValidationConfig.eval_at_start for step-0 baseline eval (#3996)

### LINKED ISSUE #3996: [feature] Eval at start (step-0 eval)
### User problem

There's no way to run a validation pass *before* the training loop starts. The only call site for `evaluate_and_print_results` inside `train()` (`megatron/bridge/training/train.py:511`) is gated on `global_state.train_state.step % val_config.eval_interval == 0`, evaluated *after* `state.step += 1`, so the first eval lands at `step == eval_interval` — not step 0. For long runs we lose the baseline number that anchors every "did training help?" plot in W&B/TensorBoard, and for short runs (smoke tests, resumes from a fresh init) we may never get a single eval at all.

A `Callback.on_train_start` hook isn't sufficient because `CallbackContext` (`callbacks.py:104`) doesn't expose `valid_data_iterator` or `forward_step_func`, both of which `evaluate_and_print_results` requires.

### Desired outcome

A first-class toggle, e.g. `ValidationConfig.eval_at_start: bool = False`, that runs exactly one validation pass (and optionally test pass) before the first training step, using the same val iterator and callback wiring as the in-loop eval. No user-side wrapping of `pretrain()` or monkey-patching of `train()`.

### Alternatives considered

_No response_

### Affected area

area:training

### Urgency / use case

Important but not blocking

### Extra context

The cleanest place is inside `train()` itself, not `_pretrain`, because all the needed args are already in scope and `inprocess_restart` re-entries land here with `state.step > 0` for free:

1. Extract the existing eval block at `train.py:511-540` into a private `_run_validation(global_state, forward_step_func, valid_data_iterator, model, model_config, callback_manager, ...)` helper.
2. Right before the `while global_state.train_state.step < train_config.train_iters:` loop, add:

   ```python
   if (val_config.eval_at_start
       and global_state.train_state.do_valid
       and global_state.train_state.step == 0):
       _run_validation(...)
   ```
3. Inside the loop, the existing eval becomes the same helper call, guarded by the interval check as today.

That's symmetric, restart-safe, composes with any callback-based eval extensions (multi-blend logging, etc.), and is roughly a 15-line change. Happy to send a PR.

lonexreb:
## What does this PR do?

Fixes #3996 (maintainer-approved ask: https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/3996#issuecomment).

There is currently no way to run a validation pass *before* the training loop starts: the only `evaluate_and_print_results` call site inside `train()` is gated on `step % eval_interval == 0` evaluated after the step increment, so the first eval lands at `step == eval_interval`. Long runs lose the baseline number that anchors W&B/TensorBoard plots, and short smoke runs may never eval at all.

## Implementation

As suggested in the issue, the toggle lives in the validation config and the eval runs inside `train()` where all required args are in scope:

- **`ValidationConfig.eval_at_start: bool = False`** — added on a new Bridge `ValidationConfig` subclass of the MCore training `ValidationConfig` (same pattern as `TrainingConfig`/`SchedulerConfig`/`TokenizerConfig`). All existing imports from `megatron.bridge.training.config` transparently pick up the subclass; configs constructed from the upstream class still work (`getattr(..., False)` guard).
- **`train()`**: immediately before the training loop, when `eval_at_start` is set, `do_valid` is true, and `step == 0`, run exactly one validation pass with the same val iterator, `eval-time`/`interval-time` timer handling, `manual_gc_eval` behavior, and callback wiring as the in-loop eval. Forward pre-hooks are still disabled at this point in `train()`, matching the in-loop eval's requirement.
- **Resume-safe**: `step > 0` (checkpoint resume, `inprocess_restart` re-entries) skips the step-0 eval for free, as the issue requested.

## Tests

`TestValidationConfigEvalAtStart` in `tests/unit_tests/training/test_config.py`: default-off, enable, MCore field inheritance, and no spurious deprecation warning from the TrainingConfig→ValidationConfig sync loop.

## Changelog

- feat(training): `ValidationConfig.eval_at_start` runs one validation pass before the first training step

### DIFF
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -38,10 +38,11 @@ from megatron.training.config import DistributedInitConfig as MTrainDistributedI
 from megatron.training.config import LoggerConfig as MTrainLoggerConfig
 from megatron.training.config import ProfilingConfig as MTrainProfilingConfig
 from megatron.training.config import RerunStateMachineConfig as MTrainRerunStateMachineConfig
-from megatron.training.config import RNGConfig, ValidationConfig
+from megatron.training.config import RNGConfig
 from megatron.training.config import SchedulerConfig as MTrainSchedulerConfig
 from megatron.training.config import StragglerDetectionConfig as MTrainStragglerDetectionConfig
 from megatron.training.config import TrainingConfig as MTrainTrainingConfig
+from megatron.training.config import ValidationConfig as MTrainValidationConfig
 
 from megatron.bridge.data.base import (
     DataloaderConfig,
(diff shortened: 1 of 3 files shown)


yaoyu-33:

- For this change specifically: a short real training run showing exactly one validation pass at step 0, plus a checkpoint-resume run showing that the initial validation is not repeated.

PR #4835, opened 2026-07-12
fix(data): honor --packed_sequence for llm-finetune-preloaded

### LINKED ISSUE #4830: [bug] `llm-finetune-preloaded` silently ignores `--packed_sequence`
### Problem

`apply_dataset_override` in `src/megatron/bridge/recipes/utils/dataset_utils.py` accepts a `packed_sequence` argument, but only the `llm-finetune` preset branch forwards it into the dataset config (`kwargs = {"packed_sequence": packed_sequence, "pad_seq_to_mult": 1}`).

The `llm-finetune-preloaded` branch builds `GPTSFTDatasetConfig` without touching `enable_offline_packing` / `offline_packing_specs` (both exist on `GPTSFTDatasetConfig`, `src/megatron/bridge/data/builders/gpt_sft.py`), so `--packed_sequence` is silently dropped and the run trains **unpacked** with no warning. Users get a correct-looking loss curve at meaningfully lower real-token throughput, and nothing indicates the flag was ignored.

### Minimal repro

```shell
# Any SFT recipe over a local JSONL, requesting packing:
uv run python scripts/training/run_recipe.py \
  --recipe <any_sft_recipe> \
  --dataset llm-finetune-preloaded \
  --packed_sequence --seq_length 16384 \
  dataset.dataset_root=/data/my_sft.jsonl

# Observe: the resulting GPTSFTDatasetConfig has enable_offline_packing=False,
# offline_packing_specs=None; samples are variable-length, never packed to 16384.
# The same command with --dataset llm-finetune (preset) packs correctly.
```

### Expected behavior

Either honor the flag — build the preloaded `GPTSFTDatasetConfig` with `enable_offline_packing=True` and `offline_packing_specs` sized to the resolved `seq_length` (mirroring what the `llm-finetune` preset branch does) — or fail loudly if the combination is unsupported. Silent fallback to unpacked is the worst of both.

### Affected area

area:data (recipes/dataset override)

### Regression?

No — the wiring appears to have never existed for the preloaded branch (the preset branch got it; preloaded didn't).

### Environment

Reproduced at runtime against unmodified `main` @ `b864189` (2026-07-11) with its pinned Megatron-LM submodule (`cf2f07d7`), Python 3.12. The gap also existed on the earlier shape of this branch (when it built `FinetuningDatasetConfig`) — it survived that refactor.

### Additional context

* We (AI21) hit this in production SFT runs; packing recovered \~+23% real tokens/s once wired. We have a working fix against the earlier `FinetuningDatasetConfig` shape and are happy to open a PR reworked for the current `GPTSFTDatasetConfig` shape.
* Related but distinct: NVIDIA-NeMo/Megatron-Bridge#1106 tracks packing for the **VLM** preloaded/HF paths; this is the LLM preloaded analog. NVIDIA-NeMo/Megatron-Bridge#4351 (index-only packed mode) and NVIDIA-NeMo/Megatron-Bridge#4593 (packed THD/PP edge cases) don't cover this wiring gap.

Achyuthan-S:
## What & why

`apply_dataset_override` accepts a `packed_sequence` flag, but the
`llm-finetune-preloaded` branch built `GPTSFTDatasetConfig` without forwarding it.
So `--packed_sequence` was silently dropped on the preloaded path — runs trained
unpacked at meaningfully lower real-token throughput, with no error or warning.
The `llm-finetune` preset branch already wired packing correctly; this closes the
gap for the preloaded branch.

Closes #4830.

## Change

When `packed_sequence` is requested for `llm-finetune-preloaded`, forward it into
`GPTSFTDatasetConfig`, mirroring the `llm-finetune` preset:

- `enable_offline_packing=packed_sequence`
- `offline_packing_specs=PackedSequenceSpecs(packed_sequence_size=<resolved seq_length>, pad_seq_to_mult=1)`
- `dataset_kwargs={"pad_to_max_length": True}`

With `packed_sequence` unset, behavior is unchanged (packing stays disabled).
No public API change.

## Testing

Added unit tests in `tests/unit_tests/recipes/utils/test_dataset_utils.py`:

- `test_llm_finetune_preloaded_honors_packed_sequence` — packing enabled, specs
  sized to the resolved `seq_length`, `pad_to_max_length` set.
- `test_llm_finetune_preloaded_packed_false_leaves_packing_disabled` — flag-off
  path leaves packing disabled.
- Hardened the existing preloaded test to assert packing defaults off.

Verified in a NeMo 25.09 GPU container: the four preloaded/packed cases pass.

cc @yaoyu-33

### DIFF
--- a/src/megatron/bridge/recipes/utils/dataset_utils.py
+++ b/src/megatron/bridge/recipes/utils/dataset_utils.py
@@ -26,6 +26,7 @@ from megatron.bridge.data.builders import (
 )
 from megatron.bridge.data.energon.energon_provider import EnergonProvider
 from megatron.bridge.data.loaders import get_blend_and_blend_per_split
+from megatron.bridge.data.packing import PackedSequenceSpecs
 from megatron.bridge.data.vlm_datasets.preloaded_provider import PreloadedVLMConversationProvider
 from megatron.bridge.recipes.utils.finetune_utils import (
     default_gsm8k_config,
@@ -226,6 +227,14 @@ def apply_dataset_override(
             raise ValueError(
                 "llm-finetune-preloaded requires dataset.dataset_root=<path> to select the local JSONL source."
             )
+        offline_packing_specs = None
+        dataset_kwargs = None
+        if packed_sequence:
+            offline_packing_specs = PackedSequenceSpecs(
+                packed_sequence_size=resolved_seq_length,
+                pad_seq_to_mult=1,
+            )
+            dataset_kwargs = {"pad_to_max_length": True}
         config.dataset = GPTSFTDatasetConfig(
             seq_length=resolved_seq_length,
             dataset_root=dataset_root,
@@ -236,6 +245,9 @@ def apply_dataset_override(
             ),
             dataloader_type="batch",
             seed=5678,
+            enable_offline_packing=packed_sequence,
+            offline_packing_specs=offline_packing_specs,
+            dataset_kwargs=dataset_kwargs,
         )
 
     elif dataset_type == "vlm-energon":
--- a/tests/unit_tests/recipes/utils/test_dataset_utils.py
+++ b/tests/unit_tests/recipes/utils/test_dataset_utils.py
@@ -382,12 +382,51 @@ class TestApplyDatasetOverride:
         assert result.dataset.dataloader_type == "batch"
         assert "dataset.dataset_root=/data/sft" not in overrides
         assert "train.train_iters=10" in overrides
+        # Packing must stay disabled unless explicitly requested.
+        assert result.dataset.enable_offline_packing is False
+        assert result.dataset.offline_packing_specs is None
 
     def test_llm_finetune_preloaded_requires_local_source(self):
         config = _make_mock_config()
 
         with pytest.raises(ValueError, match="requires dataset.dataset_root"):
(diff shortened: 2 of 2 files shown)


yaoyu-33:
Thanks for the contribution and for validating the original failure mode.

Since this PR was opened, #4856 has replaced the legacy dataset-override interface on `main`. `apply_dataset_override`, `llm-finetune-preloaded`, and `--packed_sequence` are no longer part of the launcher flow. Local prompt-completion JSONL is now selected with `--dataset local-jsonl`, and offline packing is enabled independently with `dataset.enable_offline_packing=true`, for example:

```bash
uv run python -m torch.distributed.run --nproc_per_node=1 scripts/training/run_recipe.py \
  --recipe <recipe> --mode sft --dataset local-jsonl \
  dataset.dataset_root=/data/sft_jsonl \
  dataset.seq_length=16384 \
  dataset.enable_offline_packing=true \
  train.micro_batch_size=1
```

The shared runner creates `PackedSequenceSpecs` when needed, sets `packed_sequence_size` from the resolved dataset sequence length, and aligns padding for CP, TP, and sequence parallelism. This covers the intended behavior on the supported current interface. As noted in #4830, we are not extending the legacy `llm-finetune-preloaded` path.

Because this patch now conflicts with `main` and targets code removed by #4856, I am closing the PR as superseded. Thank you again for surfacing the silent-ignore issue and providing tests.

PR #4805, opened 2026-07-10
fix(model): improve conversion and inference examples

yaoyu-33:
## Summary

- land a focused subset of conversion and inference fixes discovered during broader model validation
- add or update portable examples and documentation for selected model families
- improve the model bridges and shared VLM utilities needed by those flows

## Scope

This PR intentionally merges only the ready, independently useful fixes and examples. It no longer tracks or claims complete validation across every supported model. The broad model inventory and exhaustive per-model checklist have been removed from this PR.

## Follow-up validation

Full-checkpoint conversion, inference, export, round-trip, numerical-parity, and training verification will continue through model verification cards. Each card will capture the model-specific configuration, complete commands, environment, and validation evidence so that coverage can be added incrementally and reviewed independently.

## Validation

- targeted unit tests for the touched conversion, model, and VLM paths
- pre-commit checks
- CI

PR #4883, opened 2026-07-15
fix(conversion): detect tied embeddings by Megatron param name, not hard-coded HF name

ErenAta16:
## What changed

`stream_weights_megatron_to_hf`'s tied-embedding branch (`model_bridge.py`, with the existing `# TODO(yuya): fix this hard coded naming` comment on it) decides whether to re-emit a duplicate `lm_head.weight` during HF export by checking `hf_name == "model.embed_tokens.weight"`. This re-emission is necessary because `build_conversion_tasks` unconditionally drops the Megatron `output_layer.weight` task whenever embeddings are tied, so the tied-embedding branch is the only place that can still produce `lm_head.weight` in the output.

That literal string only matches the plain-LLM naming convention. At least ten bridges map the tied embedding to a differently-prefixed HF name instead, e.g.:
- `qwen_vl/qwen3_vl_bridge.py`, `glm_vl/glm_45v_bridge.py`, `gemma_vl/gemma4_vl_bridge.py`: `"model.language_model.embed_tokens.weight"`
- `gemma_vl/gemma3_vl_bridge.py`: `"language_model.model.embed_tokens.weight"` (different prefix order)
- plus `qwen/qwen35_bridge.py`, `qwen_omni/*`, `qwen_audio/qwen2_audio_bridge.py`, `qwen3_asr/qwen3_asr_bridge.py`, `nemotron_vl/nemotron_vl_bridge.py`, `ministral3/ministral3_bridge.py`

For any of these with `tie_word_embeddings=True` (a real, common config, e.g. the smaller dense Qwen VL variants), export falls into the generic `else` branch: the embedding tensor is emitted once under its own name, and `lm_head.weight` is silently never emitted, even if the source checkpoint had it. No error, no warning, just a missing key in the exported HF checkpoint.

## Fix

Detect the tied embedding task via `task.global_param_name.endswith("embedding.word_embeddings.weight")` instead of matching the HF-side name. This is the same convention the file already uses a few hundred lines up, in `_should_skip_mtp_duplicate_embedding_export`. I grepped every bridge under `models/` for its Megatron-side embedding key and confirmed this holds everywhere (`embedding.word_embeddings.weight` for plain LLMs, `language_model.embedding.word_embeddings.weight` or similar for VLMs), so the new condition is a strict superset of the old one and plain-LLM behavior is unchanged.

## Verification

Added `test_stream_weights_megatron_to_hf_emits_lm_head_for_nested_vlm_embedding_name`, mirroring the existing `test_stream_weights_megatron_to_hf_transforms_tied_aliases_independently` but with a VLM-style nested HF name (`"model.language_model.embed_tokens.weight"`), to cover the case the old code missed.

I could not run the test suite in my environment: importing `megatron.core` pulls in `triton`, which has no Windows wheel available. I traced the exact control flow by hand and dry-ran the old vs. new condition against the real Megatron-side/HF-side name pairs from three representative bridges (plain LLM, qwen3_vl-style VLM, gemma3_vl's differently-ordered prefix) to confirm the logic, but this needs a real CI run to confirm the test itself passes.

### DIFF
--- a/src/megatron/bridge/models/conversion/model_bridge.py
+++ b/src/megatron/bridge/models/conversion/model_bridge.py
@@ -1391,9 +1391,16 @@ class MegatronModelBridge(
                     else:
                         hf_name = hf_name[:suffix_pos] + ".base_layer" + hf_name[suffix_pos:]
 
-                # Handle tied embeddings case
-                # TODO(yuya): fix this hard coded naming
-                if embeddings_are_tied and hf_name == "model.embed_tokens.weight":
+                # Handle tied embeddings case. Detect via the Megatron-side param
+                # name, which is consistently "...embedding.word_embeddings.weight"
+                # across every bridge (see _should_skip_mtp_duplicate_embedding_export
+                # above for the same convention), rather than the HF-side name: that
+                # varies per bridge, e.g. "model.embed_tokens.weight" for plain LLMs
+                # vs. "model.language_model.embed_tokens.weight" for VLM bridges like
+                # qwen3_vl/qwen35/gemma3_vl/glm_45v/etc. The literal HF-name match
+                # this replaced silently skipped lm_head.weight export for all of
+                # those when embeddings are tied.
+                if embeddings_are_tied and task.global_param_name.endswith("embedding.word_embeddings.weight"):
                     emit_lm_head = isinstance(hf_pretrained, PretrainedConfig)
                     if hasattr(hf_pretrained, "state") and hasattr(hf_pretrained.state, "source"):
                         expected_keys = hf_pretrained.state.source.get_all_keys()
--- a/tests/unit_tests/models/test_model_bridge.py
+++ b/tests/unit_tests/models/test_model_bridge.py
@@ -413,3 +413,61 @@ def test_stream_weights_megatron_to_hf_transforms_tied_aliases_independently(mon
         "lm_head.weight.scale",
     ]
     assert weights[0].weight.data_ptr() != weights[2].weight.data_ptr()
+
+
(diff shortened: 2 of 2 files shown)


yaoyu-33:
Thanks for reporting this and for adding the nested-VLM regression test.

We extracted the related tied-alias fix from #4805 into the focused #4884 and generalized it through the bridge mapping registry. The comparison showed that the two original approaches covered complementary cases:

- the #4805 suffix logic handled `thinker.model.embed_tokens.weight` → `thinker.lm_head.weight`, but mis-derived `model.language_model.embed_tokens.weight`
- this PR detects the correct Megatron embedding task, but still checks/emits only the root `lm_head.weight`, so it misses `thinker.lm_head.weight`, `language_model.lm_head.weight`, and other mapped output names

#4884 now identifies the Megatron embedding task, derives its sibling `output_layer.weight`, and resolves the actual HF output name from the model bridge. Its tests cover the nested VLM case reported here as well as component-prefixed and nonstandard embedding names. I also credited your report in the PR description.

To avoid landing two conflicting changes in the same common conversion path, let us track the consolidated fix in #4884. I will leave this PR open until #4884 completes review and CI.

ErenAta16:
Appreciate the thorough follow-up, and you're right about the gap. My fix got the detection side generalized (matching on global_param_name instead of the HF-side string) but still hardcoded the emitted alias to "lm_head.weight" itself, which breaks for anything with a component prefix like thinker.lm_head.weight. Resolving the actual alias through the mapping registry in #4884 is the correct fix, not a workaround around the same limitation. Reviewed and approved #4884, happy to have this one close in its favor once it lands.

ErenAta16:
#4884 merged, closing this in its favor rather than resolving the conflict — no need to carry two PRs for the same underlying issue now that the more complete fix has landed.

PR #4884, opened 2026-07-15
fix(conversion): resolve tied output aliases from parameter mappings

yaoyu-33:
## Summary

- extract the tied-embedding export fix from #4805 into a focused change
- identify tied embeddings by their Megatron parameter name and resolve the corresponding HF output alias through the bridge mapping registry
- preserve component-prefixed, nested-language-model, root-head, and nonstandard HF naming schemes without inventing aliases that are absent from the source checkpoint

This also covers the nested VLM naming problem reported in #4883. Thanks to @ErenAta16 for identifying that additional failure mode.

## Tests

- `git diff --check`
- targeted unit-test collection was attempted locally, but macOS ARM cannot install the pinned Linux-only `nvidia-resiliency-ext`/Triton runtime; Linux CI is required for the actual pytest run

## Scope

Only the `model_bridge.py` fix and its unit tests were extracted. The unrelated validation-document and generation-config changes from the original #4805 commit are intentionally excluded.

### DIFF
--- a/src/megatron/bridge/models/conversion/model_bridge.py
+++ b/src/megatron/bridge/models/conversion/model_bridge.py
@@ -1287,6 +1287,9 @@ class MegatronModelBridge(
             megatron_model = [megatron_model]
 
         unwrapped_model_list = unwrap_model(megatron_model)
+        self.hf_pretrained = hf_pretrained
+        self.hf_config = hf_pretrained.config if hasattr(hf_pretrained, "config") else hf_pretrained
+
         # Use provided conversion tasks or build them
         if conversion_tasks is None:
             conversion_tasks = self.build_conversion_tasks(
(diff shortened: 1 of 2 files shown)


ErenAta16 (approved):
This is the right fix, more complete than #4883. The gap you called out is real: my PR fixed the detection side (task.global_param_name.endswith("embedding.word_embeddings.weight") correctly identifies the tied embedding task regardless of prefix) but still hardcoded the emission side to the literal string "lm_head.weight", which is wrong for anything with a component prefix like thinker.lm_head.weight.

_get_tied_output_hf_name resolving through mapping_registry().megatron_to_hf_lookup(f"{megatron_prefix}output_layer.weight") is the correct general solution, since it derives the actual expected alias from the bridge's own mapping definitions instead of assuming a fixed name. The guards look right too: returning None when there's no output mapping (nothing to invent) or when the resolved output name equals the embedding's own name (would just re-emit itself), and the existing expected_keys check still gates whether the alias actually gets emitted, so this doesn't invent keys absent from the source checkpoint either.

Test coverage is thorough, five naming scenarios (plain, thinker.-prefixed, nested-root-head, nested-language-model, nonstandard-embedding-name) plus the two no-output-mapping/output-not-in-source cases. Only note I don't have a change to request: the ValueError branch's condition (task.global_param_name.endswith("output_layer.weight") or hf_name == tied_output_hf_name) is worth double-checking against a real MTP-duplicate-embedding scenario if one exists in the test suite, since that path also touches global_param_name endings, but that's a "did you already think about this" question, not a blocker.

Thanks for tracking this down properly and for the credit in the description. Good with #4883 staying open only until this lands.

Issue #4351, opened 2026-06-14
[feature] Support offline tokenized + index-only packed mode for SFT datasets


muyihao:
### User problem

  Megatron-Bridge currently supports SFT sequence packing through `FinetuningDatasetBuilder` and `prepare_packed_sequence_data()`.

  The current flow prepares packed data before training by:

  1. loading the raw SFT dataset;
  2. tokenizing examples through the SFT dataset path;
  3. building a first-fit packing strategy;
  4. materializing packed rows to disk.

  The materialized packed format currently stores packed token rows, for example:

  - `input_ids`
  - `loss_mask`
  - `seq_start_id`

  in packed parquet format, with `.npy` also supported as a legacy format.

  This works well for normal SFT use cases, but for large-scale SFT / mid-training / CPT-style workloads, materializing every packed token row has several limitations:

  - the packed output duplicates token data that may already exist in an offline tokenized dataset;
  - changing `packed_sequence_size`, packing seed, or packing algorithm requires regenerating large packed token files;
  - it is harder to reuse the same tokenized corpus across different packing strategies;

### Desired outcome

Add an `index_only` output mode for SFT sequence packing, splitting preparation into two reusable stages plus a lazy training-time reader.

  #### Stage 1: offline tokenization

  Add or expose a preprocessing path such as:

  ```python
  prepare_tokenized_data(...)
  ```

  This stage writes a reusable tokenized dataset, for example in:

  - a Bridge-native mmap format;
  - an AIRecord-compatible tokenized format;
  - or another random-access tokenized sample store.

  Each tokenized sample would contain the fields needed by SFT training, such as:

  - `input_ids`
  - `loss_mask`
  - optional metadata required for chat / answer-only-loss semantics

  #### Stage 2: index-only packing

  Add a packing preparation path such as:

  ```python
  prepare_packed_index_data(...)
  ```

  When `output_format="index_only"` is enabled, this stage should only write packing metadata / indices instead of materialized packed token rows. For each packed sample, the index could store entries like:

  ```text
  pack_id -> [
    (doc_index, start_offset, end_offset),
    (doc_index, start_offset, end_offset),
    ...
  ]
  ```

  The packing index should preserve the existing semantics of:

  - `packed_sequence_size`
  - `first_fit_shuffle`
  - `first_fit_decreasing`
  - deterministic seed behavior
  - `seq_start_id` / `cu_seqlens` construction
  - `loss_mask` alignment
[...]

PR #4891, opened 2026-07-16
[data] fix: honor seed during offline packing

yaoyu-33:
## What does this PR do?

Offline GPT SFT packing now uses the configured dataset seed for both randomized packing steps. Previously, `first_fit_shuffle` and the same-length permutation consumed ambient NumPy state, so identical inputs and `seed=777` could produce different groupings and even different packed-row counts depending on unrelated earlier NumPy calls. That makes offline-packed training and the cached artifact non-reproducible through supported builder and standalone-script paths.

Related context: #4351 expects deterministic seed-driven index preparation but does not fix this offline-packing path.

## Root cause and fix

`prepare_gpt_sft_packed_data` passed `seed` into dataset tokenization but not into the module-global NumPy draws used by `create_packing_strategy` and `fill_packing_strategy`.

The fix snapshots NumPy's global state, seeds the two packing operations with the configured value, and restores the caller's state in a `finally` block. This keeps the existing APIs and packing algorithms unchanged.

## Validation

The new regression test runs identical input and configured seed under two different ambient NumPy states.

- Before the fix: `uv run python -m pytest tests/unit_tests/data/packing/test_offline.py::test_configured_seed_controls_offline_packing -q` — **failed**; one run produced 2 packed rows and the other produced 3.
- After the fix: the same command — **1 passed**.
- Adjacent tests: `uv run python -m pytest tests/unit_tests/data/packing/test_algorithms.py tests/unit_tests/data/packing/test_offline.py tests/unit_tests/data/builders/test_gpt_sft_config.py tests/unit_tests/data/builders/test_gpt_sft_builder.py tests/unit_tests/scripts/test_prepare_gpt_sft_packed_data.py -q` — **67 passed**.
- `git diff --check` — passed.

## Scope

This changes only CPU-side offline GPT SFT materialization. It does not change public APIs, packing algorithms, GPU/distributed execution, or make claims about model quality, convergence, or performance.

### DIFF
--- a/src/megatron/bridge/data/packing/offline.py
+++ b/src/megatron/bridge/data/packing/offline.py
@@ -267,8 +267,13 @@ def prepare_gpt_sft_packed_data(
     )
     sequences, histogram = create_hist(dataset, max_seq_length)
 
-    assignments, packing_metadata = create_packing_strategy(histogram, packed_sequence_size, packing_algorithm)
-    output_data = fill_packing_strategy(assignments, sequences, packed_sequence_size, tokenizer.eos_id)
+    random_state = np.random.get_state()
+    np.random.seed(seed)
+    try:
+        assignments, packing_metadata = create_packing_strategy(histogram, packed_sequence_size, packing_algorithm)
+        output_data = fill_packing_strategy(assignments, sequences, packed_sequence_size, tokenizer.eos_id)
+    finally:
+        np.random.set_state(random_state)
 
     # save output data
     output_path_str = str(output_path)
--- a/tests/unit_tests/data/packing/test_offline.py
+++ b/tests/unit_tests/data/packing/test_offline.py
@@ -15,6 +15,7 @@
 from pathlib import Path
 from types import SimpleNamespace
 
+import numpy as np
 import pytest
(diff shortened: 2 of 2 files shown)


yaoyu-33:
CI triage for the failed `Launch_Unit_Tests_Core` job:

- The packing change and its new regression test did not fail.
- The job completed **7,328 passing tests** and then reported nine setup errors from the module-scoped Qwen3-VL fixture.
- Every error was the same unrelated timeout: `Qwen3VLMoeConfig.from_pretrained("Qwen/Qwen3-VL-30B-A3B-Instruct")` attempted an Hugging Face metadata request and blocked in an SSL read until the 50-second pytest timeout.
- The AWS runner started the container without the shared test-data mount, so the config was not available from the expected cache. A concurrent unrelated PR hit the same Qwen3-VL fixture timeout, confirming this is not caused by the offline-packing diff.
- `Nemo_CICD_Test` is only the downstream summary failure.

The two-file diff remains review-clean, so I am rerunning CI on the unchanged exact SHA rather than adding an unrelated Qwen test change to this data PR.

PR #4909, opened 2026-07-16
feat(model): add model support verification cards

yaoyu-33:
# What does this PR do?

Adds concise, agent-readable model verification cards under `examples/model_verification_cards/` for four variants:

- Qwen/Qwen3-8B
- Qwen/Qwen3-30B-A3B
- moonshotai/Moonlight-16B-A3B
- nvidia/NVIDIA-Nemotron-3-Nano-4B-BF16

It also adds the repo-local create-model-verification-card skill, its deterministic validator, and a helper for exact-length, byte-stable HF inference verification.

The cards distinguish bounded convergence evidence from benchmark-only performance evidence, pin immutable Hugging Face and dataset revisions, preserve recipe-owned GBS/MBS/LR contracts, require saved post-setup runtime configuration, and keep private executor wiring out of tracked cards and PR text.

# Card contract

- Keep all twelve core items; include `pretrain_performance` only when the exact variant has a canonical public performance recipe.
- Start untuned cards with the required sanity-check disclaimer; scope tuned claims to exact `pretrain_performance.<hardware>` leaves.
- Use portable public launchers and short logical `work/...` paths.
- Pin immutable model, tokenizer, and dataset revisions plus split, seeds, and packing alignment.
- Keep recipe GBS/MBS/LR unchanged in verification commands.
- Record initial/final loss plus final-ten mean step time and model TFLOP/s/GPU for every verified training item.
- Nest hardware-dependent training and SFT export evidence under public accelerator keys (for example, `pretrain.H100`), with resume/export dependencies resolved only within the same hardware target.
- Require exact-length, byte-stable inference output from two executions.
- Verify manual forward correlation with an exact next-token match and cosine similarity at least 0.99; maximum and mean absolute logit differences remain report-only diagnostics.
- Resume directly from the uninterrupted bounded-pretrain step-50 checkpoint to step 100 in a distinct output root.
- Treat submission or a short smoke as evidence-in-progress, not verification.

# Current verification status

The branch remains a draft while the final clean-contract resume/export audits finish. Current head: `fbd1ad6a4`.

| Variant | Tracked card | Completed evidence awaiting final audit/card promotion | Remaining runtime gate |
|---|---:|---|---|
| Qwen3-8B | 10/12 | Bounded source checkpoints | Direct checkpoint resume and frozen audit |
| Qwen3-30B-A3B | 10/13 | Full SFT, strict post-SFT export/inference, PEFT, supervised audit | Frozen export audit and card promotion |
| Moonlight-16B-A3B | 7/12 | Full SFT and PEFT | Direct resume, corrected post-SFT export/inference, and frozen audits |
| Nemotron 3 Nano 4B | 12/12 | Complete | None |

The three long-context items are verified from one coherent clean-run evidence cohort:

| Variant | Loss | Final-ten step time | Final-ten TFLOP/s/GPU | Packing | Actual supervised tokens |
|---|---:|---:|---:|---:|---:|
| Qwen3-8B | 1.611622 -> 1.490618 | 71668.780 ms | 34.080 | 99.97% | 3,444,917 |
| Qwen3-30B-A3B | 1.645009 -> 1.468103 | 142663.710 ms | 26.120 | 99.28% | 13,573,663 |
| Moonlight-16B-A3B | 1.306440 -> 1.228019 | 60599.310 ms | 40.840 | 99.69% | 13,765,732 |

Each completed exactly 20 optimizer steps with finite losses, zero skipped/NaN iterations, a persisted runtime config, and independently reconstructed packed-sample accounting.

# GPU export config correction

The final commit fixes GPU export for checkpoints whose training-time config differs from the original Hugging Face config. Export retains the existing state-backed reference wrapper and shard map, but replaces its config with the checkpoint-derived Hugging Face config before saving. This preserves strict/distributed streaming behavior while making serialized config shapes agree with the checkpoint tensors.

The focused GPU-export tests pass 10/10 on both macOS and the fixed Linux container. Repository-wide pre-commit passes in that Linux container. The current public docs, secrets, and DCO checks are green.

# Revision hardening and validation

- Hugging Face model and tokenizer revisions are persisted by the affected recipes instead of relying on mutable defaults.
- The revision-hardening patch passed 184 targeted tests and repository-wide pre-commit.
- The current long-context/card update passed all three card validators, tokenizer unit tests, Ruff checks, the CI-equivalent secrets scan, and repository-wide pre-commit.
- Two public immutable revision literals triggered the generic high-entropy detector; narrow inline allowlist annotations scope those known false positives to the exact literals.
- `git diff --check` passes, and independent evidence reviews reproduced every promoted metric and supervised-token count.

# Remaining work

The final bounded-pretrain direct resumes and corrected Moonlight export are queued from frozen, independently reviewed launchers. Execution state alone does not promote an item: each remains unverified until its exact optimizer-step set, finite-loss/no-skip gates, saved runtime config, checkpoint or adapter artifacts, packing/supervised-token evidence, resume sentinels, strict load, deterministic inference, and frozen evidence audit all pass. The cards, PR status, and Linear issues will be promoted only after those gates complete.


examples/model_verification_cards/qwen3-8b/card.yaml
```diff
@@ -0,0 +1,86 @@
...
+# last_verified: YYYY-MM-DD or null
+# verified requires an immutable hf_revision, a concrete expected_result,
+# non-null last_verified, and non-empty evidence.
+# unverified may retain an older last_verified date when validation is stale.
+
+title: qwen3_8b
+model:
+  hf_id: Qwen/Qwen3-8B
+  hf_revision: null
+  architecture: Qwen3ForCausalLM
```
cuichenx, line 27: add min transformers version

examples/model_verification_cards/qwen3-8b/card.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+  megatron_to_hf:
+    status: unverified
+    last_verified: null
+    expected_result: >
+      The full checkpoint exports and reloads, and the Megatron/exported-HF
+      logit delta is less than or equal to TBD.
+    evidence: []
+
+  inference:
```
cuichenx, line 110: separate between megatron inference and manual forward pass results

examples/model_verification_cards/qwen3-8b/card.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+  pretrain:
+    status: unverified
+    last_verified: null
+    expected_result: >
+      qwen3_8b_pretrain_4gpu_h100_bf16_config runs for TBD steps and reaches
+      loss TBD.
+    evidence: []
+
+  sft:
```
cuichenx, line 164: add packed seq and CP

model_cards/qwen3-8b/card.yaml
```diff
@@ -0,0 +1,344 @@
...
+# expected_result, and non-null last_verified.
+# unverified may retain an older last_verified date when validation is stale.
+# Every training item records the GPU type used for verification.
+
+title: qwen3_8b
+model:
+  hf_id: Qwen/Qwen3-8B
+  hf_revision: b968826d9c46dd6066d109eabc6255188de91218  # pragma: allowlist secret
+  architecture: Qwen3ForCausalLM
+release: "26.08"
```
yaoyu-33: no need, put in container information

yaoyu-33: put in base nemo container + mounted commit if mounted.

examples/model_verification_cards/qwen3-8b/card.yaml
```diff
@@ -0,0 +1,344 @@
...
+      ./scripts/conversion/convert.sh export --executor slurm --device gpu
+      --nodes 1 --gpus-per-node 4 --mount "$PWD:/opt/Megatron-Bridge"
+      --mount "$MODEL_CARD_WORKDIR:/workspace" --mount "$HF_HOME"
+      --env HF_HOME --env HF_TOKEN
+      --hf-model Qwen/Qwen3-8B
+      --megatron-path /workspace/qwen3-8b/imported-megatron/iter_0000000
+      --hf-path /workspace/qwen3-8b/hf-export --torch-dtype bfloat16
+      --export-weight-dtype bfloat16 --distributed-save --tp 4
+    last_verified: 2026-07-16
+    expected_result: >
```
yaoyu-33, line 87: need additional hf export inference

model_cards/qwen3-8b/card.yaml
```diff
@@ -0,0 +1,344 @@
...
+  pretrain_performance:
+    status: verified
+    gpu_type: H100
+    command: >
+      ./scripts/training/train.sh --nodes 1 --gpus-per-node 4
+      --experiment-name qwen3-8b-pretrain-performance
+      --mount "$PWD:/opt/Megatron-Bridge" --mount "$MODEL_CARD_WORKDIR:/workspace"
+      --mount "$HF_HOME" --env HF_HOME --env HF_TOKEN
+      --recipe qwen3_8b_pretrain_4gpu_h100_bf16_config --mode pretrain
+      --dataset mock --max_steps 100 --global_batch_size 32
```
yaoyu-33: should not change mbs/gbs from perf recipe

model_cards/qwen3-8b/card.yaml
```diff
@@ -0,0 +1,344 @@
...
+    enabled_features: {}
+    command: >
+      ./scripts/training/train.sh --nodes 1 --gpus-per-node 4 --time 02:00:00
+      --experiment-name qwen3-8b-pretrain-rp2-reference
+      --mount "$PWD:/opt/Megatron-Bridge" --mount "$MODEL_CARD_WORKDIR:/workspace"
+      --mount "$PRETRAIN_DATA_ROOT:/data/rp2"
+      --mount "$HF_HOME" --env HF_HOME --env HF_TOKEN
+      --recipe qwen3_8b_pretrain_4gpu_h100_bf16_config --mode pretrain
+      --dataset megatron-indexed --deterministic --seq_length 4096
+      --max_steps 400 --global_batch_size 32 --micro_batch_size 2
```
yaoyu-33: do not change gbs/mbs from recipe

examples/model_verification_cards/qwen3-8b/card.yaml
yaoyu-33: Addressed in 89375d301: added model.min_transformers_version 5.8.1 from the project dependency lower bound, and the validator now requires a semantic version.

examples/model_verification_cards/qwen3-8b/card.yaml
yaoyu-33: Addressed in 89375d301: added a separate manual_forward_pass item and kept inference for deterministic text generation. A fresh TP-aligned eight-token comparison matched next-token ID 12095 on HF and Megatron with cosine similarity 0.999969; max/mean absolute logit differences were 0.187500/0.030649.

examples/model_verification_cards/qwen3-8b/card.yaml
yaoyu-33: Addressed: sft_long_context verifies sequence length 32768 with offline packing and CP=2 together; the completed 20-step result and final-10 metrics are recorded.

model_cards/qwen3-8b/card.yaml
yaoyu-33: Addressed in 89375d301: removed release and added explicit verification provenance with the public base container and mounted Bridge commit. The validation image was based on nvcr.io/nvidia/pytorch:26.04-py3 rather than a NeMo tag, so the card records that actual public base; private image and mount details remain excluded.

examples/model_verification_cards/qwen3-8b/card.yaml
yaoyu-33: Addressed: sft_export_inference now depends on SFT and contains an ordered two-command flow: export the final step-100 full-model checkpoint, then reload it with Transformers and verify the exact deterministic 45-token HF completion.

model_cards/qwen3-8b/card.yaml
yaoyu-33: Addressed: all GBS/MBS overrides were removed from card commands. The selected recipe owns both values, and the validator rejects command-line or config-path batch-size overrides.

model_cards/qwen3-8b/card.yaml
yaoyu-33: Addressed: all GBS/MBS overrides were removed from card commands. The selected recipe owns both values, and the validator rejects command-line or config-path batch-size overrides.

yaoyu-33:
Updated the Qwen3-30B-A3B card through `0cbd7185a` with the exact performance-recipe topology and batch for functional RP2 pretraining. SFT, long-context SFT, and LoRA are unchanged.

- Pretrain now uses `qwen3_30b_a3b_pretrain_16gpu_h100_bf16_config` on 16×H100. The recipe supplies TP/PP/CP/EP = `1/1/1/16` and MBS/GBS = `1/1024`; the card does not override any of them.
- The real-data run keeps natural routing while retaining HybridEP and Transformer Engine CUDA graphs for `moe_router` and `moe_preprocess`. It completed 400/400 steps with zero skipped/NaN iterations, loss `12.41145 -> 3.917895`, last-10 `28211.570 ms / 213.780 TFLOPS/GPU`, and complete step-200/400 checkpoints.
- LR remains `3e-4`, minimum LR `3e-5`, with 40 warmup steps and cosine decay through step 400. GBS increased 32×, but scaling LR linearly (`9.6e-3`) or by square root (~`1.70e-3`) was rejected as unsupported and too aggressive; a 100-step probe crossed peak LR stably before the full run.
- Direct resume restores optimizer/RNG from the reference step-200 checkpoint and executes steps 201–400 in a distinct output root. Step 201 matches exactly; step 400 is `3.912371` vs `3.917895`, a `0.1410%` relative difference, so it passes the accepted 1% gate. `train.empty_unused_memory_level=2` is resume-only allocator hygiene needed to preserve NCCL allocation headroom after checkpoint loading; it does not change training math or the recipe batch/topology.
- Functional pretrain throughput is now `3.00×` the previous card result and reaches `89.0%` of the separate mock-data performance result (`240.110 TFLOPS/GPU`). The remaining gap is consistent with real RP2 plus natural expert imbalance and correctness/checkpoint overhead; it is not a TFLOPS accounting mismatch.

Checks: bundled card validator, `git diff --check`, a 30-test training-launcher target, independent full-diff/privacy review, and all six isolated pre-commit hooks pass. The standard project-sync pre-commit invocation remains blocked before hooks on macOS ARM by the Linux-only `nvidia-resiliency-ext` wheel.

yaoyu-33:
Superseding the earlier Qwen3-30B-A3B functional-pretrain figures with the current default-recipe verification at Bridge `5b9d9cf501193277e1ca47a99b97c640b8f39f90`.

- The generic library `qwen3_30b_a3b_pretrain_config` owns TP/PP/CP/EP/ETP = `1/1/1/16/1`, MBS/GBS = `1/1024`, HybridEP, Transformer Engine CUDA graphs for `moe_router` and `moe_preprocess`, and the BF16 Adam precision-aware optimizer behavior previously injected by the performance launcher. The H100 BF16 performance wrapper calls this default and keeps only benchmark-specific behavior.
- The bounded real-RP2 reference completed 100/100 steps with LR `3e-4 -> 3e-5`, 40 warmup steps, decay through step 100, and checkpoints at steps 50 and 100. Loss was `12.411450 -> 6.139116`; steps 91–100 averaged `30289.550 ms` and `199.120 TFLOPS/GPU`. The expected early pre-clip spike peaked at `124.849` on step 4 with `clip_grad=1.0`; skipped/NaN iterations were `0/0`.
- This final window is 3.75% above the earlier default-recipe reference without the consolidated optimizer flag and within about 1.2% of the earlier performance-runner result at the same steps. Because these runs used different node pairs, this is evidence that the apparent throughput regression is largely closed, not a clean single-node A/B attribution.
- Direct resume restored TrainState, RNG, and the `dp_reshardable` distributed optimizer from the new reference step-50 checkpoint, began at step 51, and completed step 100 in a distinct output root. Step 51 matched the uninterrupted reference exactly at loss `6.989006`; resumed step 100 was `6.145390` versus reference `6.139116`, a `0.102197%` relative difference within the 1% gate. Steps 91–100 averaged `31142.650 ms` and `193.640 TFLOPS/GPU`, with skipped/NaN iterations `0/0`.
- The reference step-50/100 checkpoints and resumed step-100 checkpoint each finalized at about 399 GiB with all 16 distributed shards plus metadata, optimizer/RNG train state, run config, and tokenizer artifacts.
- The separate 50-step mock-data `pretrain_performance` evidence remains unchanged; it continues to validate the exact performance wrapper rather than the functional RP2 workload.

Validation for this delta: two focused Qwen default/perf inheritance tests pass in the verification container, the Qwen card validator passes, all 9 resume-provenance validator tests pass, `git diff --check` passes, and full pre-commit passes. The earlier broad recipe validation remains 75 targeted Qwen recipe tests and 677 exhaustive recipe/setup factory tests.

### Qwen3-8B bounded pretrain/resume refresh

- The unchanged library `qwen3_8b_pretrain_4gpu_h100_bf16_config` owns TP/PP/CP = `4/1/1`, MBS/GBS = `2/32`, BF16 training, FP32 optimizer state, and no CUDA graphs. Dense Qwen3-8B has no canonical public performance recipe, so this remains a functional library-recipe check.
- The real-RP2 reference completed exactly 100 steps with LR `3e-4 -> 3e-5`, 40 warmup steps, cosine decay through step 100, and complete checkpoints at steps 50 and 100. Loss was `12.762130 -> 6.374147`; steps 91-100 averaged `4756.430 ms` and `337.820 TFLOPS/GPU`, with skipped/NaN iterations `0/0`.
- Direct resume restored optimizer and RNG state from step 50, executed exactly steps 51-100, and matched the uninterrupted reference exactly for samples, LR, loss, gradient norm, skipped count, and NaN count at every resumed step. Its final loss was `6.374147`; steps 91-100 averaged `4729.320 ms` and `339.760 TFLOPS/GPU`.
- Because this recipe's MBS/GBS `2/32` differs from the comparison-cohort `1/1024` contract, these figures are bounded support-verification evidence, not an apples-to-apples cross-model convergence comparison.
- Validation for commit `9937013344dd5ffa81412469966ed1ee74a13368`: the card/privacy validator passes, all 9 focused resume-provenance validator tests pass, `git diff --check` passes, and full pre-commit passes.

yaoyu-33:
> Superseded by the current convergence-contract audit in `abaf10748`. The
> completed runs below remain useful support history, but the training items no
> longer count as current `qwen3_30b_a3b_convergence_v1` evidence because their
> executed commands did not freeze the immutable revision, packing alignment,
> and current recipe contract. The PR body is authoritative; the replacement
> workloads remain in progress.

Moonlight-16B-A3B was 12/12 verified under the earlier support contract in
`cc7fcfc04`.

- CPU and GPU HF↔Megatron conversion artifacts reload and pass the recorded exact configuration/tensor gates.
- Clean-checkout HF/Megatron parity selects token ID `17374` (`" Paris"`) on both paths with cosine `0.999893`; max/mean absolute logit differences are `0.472656 / 0.074414`.
- Two independent Megatron generations produce the same exact 32-token completion recorded in the card.
- Full SFT, post-SFT export/HF inference, long-context SFT with CP2+offline packing, and LoRA all completed with finite losses, zero skipped/NaN iterations, and final-ten metrics under that earlier contract.
- Uninterrupted RP2 pretraining completed 400 steps at `12.41085 -> 3.413953`, with final-ten `25,895.91 ms / 86.84 TFLOPS/GPU` and complete step-200/400 checkpoints.
- Direct resume restored optimizer/RNG from step 200 and executed exactly steps 201–400. Step 201 matched exactly; step 400 was `3.413714` vs `3.413953` (`0.007001%` relative difference), well inside the 1% gate. Resumed final-ten was `25,707.45 ms / 87.47 TFLOPS/GPU`.

Moonlight has no exact canonical performance recipe, so its card intentionally has no `pretrain_performance` item. Public commands retain recipe-owned GBS/MBS and contain no private runtime wiring.

yaoyu-33:
Qwen3-30B-A3B pretrain/resume refresh is complete on the current default recipe.

- Audited the historical 400-step reference + 200→400 resume and the previous 100-step run.
- Re-ran the current recipe at `5b9d9cf501193277e1ca47a99b97c640b8f39f90`, including its precision-aware optimizer, for 100 RP2 steps with checkpoints at 50/100.
- Reference: loss `12.411450 → 6.139116`; steps 91–100 average `30289.550 ms / 199.120 TFLOPS/GPU`; no skipped/NaN steps.
- Direct step-50→100 resume restored optimizer and RNG state. Step 51 matched exactly; step 100 was `6.145390` vs `6.139116` (`0.102197%`), within the 1% gate. Resumed steps 91–100 average `31142.650 ms / 193.640 TFLOPS/GPU`.
- Reference and resumed checkpoints contain all 16 distributed shards plus metadata, train state, run config, and tokenizer artifacts.
- Sequential verification time drops from `5:11:36` for the old 400/200 scheme to `1:37:37`, a `68.7%` reduction.

The public card is updated, and the model-card skill now documents 100-step reference + direct 50→100 resume as an expensive-model support-verification option (not a full-convergence claim). Full EOS details are recorded in [MB-863](https://linear.app/nvidia/issue/MB-863/qwen3-30b-a3b).

Validation: all three card validators and privacy deny terms pass; isolated pre-commit passes; `git diff --check` passes.

yaoyu-33:
Updated the model-verification and recipe-recommender skills with a shared convergence protocol before any Moonlight rerun.

- Convergence contract: data/preprocessing, packing, sequence length, GBS/MBS, tokens/steps, optimizer/LR/schedule, precision, routing/loss semantics, and PEFT definition.
- Execution/performance contract: topology/parallelism, recompute/offload, dispatcher transport, overlap, validated fusions/backends, and CUDA graphs. These are intended semantics-preserving but still require loss-sentinel checks.
- Benchmark-only: mock data, forced load balancing, changed batch/LR or timing-only schedules, and disabled correctness checks. These cannot provide convergence evidence.

Default comparison cohort:
- Pretrain: RP2 raw selection, seq 4096, GBS/MBS 1024/1, BF16, LR 3e-4 to 3e-5, warmup 40, decay 100, save 50/100, direct resume 50 to 100.
- SFT: Tulu 3 train[:10000], offline packing, seq 2048, GBS/MBS 32/1, BF16, LR 5e-6 to 0, warmup 10, decay 100.
- PEFT: same SFT data, GBS/MBS 32/1, LR 1e-4 to 0, warmup 10, decay 100, with a frozen adapter definition.

Moonlight has not been rerun yet. Its current recipes differ in pretrain GBS (2048), SFT/PEFT GBS (128), optimizer precision/hyperparameters, and PEFT adapter definition. The skills now require recipe changes to resolve those differences instead of card-level batch overrides.

Commit: ba19ea979
Internal rerun gate: https://linear.app/nvidia/issue/MB-897/moonlight-16b-a3b-model-support-card

yaoyu-33:
Nemotron 3 Nano 4B verification is complete at card commit `f04e941906e05d4ef69c4c1e5fc7b3b78e12c844`. Runtime evidence is pinned to clean Bridge commit `0e6c1837edeb442b07c9175b855b3d288c5cad33`, MCore `61f31145dfb4cfaa7d285df12aa9abffb22ff485`, and immutable HF revision `dfaf35de3e30f1867dd8dbc38a7fc9fb52d3914f`.

| Core item | Status | Result |
| --- | --- | --- |
| HF → Megatron, CPU | verified | Reloadable checkpoint; paired round trip exact |
| HF → Megatron, GPU | verified | Reloadable checkpoint; paired round trip exact |
| Megatron → HF, CPU | verified | 263/263 tensors and 3,973,556,832 parameters bitwise exact; native reload |
| Megatron → HF, GPU | verified | Same exact audit; generation config byte-exact; all 21 Mamba A_log/D/dt_bias/out_proj groups exact |
| Manual HF/Megatron parity | verified | Same next token 6993 (` Paris`); cosine 0.999982; max/mean logit difference 0.093750/0.018522 |
| Deterministic Megatron inference twice | verified | Exactly 2 tokens twice, identical completion ` Paris.` |
| Bounded random-init pretraining | verified | 100/100 steps; loss 12.432500 → 5.511951; last-10 26,775.11 ms and 402.51 TFLOPS/GPU; finite, no skipped/NaN iterations; midpoint/final artifacts and final reload verified |
| Full SFT | verified | 100/100 steps; loss 1.839282 → 1.234906; last-10 627.0 ms and 262.89 TFLOPS/GPU; 99.27% packed; final checkpoint reload verified |
| Final SFT export + HF inference twice | verified | BF16 export with 263 weights; native reload; exactly 45 identical generated tokens twice |
| 32K packed SFT with CP=2 | verified | TP2/CP2/SP, 100/100 steps; loss 1.658660 → 1.100197; last-10 5,575.59 ms and 142.06 TFLOPS/GPU; 99.35% packed; final reload verified |
| PEFT | verified | LoRA 100/100 steps; loss 1.769892 → 1.256916; last-10 381.69 ms and 431.90 TFLOPS/GPU; 99.36% packed; adapter reload verified |
| Direct midpoint resume | verified | Restored step 50 and ran exactly 51–100; step 51 matched exactly; step-100 loss 5.512038 vs 5.511951 reference (0.00158%); final checkpoint reload verified |

The card is at `model_cards/nemotron-3-nano-4b/card.yaml`. The exact 4B variant has no canonical public performance recipe, so `pretrain_performance` is intentionally omitted.

Sanitized fixes covered immutable HF-revision plumbing, strict generation-config preservation, native Transformers loading for the fixed Nemotron-H implementation, Mamba chunk-size mapping, real Megatron inference context/InferenceMode with TP-aware output handling, and opt-in synchronous training via `--wait`. The first TP1/CP2 32K attempt exceeded memory; the semantics-preserving TP2/CP2/SP topology completed all 100 steps. Continual pretraining from the released HF checkpoint and the damaged/rare embedding-row investigation remain out of scope in #4967.

Final clean gates: model-card validator plus privacy deny terms passed; targeted secret scan passed; 136 focused tests passed; `git diff --check` passed; full pre-commit passed.

scripts/conversion/run_conversion.py
```diff
@@ -120,6 +121,9 @@ def main(argv: list[str] | None = None) -> None:
     """Parse worker arguments and run checkpoint conversion."""
     args = build_parser(include_execution=False).parse_args(argv)
     _validate_args(args)
+    if args.hf_revision is not None:
+        args.hf_model = resolve_hf_model_revision(args.hf_model, args.hf_revision)
```
yaoyu-33: This pins the download by replacing the public Hub ID with a local snapshot path. The subsequent import then persists that cache path and loses the public model ID plus immutable revision in provider/tokenizer metadata. Please resolve the supplied ref once to a commit SHA, retain the public ID, load all import artifacts with that SHA, and preserve the existing rejection of `--hf-revision` for local paths.

tests/unit_tests/test_model_verification_card_validator.py
```diff
@@ -0,0 +1,272 @@
...
+from typing import Any
+
+import pytest
+import yaml
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+VALIDATOR_PATH = REPO_ROOT / "skills/create-model-verification-card/scripts/validate_card.py"
+COMPARE_PATH = REPO_ROOT / "examples/conversion/compare_hf_and_megatron/compare.py"
+CARD_PATH = REPO_ROOT / "model_cards/nemotron-3-nano-4b/card.yaml"
```
yaoyu-33: This regression gate validates only the new Nemotron card even though this PR also changes the shared schema/validator and three existing cards. Parameterize the test over every checked-in `model_cards/*/card.yaml` so shared contract changes cannot leave another card stale.

src/megatron/bridge/models/qwen/qwen3_moe_bridge.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2025-2026, NVIDIA CORPORATION.  All rights reserved.
```
yaoyu-33: Please preserve the existing copyright header exactly. Project instructions prohibit extending the year range in an existing header; only newly generated files/content should use the current year.

yaoyu-33 (commented):
Pre-fix review submitted as COMMENTED. I am addressing all three findings now and will re-review the final pushed diff; I will not approve my own PR.

src/megatron/bridge/models/hf_pretrained/base.py
```diff
@@ -211,7 +211,28 @@ def save_artifacts(
...
+
+                    source_path = original_source_path or getattr(self, "model_name_or_path", None)
+                    if source_path is None:
+                        raise
+
+                    copied_files = self._copy_custom_modeling_files(
+                        source_path=source_path,
+                        target_path=save_path,
+                        file_patterns=["generation_config.json"],
+                    )
```
yaoyu-33, line 235: This remote fallback does not forward the wrapper’s pinned revision, so a validation failure can silently copy `generation_config.json` from the repository’s default branch instead of the checkpoint revision. Thread the immutable revision through this fallback and the other remote artifact-copy calls, with a regression test asserting the Hub list/download calls receive it.

yaoyu-33 (commented):
Additional pre-fix finding submitted as COMMENTED; it is included in the same fix-and-revalidate pass.

tests/unit_tests/test_model_verification_card_validator.py
yaoyu-33: Fixed in d3808a49285e66c616c9312ea5e1187ba1e7f03b. The repository gate is now parameterized over every checked-in model_cards/*/card.yaml; the EOS focused suite validated all cards.

scripts/conversion/run_conversion.py
yaoyu-33: Fixed in d3808a49285e66c616c9312ea5e1187ba1e7f03b. Import now resolves a branch/tag/commit once to an immutable Hub SHA, retains the public model ID, forwards that SHA through model/tokenizer loading and checkpoint metadata, and preserves the local-path rejection with regression coverage.

src/megatron/bridge/models/hf_pretrained/base.py
yaoyu-33: Fixed in d3808a49285e66c616c9312ea5e1187ba1e7f03b. The pinned revision now reaches generation-config fallback plus additional/custom remote artifact listing and downloads, with a regression test asserting both Hub calls use the immutable revision.

src/megatron/bridge/models/qwen/qwen3_moe_bridge.py
yaoyu-33: Fixed in d3808a49285e66c616c9312ea5e1187ba1e7f03b. All existing copyright headers touched by the pre-fix head are restored exactly; only the newly added test file has a 2026 header.

yaoyu-33 (commented):
Final re-review of the complete PR diff against current main is complete. The four actionable pre-fix findings are fixed in d3808a49285e66c616c9312ea5e1187ba1e7f03b, their threads are resolved, the current-main overlap preserves both sides, and the authoritative focused EOS suite plus repository-wide pre-commit are green. I found no remaining actionable issue, so no honest additional inline comment is warranted.

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4909/commits/52c521427a53d1d0a7a8c0b2d07c9860a4ed7f88

PR #4946, opened 2026-07-19
fix: preserve Qwen3 routing and packed loss masks

yaoyu-33:
# What does this PR do?

Splits two runtime fixes discovered during Qwen3 model-card verification out of [#4909](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4909), so they can land independently before the card and its tooling.

- Normalize packed-Parquet `loss_mask` values to an explicit `list<int8>` Arrow column, including mixed boolean/integer inputs.
- Preserve Hugging Face Qwen3-MoE `norm_topk_prob` symmetrically through Megatron's `moe_router_pre_softmax` setting.
- Add focused unit coverage for both behaviors.

The model cards, card skill, validator, and deterministic HF inference helper remain in #4909.

# Validation

- `uv run python -m pytest tests/unit_tests/data/packing/test_parquet.py`
- `uv run python -m pytest tests/unit_tests/models/qwen/test_qwen3_moe_bridge.py`
- `uv run --no-project --with pre-commit pre-commit run --all-files`
- `git diff --check`

The targeted unit tests passed during the original Linux verification for #4909. The split branch is byte-identical for these four files; the full pre-commit suite and diff check also pass after the split.


yaoyu-33:
CI note: the non-required `copyright-check / main` failure comes from the unchanged main-branch file `tests/unit_tests/examples/test_glm_45v_slurm.py`, which lacks a copyright header. All Python files added or modified by this PR contain NVIDIA copyright notices, and the required `copyright-check-summary` check passes.

PR #4964, opened 2026-07-20
Add CLI configurability of num_moe_experts

scsudhakaran:
This PR adds `num_moe_experts` as a CLI configuration, which is required for DeepSeek-V3 proxy models running on 72×N GPUs.


bdubauski (approved):
Looks great

PR #4567, opened 2026-06-29
[feature, model]Fix Qwen3-Omni Megatron HF corr checks

hbhflw2000:
## What

This PR tightens Qwen3-Omni Megatron/HF parity support for fixed-score and local-attention validation paths.

- Select the local GPT layer spec when Qwen3-Omni is configured with the local attention backend, while keeping the Transformer Engine spec as the default when TE is available.
- Install the mRoPE-aware `Qwen3VLSelfAttention` module in the Qwen3-Omni language transformer spec.
- Add a local causal attention fallback for padded fixed-score paths that use `DotProductAttention`.
- Update dense context-parallel batch slicing to pass the current `get_batch_on_this_cp_rank(..., is_hybrid_cp=False, ...)` API.
- Avoid view aliasing in the Qwen3-VL deepstack visual update path.
- Add a small no-Ray fixed-score correlation checker for Megatron actor JSON vs. HF JSON.

## Why

Qwen3-Omni Megatron/HF parity depends on two independent pieces being correct:

- MoE EP expert parameters must map rank-local expert indices back to global expert indices before comparing or exporting against HF.
- Attention execution must preserve Qwen3-VL's mRoPE behavior in both the TE-backed path and the local/fallback path.

The EP local-to-global expert mapping is already present in the base used by this branch. This PR focuses on the second part: local-attention/fixed-score paths and a lightweight checker that makes parity regressions easier to catch.

Qwen3-Omni parity checks can exercise local attention and dense CP code paths even when the main training environment uses the regular TE-backed path. The affected paths need to:

- preserve Qwen3-VL's absolute/mRoPE attention behavior,
- match the current Megatron-Core context-parallel helper signature,
- avoid view aliasing when updating visual tokens in-place, and
- provide a lightweight way to validate actor/HF fixed-score correlation without starting the RL stack.

## Compatibility / Risk

The local-attention changes are gated by backend and module type:

- In the default community environment with TE available and a non-local attention backend, Qwen3-Omni still builds from `get_gpt_layer_with_transformer_engine_spec`.
- The local spec is selected only when `attention_backend` is `AttnBackend.local` or `"local"`, or when TE is unavailable.
- The new local causal attention fallback only runs for `DotProductAttention` with no packed-sequence params, no attention bias, and static batching. Other attention implementations continue through the existing core-attention path.

This means an environment that has TE/apex installed should continue using the TE-backed default path. The fallback mainly covers local/fixed-score validation paths and environments where TE is not available.

[...]

PR #4988, opened 2026-07-21
[model, bugfix] qwen3-omni: fix Megatron parity for external RL integration

hbhflw2000:
## Summary

This PR fixes Qwen3-Omni Megatron/HF parity issues observed in the fully-async RL path and exposes the narrow Bridge extension hooks needed by external RL runtimes.

Changes include:

- Fix Qwen3-Omni provider/config conversion parity.
- Allow runtime provider overrides and external DDP config construction.
- Expose the model hooks required by external RL training code.
- Correct Qwen3-VL RoPE/transformer-block compatibility used by the Qwen3-Omni thinker path.
- Normalize Qwen's 2D valid-token attention mask before passing it to Transformer Engine attention:
  - all-valid masks become `None`;
  - valid-token convention is converted to TE masked-token convention;
  - packed-sequence inputs retain their existing mask contract.

## Scope

This PR is independent of [#4304](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4304).

It does not enable or modify Qwen3-Omni packed-sequence or context-parallel behavior, and does not touch the packed-sequence step/model code or shared sequence utilities introduced in that PR.

## Validation

- Added unit coverage for:
  - Qwen3-Omni bridge/provider parity;
  - runtime provider override and external DDP config construction;
  - external RL model hooks;
  - Qwen TE attention-mask conversion and non-regression contracts.
- Ran a 4-node / 32-GPU fixed-sequence parity gate:
  - TP=2, PP=2, EP=4, CP=1;
  - actor vs HF logprob Pearson correlation: `0.99048`;
  - record vs replay Pearson correlation: `0.99873`;
  - all checked values were finite;
  - all four nodes completed successfully.

## Notes

The 32-GPU gate uses the non-packed CP=1 path intentionally, so it validates the parity fix without depending on packed-sequence work still under review.

## Related work

This PR follows the merged [#4567](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4567).

#4567 fixes Qwen3-Omni local-attention and fixed-score parity paths.
This PR addresses the remaining TE-backed multi-node parity and external RL integration path.

This branch has been rebased onto current `main`; it does not reintroduce the fixes from #4567.


src/megatron/bridge/models/qwen_omni/qwen3_omni_provider.py
```diff
@@ -98,9 +104,12 @@ class Qwen3OmniModelProvider(GPTModelProvider):
     moe_router_load_balancing_type: str = "aux_loss"
     moe_aux_loss_coeff: float = 1e-3
     moe_router_pre_softmax: bool = False
+    moe_router_dtype: str = "fp32"
+    moe_router_score_function: str = "softmax"
     moe_token_dispatcher_type: str = "alltoall"
     moe_permute_fusion: bool = True
     masked_softmax_fusion: bool = False
+    enable_routing_replay: bool = False
```
yaoyu-33: [P1] Use MCore's actual routing-replay config field

At the current main MCore pin, the router and inference engines read `TransformerConfig.moe_enable_routing_replay` exclusively. This new `enable_routing_replay` field is never consulted, so an external runtime can set it successfully while record/replay remains disabled—the exact path this PR intends to expose. Please remove this alias or map it to `moe_enable_routing_replay`, and add coverage that asserts the field consumed by MCore is enabled.

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/attention.py
```diff
@@ -31,6 +32,25 @@
...
+    """Translate Qwen's 2D valid-token mask to TE's masked-token convention."""
+    if (
+        not isinstance(core_attention, TEDotProductAttention)
+        or packed_seq_params is not None
+        or not isinstance(attention_mask, Tensor)
+        or attention_mask.ndim != 2
+        or attention_mask.dtype != torch.bool
+    ):
+        return attention_mask
+    if bool(attention_mask.all().item()):
```
yaoyu-33: [P1] Avoid a host scalar read in every attention layer

`attention_mask` is a CUDA tensor in real training. Calling `.all().item()` here forces a device-to-host synchronization once per transformer layer (and again during activation-checkpoint recompute), and the data-dependent scalar read makes this forward path unsafe during CUDA graph capture. Please handle the all-valid case without a host read—for example, pass the broadcast all-false mask if TE accepts it, or normalize/drop the mask once before entering the layer stack.

yaoyu-33 (commented):
Post-conflict review of the complete final diff. I found two actionable issues in the external-RL path; details are inline.

src/megatron/bridge/models/qwen_omni/qwen3_omni_provider.py
hbhflw2000: Addressed in 66e3738. The provider now exposes MCore's consumed `moe_enable_routing_replay` field directly; the unused alias was removed. I also added a focused provider test that constructs Qwen3-Omni with this field enabled.

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/attention.py
hbhflw2000: Addressed in 66e3738. The layer-local `all().item()` branch is removed. All-valid Qwen masks are now converted to the broadcast all-false mask, so the path stays on device and is safe for CUDA graph capture. The updated unit test fails if `Tensor.item()` is called.

yuekaizhang:
/ok to test [358fc8b](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4988/commits/358fc8be05853f0978428cc1fcd9c1b9e408292b)

yuekaizhang:
/ok to test [ed71178](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4988/commits/ed71178d1fba6f9046dad20a0f1c5907c3d7084d)

hbhflw2000:
Hi @yaoyu-33, this has been rebased, all comments addressed, and the full test suite is green. Could you please take a look when convenient？ Thanks

yaoyu-33:
@hbhflw2000 checking now

yaoyu-33 (approved):

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/attention.py
```diff
@@ -31,6 +32,23 @@
...
+    attention_mask: Tensor | None,
+    packed_seq_params: PackedSeqParams | None,
+) -> Tensor | None:
+    """Translate Qwen's 2D valid-token mask to TE's masked-token convention."""
+    if (
+        not isinstance(core_attention, TEDotProductAttention)
+        or packed_seq_params is not None
+        or not isinstance(attention_mask, Tensor)
+        or attention_mask.ndim != 2
+        or attention_mask.dtype != torch.bool
```
yaoyu-33, line 46: [P1] Normalize the actual integer Qwen mask before TE

The normal Qwen3-Omni path does not reach this conversion: `Qwen3OmniMoeProcessor`/the collator emits a 2-D integer valid-token mask (the fallback is [`torch.ones_like(input_ids)`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/2086ee62cf2433c132b78778d43a99ea0ac1d2bf/src/megatron/bridge/models/qwen_omni/data/collate_fn.py#L111-L112)), and the recipe forwards that mask. This dtype guard therefore returns it unchanged, so TE receives `[B, S]` integer valid-mask semantics instead of the intended `[B, 1, 1, S]` boolean masked-token semantics.

The new unit test currently codifies the gap by requiring a `torch.long` mask to remain unchanged, and the Megatron smoke only asserts `outputs is not None`, so the green suite does not establish finite outputs or mask parity. Please follow up by normalizing integral 0/1 masks (for example, `valid_mask = attention_mask != 0` before inversion/unsqueeze), changing the unit case to a padded int64 mask with the expected TE result, and asserting finite logits/parity in the smoke. This is now a post-merge follow-up because #4988 merged while the review was running.

tests/unit_tests/models/qwen_omni/test_qwen3_omni_bridge.py
```diff
@@ -145,6 +151,56 @@ def test_mapping_registry(self):
         )
         assert any("thinker.language_model.decoder.layers.*.mlp.router.weight" in name for name in mapping_names)
```
yaoyu-33, line 152: [P2] Exercise mapping semantics, not only resolved names

This test verifies registry lookup strings, but it does not execute any tensor transform. The L1 conversion is also a symmetric HF→MCore→HF round trip through the same registry, so a consistently swapped pair (for example Q/K/V or gate/up) can still round-trip; it currently covers TP=2 with PP=EP=1. The multimodal smoke does not compare HF and MCore logits. As a result, the checked-in tests do not independently support the PR body's TP=2/PP=2/EP=4 actor-vs-HF parity claim or catch swapped expert projections.

Please add an asymmetric test with distinct sentinel tensors that asserts the actual MCore fused QKV/expert slices after import (and the reverse HF tensors separately), plus at least one EP>1 mapping/forward gate. A small HF-vs-MCore finite-logit/cosine test would cover the rotary/router architecture flags. The reported 32-GPU correlations are useful external evidence, but there is no durable command/log or checked assertion here from which to reproduce them.

src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/attention.py
yaoyu-33: Additional contract evidence: the Transformer Engine API used by the pinned dependency requires `attention_mask` to be Boolean with shape `[batch, 1, 1, seqlen]`, where `True` means masked. The current integer `[batch, seqlen]` path therefore violates both the dtype and shape contract; it is not merely a theoretical dtype mismatch. A focused regression should pass a padded `torch.long` processor-style mask and assert the converted Boolean mask and finite output.

PR #5023, opened 2026-07-22
feat(training): add NVFP4 training callback with FP8 healing

michal2409:
# What does this PR do ?

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)


src/megatron/bridge/training/nvfp4_healing.py
```diff
@@ -0,0 +1,423 @@
...
+        self._restore_fp4_recipe()
+
+    # ---------------------------------------------------------------- healing
+
+    def _apply_healing(self, context: CallbackContext) -> None:
+        """Switch training to the FP8 healing recipe (and pre-quantized FP8 weights)."""
+        model = context.model
+        model_config = self._get_model_config(model)
+        self._reset_cuda_graphs()
+        if self.config.pre_quantize_base_weights:
```
yaoyu-33, line 193: Blocking: `bf16_with_nvfp4_mixed` sets `fp4_param_gather=True` (and synchronizes `fp4_param`), so the default `pre_quantize_base_weights=False` path skips the weight swap and then makes the FP4 autocast path return an FP8 recipe while parameters remain NVFP4. The latest commit message says TE rejects exactly this weight/recipe mismatch; the failing GB200 functional test was removed, while `test_recipe_switch_only_without_prequantize` still mocks this path as valid. Please either reject this configuration / require pre-quantization for FP4-stored parameters, or add a supported BF16-stored-parameter path, and cover it with a real GB200 smoke test.

src/megatron/bridge/training/nvfp4_healing.py
```diff
@@ -0,0 +1,423 @@
...
+            return
+        if context.state.train_state.step + 1 != self.config.healing_iter:
+            return
+        print_rank_0(f"NVFP4 healing: switching to FP8 ('{self.config.healing_recipe}') recipe...")
+        self._apply_healing(context)
+        self._healed = True
+
+    def on_train_end(self, context: CallbackContext) -> None:
+        """Restore the original Megatron-Core FP4 recipe function."""
+        self._restore_fp4_recipe()
```
yaoyu-33, line 184: Blocking: Bridge fires `on_train_end` at the end of `train()`, but `_pretrain()` runs the final validation/test after `train()` returns. Restoring the NVFP4 recipe here therefore evaluates the healed model under the wrong recipe; with `pre_quantize_base_weights=True`, the model now holds FP8 weights and may also hit the same recipe/weight mismatch. The patch is likewise not guaranteed to be restored on exceptions. Please scope this global monkey-patch across the full train + final-eval lifecycle and restore it in a guaranteed `finally`/fit-end cleanup point.

src/megatron/bridge/training/nvfp4_healing.py
```diff
@@ -0,0 +1,423 @@
...
+            return
+        gc.collect()
+        torch.cuda.empty_cache()
+        self._pre_quantize(context.model)
+
+    def on_train_step_end(self, context: CallbackContext) -> None:
+        """Apply the FP8 healing switch once, when ``healing_iter`` is reached."""
+        if self._healed:
+            return
+        if context.state.train_state.step + 1 != self.config.healing_iter:
```
yaoyu-33, line 176: Blocking: this exact-equality trigger is not checkpoint-resume safe. If a checkpoint restores `train_state.step >= healing_iter`, this callback starts with `_healed=False` but the condition can never be true again, so the remainder silently runs NVFP4. `on_data_init_start` runs after checkpoint load and is the right place to detect the restored step and re-establish the FP8 recipe/weights before the next forward. The existing `test_does_not_fire_past_healing_iter` currently codifies the broken behavior and should become a resume test. Please also reset/restore `_healed`, `_patched`, and `_fp8_stash` for in-process restart, where the same callback instance survives retries.

src/megatron/bridge/training/nvfp4_healing.py
```diff
@@ -0,0 +1,423 @@
...
+        try:
+            from megatron.core.full_cuda_graph import FullCudaGraphWrapper
+        except ImportError:
+            return
+        if not hasattr(FullCudaGraphWrapper, "cuda_graph"):
+            return
+        for stage in ("training", "validation"):
+            FullCudaGraphWrapper.cuda_graph[stage] = None
+            FullCudaGraphWrapper.result[stage] = None
+            if self.config.reset_cuda_graph_warmup:
```
yaoyu-33, line 248: Blocking for full-iteration CUDA graphs: after a graph has been captured, `curr_iteration` is already greater than `cuda_graph_warmup_steps`. Clearing `cuda_graph` while leaving that counter unchanged (the default here) means `FullCudaGraphWrapper.__call__` will never satisfy `curr_iteration == cuda_graph_warmup_steps`, so it runs eagerly forever instead of re-capturing. The current unit test explicitly asserts this stale counter. Please use the wrapper's reset API or always put the counter into a state that guarantees warmup/re-capture. Also reject or reset local/scoped TE CUDA graphs; this code only touches `FullCudaGraphWrapper`, so existing scoped graphs may continue replaying the NVFP4 path.

yaoyu-33 (changes requested):
Requesting changes because the current code has four correctness/operability blockers: the documented default weight-storage path is known to fail TE recipe/weight validation; recipe restoration occurs before final validation; checkpoint/in-process restart can silently skip healing; and full-iteration CUDA graphs do not re-capture with the default counter handling. CI is green only after the feature-specific GB200 functional test was removed, so a supported end-to-end configuration needs real Blackwell coverage before merge.

PR #5025, opened 2026-07-22
chore(model): deprecate legacy model support

yaoyu-33:
## Summary

- deprecate Gemma 1, Gemma 2, Llama 2, Mistral 7B, Mistral Small 3 24B, Nemotron H v1, and the legacy `NemotronForCausalLM` bridge
- emit rank-zero `FutureWarning` notices from `AutoBridge` and legacy Nemotron H recipe entry points, with removal scheduled for Megatron Bridge 0.7.0
- warn for `nvidia/Nemotron-4-340B-Instruct` before config loading because the public repository does not expose an AutoBridge-readable HF config
- mark the same support status in the main model table, model guides, and Fern nightly docs
- remove model-specific unit tests, functional tests, checkpoint/quantization coverage, and H100/GB200 launch scripts for deprecated support
- exclude deprecated recipes and variants from shared test matrices while retaining coverage for Llama 3, Ministral 3, Nemotron Nano v2, and Nemotron 3
- keep bridges and recipes in place for the deprecation window; they are no longer actively maintained or validated

Llama 3, Ministral 3, Nemotron Nano v2, and Nemotron 3 are not deprecated.

## Testing

- focused `ruff check`, `ruff format --check`, Python compile, and deprecated-reference audits
- `PYTHONPATH=src:3rdparty/Megatron-LM uv run --no-sync mypy --strict --follow-imports=skip src/megatron/bridge/models/_deprecation.py`
- focused warning/detection smoke test

Focused pytest collection could not run locally on macOS. Normal `uv run` dependency sync is blocked by the Linux-only `nvidia-resiliency-ext` wheel; the no-sync worktree lacks an importable Megatron-Core installation, and the reduced environment reaches a Triton dependency unavailable on macOS. Linux CI covers the remaining validation.


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5025/commits/16d2f1060bb019c63e2320dfe46b6cbe293646a9

PR #5030, opened 2026-07-22
fix(flops): use parquet packed dataset for LoRA SQuAD FLOP accounting

rsalagame-nvidia:
Backport to `llmb-r0.5.1`: LoRA Llama-3-70B SQuAD over-reports `MODEL_TFLOP/s/GPU` (~1755 vs ~1150) because `flop_utils.py` only detects the deprecated `.npy` packed dataset, while packing now emits `.parquet`, so it falls back to full-model 6ND accounting. This fix makes the resolver prefer `.idx.parquet` (fall back to `.npy`) and adds a parquet-aware seqlen loader — metrics-only, no step-time change.


bdubauski (approved):
LGTM

PR #5034, opened 2026-07-22
feat(model): add GLM-5.2 verification card

HollowMan6:
# What does this PR do?

Adds a GLM-5.2 model verification card together with H100 and GB200 pretraining, SFT, PEFT, and long-sequence recipe variants, plus targeted unit coverage.

# Changelog

- Records verified H100 BF16 CPU/GPU checkpoint import and export, manual forward correlation, and deterministic Megatron inference.
- Adds bounded H100 pretraining, full SFT, 200K packed SFT, and LoRA recipes; the card also records SFT export/reload and checkpoint-resume verification.
- Adds direct 192-GPU GB200 pretraining, full SFT, 128K packed SFT, and PEFT recipes instead of launching H100 recipes with platform override blocks.
- Keeps public recipe names within the established pretrain/SFT/PEFT categories; sequence variants use 200K and 128K suffixes.
- Keeps platform recipe overrides flat and explicit instead of hiding topology, optimizer, or dataset settings behind private configuration helpers.
- Configures the 192-GPU GB200 128K recipe to use the flex token dispatcher with HybridEP.
- Derives the RoPE head dimension from the normalized Hugging Face GLM-5.2 config fields, avoiding a provider mismatch when Transformers exposes the normalized qk head dimension.
- Adds unit coverage for H100 and GB200 recipe topologies, recipe exports, packed-sequence variants, the default PP13/VPP2 layout, the pretraining gradient path, PEFT targets, the normalized RoPE dimension, and the GB200 HybridEP environment.

# GLM-5.2 GB200 128K performance reproduction

The completed full-architecture experiment used mock inputs and the ordinary non-THD sequence path:

| Item | Value |
| --- | --- |
| Hardware | 192 GB200 GPUs |
| Model | 78 layers / 256 experts |
| Precision | BF16 |
| Sequence length | 131,072 |
| Parallelism | TP1 / PP6 / CP32 / EP32 |
| Batch sizes | MBS1 / GBS56 |
| Dispatcher | Flex / HybridEP |
| Attention and routing | cuDNN, top-k 2048, indexer auxiliary loss off |
| Memory | Full activation recompute |
| Completed run | Lyris job 2513242, 12/12 steps |
| Steady-state result | Steps 2-12: 380.62 TFLOPS/GPU, 127.259 seconds/step |

The successful run used the non-packed mock-data path; exact pseudo-THD attempts stopped in the CP all-gather path. The remaining throughput gap is therefore being tracked as a packed-THD/cuDNN/CP-path issue rather than attributed to model size or token load alone. This performance experiment is separate from model-card correctness verification.

Source disclosure for the completed run:

- Megatron Bridge base: `a36fa26e951fafe63ac06cd13c713934d4de16e3`; the Bridge-side recipe/provider fixes are included in this PR at `113da339585e7997276a6443c8db1ac042bc5b82`.
- Megatron-Core base: `6cd6ea530e18776da54297bbf88292264264bcd3`.
- The run also used one external Megatron-Core bring-up patch in `dsa_cudnn_kernels.py` to convert packed maximum-sequence-length scalar tensors to Python integers before the cuDNN call. That upstream-only patch is intentionally not vendored into this Bridge PR.
- Cluster submission orchestration remains out of tree. The full internal launch script and run log are tracked in [MB-997](https://linear.app/nvidia/issue/MB-997).

# Validation

- `uv run --no-sync python -m pytest tests/unit_tests/recipes/test_glm5_recipes.py -q`: 14 passed.
- `uv run --no-sync python -m pytest tests/unit_tests/recipes/test_recipe_environment_defaults.py tests/unit_tests/recipes/test_glm5_recipes.py -q`: 20 passed.
- Model verification card YAML parses successfully.
- `uv run --no-sync pre-commit run --all-files`: all hooks passed on Lyris.

# Additional Information

- Related performance tracking: [MB-997](https://linear.app/nvidia/issue/MB-997)


examples/model_verification_cards/glm5-2/card.yaml
```diff
@@ -0,0 +1,337 @@
...
+  targets. All card items remain unverified under the model-verification
+  completion gates.
+verification_index:
+  model_level:
+    unverified: all
+  training:
+    GB200:
+      unverified: all
+    H100:
+      unverified: all
```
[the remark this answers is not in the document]
HollowMan6: The current scalar all entries are intentional and required by the bundled validator. SKILL.md says to use scalar all when every item in a scope has the same status, and _validate_index_status_buckets emits use scalar all when every item in the scope has the same status if all items are enumerated explicitly. I ran the mandatory validator locally against this card, and it passes: valid model verification card. Therefore I’m leaving these entries unchanged.

yaoyu-33 (dismissed):

examples/model_verification_cards/glm5-2/card.yaml
```diff
@@ -0,0 +1,507 @@
...
+        'dataset.blend=[["work/data/glm5-2/pretrain-tulu3-text-document/tulu3_glm52_text_document_text_document"],null]'
+        dataset.path_to_cache=work/cache/glm5-2/pretrain-tulu3-gb200
+        dataset.random_seed=1234 rng.seed=1234 scheduler.lr_decay_iters=100
+        --save_dir work/model-verification/glm5-2/pretrain-gb200-cudnn/checkpoints
+        --save_interval 50 logger.log_interval=1 logger.log_throughput=true
+        logger.tensorboard_dir=null
+      last_verified: 2026-07-24
+      metrics:
+        initial_loss: 13.19136
+        final_loss: 8.171059
```
cuichenx, line 208: loss doesn't match with H100?

examples/model_verification_cards/glm5-2/card.yaml
```diff
@@ -0,0 +1,507 @@
...
+        --recipe glm52_sft_192gpu_gb200_bf16_config --mode sft
+        --pretrained_checkpoint work/model-verification/glm5-2/gpu-megatron/iter_0000000
+        tokenizer.chat_template_path=work/templates/glm52_chat_template_generation.jinja
+        --save_dir work/model-verification/glm5-2/sft-gb200-cudnn/checkpoints
+        --save_interval 100 logger.log_interval=1 logger.log_throughput=true
+        logger.tensorboard_dir=null
+      last_verified: 2026-07-24
+      metrics:
+        initial_loss: 1.831100
+        final_loss: 0.5953441
```
cuichenx, line 275: ditto

examples/model_verification_cards/glm5-2/card.yaml
```diff
@@ -0,0 +1,507 @@
...
+        --recipe glm52_pretrain_416gpu_h100_bf16_config --mode pretrain
+        --dataset megatron-indexed --deterministic --seq_length 2048 --max_steps 100
+        --lr 3e-4 --min_lr 3e-5 --warmup_iters 40
+        'dataset.blend=[["work/data/wikitext103/wikitext103_glm52_text_document"],null]'
+        dataset.path_to_cache=work/cache/glm5-2/wikitext-pretrain
+        dataset.random_seed=1234 rng.seed=1234 scheduler.lr_decay_iters=100
+        model.pipeline_model_parallel_size=11
+        model.virtual_pipeline_model_parallel_size=null
+        model.microbatch_group_size_per_vp_stage=null
+        model.pipeline_model_parallel_layout='Etttttt|tttttttt|tttttttt|tttttttt|tttttttt|tttttttt|tttttttt|tttttttt|tttttttt|tttt|ttttmL'
```
cuichenx, line 458: is this required for checkpoint resume?

examples/model_verification_cards/glm5-2/card.yaml
HollowMan6: Mostly it's because they didn't really use the same dataset, seq length, etc.

examples/model_verification_cards/glm5-2/card.yaml
HollowMan6: Mostly it's because they didn't really use the same dataset, seq length, etc.

examples/model_verification_cards/glm5-2/card.yaml
HollowMan6: Ah probably not since we have already specified those parallelism configs under `glm52_pretrain_416gpu_h100_bf16_config`

yaoyu-33 (dismissed):

yaoyu-33 (dismissed):

examples/model_verification_cards/glm5-2/card.yaml
cuichenx: @yaoyu-33 we should change to RP2 in a followup PR

scripts/inference/setup_inference.py
```diff
@@ -32,6 +32,7 @@
 CONTAINER_REPO_ROOT = Path("/opt/Megatron-Bridge")
 INFERENCE_TASKS = {
     "text-generation": Path("scripts/inference/text_generation.py"),
+    "full-prefix-generation": Path("examples/conversion/hf_to_megatron_generate_text.py"),
```
cuichenx: we should clearly mark this as slow and non-optimized

skills/create-model-verification-card/scripts/validate_card.py
```diff
@@ -1097,6 +1097,7 @@ def _validate_inference(
         task_values = _argument_values(command, "--task")
         if len(task_values) != 1 or task_values[0] not in {
             "text-generation",
+            "full-prefix-generation",
```
cuichenx: same here

src/megatron/bridge/recipes/glm/h100/glm5.py
```diff
@@ -0,0 +1,252 @@
...
+        load_weights=False
+    )
+    cfg.tokenizer.tokenizer_model = _GLM52_MODEL_ID
+    cfg.tokenizer.hf_tokenizer_kwargs = {"revision": _GLM52_MODEL_REVISION}
+
+    cfg.model.seq_length = seq_length
+    cfg.model.tensor_model_parallel_size = 1
+    cfg.model.pipeline_model_parallel_size = 13
+    cfg.model.virtual_pipeline_model_parallel_size = virtual_pipeline_size
+    cfg.model.pipeline_model_parallel_layout = _GLM52_VPP2_LAYOUT if virtual_pipeline_size == 2 else _GLM52_PP13_LAYOUT
```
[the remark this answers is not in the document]
yaoyu-33: Addressed: test_glm52_vpp2_default_layout_shape now asserts the default 78-layer VPP2 layout. No model rerun was needed.

examples/model_verification_cards/glm5-2/card.yaml
yaoyu-33: Clarified in e1dc9c0b2: the H100 run used Wikitext at sequence length 2,048, while the GB200 run used Tulu 3 at sequence length 4,096, so their losses are not directly comparable.

examples/model_verification_cards/glm5-2/card.yaml
yaoyu-33: Agreed. I kept the current verified evidence unchanged; migrating this workload to RP2 belongs in a follow-up PR and will require fresh verification.

examples/model_verification_cards/glm5-2/card.yaml
yaoyu-33: These overrides are required to reproduce the reference/resume topology. The recipe default is 416 GPUs with PP13/VPP2, while the verified H100 reference and resume runs use 352 GPUs with PP11 and VPP disabled, so both commands must retain the same execution overrides.

scripts/inference/setup_inference.py
yaoyu-33: Addressed in e1dc9c0b2: renamed the task to legacy-full-prefix-generation, labeled it as a slow, non-optimized compatibility path in launcher help and the README, and required --legacy-full-prefix.

skills/create-model-verification-card/scripts/validate_card.py
yaoyu-33: Addressed in e1dc9c0b2: the validator and model-card skill now use the same legacy-full-prefix-generation name and require --legacy-full-prefix for that task.

cuichenx (approved):

PR #5036, opened 2026-07-22
fix(models): verify Gemma 4 VL BF16 conversion

yaoyu-33:
## Summary

- add an agent-readable verification card for google/gemma-4-26B-A4B-it at an immutable revision
- pin VLM inference model, config, tokenizer, and processor loads to that revision
- remove the Gemma 4 VL forced-FP32 override so the public BF16 checkpoint remains BF16
- preserve the Gemma 4 attention pattern and rotary bases after checkpoint config reload
- filter runtime-only synthesized KV projections from config-only CPU export
- preserve the public Gemma 4 block-spec target used by serialized checkpoint configs

## Validation

- 160 focused Gemma 4 and Gemma 4 VL bridge unit tests passed at e42bb189a6b1252725231bd86e56c47fad82a39e
- 81 Gemma 4 modeling tests, including serialized public-target resolution, passed at c4250ac1382eccc8e2b58c2424b3db85415fbf6f
- model-card validator passed at 1a292990426b36fc4c5a3035c73f74db6db9fdc7
- full pre-commit passed
- CPU and GPU import artifacts are BF16 and reload successfully
- CPU and GPU export each reproduce all 1,013 source tensors exactly in key, shape, dtype, and value and reload natively as Gemma4ForConditionalGeneration
- one deterministic BF16 multimodal inference run completed at c4250ac1382eccc8e2b58c2424b3db85415fbf6f with the existing candy image and prompt from inference.sh; a 32-token limit terminated on EOS after 29 tokens and identified the animal as a turtle
- the public input is p-blog/candy.JPG from huggingface/documentation-images at revision 44c3249461c5a5336585d1e31e0a4ce7fce06a5a with SHA-256 fc417c899e94f8df465b7541c5a70f0eebb85c414d06345f0b290c061eccc84c

## Scope

- no example-folder documentation, recipe, generic parity-helper, FP32 reference, generic HF-inference-helper, or inference.sh changes remain in the PR diff
- manual forward correlation remains unverified because the recorded TP2 run compared a padding position rather than the prompt next-token position
- full SFT, long-context SFT, PEFT, and post-SFT export/inference remain unverified and are not claimed by this PR

This PR remains draft and does not claim pending model-card items as verified.


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5036/commits/1a292990426b36fc4c5a3035c73f74db6db9fdc7

PR #5068, opened 2026-07-24
[training] fix: Respect disabled exit signal handling

yaoyu-33:
## Problem

`train.exit_signal_handler` defaults to `False`, but assigning the training
config to `GlobalState` still installed Bridge's SIGTERM trap unconditionally.
The trap only records a flag, while the training loop polls that flag only when
the option is enabled.

As a result, ordinary training with the documented/default disabled setting
could swallow an operator or scheduler SIGTERM, continue until a hard kill,
and lose work since the last periodic checkpoint.

## Root cause and fix

The config setter always called `_set_signal_handler()`, and
`_set_signal_handler()` checked only that `cfg.train` existed. Guard handler
installation with `cfg.train.exit_signal_handler` so the disabled setting
preserves the process's existing/default SIGTERM disposition.

The enabled graceful-checkpoint path is unchanged.

## Regression evidence

Focused regression, unchanged before and after:

```bash
uv run --no-sync python -m pytest \
  tests/unit_tests/training/test_state.py::TestGlobalState::test_disabled_signal_handler_does_not_intercept_sigterm \
  -q -p no:cacheprovider
```

- Before: `1 failed`; the child logged that it received SIGTERM and exited
  normally with code 0 instead of being terminated by SIGTERM.
- After: `1 passed`.

Adjacent signal/preemption coverage:

```bash
uv run --no-sync python -m pytest \
  tests/unit_tests/training/utils/test_sig_utils.py \
  tests/unit_tests/training/test_state.py \
  tests/unit_tests/training/test_train.py \
  tests/unit_tests/recipes/test_run_plugins.py \
  -k "signal or preemption" -q -p no:cacheprovider
```

Result: `17 passed, 137 deselected`.

Additional checks:

- `git diff --check` — passed
- `uv run --no-sync pre-commit run --all-files` — every hook passed

## Scope

This is a CPU-only signal-disposition regression. It does not change the
enabled save-on-SIGTERM flow, checkpoint implementation, distributed
collectives, dependencies, CI workflows, public API signatures, or
Megatron-Core.

### DIFF
--- a/src/megatron/bridge/training/state.py
+++ b/src/megatron/bridge/training/state.py
@@ -464,7 +464,7 @@ class GlobalState:
 
     def _set_signal_handler(self) -> None:
         """Initializes the distributed signal handler based on the configuration."""
-        if self.cfg.train is not None:
+        if self.cfg.train is not None and self.cfg.train.exit_signal_handler:
             self._signal_handler = DistributedSignalHandler(self.cfg.train.exit_signal).__enter__()
 
     def reset_for_restart(self) -> None:
--- a/tests/unit_tests/training/test_pg_collection_wiring.py
+++ b/tests/unit_tests/training/test_pg_collection_wiring.py
@@ -39,6 +39,7 @@ def test_should_skip_iteration_uses_passed_pg_collection(monkeypatch):
         train=SimpleNamespace(
             iterations_to_skip={1},
             micro_batch_size=4,
+            exit_signal_handler=False,
             exit_signal=signal.SIGTERM,
         )
     )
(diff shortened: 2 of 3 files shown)


yaoyu-33:
CI follow-up: the regression child used the fork start method, so it inherited the full test runner signal disposition. That made the test depend on ambient process state rather than only the Bridge setting. The child now explicitly establishes SIGTERM default disposition before assigning the disabled config; the pre-fix Bridge path would still replace that baseline, while the fixed path preserves it.

Validation:
- focused regression under the CI-style coverage runner: 1 passed
- complete test_state.py: 69 passed
- all-files pre-commit: passed

Issue #2101, opened 2026-01-28
[BUG/CLARIFICATION] enabling `multiple_validation_sets` in `GPTDatasetConfig` results in a `ZeroDivisionError` when entering in a validation cycle


LucaTedeschini:
**Describe the bug**

Training a model with `GPTDatasetConfig` configured to use multiple, separate validation sets via the `multiple_validation_sets` flag and the `blend_per_split` attribute results in an error when the validation phase starts.

```log
[rank9]: Traceback (most recent call last):
[rank9]:   File "/media/storage/store1/ltedeschini/bridge_config.py", line 286, in <module>
[rank9]:     run_pretraining_bridge()
[rank9]:   File "/media/storage/store1/ltedeschini/bridge_config.py", line 281, in run_pretraining_bridge
[rank9]:     pretrain(cfg, forward_step)
[rank9]:   File "/media/storage/store1/ltedeschini/.venv_b/lib/python3.12/site-packages/megatron/bridge/utils/decorators.py", line 39, in wrapper
[rank9]:     return func(*args, **kwargs)
[rank9]:            ^^^^^^^^^^^^^^^^^^^^^
[rank9]:   File "/media/storage/store1/ltedeschini/.venv_b/lib/python3.12/site-packages/megatron/bridge/training/pretrain.py", line 92, in pretrain
[rank9]:     _pretrain(state=state, forward_step_func=forward_step_func)
[rank9]:   File "/media/storage/store1/ltedeschini/.venv_b/lib/python3.12/site-packages/megatron/bridge/training/pretrain.py", line 134, in _pretrain
[rank9]:     train(
[rank9]:   File "/media/storage/store1/ltedeschini/.venv_b/lib/python3.12/site-packages/megatron/bridge/training/train.py", line 468, in train
[rank9]:     evaluate_and_print_results(
[rank9]:   File "/media/storage/store1/ltedeschini/.venv_b/lib/python3.12/site-packages/megatron/bridge/training/eval.py", line 260, in evaluate_and_print_results
[rank9]:     total_loss_dict, collected_non_loss_data, timelimit = evaluate(
[rank9]:                                                           ^^^^^^^^^
[rank9]:   File "/media/storage/store1/ltedeschini/.venv_b/lib/python3.12/site-packages/megatron/bridge/training/eval.py", line 131, in evaluate
[rank9]:     loss_dicts = forward_backward_func(
[...]

PR #5076, opened 2026-07-24
feat(training): support multiple validation sets and eval at start

### LINKED ISSUE #618: Support for multiple validation sets + separate loss validation reporting

### LINKED ISSUE #3996: [feature] Eval at start (step-0 eval)

MoPl90:
## What does this PR do ?

Implements #3996 (eval at start; approved by @yaoyu-33) and #618 (multiple validation sets + separate loss reporting), bringing Bridge's training loop to parity with Megatron-LM's `multiple_validation_sets` / `validation_set_names` convention. Bridge inherits both fields in `ValidationConfig`, and `BlendedMegatronDatasetBuilder` already returns a list of validation datasets when the flag is set, but Bridge's loaders and eval cannot consume it today (likely related to the crash reported in #2101).

## Changelog

- `training/config.py`: add Bridge `ValidationConfig` subclass (same pattern as `TrainingConfig`) with `eval_at_start: bool = False`
- `training/train.py`: extract the in-loop eval into `_run_validation()` (all existing logic preserved one-to-one); called once before the loop when  `eval_at_start` and `do_valid` are set
- `training/eval.py`: per-set evaluation and suffixed logging in `evaluate_and_print_results()` for the supported loggers (TensorBoard / W&B / MLflow / Comet); cross-rank dataset-count guard; raises `ValueError` on `validation_set_names` length mismatch
- `data/loaders.py`: folds the three duplicated valid/test dataloader build sites into `_build_eval_test_dataloaders()`, mapping over dataset lists; errors when a provider returns multiple validation datasets without the flag
- Tests: see Testing section below

## Design notes

1. **No combined loss in multi-val mode.** Megatron-LM's convention evaluates each set separately and logs no aggregate; we match it here
2. **Collective balance under PP.** Every rank runs the same number of fully-flushed `evaluate()` calls per eval round, so the per-set loop cannot desync pipeline collectives
3. **Fail loud instead of hanging.** A min/max all-reduce on the set count raises on every rank if ranks ever disagree (e.g. a provider deriving sets from per-rank file reads), as otherwise the mismatch surfaces as a silent NCCL hang
4. **Why one PR and not two:** both features reshape `evaluate_and_print_results` and the eval plumbing in `train()`/`loaders.py`, so most of the diff (the `_run_validation()` extraction, the per-set eval restructuring, the test scaffolding) would be needed for either feature alone. They are also used together in practice: the motivating use case is a step-0 baseline *per validation set*, and the end-to-end evidence below exercises exactly that combination

### Unit tests

- `test_eval.py`: indexed and named suffixes; exact logger calls (metric names, ppl, step) on TensorBoard/W&B; `validation_set_names` length mismatch raises before any evaluation; timelimit mid-loop aborts remaining sets; a list without the flag is evaluated once (VPP); rank set-count agreement passes / mismatch raises; `is_test` never splits a list.
- `test_loaders.py`: one dataloader per set with correct per-set contents; iterator layer yields one `RerunDataIterator` per set in blend order; guard error without the flag.
- `test_config.py`: `eval_at_start` default/settable; `multiple_validation_sets` validates under PP 1 and 2.

### End-to-end training 

Ran 6 short training jobs (train_iters=4, eval_interval=2, eval_iters=2) with eval_at_start=true and multiple_validation_sets=true on a blend of 5 validation datasets, using dense Qwen3.5-VL models (2B and 9B sizes). All 6 finished and logged one loss series per validation set, starting at iteration 0. The matrix covers:

- Model size. 2B on 4 GPUs and 9B on 8 GPUs.
- Set naming. validation_set_names=["set-A", ..., "set-E"] vs unset (falls back to the index suffix -0 ... -4).
- Parallelism. Pure DP (TP1/PP1/DP4), TP-only (TP4/DP2), and pipeline parallel (TP2/PP2/DP2), confirming per-set loss reduction and logging work when the loss lives on the last PP stage.

**Validation on start.** With eval_at_start=true, all validation sets are evaluated before the first training step and logged at iteration 0. From the 2B run with named sets:

```bash
Evaluating on 64 samples
Evaluating iter 1/2
Evaluating iter 2/2
--------------------------------------------------------------------------------------------------------------------------
 validation-set-A loss at iteration 0 (pre-train validation) | lm loss value: 8.543873E-01 | lm loss PPL: 2.349934E+00 |
--------------------------------------------------------------------------------------------------------------------------
Evaluating on 64 samples
Evaluating iter 1/2
Evaluating iter 2/2
--------------------------------------------------------------------------------------------------------------------------
 validation-set-B loss at iteration 0 (pre-train validation) | lm loss value: 1.473683E+00 | lm loss PPL: 4.365282E+00 |
--------------------------------------------------------------------------------------------------------------------------
Evaluating on 64 samples
Evaluating iter 1/2
Evaluating iter 2/2
--------------------------------------------------------------------------------------------------------------------------
 validation-set-C loss at iteration 0 (pre-train validation) | lm loss value: 9.426797E-01 | lm loss PPL: 2.566851E+00 |
--------------------------------------------------------------------------------------------------------------------------
Evaluating on 64 samples
Evaluating iter 1/2
Evaluating iter 2/2
--------------------------------------------------------------------------------------------------------------------------
 validation-set-D loss at iteration 0 (pre-train validation) | lm loss value: 8.461825E-01 | lm loss PPL: 2.330732E+00 |
--------------------------------------------------------------------------------------------------------------------------
Evaluating on 64 samples
Evaluating iter 1/2
Evaluating iter 2/2
--------------------------------------------------------------------------------------------------------------------------
 validation-set-E loss at iteration 0 (pre-train validation) | lm loss value: 9.909064E-01 | lm loss PPL: 2.693675E+00 |
--------------------------------------------------------------------------------------------------------------------------
```

**Multi-dataset validation.** Each set is evaluated independently at every eval_interval and at the end of training, printed as its own block with the set suffix. From the 9B TP2/PP2 run (default index naming, final validation):

```
Evaluating on 64 samples
Evaluating iter 1/2
Evaluating iter 2/2
-----------------------------------------------------------------------------------------------------------------
 validation-0 loss at iteration 4 on validation set | lm loss value: 9.499896E-01 | lm loss PPL: 2.585683E+00 |
-----------------------------------------------------------------------------------------------------------------
Evaluating on 64 samples
Evaluating iter 1/2
Evaluating iter 2/2
-----------------------------------------------------------------------------------------------------------------
 validation-1 loss at iteration 4 on validation set | lm loss value: 1.331964E+00 | lm loss PPL: 3.788476E+00 |
-----------------------------------------------------------------------------------------------------------------
Evaluating on 64 samples
Evaluating iter 1/2
Evaluating iter 2/2
-----------------------------------------------------------------------------------------------------------------
 validation-2 loss at iteration 4 on validation set | lm loss value: 8.635647E-01 | lm loss PPL: 2.371600E+00 |
-----------------------------------------------------------------------------------------------------------------
Evaluating on 64 samples
Evaluating iter 1/2
Evaluating iter 2/2
-----------------------------------------------------------------------------------------------------------------
 validation-3 loss at iteration 4 on validation set | lm loss value: 7.436793E-01 | lm loss PPL: 2.103661E+00 |
-----------------------------------------------------------------------------------------------------------------
Evaluating on 64 samples
Evaluating iter 1/2
Evaluating iter 2/2
-----------------------------------------------------------------------------------------------------------------
 validation-4 loss at iteration 4 on validation set | lm loss value: 7.696714E-01 | lm loss PPL: 2.159057E+00 |
-----------------------------------------------------------------------------------------------------------------
```

# GitHub Actions CI

**Pre checks**:

- [ ] <s>Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)</s>

# Additional Information

- Closes  #3996 
- Closes #618
- Related to #2101
- Suggested labels: `feature`, `area:training`, `needs-review`


src/megatron/bridge/data/loaders.py
```diff
@@ -345,59 +351,54 @@ def worker_init_fn(_):
...
-            worker_init_fn=maybe_worker_init_fn,
-            collate_fn=valid_ds.collate_fn if hasattr(valid_ds, "collate_fn") else None,
-            pin_memory=cfg.dataset.pin_memory,
-            persistent_workers=cfg.dataset.persistent_workers,
-            data_parallel_rank=eval_dp_rank,
-            data_parallel_size=eval_dp_size,
-            global_batch_size=eval_gbs,
-            seed=sampler_seed,
+        valid_dataloader = _build_eval_test_dataloaders(
+            valid_ds, train_state.consumed_valid_samples, val_dataloader_type
```
yaoyu-33: This applies the single aggregate `consumed_valid_samples` offset to every validation set. `evaluate()` increments that scalar for every set, so after N sets each individual loader has consumed only 1/N of the recorded total. After checkpoint resume, every loader skips the full aggregate offset (for example, 5 sets x 2 eval iterations makes each set resume at 10 iterations rather than 2), silently changing the evaluated samples or repeating cyclic data. MCore currently refuses multiple validation sets without full validation for this exact reason. Please either persist per-set offsets, or require and fully implement `full_validation`, and add save/resume coverage.

src/megatron/bridge/data/loaders.py
```diff
@@ -475,7 +476,10 @@ def _get_iterator(dataloader_type, dataloader):
 
     if valid_dataloader is not None:
         val_dataloader_type = "cyclic" if isinstance(cfg.dataset, GPTDatasetConfig) else cfg.dataset.dataloader_type
-        valid_data_iterator = _get_iterator(val_dataloader_type, valid_dataloader)
+        if isinstance(valid_dataloader, list):
+            valid_data_iterator = [_get_iterator(val_dataloader_type, dl) for dl in valid_dataloader]
+        else:
+            valid_data_iterator = _get_iterator(val_dataloader_type, valid_dataloader)
```
yaoyu-33, line 538: When `multiple_validation_sets=True` but the provider returns a single validation dataset, this branch produces one `RerunDataIterator`. `evaluate_and_print_results()` then assumes the flag means a sequence of sets and calls `len()` on it, but `RerunDataIterator` has no `__len__` or `__iter__`, so the first validation fails. The built-in blended builder does return a single dataset when the validation blend has one prefix. Please normalize it to a one-element list whenever the flag is enabled, or reject that configuration early, and cover the singleton case.

src/megatron/bridge/training/train.py
```diff
@@ -369,6 +369,55 @@ def train(
...
+            gc.collect(generation=0)
+        if toggle_pre_hook and should_toggle_forward_pre_hook:
+            enable_forward_pre_hook(model)
+            pre_hook_enabled = True
+        timers("interval-time", log_level=0).start(barrier=True)
+        if energy_monitor is not None:
+            energy_monitor.resume()
+
+    if val_config.eval_at_start and global_state.train_state.do_valid:
+        _run_validation(
```
yaoyu-33, line 431: This extra validation pass is not included in `get_train_valid_test_num_samples()`: the existing formula reserves samples for interval evaluations plus the final evaluation only. For `train_iters=10`, `eval_interval=5`, and `eval_iters=2`, it builds 6 validation iterations but this change executes 8. A cyclic loader hides the shortage by reusing early samples, while a single-pass loader can exhaust early. Please add one `eval_iters * eval_global_batch_size` allocation when the pre-train pass can run.

src/megatron/bridge/training/train.py
```diff
@@ -369,6 +369,55 @@ def train(
...
+            f"iteration {global_state.train_state.step} (pre-train validation)",
```
yaoyu-33, line 432: This condition also runs at a nonzero resumed step and on every NVRx in-process retry, because recovery re-enters `train()` with the restored step. That duplicates validation/logging and advances validation state during recovery. Issue #3996 proposed guarding this with `train_state.step == 0`. If ordinary checkpoint resume should intentionally evaluate, please distinguish it from an automatic in-process retry; otherwise add the step-zero guard and a nonzero-step test.

yaoyu-33 (changes requested):
Thanks for adding these two features. I found four correctness issues around validation iterator shape, checkpoint resume, and step-zero evaluation bookkeeping; details are inline.

MoPl90:
Thank you for the review and spotting those correctness issues. I will address your inline comments directly. 

Regarding the `consumed_valid_samples` offset, I considered following the MCore logic (`full_validation`), but it derives per-set eval iterations from `len(dataloader)`, which doesn't exist for external dataloaders. This is therefore orthogonal to the offset book keeping approach and doesn't fit our use case. Happy to address as a follow-up. Instead I: 

- added `TrainState.consumed_valid_samples_per_set`: one offset counter per validation set, persisted in the checkpoint train state. Checkpoints written before this field existed load as an empty list and the counters initialize to zeros, so checkpoints resume cleanly
- enabled per-set resume in the loader builder. `build_train_valid_test_data_loaders` sizes the per-set list once per build and resumes each validation loader from its own offset. If the number of sets changed since the checkpoint, the counters reset to zero with a warning
- advance the aggregate: `evaluate()` advances `consumed_valid_samples` for every set, and additionally the indexed per-set counter when `multiple_validation_sets` is on. Rationale: preserve the field's existing semantics (total across sets); `consumed_valid_samples == sum(consumed_valid_samples_per_set)` holds, and a multi-set checkpoint resumed with the flag disabled falls back to the aggregate offset with a warning, i.e. the pre-PR behavior for that configuration
- note: External dataloaders (`dataloader_type="external"`) pass offsets through as before; for them the per-set counters are bookkeeping only, since sampler offsets never applied to external loaders in the first place

Moreover, I added unit tests and `test_pretrain_resume_multi_val.py`, a functional end-to-end test (L1) that checkpoints mid-run, resumes, and verifies the per-set counters persist and continue from their restored values (verified on a 2-GPU machine). Two notes:

1. It reads the tracker with raw `torch.load` on all ranks rather than the `read_train_state` helper, which is lru_cached (returns stale state on a second read) and broadcasts its result across ranks
2. It sets `async_save=False` because it reads `latest_train_state.pt` immediately after `pretrain()` returns, which otherwise races the async finalization write


Happy to address any remaining concerns!

src/megatron/bridge/data/loaders.py
MoPl90: As I outline below, I opted for the full bookkeeping approach instead of `full_validation`. See my comment below for details.

src/megatron/bridge/data/loaders.py
MoPl90: Addressed in [5d93fab](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5076/commits/5d93fab8fe2506bf21238a06e990d259f762601d)

src/megatron/bridge/training/train.py
MoPl90: Addressed in [b045fe2](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5076/commits/b045fe232d284c100da4d84acdd20729d5857e33)

src/megatron/bridge/training/train.py
MoPl90: With [cb5601f](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5076/commits/cb5601fea538d813715bde206050ba866c00e90b)  this is now correctly calculated.

src/megatron/bridge/data/loaders.py
```diff
@@ -268,6 +271,14 @@ def build_train_valid_test_data_loaders(
         cfg=cfg, build_train_valid_test_datasets_provider=build_train_valid_test_datasets_provider
     )
 
+    if isinstance(valid_ds, list) and not cfg.validation.multiple_validation_sets:
+        raise ValueError(
+            "The dataset provider returned multiple validation datasets but "
+            "ValidationConfig.multiple_validation_sets is not set."
+        )
+    if cfg.validation.multiple_validation_sets and valid_ds is not None and not isinstance(valid_ds, list):
```
yaoyu-33, line 279: **Blocking: the MegatronMIMO path bypasses all of this.**

`build_train_valid_test_data_loaders()` returns early for `MegatronMIMOProvider` at line 259, which is *above* this normalization block and above the `consumed_valid_samples_per_set` sizing at line 400. If a user sets `multiple_validation_sets=True` on a MIMO model, three things go wrong:

1. `train_state.consumed_valid_samples_per_set` is never sized, so it stays `[]`. `evaluate()` then does `state.train_state.consumed_valid_samples_per_set[0] += eval_batch_size` → `IndexError`.
2. `train_megatron_mimo.py:390-393` gates evaluation on a **per-rank** condition (`valid_data_iterator is not None`), because heterogeneous MIMO legitimately leaves some ranks without a validation loader. The new set-count `all_reduce` in `evaluate_and_print_results()` is unconditional, so only a subset of ranks enters it → **collective hang**. That is the exact failure mode the all_reduce was added to prevent.
3. `len(data_iterators)` is called on a single `RerunDataIterator`, which has no `__len__` → `TypeError`.

Given design note 3 in the PR description ("fail loud instead of hanging"), the consistent fix is to reject the combination up front — e.g. in `ConfigContainer.validate()`, raise when `validation.multiple_validation_sets` is set and `cfg.model` is a `MegatronMIMOProvider`. Supporting it properly is fine too, but out of scope for this PR; an explicit error is enough.

Please add a unit test for whichever guard you choose.

src/megatron/bridge/training/train.py
```diff
@@ -369,6 +373,56 @@ def train(
...
+            gc.collect(generation=0)
+        if toggle_pre_hook and should_toggle_forward_pre_hook:
+            enable_forward_pre_hook(model)
+            pre_hook_enabled = True
+        timers("interval-time", log_level=0).start(barrier=True)
+        if energy_monitor is not None:
+            energy_monitor.resume()
+
+    # Skipped on in-process restart retries so recovery does not re-run validation.
+    if val_config.eval_at_start and global_state.train_state.do_valid and not is_inprocess_restart_retry:
```
yaoyu-33: **`eval_at_start` re-runs on every ordinary checkpoint resume, but only one extra pass is budgeted.**

`is_inprocess_restart_retry` only suppresses NVRx in-process retries. An ordinary resume from a checkpoint is a fresh process with `restart_attempt == 0`, so this branch runs again.

Meanwhile `get_train_valid_test_num_samples()` (loaders.py:133-135) reserves exactly **one** extra `eval_iters` worth of samples. A job that is preempted and resumed K times executes K+1 pre-train validation passes and consumes `K * eval_iters * eval_global_batch_size` more validation samples than the dataset was sized for.

Concrete failure: `train_iters=10000, eval_interval=1000, eval_iters=2, dataloader_type="single"` (non-cyclic — the same type your new functional test uses). Five preemptions ⇒ six pre-train passes ⇒ ten eval iterations of validation data beyond the reservation. The single-pass sampler drifts forward and can exhaust the validation set / raise `StopIteration`.

Issue #3996 proposed guarding on `train_state.step == 0` specifically, and I think that matches the actual use cases: the value of this feature is a baseline for the *loaded checkpoint at the start of training* (SFT from a converted HF checkpoint, sanity-checking a HF↔MCore conversion before burning GPU-hours, anchoring an A/B of data mixes). On resume you already have that number from the last in-loop eval, so re-running it adds cost and puts a duplicate point at an already-logged step in TensorBoard/W&B.

Either add the `step == 0` guard, or — if you deliberately want resume to re-evaluate — say so in the field docstring and widen the sizing formula accordingly. Right now it does the widened behaviour with the narrow budget. Please also add a nonzero-step test for whichever semantics you land on.

src/megatron/bridge/training/eval.py
```diff
@@ -436,71 +450,121 @@ def evaluate_and_print_results(
...
+            )
+            torch.distributed.all_reduce(counts, op=torch.distributed.ReduceOp.MAX)
+            max_sets, min_sets = counts[0].item(), -counts[1].item()
+            if max_sets != min_sets:
+                raise RuntimeError(
+                    f"multiple_validation_sets: ranks disagree on the number of validation sets "
+                    f"(min {min_sets}, max {max_sets}; this rank has {num_sets}). The dataset "
+                    "provider must return the same number of validation sets on every rank."
+                )
+        if val_config.validation_set_names is not None and len(val_config.validation_set_names) != len(data_iterators):
```
yaoyu-33: `validation_set_names` set *without* `multiple_validation_sets` is silently ignored — this length check only runs inside the flag branch, and the suffix logic at line 486 is likewise gated on the flag. The user gets no names and no warning.

MLM asserts on this (`megatron/training/training.py:3633`: `assert args.multiple_validation_sets, "--validation-set-names requires --multiple-validation-sets"`). Worth matching.

Ideally both this check and the names/count check move into `ConfigContainer.validate()` — a static config error should fail at startup, not at the first `eval_interval` after the user has already paid for model init and dataset build.

src/megatron/bridge/training/eval.py
```diff
@@ -436,71 +450,121 @@ def evaluate_and_print_results(
...
+        data_iterators = data_iterator
+        if torch.distributed.is_initialized():
+            # A rank with fewer sets would run fewer evaluate() calls (each one
+            # runs collectives) and deadlock the others; fail loudly on every
+            # rank instead.
+            num_sets = len(data_iterators)
+            counts = torch.tensor(
+                [num_sets, -num_sets], dtype=torch.long, device="cuda" if torch.cuda.is_available() else "cpu"
+            )
+            torch.distributed.all_reduce(counts, op=torch.distributed.ReduceOp.MAX)
```
yaoyu-33: Two concerns about this all_reduce:

1. **Process group.** It uses the default world group, while the rest of this function routes collectives through `pg_collection` (and `evaluate()` takes an explicit `p2p_communicator`/`pg_collection`). For multi-module / VLM setups that is inconsistent.
2. **Placement.** The number of validation sets is a static property of the config and the dataset provider — it cannot change between eval rounds. Validating it costs a full-world collective on *every* eval round, on the hot path. `build_train_valid_test_data_loaders()` already knows the count (line 400) and is where the per-set counters get sized; doing the agreement check once there and trusting it afterwards is both cheaper and closer to the source of truth.

The check itself is a good idea — I just think it belongs at dataloader-build time.

src/megatron/bridge/training/eval.py
```diff
@@ -436,71 +450,121 @@ def evaluate_and_print_results(
...
-    print_rank_last("-" * length)
-    print_rank_last(string)
-    print_rank_last("-" * length)
+        length = len(string) + 1
+        print_rank_last("-" * length)
+        print_rank_last(string)
+        print_rank_last("-" * length)
 
-    return total_loss_dict
+    return set_loss_dict
```
yaoyu-33: Non-blocking, but worth pinning down: the return value is now the **last** set's loss dict when `multiple_validation_sets` is on. No current caller uses the return value (`pretrain.py:165/178`, `train.py:612`, `train_megatron_mimo.py:396` all discard it), so nothing breaks today — but this is a silent trap for any future early-stopping / save-best-checkpoint logic, which would end up keying off whichever set happens to be last in the blend.

Consider returning a dict keyed by the set suffix (or `None` in multi-set mode) so the ambiguity is impossible to consume accidentally. The docstring update is good; making it unrepresentable would be better.

Separately: when the timelimit fires mid-loop you `return None` after sets `0..k` have already advanced their per-set counters while `k+1..N-1` have not, so those sets resume from different offsets. That is defensible since the sets are independent, but it deserves a sentence in the docstring.

yaoyu-33 (changes requested):
Thanks for the thorough revision — all four items from my previous round are genuinely addressed, and the approach you picked for the per-set offsets (real bookkeeping in `TrainState` rather than requiring `full_validation`) is the right call. The save/resume functional test with two sets and `dataloader_type="single"` is exactly the coverage that was missing.

I also checked two things that are easy to get wrong here, and you got both right, so no action needed:

- **PP collective balance holds.** `evaluate()` returns `{}` rather than `None` off the last pipeline stage, and the only `return None` path is the timelimit one, which is gated on an `all_reduce(MAX)`. So the early `return None` inside the per-set loop cannot leave ranks with unequal `evaluate()` counts.
- **`_run_validation(toggle_pre_hook=False)` is correct for the pre-train pass.** The forward pre-hook is already disabled at `train.py:338` before your insertion point, and `interval-time` is started at `train.py:189`, so the `timers("interval-time").stop()` inside the helper will not trip MCore's `assert self._started`.

One blocking issue and a few smaller ones inline. The PR is also currently conflicting with `main` and will need a rebase.

src/megatron/bridge/training/train.py
yaoyu-33: Following up on my own comment — I agree the `step == 0` semantics is the one that actually carries meaning, and I checked that it covers the motivating use cases cleanly:

`load_checkpoint()` resets `state.train_state.step = 0` when `cfg.checkpoint.finetune` or `release` is set (`checkpointing.py:2322-2323`). So "SFT / continued pretraining from a converted HF checkpoint" — the strongest reason to want this feature — already lands at step 0 and gets its baseline. A plain mid-run resume does not, and that is exactly the case where the extra pass buys nothing (the previous in-loop eval already produced that number) while costing unbudgeted validation samples and putting a duplicate point at an already-logged step. So `step == 0` is not a compromise here; it is the semantics that matches the use cases.

**Given that, I think the field should be renamed while it is still unreleased and free to change.** `eval_at_start` invites exactly the ambiguity this thread is about — "start" of the run, or of training? Two candidates, in my order of preference:

1. **`eval_at_step_zero`** — states the guard itself, so the behaviour on resume is unambiguous from the name alone and the docstring does not have to carry the disambiguation.
2. **`baseline_eval`** — intent-revealing: a baseline is by definition taken once, at the beginning, which makes "why doesn't this re-run on resume?" a non-question.

There is an additional argument for renaming: `ValidationConfig` already inherits MCore's **`start_eval_at_iter`** (`megatron/training/config/training_config.py:125`), which delays the *first* evaluation. Having `eval_at_start` and `start_eval_at_iter` sit next to each other in the same config — near-identical wording, essentially opposite meanings — is a real readability trap for anyone scanning the config.

Happy to defer to your preference on the exact name; the substantive ask is the `step == 0` guard plus a name that does not require a docstring to disambiguate.

MoPl90:
Thank you for the re-review and the internal consistency verification. Glad to hear that you agree with the chosen approach. 

I will respond directly to the inline comments. I hope this addresses your concerns! Branch is updated.

src/megatron/bridge/data/loaders.py
MoPl90: Agree that this is out of scope. I added a guard in [1a80d21](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5076/commits/1a80d218725d0944f595389c45509f321ce66c4e), which also loudly fails for one more potential silent failure modes: `cfg.validation.full_validation` [can be set](https://github.com/NVIDIA/Megatron-LM/blob/83392004c81b6e477002ae84d4f0b1271518c1d1/megatron/training/config/training_config.py#L147), but has no effect. Now raises a ValueError at startup.

src/megatron/bridge/training/train.py
MoPl90: I think I can agree with this argument. The checkpoint-resume functionality was valuable for development, but practically adds little value. 

The suggested re-naming is also reasonable given that Megatron-core already has a near-identical config field.

Fixed in [8f8fa1a](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5076/commits/8f8fa1a5192d138a18e0ff1c3ed7a9a15c48e56b)

src/megatron/bridge/training/eval.py
MoPl90: Added a config level check in [e176de0](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5076/commits/e176de0e61dea40b623a89f7c0cc69c9d406c657)

src/megatron/bridge/training/eval.py
MoPl90: Excellent point! Moved to dataloader build in [666aefc](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5076/commits/666aefc8bd7dba5fe4ffe792c9db582a6e2e390b)

src/megatron/bridge/training/eval.py
MoPl90: Valid concern, addressed in [a8b35d3](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5076/commits/a8b35d3ffd973f49a0382874403483fe099e4792)

MoPl90:
Merged main (one conflict): #5239 added `drop_last=...` to the three duplicated valid/test dataloader build sites. The same three this PR folds into one helper `_build_eval_test_dataloaders()`. Resolved as one drop_last line in the helper, applied per validation set for free. Three-copy changes on main become one-line changes here.

MoPl90:
@yaoyu-33 is anything needed from my side to run those additional tests?

MoPl90:
@yaoyu-33 can we move forward with this?

yaoyu-33:
Checking now

src/megatron/bridge/training/pretrain.py
```diff
@@ -180,6 +182,7 @@ def _pretrain(
                 verbose=True,
                 write_to_tensorboard=not config.validation.skip_train,
                 callback_manager=callback_manager,
+                is_inprocess_restart_retry=is_inprocess_restart_retry,
```
yaoyu-33: **Blocking:** is_inprocess_restart_retry is wired to the wrong call. evaluate_and_print_results() has no such parameter, so a run with do_valid=True reaches post-training validation and raises TypeError; meanwhile train() never receives the retry flag, so step-zero validation can repeat on an in-process restart. Please remove this keyword here, pass it to train(...) above, and keep the focused retry coverage.

src/megatron/bridge/data/loaders.py
```diff
@@ -268,6 +271,40 @@ def build_train_valid_test_data_loaders(
...
```
yaoyu-33, line 289: **Blocking:** the built-in GPT path has two separate owners for this flag. BlendedMegatronDatasetBuilder branches on cfg.dataset.multiple_validation_sets, while this loader/evaluation path branches on cfg.validation.multiple_validation_sets. With the advertised ValidationConfig-only setup, the builder returns one blended validation dataset; conversely, existing dataset-only configurations now fail here. Please establish one source of truth (including backward compatibility and the duplicate full_validation field) before dataset construction, and add a test using the built-in GPT builder with more than one validation prefix. The custom-provider test bypasses this integration.

src/megatron/bridge/data/loaders.py
```diff
@@ -130,6 +130,9 @@ def get_train_valid_test_num_samples(cfg: ConfigContainer) -> tuple[int, int, in
         eval_iters = (cfg.train.train_iters // cfg.validation.eval_interval + 1) * cfg.validation.eval_iters
     else:
         eval_iters = 0
+    # Reserve one extra evaluation for the pre-train pass when eval_at_step_zero is True.
+    if cfg.validation.eval_at_step_zero and cfg.validation.eval_iters:
```
yaoyu-33: **Blocking:** eval_interval=None under-budgets this path. The base count is zero and this reserves one evaluation, but a normal run evaluates once at step zero and again in _pretrain() after training. A finite dataloader_type="single" iterator therefore gets two passes from a one-pass allocation and can exhaust. Please either reserve both passes or suppress the final pass for this case, and add a finite-loader regression.

src/megatron/bridge/data/loaders.py
```diff
@@ -345,62 +382,79 @@ def worker_init_fn(_):
...
-            data_parallel_size=eval_dp_size,
-            global_batch_size=eval_gbs,
-            drop_last=not (isinstance(cfg.dataset, GPTSFTDatasetConfig) and val_dataloader_type == "batch"),
-            seed=sampler_seed,
-        )
+        if cfg.validation.multiple_validation_sets:
+            # Resume each validation set from its own consumed-samples offset.
+            valid_consumed_samples = train_state.consumed_valid_samples_per_set
+        else:
+            if train_state.consumed_valid_samples_per_set:
```
yaoyu-33, line 457: **Blocking:** the aggregate is not a valid fallback offset for one dataset after disabling multi-set validation. For example, per-set counters [100, 200] produce aggregate 300; resuming one validation loader at 300 skips samples from both former sets and can exhaust the loader. Since dataset identity is not tracked, a warning cannot make this mapping correct. Please fail explicitly, reset to zero, or require an unambiguous user-selected mapping.

tests/unit_tests/training/test_config.py
```diff
@@ -556,6 +556,72 @@ def _mock_use_deterministic(flag):
...
+        """eval_at_step_zero is a bridge addition defaulting to False and is settable."""
+        assert ValidationConfig().eval_at_step_zero is False
+        assert ValidationConfig(eval_at_step_zero=True).eval_at_step_zero is True
+        # Inherited from Megatron-LM's ValidationConfig.
+        assert ValidationConfig().multiple_validation_sets is False
+        assert ValidationConfig().validation_set_names is None
+
+    def test_full_validation_rejected(self):
+        """full_validation is inherited from MCore but unimplemented; setting it must fail loud."""
+        container, og_ws, cfg_mod = create_test_config_container(world_size_override=8)
```
yaoyu-33: **Blocking test failure:** these three new tests no longer construct their fixture after the latest main merge. create_test_config_container() requires model_config, but this call and the calls in the two tests below omit it, so they raise TypeError before reaching the intended assertions. Please pass a concrete create_test_gpt_config(...) (or supply the desired validation config through the helper) and rerun this file.

yaoyu-33 (changes requested):
Re-reviewed the current head after the latest main merge. Several earlier requests are represented, but the blocking runtime, configuration-ownership, data-sizing/resume, and test-fixture issues below remain. Please address these before CI is authorized; I am withholding /ok to test for this SHA.

MoPl90:
Thank you for reviewing again! It seems like I missed some crucial changes when merging and resolving conflicts. To make sure this is not the case now, I ran the following tests and made sure they are all green: Unit suite (311 tests) and the L0/L1 pretrain functional tests (2 GPUs) pass on the branch tip.

Addressing your inline comments above. Thanks!

src/megatron/bridge/training/pretrain.py
MoPl90: Fixed in b8eefd9c2: the keyword now goes to `train` and is removed from the `evaluate_and_print_results()` call. The existing unit test (`test_forwards_is_inprocess_restart_retry`) already asserts `train` receives the flag, so it now guards this wiring.

src/megatron/bridge/data/loaders.py
MoPl90: Fixed in 34bda08a0. `ConfigContainer.validate` now reconciles the two before dataset construction: `ValidationConfig` is the single source of truth and is mirrored onto the dataset config's field, a dataset-only (Megatron-LM style) setting is adopted for backward compatibility, and an explicit contradiction raises. The dataset-side `full_validation` duplicate is rejected with the same error as the validation-side one. Added the requested integration test: `test_multiple_validation_sets_built_in_gpt_builder` builds two indexed validation prefixes and asserts the built-in GPT builder path yields one dataloader per prefix.

src/megatron/bridge/data/loaders.py
MoPl90: Fixed in 04686bd8a. With `eval_at_step_zero` and no `eval_interval`, both the step-zero pass and the post-training pass are now reserved. A regression test asserts the budget for both cases.

src/megatron/bridge/data/loaders.py
MoPl90: Fixed in e10870ce8: the aggregate fallback is gone. When per-set counters exist but `multiple_validation_sets` is disabled, validation restarts from offset 0 with a warning, consistent with the existing set-count-change reset. Covered by `test_disabling_multiple_validation_sets_resets_resume_offset`.

tests/unit_tests/training/test_config.py
MoPl90: Fixed in 41a1629bb.

MoPl90:
A few additional hardening commits on top of the requested changes:

- Reject a list of test datasets (56221bd68): previously the list would nest into the dataloader machinery and fail far from the cause
- Clear stale per-set counters when multi-val is disabled (c44426349)
- Reject duplicate `validation_set_names` (7a121101e) that would collide in logging keys and in the per-set loss dict
- Fail fast on a non-list `data_iterator` with `multiple_validation_sets` set (954c40a77)

One behavior noted but intentionally unchanged: weights in the validation blend are ignored under `multiple_validation_sets`, matching Megatron-LM's builder.

The branch is also merged with the latest main. At the current tip, the unit suites for `test_pretrain`, `test_config`, `test_loaders`, and `test_eval`, the `test_pretrain_resume_multi_val` functional test, and the L0 pretrain functional group pass on 2 GPUs.

MoPl90:
Happy to address any further concerns if there are any @yaoyu-33

MoPl90:
@yaoyu-33 I've merged with main and incorporated your update to the eval_iters counter in #5810. All still green on the GPU tests.

MoPl90:
@yaoyu-33 should we proceed with this feature?

PR #5084, opened 2026-07-24
fix(model): add Gemma 3 VL verification coverage

yaoyu-33:
## Summary

- add an agent-readable verification card for `google/gemma-3-4b-it` at its canonical immutable revision
- preserve Gemma 3 VL image-bidirectional attention by selecting the arbitrary unfused mask path, using a symmetric local window, and rejecting incompatible context parallelism/runtime backends
- disable incompatible in-batch THD packing in all six Gemma 3 VL recipes
- retain regression coverage for legacy vision-tower checkpoint namespaces, pipeline ownership, mask propagation, and recipe invariants

## Validation

- targeted Gemma 3 VL provider/modeling/recipe tests — 62 passed
- card validator — passed
- `uvx --from pre-commit pre-commit run --all-files` — passed
- parity root-cause check: RoPE parameters matched HF; first material divergence was layer-0 attention output because the custom multimodal mask was not consumed

## Exact-checkpoint evidence

Validated against Bridge fix commit `2d0175b116af6cb785bb8efab7ee053435496266`, MCore `6cd6ea530e18776da54297bbf88292264264bcd3`, and HF revision `093f9f388b31de276ce2de164bdc2081324b9767`.

- `hf_to_megatron_cpu` — PASS: 781/781 Bridge tensors; independent Gloo reload matched all 883 source tensors exactly
- `hf_to_megatron_gpu` — PASS: TP2/PP2 four-shard reload, correct PP ownership, and 883/883 exact reshard comparison
- `megatron_to_hf_cpu` — FAIL/unverified: 437 legacy `vision_tower.vision_model.*` keys became 437 flat `vision_tower.*` keys
- `megatron_to_hf_gpu` — PASS: 883/883 exact keys/shapes/dtypes/values, legacy namespace retained, native strict reload passed
- `manual_forward_pass` — PASS: same token `22515` (`Based`), cosine `0.990209`; max/mean absolute differences `6.046631`/`1.016037` are diagnostic only
- `inference` — PASS: one fresh TP2/PP2 multimodal run generated exactly 32 tokens and a non-empty literal completion

Long-context SFT is marked unsupported because the required arbitrary multimodal mask currently excludes CP>1 and in-batch THD packing. Other training leaves remain unchanged/pending.


yaoyu-33:
Focused Linux/container validation completed on the clean PR head:

- `tests/unit_tests/models/gemma_vl/test_gemma3_vl_bridge.py`: 29 passed
- `tests/unit_tests/test_compare_mask_handling.py`: 14 passed

These are regression/helper results only; exact-checkpoint card leaves remain unverified pending gated model access.

yaoyu-33:
The fused-attention follow-up is stacked in #5203.

It addresses the two review points here as part of the same change: the contradictory packed-sequence smoke is removed while dense multimodal bias + THD remains unsupported, and the stale `qwen_vl_utils` fallback comment is corrected. Exact H100 parity now selects TE FusedAttention and passes the `0.99` cosine gate.

The existing GB200 converter CI failure on this PR is an unrelated NVIDIA PyPI HTTP 503 while fetching `inquirerpy`; the corresponding H100 converter job passed.

PR #5089, opened 2026-07-25
[Fix] Fix Qwen3-VL CUDA graph helper aliases without changing checkpoint.

Baibaifan:
Qwen3VLModel previously assigned the language model decoder and rotary
embedding modules onto the VLM root when CUDA graphs were enabled. Because
Qwen3VLModel is an `nn.Module`, those assignments registered root-level module
aliases and could make checkpoint state dicts expose decoder. `rotary_pos_emb.`
keys instead of preserving the expected language_model.* prefix.

This change exposes decoder and `rotary_pos_emb` through read-only properties
instead. CUDA graph helpers can still access `model.decoder` and
model.`rotary_pos_emb`, while PyTorch module registration and checkpoint naming
remain anchored under language_model.*.

Also updates the `Qwen3-VL` model tests to verify that the CUDA graph helper
aliases are accessible but are not registered as root-level modules.

**Pre checks**:

- [v ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [v] Did you write any new necessary tests?
- [v] Did you add or update any necessary documentation?
- [v] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)


yaoyu-33 (approved):

Baibaifan:
hi @yaoyu-33
Is there a problem with the CI status? It hasn't been refreshing the progress. Could you please rerun it?

Baibaifan:
> /ok to test [efd3719](https://github.com/NVIDIA-NeMo/Megatron-Bridge/commit/efd3719cd91be3e5aac8a7785b6b8471492f885e)

hi @yaoyu-33 @meatybobby 

I modified a file with over 80% code coverage. What could be causing the problem?

image

PR #5119, opened 2026-07-27
fix(ci): AUT-1100 keep verl and MCore on main

svcnemo-autobot:
## Background

Run 30245218834 failed because verl could not import a removed Megatron-LM API and skipped Megatron backend registration.

## What changed

- Keep verl and Megatron-LM on `main`.
- Make Energon optional for non-Energon Bridge use.
- Pin Transformers to `5.12.1` in verl suites.
- Respect Megatron-LM’s Python metadata.

## Details

Energon runtime code is imported lazily, and its builder tests patch that runtime module. Transformers 5.3 lacked `Exaone4_5_ForConditionalGeneration`; `5.12.1` is the immutable version installed before the successful local suites. The MCore API fix is NVIDIA/Megatron-LM#6065.

## Tested

- Full pre-commit gate, shell syntax, and `git diff --check`
- Energon builder + module-layout tests: 19 passed
- Clean Python 3.12 pip dry run for Megatron-LM `main`
- Exact 8-GPU Qwen3, DeepSeek, and MoE suites completed successfully with Transformers 5.12.1 installed before their dependency resolves


.github/workflows/verl_ci/run_megatron_deepseek.sh
```diff
@@ -27,8 +27,10 @@ trap cleanup EXIT
 
 ## Use our checkout
 pip3 install git+https://github.com/NVIDIA-NeMo/Megatron-Bridge.git@main --no-deps --no-build-isolation
-pip3 install git+https://github.com/NVIDIA/Megatron-LM.git@main --no-deps --no-build-isolation
+pip3 install git+https://github.com/NVIDIA/Megatron-LM.git@main --ignore-requires-python --no-deps --no-build-isolation
```
ko3n1g: @svcnemo-autobot why do we need `--ignore-requires-python`?

ko3n1g (approved):

PR #5131, opened 2026-07-28
feat(training): add MegatronMIMO in-batch sequence packing

going-song:
> **Update (review round 2):** packing lengths now come from the tokenizer's `attention_mask`
> shipped with the batch — no pad-id inference in the length path, and the round-1 `pad_token_id` field is
> gone again (zero new config fields). Packing fields moved into `_build_dataset_config`.

## What

Wires **in-batch sequence packing** into the MegatronMIMO training path. After the module-DP
slice, each language shard's real tokens are concatenated into one `[1, T]` THD sequence, and the
boundaries go to the model as `packing_kwargs`. mcore's `MimoModel.forward()` already consumes
them — the Bridge-side wiring was missing. Image-placeholder tokens count as real tokens, so the
modality splice still lands inside each sample's segment.

```text
before — dense [bs, S], LM attends over bs*S positions
  row 0: t t t t <img> <img> p p p p p p     S=12, real=6
  row 1: t t <img> t p p p p p p p p         S=12, real=4
                                             -> 24 positions, 14 of them padding

after  — packed [1, T] (THD), T = sum of real lengths
  [ t t t t <img> <img> | t t <img> t ]      T=10
  cu_seqlens = [0, 6, 10]                    -> block-diagonal attention per sample
```

First of three PRs splitting #4608, as agreed in #4609 (① sequence packing, ② scalable data
loading, ③ intra-microbatch reordering). Original work by **Chanwoo Park (@sailor1493)**,
credited as co-author and signer; @terryoo and I carry the split forward.

## Configuration (per review — no new config class)

| Path | Config | Where packing happens |
|---|---|---|
| Offline packed SFT | `dataset.enable_offline_packing` | dataset materialization |
| Direct-HF / VLM in-batch packing | `dataset.enable_in_batch_packing` | collate time |
| qwen3_vl step packing | `dataset.enable_in_batch_packing` + `defer_in_batch_packing_to_step=True` | `qwen3_vl_step` |
| **MegatronMIMO packing (this PR)** | `dataset.enable_in_batch_packing` + `defer_in_batch_packing_to_step=True` | `megatron_mimo_step`, after the module-DP slice |

- `megatron_mimo_step` **requires** the defer flag and raises otherwise: MIMO slices per
  module-DP rank before packing. Inverse of the `vlm_step` guard.
- No new config fields. Per-sample lengths come from the tokenizer's `attention_mask`, shipped
  with the language batch on every PP stage; the step raises when packing is enabled and the
  mask is missing. The mask distinguishes content tokens from padding by provenance, so ids
  shared between content and padding (e.g. eos-as-pad tokenizers) cannot truncate anything.

## Design notes

- All language PP stages must pack to the same `[1, T]` as the hidden states crossing the
  pipeline, so the mask (and with it the lengths) rides on every language stage.
- MRoPE packs `[3, B, S] -> [3, 1, T]`; a unit test asserts parity with `get_rope_index` on the
  packed layout (the branch the monolithic `Qwen3VLModel.forward` takes).
- The shared collate packer can't be reused as-is: it rejects 3-D MRoPE `position_ids`, derives
  lengths only internally, and requires a 2-D token tensor. All fixable with additive arguments —
  happy to unify in a follow-up.
- `intra_microbatch_pack.py` is renamed `sequence_pack.py`: generic shard packing, no ③
  dependency (per #4609, happy to move/rename).

## Changes

| File | Change |
|---|---|
| `data/megatron_mimo/sequence_pack.py` | **new** — THD concat, `cu_seqlens`, 2-D and MRoPE position ids; `pack_language_shard` returns `packing_kwargs` |
| `training/megatron_mimo_step.py` | `resolve_step_packing` gating; lengths from the batch `attention_mask`, pack the shard, pass `packing_kwargs` |
| `examples/megatron_mimo/qwen35_vl/finetune_qwen35_vl.py` | `--pack-sequences-in-batch` maps onto the dataset fields; ships the tokenizer `attention_mask` on every language stage when packing |
| tests | packing, gating, mask-length contract, MRoPE parity, `forward_step` wiring |

Off by default; with the switch unset the step path behaves exactly as before.

## Testing

Unit tests cover packing round-trip, PP-consistency, gating, the mask-length contract, MRoPE
parity, and `forward_step` wiring — all passing. Live PP=2 run (8xA100, 5 ranks:
`language=tp=1,pp=2,dp=2` + `images=dp=1`; Qwen3.5-0.8B config, random init, `cord_v2`;
`micro_batch_size=4` → 2 samples per language DP rank, real multi-segment packs): 10/10
iterations, no NaN, no skips, loss matches the packing-off run within bf16 noise:

```text
iter    pack ON   pack OFF      diff
   1    12.5744    12.5752   -0.0008
   2    12.5314    12.5324   -0.0010
   3    11.8536    11.8536   -0.0000
   4    11.2800    11.2788   +0.0012
   5    10.4698    10.4686   +0.0012
   6    10.5833    10.5838   -0.0005
   7     9.8351     9.8330   +0.0021
   8     9.6529     9.6515   +0.0014
   9     9.5024     9.5034   -0.0010
  10     9.3312     9.3316   -0.0004
```

## Known limitation

MegatronMIMO cannot run CP yet (`BridgeCommunicator` asserts `cp == 1`), so all of the above is
CP=1. Once CP lands, packed segments need `2 * cp` alignment via `cu_seqlens_padded`; the packer
does no alignment padding yet — planned follow-up.


src/megatron/bridge/training/config.py
```diff
@@ -955,6 +955,38 @@ class InProcessRestartConfig:
     """Directory for monitor process log files. If None, monitor process logging is disabled."""
 
 
+@dataclass
+class MegatronMIMOFeatureConfig:
```
yaoyu-33: not sure we want to add a new config for this, either should be paired in dataset / distributed.

going-song: Thanks for the suggestion. I dropped MegatronMIMOFeatureConfig; the switch now uses the existing dataset surface, exactly like the qwen3_vl step-packing path: dataset.enable_in_batch_packing + defer_in_batch_packing_to_step=True (the step raises an error without the deferral, as MIMO has to slice per module-DP rank before packing).

One field was added: dataset.pad_token_id, which the step needs to derive real lengths from input_ids. Details are in the updated PR description.

src/megatron/bridge/training/megatron_mimo_step.py
```diff
@@ -19,14 +19,48 @@
...
+        ``(pack_sequences_enabled, pad_token_id)``; ``(False, 0)`` when disabled.
+
+    Raises:
+        ValueError: If packing is enabled without step deferral or without a pad id.
+    """
+    if not bool(getattr(dataset_cfg, "enable_in_batch_packing", False)):
+        return False, 0
+    if not bool(getattr(dataset_cfg, "defer_in_batch_packing_to_step", False)):
+        raise ValueError(
+            "megatron_mimo_step requires step-time in-batch packing; set defer_in_batch_packing_to_step=True"
```
going-song, line 50: MIMO slices the batch per module-DP rank inside the step, so collate-time packing would leave nothing to slice — hence the inverse of the vlm_step guard.

src/megatron/bridge/data/megatron_mimo/sequence_pack.py
```diff
@@ -0,0 +1,240 @@
...
+        if src is None:
+            return None
+        out = torch.full((1, total), pad, dtype=src.dtype, device=device)
+        offset = 0
+        for i, seg in zip(group, seglens):
+            out[0, offset : offset + seg] = src[i, :seg]
+            offset += seg
+        return out
+
+    def _concat_mrope(src: torch.Tensor) -> torch.Tensor:
```
going-song, line 103: This [3, B, S] -> [3, 1, T] transform is the main reason the shared collate packer can't be reused as-is — it only accepts 2-D position_ids. A parity unit test asserts the result matches get_rope_index on the packed layout.

src/megatron/bridge/data/builders/direct_hf_sft.py
```diff
@@ -74,6 +74,9 @@ class DirectHFSFTDatasetConfig(DataloaderConfig):
     pad_to_max_length: bool = False
     pad_to_multiple_of: int = 128
     in_batch_packing_pad_to_multiple_of: int = 1
+    pad_token_id: int | None = None
```
going-song: The MIMO language batch carries attention_mask=None (required by MimoModel under CP), so input_ids != pad_token_id is the step's only length source — this must match the collator's actual pad id.

examples/megatron_mimo/qwen35_vl/finetune_qwen35_vl.py
```diff
@@ -1271,6 +1303,11 @@ def main() -> None:
         _log(f"checkpoint dir: {args.checkpoint_dir}")
         _log("building training config")
         cfg = _build_config(model_provider=model_provider, dataset_config=dataset_config, args=args)
+        if args.pack_sequences_in_batch:
+            # MegatronMIMO packs in the step, after the module-DP slice (deferred packing).
+            cfg.dataset.enable_in_batch_packing = True
+            cfg.dataset.defer_in_batch_packing_to_step = True
+            cfg.dataset.pad_token_id = hf_spec.pad_token_id
```
liding-nv: this can be moved to `_build_dataset_config` func

src/megatron/bridge/data/megatron_mimo/dp_utils.py
```diff
@@ -278,3 +278,39 @@ def slice_batch_for_megatron_mimo(
...
+    Returns:
+        An ``int64`` tensor of shape ``[B]`` with each sample's real length.
+    """
+    if (
+        isinstance(attention_mask, torch.Tensor)
+        and attention_mask.dim() == 2
+        and attention_mask.shape == input_ids.shape
+    ):
+        return attention_mask.to(torch.bool).sum(dim=1).to(torch.long)
+    return (input_ids != pad_token_id).sum(dim=1).to(torch.long)
```
liding-nv: `pad_token_id` can fall back to `eos_token_id` when the tokenizer has no dedicated pad token, as in `_tokenizer_pad_token_id()`. This could silently truncate real EOS tokens.

examples/megatron_mimo/qwen35_vl/finetune_qwen35_vl.py
going-song: Done — the packing fields are now set in `_build_dataset_config`.

src/megatron/bridge/data/megatron_mimo/dp_utils.py
going-song: Good catch — we removed the pad-id inference from the length path entirely. Lengths now come from the tokenizer's `attention_mask` shipped with the batch (the step raises without it), so a content token that shares the pad id can't be truncated; `dataset.pad_token_id` is gone as well. Re-validated with the PP=2 mb=4 pack on/off loss-parity run.

liding-nv (approved):

PR #5132, opened 2026-07-28
refactor(model): add Llama model config builder path

yaoyu-33:
## Summary

- extract the minimal builder-backed model-config foundation from #4545
- add opt-in `AutoBridge.get_model_config()` and `get_megatron_model()` APIs while preserving provider construction for unmigrated families
- migrate Llama HF config mapping to a serializable `BridgeGPTModelConfig` backed by Megatron Core `GPTModelBuilder`
- reject phantom config fields, preserve activation serialization, and restore weight-loading state on success or failure
- document the incremental Llama builder workflow

## Scope

This is PR 1 of the staged migration discussed for #4545. Only Llama opts into the new path. Other model families continue to use the existing provider path unchanged.

## Validation

- `uv run --no-sync python -m pytest tests/unit_tests/models/test_gpt_model_config.py tests/unit_tests/models/common/test_model_base.py tests/unit_tests/models/llama/test_llama_bridge.py tests/unit_tests/models/test_model_bridge.py` plus the five new focused `TestAutoBridge` cases: **71 passed**
- includes a real tiny CPU Llama construction through Megatron Core `GPTModelBuilder`
- `uv run --no-sync pre-commit run --all-files`: **passed**

No `3rdparty/Megatron-LM`, CI workflow, or `uv.lock` changes.


docs/bridge-guide.md
```diff
@@ -297,6 +324,8 @@ AutoBridge.list_supported_models() -> list[str]
 AutoBridge.supports(config: Any) -> bool
 
 # Provider/model construction
+AutoBridge.get_model_config() -> ModelConfig  # Builder-backed Llama config
+AutoBridge.get_megatron_model(model_config: ModelConfig | None = None, load_weights: bool = True, hf_path: str | Path | None = None, **kwargs) -> list[MegatronModule]
```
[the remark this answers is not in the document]
yaoyu-33: Updated in 331009b8996947f93dc641ff708b0338da92b6f2: the builder API is now `get_model(...)`, and the API reference includes the keyword-only `pg_collection: ProcessGroupCollection | None = None` parameter to match the implementation.

src/megatron/bridge/models/conversion/auto_bridge.py
```diff
@@ -1621,6 +1626,153 @@ def _load_and_export_adapter(model):
...
+            if isinstance(self.hf_pretrained, _PRETRAINED_WRAPPER_TYPES)
+            else self.hf_pretrained
+        )
+        model_config = self._model_bridge.hf_config_to_model_config(hf_config)
+
+        hf_name_or_path = getattr(self.hf_pretrained, "model_name_or_path", None)
+        if hf_name_or_path is None and isinstance(self.hf_pretrained, PretrainedConfig):
+            hf_name_or_path = getattr(self.hf_pretrained, "name_or_path", None)
+        if hf_name_or_path and hasattr(model_config, "hf_model_id"):
+            model_config.hf_model_id = str(hf_name_or_path)
```
maanug-nv: didn't we decide to put this stuff in `extra_checkpoint_metadata` instead of explicit attributes?

src/megatron/bridge/models/gpt/model_config.py
```diff
@@ -0,0 +1,107 @@
...
+
+from megatron.training.models.gpt import GPTModelConfig
+
+from megatron.bridge.models.common.base import ModelConfig
+from megatron.bridge.utils.activation_map import callable_to_str, str_to_callable
+from megatron.bridge.utils.instantiate_utils import _resolve_target
+
+
+@dataclass(kw_only=True)
+class BridgeGPTModelConfig(GPTModelConfig):
```
maanug-nv, line 28: the implementations for `__setattr__`, `get_builder_cls`, `as_dict`, `from_dict` seem safer/more robust than base classes' impls. The impls here avoid dropping callables when serializing, and also don't allow silently setting phantom fields. 
i think we could upstream them in the future. there's nothing specific to MBridge afaict. 

Also I think the two attributes in this class (`hf_model_*`) should be set in `extra_checkpoint_metadata`.

src/megatron/bridge/models/conversion/auto_bridge.py
yaoyu-33: actually I think this can be removed, no need anymore.

src/megatron/bridge/models/gpt/model_config.py
yaoyu-33: yep agree. hf_model* can be moved to extra metadata

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5132/commits/d50c037638bbac8e10f051da0895d4c5e23f5b5c

PR #5137, opened 2026-07-28
[inference] fix: Honor VLM inference parameter dtype

yaoyu-33:
## Summary

- honor the public VLM `params_dtype` argument during provider setup and Megatron checkpoint reconstruction
- keep `params_dtype`, `pipeline_dtype`, `bf16`, and `fp16` consistent before provider finalization
- pass the requested dtype into the final inference wrapper conversion
- add a focused regression for a non-default FP16 request

## Bug

`setup_model_and_tokenizer()` documents `params_dtype` as the model-parameter dtype, but the supported Qwen VLM checkpoint path hard-coded BF16 at every downstream boundary. For example, a caller requesting `torch.float16` received a BF16 provider configuration, BF16 checkpoint override, and BF16 final model conversion. The mismatch was silent and changed the requested numeric and hardware-compatibility contract.

## Fix

Propagate the requested dtype through the provider and checkpoint precision fields and into `setup_inference_wrapper()`. This follows the existing text-inference precision contract while leaving the BF16 default unchanged.

## Validation

Fail before:

```text
uv run --active --no-sync python -m pytest \
  tests/unit_tests/inference/vlm/test_base.py::TestSetupModelAndTokenizer::test_setup_model_and_tokenizer_propagates_params_dtype \
  -q
```

Result: `1 failed` because an explicit FP16 request observed `torch.bfloat16`.

Pass after:

```text
uv run --active --no-sync python -m pytest \
  tests/unit_tests/inference/vlm/test_base.py::TestSetupModelAndTokenizer::test_setup_model_and_tokenizer_propagates_params_dtype \
  -q
```

Result: `1 passed`.

Adjacent:

```text
uv run --active --no-sync python -m pytest tests/unit_tests/inference/vlm -q
```

Result: `38 passed`.

Static checks:

```text
git diff --check
```

Both passed; every applicable pre-commit hook passed.

## Scope

This verifies configuration and dtype propagation through the public VLM setup path. It does not claim GPU kernel, model-quality, convergence, or performance validation. No dependency, lockfile, CI workflow, public signature, or Megatron-Core source is changed.

### DIFF
--- a/src/megatron/bridge/inference/vlm/base.py
+++ b/src/megatron/bridge/inference/vlm/base.py
@@ -76,7 +76,10 @@ def setup_model_and_tokenizer(
     model_provider = bridge.to_megatron_provider(load_weights=False)
     model_provider.tensor_model_parallel_size = tp
     model_provider.pipeline_model_parallel_size = pp
-    model_provider.pipeline_dtype = torch.bfloat16
+    model_provider.params_dtype = params_dtype
+    model_provider.pipeline_dtype = params_dtype
+    model_provider.bf16 = params_dtype == torch.bfloat16
+    model_provider.fp16 = params_dtype == torch.float16
     model_provider.parallel_output = False
     model_provider.finalize()
     model_provider.initialize_model_parallel(seed=0)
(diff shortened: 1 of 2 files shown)


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5137/changes/7132e078b7be650bf2007e8b6abfa181a61c17a6

Issue #2652, opened 2026-03-04
feat: Add first-class Comet ML experiment tracking


shanecmoran:
## Description

Megatron Bridge has first-class integrations for TensorBoard, Weights & Biases, and MLflow as experiment tracking backends. All three follow a consistent pattern: config fields in `LoggerConfig`, lazy-initialized properties in `GlobalState`, and symmetric inline logging calls in `training_log()`, `eval.py`, and `checkpointing.py`.

This issue proposes adding **Comet ML** as a fourth first-class backend, following the exact same structural pattern.

## Motivation

Users who use Comet ML for experiment tracking currently rely on Comet's TensorBoard auto-patcher (`COMET_AUTO_LOG_TENSORBOARD=1`), which:

- Only captures metrics written to TensorBoard, missing WandB-only or MLflow-only metrics
- Depends on Comet intercepting `SummaryWriter` calls, which is fragile
- Does not capture config/hyperparameters as Comet parameters
- Does not support checkpoint artifact tracking
- Does not integrate with the DLFw tensor inspect system

A first-class integration would provide the same level of metrics, parameters, and artifact logging that WandB and MLflow users already have.

## Proposed Changes

Following the existing WandB/MLflow pattern:

1. **Config fields** — Add `comet_project`, `comet_experiment_name`, `comet_workspace`, `comet_api_key`, `comet_tags` to `LoggerConfig`
2. **Lazy init** — Add `GlobalState.comet_logger` property (last-rank init, stores `comet_ml.Experiment` instance)
3. **Training metrics** — Add Comet logging calls alongside existing WandB/MLflow calls in `training_log()`
4. **Validation metrics** — Add Comet logging in `eval.py`
5. **Timer metrics** — Add `_timers_write_to_comet` monkey-patch
6. **Checkpoint tracking** — Add `comet_utils.py` with save/load callbacks
7. **Tensor inspect** — Add `_CometExperimentLogger` wrapper
8. **Run plugin** — Add `CometPlugin` for NeMo Run launching
9. **Cleanup** — Add `experiment.end()` at shutdown
10. **Documentation** — Add Comet section to `docs/training/logging.md`

## Config Example

```python
cfg.logger.comet_project = "my-project"
cfg.logger.comet_experiment_name = "qwen3-14b-sft"
cfg.logger.comet_workspace = "my-workspace"
cfg.logger.comet_tags = ["sft", "qwen3"]
```

## Reference

[...]

PR #5141, opened 2026-07-28
fix(training): preserve timer samples across logging backends

yaoyu-33:
## What changed

Detailed training timers are stored in one shared Megatron-Core timer set. Bridge previously read that set for TensorBoard without resetting, then invoked the always-bound W&B, MLflow, and Comet timer writers in order with `reset=True` for each. The W&B slot therefore consumed the interval even when W&B was disabled, leaving MLflow and Comet with empty or near-zero detailed timer values.

The timer fanout now reads non-destructively for every available backend except the last one, which resets the interval exactly once. This preserves the same all-rank call order while allowing each enabled backend to receive the same completed sample.

## Supported trigger and impact

This affects ordinary training with:

- `log_timers_to_tensorboard=True`
- `timing_log_level=1` or `2`
- MLflow and/or Comet enabled

W&B does not need to be enabled. Training itself continues, but documented detailed metrics such as `forward-backward-time` and `optimizer-time` are silently absent or wrong in later logging backends.

Related logging-integration context: #2652.

## Regression evidence

The regression test binds the real Bridge timer-backend helpers to one stateful timer sample, disables W&B, enables MLflow and Comet, and asserts both later backends receive the same sample before it is cleared.

Fail before:

```bash
uv run --no-sync python -m pytest tests/unit_tests/training/utils/test_train_utils.py::TestTrainingLog::test_detailed_timers_reach_all_enabled_backends -q
```

Result: `1 failed`; MLflow received `{}` instead of `{"forward-backward-time": 6.0}`.

Pass after: the unchanged command reports `1 passed`.

Adjacent validation:

```bash
uv run --no-sync python -m pytest \
  tests/unit_tests/training/utils/test_train_utils.py::TestTrainingLog::test_detailed_timers_reach_all_enabled_backends \
  tests/unit_tests/training/utils/test_train_utils.py::TestTrainingLog::test_comet_only_logs_general_training_scalars \
  tests/unit_tests/training/test_state.py::TestGlobalState::test_timers_property_lazy_initialization \
  tests/unit_tests/training/test_state.py::TestGlobalState::test_timers_property_has_write_to_mlflow \
  tests/unit_tests/training/test_state.py::TestTimersWriteToMlflow \
  tests/unit_tests/training/test_state.py::TestCometLoggerProperty::test_timers_property_has_write_to_comet \
  tests/unit_tests/training/test_state.py::TestTimersWriteToComet -q
```

Result: `17 passed`.

Additional checks:

```bash
git diff --check
uv run --no-sync pre-commit run --all-files
```

Both passed; every applicable pre-commit hook passed.

## Scope

This changes timer telemetry fanout only. It does not change model computation, optimizer behavior, public APIs, dependencies, CI configuration, or the pinned Megatron-Core revision. Validation is a deterministic CPU contract test plus adjacent unit tests; no full-suite, model-quality, convergence, performance, or live tracking-service validation is claimed.

### DIFF
--- a/src/megatron/bridge/training/utils/train_utils.py
+++ b/src/megatron/bridge/training/utils/train_utils.py
@@ -821,14 +821,21 @@ def training_log(
     # Tensorboard values.
     # Timer requires all the ranks to call.
     if logger_config.log_timers_to_tensorboard and (iteration % logger_config.tensorboard_log_interval == 0):
-        reset_in_tb = False if hasattr(timers, "write_to_wandb") else True
-        timers.write(timers_to_log, writer, iteration, normalizer=total_iterations, reset=reset_in_tb)
-        if hasattr(timers, "write_to_wandb"):
-            timers.write_to_wandb(timers_to_log, wandb_writer, iteration, normalizer=total_iterations, reset=True)
-        if hasattr(timers, "write_to_mlflow"):
-            timers.write_to_mlflow(timers_to_log, mlflow_logger, iteration, normalizer=total_iterations, reset=True)
-        if hasattr(timers, "write_to_comet"):
-            timers.write_to_comet(timers_to_log, comet_logger, iteration, normalizer=total_iterations, reset=True)
+        timer_backends = [
+            (getattr(timers, "write_to_wandb", None), wandb_writer),
+            (getattr(timers, "write_to_mlflow", None), mlflow_logger),
+            (getattr(timers, "write_to_comet", None), comet_logger),
+        ]
+        timer_backends = [(write_fn, backend) for write_fn, backend in timer_backends if write_fn is not None]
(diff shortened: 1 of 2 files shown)


yaoyu-33:
Rebased onto latest main (was 20 commits behind). The failing `Launch_Unit_Tests_Core` was an unrelated pre-existing breakage in `tests/unit_tests/models/nemotronh/test_nemotron_h_bridge.py` (missing fixture `nemotron_nano_v2_config_dict`), since fixed upstream — the timer-fanout test itself passed. Re-triggering CI on the rebased HEAD.

/ok to test d7ea6c1bc57da6c714dd324508ed63241faeeea7

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/commit/d7ea6c1bc57da6c714dd324508ed63241faeeea7

PR #4601, opened 2026-06-30
fix(data): make finetuning batch sampler epoch-aware on checkpoint resume

Achyuthan-S:
# What does this PR do ?

Fixes abnormal loss when resuming fine-tuning from a checkpoint (#4565). The
fine-tuning batch sampler froze `consumed_samples` and re-applied the resume
offset on every epoch, so after a mid-epoch resume each later epoch skipped the
start of the dataset and replayed only the tail — repeated data that depressed
the loss. This makes `MegatronPretrainingBatchSampler` epoch-aware (and adds
deterministic per-epoch reshuffling), mirroring `MegatronPretrainingRandomSampler`.

# Changelog

- `data/samplers.py`: `MegatronPretrainingBatchSampler.__iter__` now advances
  `consumed_samples` as it yields and computes the within-epoch offset against the
  whole-global-batch multiple, so each `cyclic_iter` re-iteration starts a fresh
  full epoch instead of permanently re-applying the resume offset and replaying
  the tail.
- `data/samplers.py`: add `shuffle`/`seed` params for deterministic, seed- and
  epoch-derived per-epoch reshuffling (resume-reproducible). Threaded through
  `build_pretraining_data_loader`. Both default to off → backward compatible.
- `data/loaders.py`: enable shuffle for the fine-tuning (`batch`) **train**
  dataloader, seeded by the dataset seed. Eval/test and pretraining paths are
  unchanged.
- `tests/.../data/test_samplers.py`: regression tests for resume-serves-full-epoch,
  epoch-boundary resume, per-epoch reshuffle, seed determinism, and shuffle resume
  parity. Existing batch-sampler assertions are unchanged.

**Behavior change:** fine-tuning data order now reshuffles each epoch
(deterministic and resume-reproducible). SFT loss curves will shift versus the
previous sequential order. New params are optional and default off.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci) in the Contributing doc for how to trigger the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

# Before your PR is "Ready for review"

**Pre checks**:

- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests? (sampler resume-parity + reshuffle regression tests)
[...]

PR #5142, opened 2026-07-28
feat(data): expose `shuffle` in DataloaderConfig

filaretov:
# What does this PR do ?

This PR exposes `shuffle` as an argument in the top-level dataset configuration API so it can be set in `MegatronPretrainingBatchSampler`. It was added in #4601 and it changed the convergence behavior of our Llama 2 70b LORA benchmark.

# Changelog

- Add `shuffle` to necessary config classes to expose it through `DataloaderConfig`.

**Pre checks**:

### DIFF
--- a/src/megatron/bridge/data/base.py
+++ b/src/megatron/bridge/data/base.py
@@ -43,8 +43,12 @@ class DataloaderConfig:
     drop_last: bool = True
     """Whether dataloaders drop the last incomplete batch."""
 
-    shuffle: bool = True
-    """Whether the global-batch sampler used by ``dataloader_type="batch"`` reshuffles samples each epoch."""
+    train_shuffle: bool = True
+    """Whether the training global-batch sampler reshuffles samples each epoch.
+
+    Only used when ``dataloader_type="batch"``. Validation and test samples are
+    never shuffled.
+    """
 
     persistent_workers: bool = True
     """Whether dataloader workers persist between iterations."""
--- a/src/megatron/bridge/data/base.py
+++ b/src/megatron/bridge/data/base.py
@@ -43,6 +43,9 @@ class DataloaderConfig:
     drop_last: bool = True
     """Whether dataloaders drop the last incomplete batch."""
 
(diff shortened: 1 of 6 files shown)


yaoyu-33:
@filaretov  this is still in draft, kindly remind to "ready for review" if need merge

filaretov:
@yaoyu-33 Thank you for the reminder. Yeah, it'd be nice to merge it if possible, it's ready for review.

yaoyu-33:
I think this is addressing the right user need at the wrong configuration layer.

After #4601, `MegatronPretrainingBatchSampler` already performs deterministic per-epoch shuffling by default (`shuffle=True`, seeded by epoch), while `dataloader_type="cyclic"` selects `MegatronPretrainingRandomSampler`. For `GPTSFTDatasetConfig`, however, `batch` is intentional: its `collate_fn` must see the full global batch to determine variable-length padding before the training loop splits it into microbatches. Switching SFT to `cyclic` is therefore not an equivalent way to control shuffling.

If the goal is simply the normal policy, I do not think we need a public `DataloaderConfig.shuffle`: keep the batch sampler, enable shuffle for the train loader, and explicitly disable it for validation/test. If users genuinely need to opt out of the new #4601 training order (for example, to reproduce the older LoRA benchmark), the option should be train-only or sampler-specific rather than a shared dataset flag. The current patch passes the same value to train, validation, and test, so the default `True` also explicitly shuffles evaluation, while setting it to `False` changes all three streams together.

Could we revise this to one of these shapes?

1. No public option: `shuffle=True` for the batch train loader and `shuffle=False` for batch validation/test; or
2. A clearly train-scoped option such as `train_shuffle`, used only for the train loader, with validation/test always false.

The focused tests should cover the resulting contract: shuffled train order remains deterministic and resume-equivalent, the train-only opt-out (if retained) is sequential, and validation/test order is deterministic and unshuffled.

filaretov:
Thanks for the review. Is the updated PR what you had in mind with option 2?

PR #5150, opened 2026-07-28
feat(misc): add memory snapshot analysis skill

minitu:
Adds `nemo-mbridge-memory-snapshot-analysis`, a skill for analyzing the CUDA memory snapshots Bridge already writes via
`ProfilingConfig(record_memory_history=True)`.

Today `docs/training/profiling.md` points users at PyTorch's Memory Viz and stops there. This skill bundles five stdlib-only scripts that go further: replay the recorded allocation timeline, diff two runs to attribute a peak memory difference to specific source locations, plot an interactive timeline, and drill into a single moment. It also handles the OOM snapshot Bridge dumps automatically on failure.

The SKILL.md documents Bridge-specific capture semantics that are easy to get wrong:

- `external_annotations` is populated with NCCL/optimizer/dataloader markers but carries no `ProfilerStep#N` entries unless `use_pytorch_profiler=True` is also set, so per-step replay is unavailable by default and `compare_snapshots.py` falls back to a full-trace replay.
- Since 9e9e58b9f, recording starts before model construction, so weights land inside the trace and `baseline_at_start` is small; older snapshots behave the opposite way.
- `trace_alloc_max_entries` is capped at 100k, so long runs can lose their beginning.
- Snapshots are written per rank as `snapshot_{rank}.pickle`, with OOM dumps as `snapshot_oom_rank-{N}.pickle`.

Scripts use only the Python standard library; the generated HTML timeline loads Plotly from a CDN in the browser. Verified against real Bridge snapshots, including a matched CUDA-graphs on/off pair.

Registers the skill in `docs/skills-index.md` and the Fern nav, and links it from the memory-profiling section of the profiling docs.

`BENCHMARK.md` and `skill.oms.sig` are produced by NVSkills CI.

# What does this PR do ?

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/docs/training/profiling.md
+++ b/docs/training/profiling.md
@@ -167,7 +167,13 @@ cfg.profiling = ProfilingConfig(
 
 ### Analyze Memory Usage
 
-After the run completes, memory snapshots for each specified rank are saved to the designated path. Load these traces using the PyTorch Memory Viz tool to plot memory usage over time and detect bottlenecks or leaks in your training pipeline.
+After the run completes, memory snapshots for each specified rank are saved to the designated path, with the rank inserted before the extension (for example, `memory_trace_0.pickle`). Load these traces using the PyTorch Memory Viz tool to plot memory usage over time and detect bottlenecks or leaks in your training pipeline.
+
+For comparing two runs, attributing peak memory to specific allocation sites, or drilling into a captured out-of-memory snapshot, see the [Memory Snapshot Analysis skill](../../skills/nemo-mbridge-memory-snapshot-analysis/SKILL.md), which bundles scripts for replaying and diffing these files.
+
+```{note}
+`record_memory_history` and `use_pytorch_profiler` are independent settings. `ProfilerStep#N` annotations inside the snapshot come from the PyTorch profiler, so a snapshot captured with `record_memory_history=True` alone contains no step boundaries. Enable both if you want per-step memory analysis.
+```
 
 ## Optimize Profiling Accuracy
 
--- a/docs/training/profiling.md
+++ b/docs/training/profiling.md
@@ -169,7 +169,7 @@ cfg.profiling = ProfilingConfig(
 
 After the run completes, memory snapshots for each specified rank are saved to the designated path, with the rank inserted before the extension (for example, `memory_trace_0.pickle`). Load these traces using the PyTorch Memory Viz tool to plot memory usage over time and detect bottlenecks or leaks in your training pipeline.
 
-For comparing two runs, attributing peak memory to specific allocation sites, or drilling into a captured out-of-memory snapshot, see the [Memory Snapshot Analysis skill](../../skills/nemo-mbridge-memory-snapshot-analysis/SKILL.md), which bundles scripts for replaying and diffing these files.
+For comparing two runs, attributing peak memory to specific allocation sites, or drilling into a captured out-of-memory snapshot, see the [Memory Snapshot Analysis skill](../skills/nemo-mbridge-memory-snapshot-analysis/SKILL.md), which bundles scripts for replaying and diffing these files.
 
 ```{note}
 `record_memory_history` and `use_pytorch_profiler` are independent settings. `ProfilerStep#N` annotations inside the snapshot come from the PyTorch profiler, so a snapshot captured with `record_memory_history=True` alone contains no step boundaries. Enable both if you want per-step memory analysis.
--- a/skills/nemo-mbridge-memory-snapshot-analysis/scripts/common.py
+++ b/skills/nemo-mbridge-memory-snapshot-analysis/scripts/common.py
@@ -1,0 +1,404 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Shared utilities for CUDA memory snapshot analysis."""
+
+import logging
+import os
+import pickle
+import sys
+from collections import defaultdict
+from dataclasses import dataclass, field
+
+
+# ---------------------------------------------------------------------------
+# Frame filtering – paths considered "internal" (not user code)
+# ---------------------------------------------------------------------------
+_INTERNAL_PATH_FRAGMENTS = (
+    "/torch/",
+    "/torch_",
+    "torch/nn/modules/module.py",
+    "torch/autograd/",
+    "torch/amp/",
+    "torch/cuda/",
+    "torch/distributed/",
+    "/cuda/",
+    "/hydra/",
+    "/python3.",
+    "/importlib/",
+    "/runpy.py",
+)
+
+_INTERNAL_FUNC_NAMES = {"<module>", "<lambda>", "_call_impl", "_wrapped_call_impl", "inner"}
+
+
+logger = logging.getLogger(__name__)
+
+
+def format_size(num_bytes: int) -> str:
+    """Human-readable byte size, e.g. '37.91 GB', '58.72 MB'."""
+    abs_b = abs(num_bytes)
+    sign = "-" if num_bytes < 0 else ""
+    if abs_b >= 1e9:
+        return f"{sign}{abs_b / 1e9:.2f} GB"
+    if abs_b >= 1e6:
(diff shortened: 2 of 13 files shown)


skills/nemo-mbridge-memory-snapshot-analysis/scripts/compare_snapshots.py
```diff
@@ -0,0 +1,457 @@
...
+    base_b = baseline_b.baseline_at_start if baseline_b else 0
+
+    # --- Per-step starting deltas ---
+    deltas_a = compute_step_start_deltas(traces_a, steps_a)
+    deltas_b = compute_step_start_deltas(traces_b, steps_b)
+
+    # --- Replay each matching step ---
+    step_results = []
+    for num in matching:
+        events_a = get_step_events(traces_a, steps_a[num]["start"], steps_a[num]["end"])
```
yaoyu-33, line 158: **High:** This replays only events inside the step into `replay_events()`, whose live set starts empty. Allocations still live at step start (weights, optimizer state, graph pools, or earlier activations) therefore contribute to `absolute_peak` through `step_start_delta` but are absent from `peak_live_set` and the source-attribution table. I reproduced a step with a 100-byte persistent allocation plus a 20-byte temporary allocation: the reported absolute peak was 120 bytes, while the table attributed only 20 bytes. Please seed the replay from the trace prefix (or replay the full trace while snapshotting within the step), and choose the drill-down using `abs_peak_diff` rather than only the within-step `peak_diff`.

skills/nemo-mbridge-memory-snapshot-analysis/scripts/parse_snapshot.py
```diff
@@ -0,0 +1,202 @@
...
+        path: Path the snapshot was loaded from, used for the file-size line.
+        as_json: Emit machine-readable JSON instead of a text report.
+    """
+    file_size = os.path.getsize(path)
+
+    # --- Segments ---
+    segments = snapshot.get("segments", [])
+    total_size = sum(s.get("total_size", 0) for s in segments)
+    active_size = sum(s.get("allocated_size", 0) for s in segments)
+    inactive_size = total_size - active_size
```
yaoyu-33, line 63: **High:** `total_size - allocated_size` is not allocator fragmentation. PyTorch snapshots distinguish `allocated_size` from `active_size`; the latter also includes `active_awaiting_free`, while truly inactive reusable memory is `total_size - active_size`. The current calculation counts awaiting-free blocks as fragmentation and can substantially over-report it. Please report allocated/active/reserved separately and calculate inactive/fragmentation from `active_size`. The same accounting/naming should be corrected in `compare_snapshots.py`.

skills/nemo-mbridge-memory-snapshot-analysis/scripts/plot_timeline.py
```diff
@@ -0,0 +1,507 @@
...
+            raw_deltas,
+            time_points,
+        )
+
+        times_s = [(t - t0) / 1e6 for t in ds_times]
+        delta_gb_time = [d / GB for d in ds_deltas_t]
+        event_idx = list(ds_indices)
+        delta_gb_event = [d / GB for d in ds_deltas_e]
+
+        peak_delta_gb = max(d / GB for d in raw_deltas) if raw_deltas else 0
```
yaoyu-33: **Medium:** The initial zero-delta state is excluded from this maximum. For an older or ring-buffer-truncated trace whose first visible event is a free, all `raw_deltas` can be negative even though the reconstructed baseline was the true peak at trace start. A synthetic unmatched 100-byte free produced `baseline=100` but `peak_abs=0`. Please seed the time series with a zero-delta point at trace start, or at minimum calculate `max(0, *raw_deltas)` and include that initial point in the plot.

skills/nemo-mbridge-memory-snapshot-analysis/scripts/common.py
```diff
@@ -0,0 +1,404 @@
...
+
+def load_snapshot(path: str) -> dict:
+    """Load a CUDA memory snapshot pickle file."""
+    if not os.path.isfile(path):
+        logger.error(f"file not found: {path}")
+        sys.exit(1)
+    size_mb = os.path.getsize(path) / 1e6
+    logger.info(f"Loading {path} ({size_mb:.1f} MB) ...")
+    with open(path, "rb") as f:
+        data = pickle.load(f)
```
yaoyu-33, line 96: **Security:** `pickle.load()` can execute arbitrary code. The trust warning currently appears only in `skill-card.md`, which is not normally loaded when an agent follows `SKILL.md`; the workflow directly tells users to run these scripts. Please put a prominent "trusted snapshots only" warning in `SKILL.md` before the first command (and ideally in the CLI help/error path) so users are warned at the execution point.

skills/nemo-mbridge-memory-snapshot-analysis/scripts/compare_snapshots.py
```diff
@@ -0,0 +1,457 @@
...
+logger = logging.getLogger(__name__)
+
+
+def diff_settings(settings_a: dict, settings_b: dict) -> list:
+    """Return list of (key, val_a, val_b) for differing settings."""
+    all_keys = sorted(set(list(settings_a.keys()) + list(settings_b.keys())))
+    diffs = []
+    for k in all_keys:
+        if k == "roundup_power2_divisions":
+            continue
```
yaoyu-33: **Medium:** Please do not silently omit `roundup_power2_divisions` from the allocator-settings diff. It directly changes allocator rounding and can be the reason two runs have different reserved memory or fragmentation—the exact behavior this tool is meant to diagnose. If its raw representation is noisy, normalize it before comparison instead of suppressing it.

skills/nemo-mbridge-memory-snapshot-analysis/skill-card.md
```diff
@@ -0,0 +1,67 @@
...
+Mitigation: Review and scan skill before deployment. <br>
+
+Risk: Memory snapshot pickle files are deserialized with `pickle.load`, which executes arbitrary code for untrusted input. <br>
+Mitigation: Only analyze snapshot files produced by your own training runs or from a trusted source. <br>
+
+Risk: Snapshot files embed absolute source paths and Python stack frames from the machine that produced them, which may reveal internal directory layouts. <br>
+Mitigation: Review before sharing snapshot files or script output outside your organization. <br>
+
+## Reference(s): <br>
+- [Profiling Documentation](docs/training/profiling.md) <br>
```
yaoyu-33: These local references are broken when rendered from this file: GitHub resolves `docs/...` and `skills/...` relative to `skills/nemo-mbridge-memory-snapshot-analysis/`. Please use `../../docs/training/profiling.md` for the docs link and `../nemo-mbridge-*/SKILL.md` for sibling skills (or repository-absolute URLs).

yaoyu-33 (changes requested):
The snapshot-analysis workflow is useful, but the replay and allocator accounting need correction before merge. Please also add focused synthetic unit tests for baseline reconstruction, step replay, and traces that begin with unmatched frees.

malay-nagda (dismissed):

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5150/commits/e9f9c3f70a6c8d57872c4df216311487de1fd080

docs/training/profiling.md
```diff
@@ -167,7 +167,13 @@ cfg.profiling = ProfilingConfig(
 
 ### Analyze Memory Usage
 
-After the run completes, memory snapshots for each specified rank are saved to the designated path. Load these traces using the PyTorch Memory Viz tool to plot memory usage over time and detect bottlenecks or leaks in your training pipeline.
+After the run completes, memory snapshots for each specified rank are saved to the designated path, with the rank inserted before the extension (for example, `memory_trace_0.pickle`). Load these traces using the PyTorch Memory Viz tool to plot memory usage over time and detect bottlenecks or leaks in your training pipeline.
+
+For comparing two runs, attributing peak memory to specific allocation sites, or drilling into a captured out-of-memory snapshot, see the [Memory Snapshot Analysis skill](../../skills/nemo-mbridge-memory-snapshot-analysis/SKILL.md), which bundles scripts for replaying and diffing these files.
```
malay-nagda: @minitu I think this needs to change to-
```suggestion
For comparing two runs, attributing peak memory to specific allocation sites, or drilling into a captured out-of-memory snapshot, see the [Memory Snapshot Analysis skill](../skills/nemo-mbridge-memory-snapshot-analysis/SKILL.md), which bundles scripts for replaying and diffing these files.
```

From agent-
From docs/training/, that points outside Sphinx’s source tree to the repository-level skills/ directory. During the build, docs/conf.py copies skills into docs/skills/, so Sphinx expects the document there. It reports:

```
WARNING: Unknown source document '.../skills/nemo-mbridge-memory-snapshot-analysis/SKILL'
[myst.xref_missing]
```

Because CI runs Sphinx with --fail-on-warning, that single warning causes exit code 1.

minitu: Thanks, pushed the fix. Hopefully the CI will pass this time.

malay-nagda (approved):

yaoyu-33 (approved):

PR #5151, opened 2026-07-28
 feat(conversion): enable Qwen3.5 MoE MIMO round trips

liding-nv:
Bind the HF source before resolving checkpoint-dependent mappings and use route-local expert parallelism when accumulating grouped expert exports.

Save an MTP-disabled HF configuration when MIMO omits MTP weights, and add Qwen3.5 MoE route, EP wrapper, and regression coverage.

### DIFF
--- a/src/megatron/bridge/models/megatron_mimo/conversion/orchestrator.py
+++ b/src/megatron/bridge/models/megatron_mimo/conversion/orchestrator.py
@@ -30,7 +30,12 @@ import torch.nn as nn
 from megatron.core.transformer.module import MegatronModule
 from transformers.configuration_utils import PretrainedConfig
 
-from megatron.bridge.models.conversion.auto_bridge import AutoBridge
+from megatron.bridge.models.conversion.auto_bridge import (
+    MTP_CONFIG_FIELDS,
+    AutoBridge,
+    _model_omits_mtp,
+    _mtp_source_key_prefixes,
+)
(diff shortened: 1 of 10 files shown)


src/megatron/bridge/models/megatron_mimo/conversion/orchestrator.py
```diff
@@ -279,13 +284,32 @@ def build_route_local_registry(
     return MegatronMappingRegistry(*stripped)
 
 
+def bind_hf_source_to_bridge(source_bridge: "MegatronModelBridge", hf_pretrained: Any) -> None:
```
yaoyu-33, line 287: why this is needed?

liding-nv: The standard path establishes this ordering in `build_conversion_tasks()` https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/dcd7ebb7fb87c7e2cec158bdaa200466822c964a/src/megatron/bridge/models/conversion/model_bridge.py#L1670-L1671 
it binds hf_pretrained and hf_config before `mapping_registry()`

MIMO constructs its component-specific registry before calling `build_conversion_tasks()`, so it must establish the same binding earlier.

cuichenx (approved):

cuichenx:
tests passed https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/30418067986/job/90485898348

PR #5162, opened 2026-07-29
perf(diffusion): use shared segment-tree first-fit for sequence packing

bzantium:
## What does this PR do?

Replace the linear-scan first-fit in the diffusion sequence packer with the segment-tree implementation already used by offline SFT packing. Bin assignments are unchanged.

`find_first_bin_that_fits()` scans every open bin and recomputes `sum(abin)` on each probe, so `first_fit()` is quadratic in the number of samples. `DiffusionTaskEncoder.select_samples_to_pack()` calls it once per packing buffer during training, so this cost recurs for the whole run rather than being paid once offline. `data/packing/algorithms.py` already carries an O(N log N) segment-tree first-fit, but it packs integers and returns integers, while the diffusion path packs `DiffusionSample` objects and needs those same objects back in the bins.

The fix:

- adds an optional keyword-only `item_lengths` to `first_fit()` in `data/packing/algorithms.py`, so capacity accounting can key on a supplied length while the original entries land in the bins;
- keeps `item_lengths=None` meaning "each entry is its own length", leaving existing two-argument calls untouched, including `create_packing_strategy()`, which resolves the packing function by name through `globals()` and calls it positionally;
- routes the diffusion `first_fit()` through that implementation instead of its own scan;
- resolves a sample's length with `0 + item`, preserving `DiffusionSample.__radd__` and its `seq_len_q_padded` over `seq_len_q` precedence without importing the sample type into the packing utils.

`find_first_bin_that_fits()` is no longer used by `first_fit()`, but it is public and separately tested, so it is left in place and can be removed in a follow-up.

## Testing

New coverage compares full bin assignments against the original linear scan rather than bin counts, since the task encoder depends on which bin each sample lands in. It covers randomized integer and object inputs, zero-length, oversized and exact-fit edge cases, object identity and conservation, and the `item_lengths` argument itself.

This checkout has an uninitialized `3rdparty/Megatron-LM` submodule, so `uv sync` fails and `megatron.core` cannot be imported. The packing tests were run against the real source files with `megatron.core` stubbed and `--noconftest`:

```text
tests/unit_tests/diffusion/data/common/test_sequence_packing_utils.py   33 passed
tests/unit_tests/data/packing/test_algorithms.py                        21 passed
tests/functional_tests/test_groups/data/packing/test_algorithms.py      17 passed
```

These are pure-CPU Python paths, but the runs under the project toolchain still need confirming:

```text
uv run pre-commit run --all-files
uv run python -m pytest tests/unit_tests/diffusion/data/common/test_sequence_packing_utils.py tests/unit_tests/data/packing/test_algorithms.py -q
```

`ruff check` and `ruff format` pass. `pre-commit` itself could not run here because its launcher is missing the `pre_commit` module, so `trailing-whitespace` and `end-of-file-fixer` were checked by hand on the changed files.

Packing time on uniformly drawn lengths with `pack_size=8192`:

```text
N=2000                  0.112s -> 0.011s
N=8000                  1.961s -> 0.057s
N=20000                13.425s -> 0.174s
N=8000 sample objects   8.878s -> 0.042s
```

The object case gains most because the previous code re-evaluated `sum(abin)` through `__radd__` on every probe.

## Scope

Limited to bin-packing order inside sequence packing. No existing signature changes incompatibly and no dependencies are added. No convergence, GPU behavior, at-scale performance, or full-suite validation is claimed.

CI needs `/ok to test <sha>` from an NVIDIA developer, since these commits are not GPG-signed.

### DIFF
--- a/src/megatron/bridge/data/packing/algorithms.py
+++ b/src/megatron/bridge/data/packing/algorithms.py
@@ -16,7 +16,7 @@
 
 import collections
 import logging
-from typing import Dict, List, Tuple
+from typing import Dict, List, Sequence, Tuple, TypeVar
 
 import numpy as np
 from tqdm import tqdm
@@ -27,6 +27,8 @@ from megatron.bridge.utils.safe_pickle import safe_load_npy
 PACKING_ALGOS = ["first_fit_decreasing", "first_fit_shuffle"]
 
 logger = logging.getLogger(__name__)
+
+_ItemT = TypeVar("_ItemT")
 
 
 class _SegmentTree:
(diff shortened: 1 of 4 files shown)


cuichenx (approved):

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5162/commits/f62ede8d856300d246c434cc6e9d0dcc1028cb20

PR #5170, opened 2026-07-29
[DO NOT MERGE] fix(model): read RADIO class_token_len from HF vision_config in NemotronOmniBridge

yaoyu-33:
## Summary

- `NemotronOmniBridge.provider_bridge()` previously hard-coded `vision_class_token_len = 10`. Nano-3.5-Omni SFT checkpoints use **8 RADIO class tokens**, causing a tensor shape mismatch at conversion time. Now reads the value from `hf_config.vision_config.class_token_len` (fallback 10).
- Adds `scripts/conversion/prepare_nemotron_omni_legacy_ckpt.py` — a zero-copy pre-processing script that makes pre-v0.6 Nano-Omni MCore checkpoints compatible with the current `convert.sh` without touching the source files.

> [!WARNING]
> **This PR is not yet usable end-to-end against the public HF model.** See
> [Known gaps](#known-gaps) below. The conversion pipeline works, but it currently
> depends on a hand-modified local HF reference config that is not part of this PR.

---

## Known gaps

These were found while re-validating on HSG and are **not fixed by this PR**.

### 1. The bridge fix is inert against the public HF config

The public `nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-BF16` `config.json` has **no
`vision_config.class_token_len` field at all**:

| field | public config | value needed for this checkpoint |
|---|---|---|
| `vision_config.class_token_len` | *absent* | `8` |
| `vision_config.separate_video_embedder` | `true` | `false` |
| `sound_config` | present | absent (`has_sound=False`) |
| `vision_config.args.register_multiple` | `10` | `8` (see gap 2) |

So `getattr(vision_config, "class_token_len", 10)` returns **10**, not 8, and the
original shape mismatch reappears. The validated run below only works because it
points at a locally hand-edited copy of the reference config.

**Needs a decision:** either add these fields to the public model repo's
`config.json`, or have the bridge read them from the checkpoint's own `common.pt`
args (which *do* carry them) instead of from the HF config.

### 2. The exported `modeling.py` cannot be loaded back unpatched

`modeling.py` is copied verbatim into the export via `ADDITIONAL_FILE_PATTERNS`. The
exported `config.json` ends up with `class_token_len=8` but `args.register_multiple=10`,
and RADIO is built from `register_multiple` — so it constructs 10 class tokens for
8-token weights. It also unconditionally creates `pg.video_embedder`, which does not
exist in a checkpoint trained with `separate_video_embedder=false`.

Loading the export requires patching the wrapper to:
- set `config.vision_config.args["register_multiple"] = class_token_len`
- only create `pg.video_embedder` when `separate_video_embedder` is true

This patch is **not in this PR**.

### 3. `convert.sh --executor slurm` needs the repo bind-mounted

`scripts/conversion/setup_conversion.py:34` hardcodes
`CONTAINER_REPO_ROOT = /opt/Megatron-Bridge`, so the repo must be mounted there.
Without it the job fails in under 5 minutes with:

```
/opt/venv/bin/python: can't open file '/opt/Megatron-Bridge/scripts/conversion/run_conversion.py': [Errno 2] No such file or directory
```

---

## How to run it (validated 2026-07-30)

Prerequisite on the HSG login node — `convert.sh` requires `uv`, which is not installed there:

```bash
python3 -m pip install --user uv
export PATH=$HOME/.local/bin:$PATH
```

### Step 1 — build the compatibility view (CPU only)

Use the `cpu` partition with the `cpu-short` QOS. The GPU `batch` partition queues for
an hour or more at this account's priority, and also rejects jobs that request no GPUs
(`Cannot find GPU specification`) while its QOS enforces a GPU *minimum* (`QOSMinGRES`).

```bash
srun --account=<ACCOUNT> --partition=cpu --qos=cpu-short \
     --nodes=1 --ntasks=1 --time=00:30:00 \
     --container-image=nvcr.io#nvidia/nemo:26.06 \
     --container-mounts=/lustre:/lustre --no-container-mount-home \
     bash -c 'cd $REPO && uv run python scripts/conversion/prepare_nemotron_omni_legacy_ckpt.py \
        --source-ckpt <LUSTRE>/users/guyueh/nano-3.5-omni-hf-ckpt/iter_0000020 \
        --hf-ref      <HF_REFERENCE_DIR> \
        --output-dir  <OUT>/compat-checkpoint'
```

`<HF_REFERENCE_DIR>` must currently be a local copy of the public config with
`class_token_len: 8`, `separate_video_embedder: false`, and `sound_config` removed
(gap 1). Allow ~25 min; most of it is `uv sync` building `megatron-core`.

### Step 2 — verify the generated `run_config.yaml` before spending a GPU

```python
from megatron.bridge.training.model_load_save import load_model_config
cfg, _ = load_model_config("<OUT>/compat-checkpoint/iter_0000020")
assert cfg.num_layers == 52
assert cfg.hidden_size == 2688
assert cfg.vision_class_token_len == 8
```

This is cheap and catches a silently-empty config before the export job is queued.

### Step 3 — export to HF (4 GPU)

```bash
./scripts/conversion/convert.sh export \
    --executor slurm --device gpu --nodes 1 --gpus-per-node 4 \
    --account <ACCOUNT> --partition batch --time 03:00:00 \
    --container-image nvcr.io/nvidia/nemo:26.06 \
    --mount /lustre \
    --mount $REPO:/opt/Megatron-Bridge \
    --hf-model      <HF_REFERENCE_DIR> \
    --megatron-path <OUT>/compat-checkpoint \
    --hf-path       <OUT>/hf-export \
    --torch-dtype bfloat16 \
    --tp 2 --pp 1 --ep 2 --etp 1 \
    --trust-remote-code --distributed-timeout-minutes 60 --not-strict \
    --distributed-save --save-every-n-ranks 1 --detach
```

`--torch-dtype`, `--distributed-timeout-minutes`, `--distributed-save` and
`--save-every-n-ranks` are carried over from the original working run; they have not
been individually re-tested for necessity.

### Step 4 — patch the export before loading it

Apply the wrapper fix from gap 2 to `<OUT>/hf-export/modeling.py`, otherwise the
exported checkpoint will not load.

---

## What the compat script fixes

| Issue | Symptom | Fix |
|---|---|---|
| `common.pt` stores `ModelType.encoder_and_decoder (2)` | `ValueError: 2 is not a valid ModelType` | Script patches `common.pt` |
| DCP keys use `language_model.*` / `vision_model.*` / … | shape/key mismatch on load | Script rewrites `.metadata` pickle |
| No `run_config.yaml` | `Could not find run_config.yaml` | Script generates one from HF reference + `common.pt` args |

The source checkpoint is **never modified** — shards are symlinked.

### Note on `run_config.yaml` generation

`run_config.yaml` must be produced via `ConfigContainer._convert_value_to_dict(provider)`.
Passing the provider directly to the YAML representers looks correct but emits only
`_target_`/`_call_` with no field values, so `load_model_config()` returns a
default-constructed provider (`num_layers=None`, `hidden_size=0`) and every
checkpoint-specific override is silently lost. Step 2 above exists to catch exactly this.

### MTP support

The nano-3 omni MTP is not currently exported. Not nano-3.5.

## Commits

1. `fix(model)` — read `vision_class_token_len` from HF `vision_config` in `NemotronOmniBridge`
2. `feat(ckpt)` — `prepare_nemotron_omni_legacy_ckpt.py` compat pre-processing script
3. `fix(ckpt, model)` — generate a `run_config.yaml` that actually loads

## Test plan

- [x] `test_nemotron_omni_provider_bridge_maps_public_config_fields` asserts `provider.vision_class_token_len == 8` when `class_token_len=8` is set
- [x] `test_nemotron_omni_provider_bridge_defaults_class_token_len_when_absent` asserts the fallback of 10 when `class_token_len` is absent
- [x] Compat view builds on HSG; generated `run_config.yaml` round-trips through `load_model_config()` (`num_layers=52`, `hidden_size=2688`, `vision_class_token_len=8`)
- [x] GPU export completes on 4×GB200 — 3677 tensors converted, no errors, 17 safetensors shards; output structurally identical to the previously validated export
- [ ] End-to-end against the **public** HF reference — blocked on gap 1
- [ ] Exported checkpoint loads without a manual `modeling.py` patch — blocked on gap 2

### DIFF
--- a/scripts/conversion/prepare_nemotron_omni_legacy_ckpt.py
+++ b/scripts/conversion/prepare_nemotron_omni_legacy_ckpt.py
@@ -139,15 +139,11 @@ def _load_common_pt(src_path: Path) -> dict:
 def _generate_run_config(common: dict, hf_ref: str, output_iter_dir: Path, *, trust_remote_code: bool) -> None:
     """Generate run_config.yaml from the HF reference + common.pt overrides."""
     from megatron.bridge.models.conversion.auto_bridge import AutoBridge
-    from megatron.bridge.models.hf_pretrained.causal_lm import PreTrainedCausalLM
+    from megatron.bridge.training.config import ConfigContainer
+    from megatron.bridge.utils.yaml_utils import dump_dataclass_to_yaml
 
     log.info("Loading HF reference config from %s ...", hf_ref)
-    hf_pretrained = PreTrainedCausalLM.from_pretrained(
-        hf_ref,
-        load_weights=False,
-        trust_remote_code=trust_remote_code,
-    )
-    bridge = AutoBridge.from_hf_pretrained(hf_pretrained)
+    bridge = AutoBridge.from_hf_pretrained(hf_ref, trust_remote_code=trust_remote_code)
     provider = bridge.to_megatron_provider(load_weights=False)
 
     # Override settings that differ between the HF reference and this checkpoint.
@@ -181,7 +177,21 @@ def _generate_run_config(common: dict, hf_ref: str, output_iter_dir: Path, *, tr
 
     run_config_path = output_iter_dir / "run_config.yaml"
     provider.finalize()
-    provider.save_run_config(run_config_path)
+
+    # Bridge has no provider-level "save run config" API: ``run_config.yaml`` is
+    # normally written by ``ConfigContainer.to_yaml()`` during training, and read
+    # back by ``load_model_config()``, which only consumes the top-level ``model``
+    # key and rebuilds the provider via ``instantiate()``. Emitting just that key
+    # is therefore sufficient, and avoids fabricating optimizer/dataset/train
+    # sections we do not have.
+    #
+    # The conversion MUST go through ``ConfigContainer._convert_value_to_dict``.
+    # Handing the provider straight to the YAML representers looks like it works
+    # but silently emits only ``_target_``/``_call_`` with no field values, so
+    # ``instantiate()`` returns a default-constructed provider (num_layers=None,
+    # hidden_size=0) and every checkpoint-specific override is lost.
+    model_dict = ConfigContainer._convert_value_to_dict(provider)
+    dump_dataclass_to_yaml({"model": model_dict}, str(run_config_path))
     log.info("Wrote run_config.yaml → %s", run_config_path)
 
 
--- a/scripts/conversion/prepare_nemotron_omni_legacy_ckpt.py
+++ b/scripts/conversion/prepare_nemotron_omni_legacy_ckpt.py
@@ -1,0 +1,327 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Prepare a legacy Nemotron-Omni MCore checkpoint for current Megatron-Bridge.
+
+Legacy checkpoints saved before Bridge v0.6 have three incompatibilities with
+the current codebase:
+
+1. ``common.pt`` stores ``ModelType.encoder_and_decoder`` (integer value 2),
+   which was removed from ``megatron.core.enums.ModelType``.  Loading the file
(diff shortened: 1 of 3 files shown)


scripts/conversion/prepare_nemotron_omni_legacy_ckpt.py
```diff
@@ -0,0 +1,327 @@
...
+        # the import path; here we propagate it into run_config.yaml so the
+        # export path also sees the correct value.
+        radio_cls = getattr(args, "vision_class_token_len", None)
+        if radio_cls is not None:
+            provider.vision_class_token_len = int(radio_cls)
+            log.info("  vision_class_token_len: %d", provider.vision_class_token_len)
+
+    run_config_path = output_iter_dir / "run_config.yaml"
+    provider.finalize()
+    provider.save_run_config(run_config_path)
```
guyueh1: there is no save_run_config API in NemotronOmniModelProvider , is this a typo?

Issue #5148, opened 2026-07-28
[Tracking] Support pip-installed Bridge conversion across the previous MCore minor release (N-1) to the current pin


yaoyu-33:
## Summary

Pip-installed Megatron Bridge conversion should support a range of Megatron-Core versions instead of only the exact editable submodule pin.

The supported window is defined **relative to the current pin, not as a fixed version**: it covers the **previous MCore minor release (N-1)** through the latest MCore commit pinned by Bridge main. This window floats — every time Bridge moves to a newer MCore minor release, the lower bound moves with it.

As of this writing, Bridge main pins the MCore submodule at `2d1fa8d37` (`25.04-alpha.rc1-2756-g2d1fa8d37`, self-reporting `0.18.0`). The concrete window today is therefore the released **0.18.x** line → that dev pin. Treat `0.18` throughout this issue as the current instance of the rule, not as a hardcoded target.

Installing Bridge must not replace or break an already installed supported version inside that window.

## Why daily MCore bumps break Bridge

- Public or imported APIs are removed or relocated.
- Checkpoint layouts and config fields change.
- MCore and Transformer Engine dependency pins drift.
- CI tests the editable submodule pin, which does not represent the pip installation path.

Recent examples: #4503, #4770, #4794, and #4932.

## Acceptance criteria

- [ ] Declare and document the supported MCore range as a **rule** (N-1 minor release → current pin), and state the concrete versions it resolves to for the current release.
- [ ] Verify the lowest supported N-1 release, the latest N-1 patch release, and the current Bridge MCore pin.
- [ ] Confirm installing Bridge preserves a compatible preinstalled MCore version.
- [ ] Run lightweight dense, MoE, and hybrid config/conversion smoke tests across the range.
- [ ] Fail installation clearly for MCore versions outside the supported range.
- [ ] Make the bound derivable from the pin so it does not need a manual edit on every MCore bump.

yaoyu-33:
## Scope: the contract is AutoBridge, not the whole repo

We do not need all of Megatron Bridge to work across an MCore range — the training loop is explicitly out of scope. What we need to hold is the **AutoBridge conversion path**. Scoping it that way makes this issue tractable, because the actual API surface is much smaller than a whole-repo count suggests.

Measured by static import closure on `megatron.bridge.models.conversion.auto_bridge` (submodule pin `2d1fa8d37`):

[...]

PR #5193, opened 2026-07-30
[ckpt, build] fix: support pip-installed conversion on Megatron-Core 0.18.x

yaoyu-33:
## What

Makes a pip-installed Megatron Bridge work against the released Megatron-Core **0.18.x** line, not just the exact source pin. This is a concrete step toward #5148.

Current main pins Megatron-Core commit `24bad8e677` (`0.20.0`), while the [public PyPI index](https://pypi.org/simple/megatron-core/) currently ends at `0.18.2`: there is no public `0.19.x` or `0.20.x` package to install. The supported pip range is therefore declared explicitly as `>=0.18.0,<0.21` and tested against both the current source pin and the actual public 0.18.x line.

On 0.18.x, `megatron.training.models.gpt` does not exist. Without this change, that missing module breaks `megatron.bridge.training` and every checkpoint import/export path that depends on it.

## How

- `megatron.bridge.compat.mcore_version` declares the public compatibility floor (`0.18.0`), current source pin (`0.20.0`), and exclusive next-minor ceiling (`0.21`). Tests keep the source gitlink, code declaration, and package metadata aligned.
- `megatron.bridge.compat.mcore_gpt` is the stable facade for GPT builder/config symbols and serialized targets. It prefers the upstream implementation when present and uses the vendored fallback only when the whole upstream module is absent; unrelated transitive import errors still propagate.
- The fallback is refreshed from the current 0.20 GPT implementation. Its model-builder call projects the current arguments onto the callable exposed by older MCore releases, including omitting the new layerwise optimizer/layout arguments on 0.18.x.
- All current-main consumers, including the newer GPT model-config path, import through the facade. Legacy upstream serialized targets are normalized when loading existing run configs.
- Package metadata and user documentation now state `megatron-core[dev,mlm]>=0.18.0,<0.21`.

No file under `3rdparty/Megatron-LM` is modified, no dependency is added, and no public conversion API signature changes.

## Validation

Standard project container with `uv`:

| Environment | Result |
|---|---|
| Current source pin (`0.20.0`) | 822 Bridge modules imported; focused compatibility/version/model load-save/conversion matrix passed (100 tests), with one separately identified workstation-driver-only NCCL mock failure |
| Exact public PyPI `0.18.0` | 822 Bridge modules imported with zero failures; 94 focused tests passed, 6 expected upstream-parity tests skipped, workstation-only NCCL case deselected |
| Exact public PyPI `0.18.2` | 822 Bridge modules imported with zero failures; 94 focused tests passed, 6 expected upstream-parity tests skipped, workstation-only NCCL case deselected |
| Built wheel | METADATA contains `megatron-core[dev,mlm]<0.21,>=0.18.0` |

The focused matrix covers support boundaries, facade/fallback selection, upstream API parity, old-builder argument projection, stable and legacy config targets, checkpoint load/save, CPU conversion launcher paths, and a tiny exact HF→MCore→HF tied/untied-logit roundtrip.

Authoritative [Qwen3-30B-A3B verification card](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/examples/model_verification_cards/qwen3-30b-a3b/card.yaml), on one 8-GPU node with exact public MCore `0.18.0` and model revision `ad44e777bcd18fa416d9da3bd8f70d33ebb85d39`:

- TP4/PP2/EP4 HF→Megatron import completed all 3,411 mappings and wrote the distributed checkpoint.
- The pinned HF/MCore forward comparison predicted token 12095 (`" Paris"`) on both paths; cosine similarity was `0.999317` (card gate `>0.99`), with the card-recorded max/mean logit diagnostics `0.593750` / `0.130064`.
- Two independent imported-checkpoint generation runs completed exactly 32 tokens with the card's byte-identical expected text.
- TP4/PP2/EP4 distributed Megatron→HF export completed all 3,411 mappings.
- Strict roundtrip verification compared all **18,867** BF16 tensors: exact key, shape, dtype, and bitwise value equality. The export reloaded as `Qwen3MoeForCausalLM`, and `norm_topk_prob` remained `true`.

`uv run pre-commit run --all-files` passed before every new DCO-signed merge/conflict-resolution commit.

## Main integration

Latest main was merged normally into this branch without rebasing or force-pushing. Conflicts were resolved by retaining current main's optional ModelOpt behavior and tests while routing its new GPT model-config consumer through the compatibility facade.

Refs #5148


docs/fern/versions/nightly/pages/index.mdx
```diff
@@ -94,6 +94,26 @@ The submodule committed to the repo always points to the **main** commit. Use th
 
 The dev branch follows Megatron-LM's upstream [dev branch philosophy](https://github.com/NVIDIA/Megatron-LM/tree/dev) — features are experimental, follow a streamlined review process, and must graduate to stable within 6 months or be deprecated.
 
+### Supported Megatron-Core versions
+
+A pip-installed Megatron Bridge is not restricted to the exact submodule pin. Bridge keeps
+compatibility with the publicly installable Megatron-Core 0.18.x line while accepting releases
+through the current pin:
+
+> `megatron-core>=0.18.0` and below the next minor after the current pin
```
cuichenx: should be current minus 1 major release, not 0.18 to current

cuichenx (commented):
similar to docs/conf.py, code coverage should exclude the compat folder too

yaoyu-33:
@cuichenx Addressed in `8a3e31902`: `[tool.coverage.run].omit` now excludes `*/src/megatron/bridge/compat/*` across source-relative and normalized absolute checkout paths. `tests/unit_tests/compat` still executes normally (26/26 passed), and an actual Coverage.py run measured 321 Bridge source files with zero compat source files reported. All-files pre-commit passed.

docs/fern/versions/nightly/pages/index.mdx
yaoyu-33: Updated in `751fc75c7`. The shortened section now states the rolling N/N-1/N-2 minor policy explicitly, notes that the lower bound advances with the pin, and gives the current N=0.20 window as `megatron-core>=0.18.0,<0.21`. The install-time bound and compatibility-shim explanation remain concise; version logic and dependency bounds are unchanged.

PR #5220, opened 2026-07-31
fix(model): map GLM-5 routed expert count

cuichenx:
# What does this PR do ?

Fix the GLM-5 expert-count conversion when affected Hugging Face configurations expose both the authoritative `n_routed_experts` value and an unrelated default `num_experts=256`.

# Changelog

- Skip the generic `num_experts` alias only when the same GLM configuration also exposes `n_routed_experts`, allowing the routed-expert value to populate Megatron's `num_moe_experts`.
- Keep the shared generic mappings unchanged for fixed and legacy configuration shapes and for reverse export.
- Document that this is a compatibility workaround for an upstream Transformers 5.11-5.12 GLM configuration bug; Transformers 5.13 removes the duplicate field and redirects legacy input.
- Add regression coverage using the real affected configuration behavior without manually supplying the conflicting default, plus tests for affected and unaffected config shapes.

# GitHub Actions CI

- `uv run python -m pytest tests/unit_tests/models/glm_moe_dsa/test_glm5_bridge.py` — 14 passed.
- Reproduced on the selected 26.08 release candidate before the fix: routed experts 8 produced provider experts 256.
- Verified with the identical causal reproducer after the fix: routed experts 8 produces provider experts 8.
- Verified with Transformers 5.13.0 that `num_experts` is absent and the unchanged generic mapping produces provider kwargs with 8 experts.

**Pre checks**:

- [x] Added focused regression tests.
- [x] Evaluated documentation impact; no documentation change is needed for this compatibility correction.
- [x] Evaluated optional-component impact; this change adds no dependency or optional import.

# Additional Information

- NVBug 6540522

### DIFF
--- a/src/megatron/bridge/models/glm_moe_dsa/glm5_bridge.py
+++ b/src/megatron/bridge/models/glm_moe_dsa/glm5_bridge.py
@@ -53,6 +53,10 @@ class GLM5Bridge(MegatronModelBridge):
         >>> provider = bridge.to_megatron_provider()
     """
 
+    CONFIG_MAPPING = [mapping for mapping in MegatronModelBridge.CONFIG_MAPPING if mapping[1] != "num_moe_experts"] + [
+        ("n_routed_experts", "num_moe_experts")
+    ]
+
     def provider_bridge(self, hf_pretrained: PreTrainedCausalLM) -> MLAModelProvider:
         provider = super().provider_bridge(hf_pretrained)
         hf_config = hf_pretrained.config
--- a/src/megatron/bridge/models/glm_moe_dsa/glm5_bridge.py
+++ b/src/megatron/bridge/models/glm_moe_dsa/glm5_bridge.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+from typing import Any
 
 from megatron.core.models.gpt.gpt_model import GPTModel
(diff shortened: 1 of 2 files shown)


src/megatron/bridge/models/glm_moe_dsa/glm5_bridge.py
```diff
@@ -53,6 +53,10 @@ class GLM5Bridge(MegatronModelBridge):
         >>> provider = bridge.to_megatron_provider()
     """
 
+    CONFIG_MAPPING = [mapping for mapping in MegatronModelBridge.CONFIG_MAPPING if mapping[1] != "num_moe_experts"] + [
```
yaoyu-33: not good idea to change like this - GLM-5.x checkpoint with a different routed-expert count exposes it immediately. Harness ruled out: the failure reproduces from a plain AutoBridge.from_hf_pretrained(...).to_megatron_provider(load_weights=False) on a locally-saved config, with no test harness, no env vars and no suite involvement. ## Description MegatronModelBridge.CONFIG_MAPPING contains two entries that both target the Megatron field num_moe_experts:
model_bridge.py:460   ("num_experts",      "num_moe_experts")
model_bridge.py:467   ("n_routed_experts", "num_moe_experts")

if num_experts not exist, it fallsback to n_routed_experts for consulting?

PR #5222, opened 2026-07-31
fix(training): restore weak scaling for benchmark recipes

### LINKED ISSUE #4759: [bug] Support weak scaling for perf recipes
### Problem

Weak scaling runs do not appear to be supported after the performance recipes were migrated from scripts to the src directory. The default Nemotron-H recipes are configured for 64 GPUs, and when attempting to run with 8 GPUs (for weak scaling), the run fails with the error mentioned below.
```
raise ValueError(f"No perf recipe {name!r} found in perf recipe packages: {searched_modules}.")
ValueError: No perf recipe 'nemotronh_56b_pretrain_8gpu_b300_fp8cs_config' found in perf recipe packages: megatron.bridge.perf_recipes.deepseek, megatron.bridge.perf_recipes.gpt_oss, megatron.bridge.perf_recipes.kimi, megatron.bridge.perf_recipes.llama, megatron.bridge.perf_recipes.nemotronh, megatron.bridge.perf_recipes.qwen, megatron.bridge.perf_recipes.qwen_vl,
```
This functionality was working as expected prior to the migration.

### Minimal repro

```shell
python3 scripts/performance/setup_experiment.py  \
     --container_image <container> \
     --compute_dtype fp8_cs \
     --gpu b300 \
     --num_gpus 8 \
     --gpus_per_node 8 \
     --model_family_name nemotronh \
     --model_recipe_name nemotronh_56b \
     --account $SLURM_ACCOUNT \
     --partition $SLURM_PARTITION \
     --log_dir <log_dir> \
     --time_limit 00:15:00
```

### Expected behavior

A successful run using 8 GPUs with the corresponding GBS.

### Affected area

area:recipe

### Regression?

Yes

### Environment

Branch: main

### Logs

```shell

```

yaoyu-33:
## Summary

- preserve exact GPU-count flat recipe selection when a tuned recipe exists
- fall back to the canonical flat recipe when the requested GPU count is not registered
- support the same weak scaling through the unified scripts/training launcher
- preserve samples per data-parallel rank by scaling GBS from canonical DP to effective requested DP
- validate both dense TP/PP/CP/GTP and MoE ETP/EP/PP/expert-GTP grids before MCore initialization
- preserve explicit global batch-size overrides and reject fractional automatic scaling
- use the same exact weak-scaling helper for performance experiment names, without integer truncation
- document canonical allocation and dry-run behavior for benchmark recipes

## Red to green

Before this change, requesting the Nemotron-H 8-GPU B300 FP8-CS workload failed because only the canonical 64-GPU recipe exists. The initial fallback also scaled by total world size, which was incorrect when TP/PP/CP overrides changed the effective DP ratio, and incompatible MoE expert grids failed only during MCore initialization.

After this change, both launch paths reuse the canonical recipe while retaining its identity. GBS follows the canonical-to-requested DP ratio, explicit GBS remains final, fractional results fail with an actionable error, and incompatible expert grids such as GPT-OSS 120B with TP=1, PP=4, EP=8 on WORLD_SIZE=8 are rejected early.

The reported Nemotron-H case still resolves the 64-GPU recipe and scales GBS from 192 to 24 at 8 GPUs with its unchanged topology.

## Tests

- red-to-green coverage for DP-ratio scaling, GPT-OSS expert-grid validation, and fractional experiment-name scaling
- targeted launcher, performance utility, and override-precedence tests: 182 passed
- focused review regressions after formatting: 6 passed
- pre-commit: passed

Fixes #4759

### DIFF
--- a/scripts/common/benchmark_parallelism.py
+++ b/scripts/common/benchmark_parallelism.py
@@ -1,0 +1,143 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Lightweight parallel-topology helpers shared by benchmark launchers."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass(frozen=True)
+class ParallelTopology:
+    """Model and expert parallel degrees that constrain a benchmark world."""
+
+    tensor_parallel: int = 1
+    pipeline_parallel: int = 1
+    context_parallel: int = 1
+    expert_parallel: int = 1
+    expert_tensor_parallel: int | None = None
+    gtp_remat: int = 1
+    expert_gtp_remat: int = 1
+
+
+def _first_attribute(config: Any, names: tuple[str, ...], default: Any) -> Any:
+    """Return the first exposed configuration attribute from ``names``."""
+    for name in names:
+        if hasattr(config, name):
+            return getattr(config, name)
+    return default
+
+
+def topology_from_config(config: Any) -> ParallelTopology:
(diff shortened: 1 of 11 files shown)


malay-nagda (dismissed):

yaoyu-33:
Addressed the review follow-up in 30b05b827: experiment-name GBS now uses the same exact DP-ratio weak-scaling helper, topology overrides scale from canonical DP to effective DP, and both unified and legacy fallback paths validate the MoE expert grid before MCore initialization. Added red-to-green coverage for all three cases; 182 targeted tests and pre-commit pass.

meatybobby:
@malay-nagda could you review the latest changes?

yaoyu-33:
approved by @malay-nagda  before, pass ci, force merge.

PR #5224, opened 2026-07-31
Rsalagame llama31 pretrain and finetune fixes

rsalagame-nvidia:
**What the issue was:** **configs/llama/__init__.py** unconditionally imported the Llama-2 70B LoRA configs, and that module imports llama2_70b_peft_config from megatron.bridge.recipes.llama. That symbol exists in newer Megatron-Bridge but not in the 26.06.01 container image — possible because scripts/performance is host-mounted while megatron.bridge comes from inside the container. Because the import sat in the package __init__, the resulting ImportError took down the entire configs.llama package, so every Llama job aborted at startup, pretrain and finetune alike, on all GPU types. The launcher only catches **ModuleNotFoundError**, so a missing symbol wasn't converted into a clean error either. 

**Fix:** wrap the Llama-2 import in try/except ImportError and export those two config names only when it succeeds. Verified on GB300 with llama3.1 8B pretrain and llama3 70B LoRA finetune.

**Steps to reproduce:**

1) clone the repo: git clone ssh://git@gitlab-master.nvidia.com:12051/unified-cloud-benchmarking/llm-benchmarking-collection-internal.git; cd llm-benchmarking-collection-internal

2) check out to my branch: 

```
git fetch origin
git checkout -b 'rsalagame_all_workloads_llmb_release_26.08.00_draft' 'origin/rsalagame_all_workloads_llmb_release_26.08.00_draft'
```

3) llmb-install (select llama pretrain and finetune workloads)

4) Run with cmd ex: `llmb-run submit -w pretrain_llama3.1 -s 8b -d fp8 --scale 8` under $LLMB_INSTALL
[log-general_infra-rd_gsw-gsw.pretrain_llama3_8b_fp8_cs_gpus16_tp1_pp1_cp1_vpNone_ep1_etpNone_mbs4_gbs256_329677_0.txt](https://github.com/user-attachments/files/30681598/log-general_infra-rd_gsw-gsw.pretrain_llama3_8b_fp8_cs_gpus16_tp1_pp1_cp1_vpNone_ep1_etpNone_mbs4_gbs256_329677_0.txt)

Log file with the issue (ex on OCI-AGA GB300 cluster where I noticed this but not chip specific):

### DIFF
--- a/scripts/performance/configs/llama/__init__.py
+++ b/scripts/performance/configs/llama/__init__.py
@@ -49,8 +49,6 @@ if HAVE_MEGATRON_BRIDGE:
         llama3_70b_pretrain_config_vr200,
     )
     from .llama31_llm_pretrain import (
-        llama31_8b_pretrain_config_gb300,
-        llama31_8b_pretrain_config_gb200,
         llama31_405b_pretrain_config_b200,
         llama31_405b_pretrain_config_b300,
         llama31_405b_pretrain_config_gb200,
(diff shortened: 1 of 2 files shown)


bdubauski (approved):
Please re-test to confirm

Issue #4664, opened 2026-07-06
[feature] Clearer "golden path" for dataset processing


nathan-az:
### User problem

Currently there are multiple dataset processors (two huggingface ones, the GPTSFT one). They use different features, render conversations differently, have varying support for online/offline packing, etc.

Right now a user has to go quite deep in the code to understand the difference between processors, and the differing behaviour can have nontrivial impact on training (e.g. one of the processors does not use the tokenizer's chat template and instead renders manually, using special tokens that aren't part of the model's standard vocabulary/pretraining).

### Desired outcome

It should be clear from documentation and existing recipes what the best supported, recommended processor is.

### Alternatives considered

I expect the solution is either to unify behind fewer processors with *clearly different use cases*, or leave all as-is, but perhaps raise *warnings* to users if they are using a non-recommended one. I will leave the decision to maintainers.

### Affected area

area:model

### Urgency / use case

Important but not blocking

yaoyu-33:
Thanks for raising this. We agree with the problem statement. The current dataset surface includes several historical carry-over paths rather than a deliberately designed set of user-facing workflows.

## Current state

| Path | Current role | Direction |
|---|---|---|
| `GPTDatasetConfig` | LLM pretraining from Megatron `.bin/.idx` | **Keep.** This remains the default pretraining path and is not chat-template based. |
| `FinetuningDatasetConfig` / `GPTSFTDataset` | Local JSONL text SFT, including packed Parquet | **Legacy.** Replace with the new pretokenized `.bin/.idx` SFT path. |
| `HFTextSFTDatasetProvider` | Converts HF rows to JSONL and feeds the legacy SFT builder | **Remove after migration.** It overlaps with the direct HF conversation path and mainly exists to connect HF data to the old offline-packing implementation. |
| `HFConversationDatasetProvider` | Direct HF conversation processing for text, vision, and audio/omni data | **Keep and make the primary direct-HF path.** |
| `PreloadedVLMConversationProvider` | Loads local VLM JSON/JSONL before using the same conversation/collate concepts | **Remove after migration.** Local files should be supported as a source or maker under the unified HF conversation provider. |
| `EnergonProvider` | WebDataset/Energon multimodal training | **Keep** as the scalable multimodal streaming path. |

Model-specific providers such as MegatronMIMO and Step-3.7 data loaders should not be presented as additional general-purpose choices. They are specialized integrations.

## Target user-facing workflows

[...]

PR #5226, opened 2026-08-01
feat(data): add indexed storage for packed SFT

yaoyu-33:
# What does this PR do?

Use Megatron Core `.bin/.idx` IndexedDataset pairs as the default storage for offline-packed text SFT and PEFT data, while retaining explicit Parquet and deprecated NumPy compatibility paths.

# Changelog

- Add a versioned packed-SFT IndexedDataset schema with one pack per item, token IDs in the lower 31 bits, a target-aligned loss-mask bit, and stored sequence boundaries.
- Route offline preparation, runtime loading, path validation, multi-shard resolution, builder-managed caching, and LoRA FLOP statistics through indexed data.
- Change builder-managed outputs from `training_<length>.idx.parquet` to the logical prefix `training_<length>.sft`, producing `.sft.bin` and `.sft.idx`.
- Keep explicit `.parquet` output and loading for migration and A/B checks; keep deprecated `.npy` loading behavior.
- Add bounded stale-NFS retries, per-shard schema checks, sample-cap mapping fixes, and parity checks that detect row-count mismatches.
- Publish local pairs under a reader/writer filesystem lock, restore the previous pair if either publication step fails, and preserve recovery backups if rollback itself fails.
- Support prebuilt `msc://` pairs with automatic MCore feature enablement, shared index-cache configuration, and remote range reads. Direct object-storage writes are rejected: prepare locally and upload `.bin` before `.idx`.
- Add `compare_packed_sft_formats.py` for row-level semantic parity and sequential-read microbenchmarks.
- Add a complete packed SFT indexed-dataset tutorial and update the text-SFT, data-preparation, and packed-sequence guides.

The `.bin/.idx` container is now shared with pretraining, but the packed-SFT payload remains a distinct versioned schema because SFT also needs loss masks and sequence boundaries.

# Validation

- 184 targeted unit/integration executions passed on CW against the original implementation, including end-to-end JSONL -> offline packing -> `.bin/.idx` -> builder load -> packed collate.
- After multiple independent review/fix rounds, focused compatibility-isolated local runs passed: 26 indexed storage tests, 40 packing algorithm/statistics tests, 3 Parquet/indexed comparison tests, 1 end-to-end builder/sample-cap test, and 10 tutorial tests.
- A locked `multi-storage-client==0.51.0` smoke test passed for local pair generation followed by `msc://default` auto-enable, remote index caching/range reading, and row decode.
- Full repository pre-commit passed after the final fixes; `git diff --check`, compile checks, and secret/path scans passed.
- CW/Lustre synthetic warm-read comparison: 512 rows x 4096 tokens (2,097,152 tokens), all `input_ids`, `loss_mask`, and `seq_start_id` values matched in three runs. Median sequential decode was about 2.32M tokens/s for Parquet and 174.9M tokens/s for bin/idx. Storage was 5.16 MiB and 8.03 MiB respectively. This is a decoder microbenchmark, not an end-to-end training-throughput claim.

The full functional builder module previously stalled while downloading legacy Megatron-LM release assets before executing a test. Fresh post-review CW attempts did not enter the Python payload because Pyxis/container initialization stalled on multiple nodes; no test failure occurred. The final changes are therefore covered by focused local tests and still require normal NVIDIA CI validation.

# GitHub Actions CI

This remains a draft PR. NVIDIA unit/L0 workflows have not run because copy-pr-bot requires additional validation before NVIDIA runners can start.

**Pre checks**:

- [x] Does the PR affect components that are optional to install? No new dependency is added. Parquet remains lazily imported on Parquet-only paths; MSC must already be installed and configured for `msc://` inputs.

# Additional Information

- Follow-up to the prepared SFT `.bin/.idx` direction discussed in #4664.
- Tutorial: `docs/training/packed-sft-indexed-dataset.md`

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -78,6 +78,7 @@ training/cpu-offloading.md
 training/moe-optimization.md
 training/peft.md
 training/packed-sequences.md
+training/packed-sft-indexed-dataset.md
 training/multi-token-prediction.md
 training/callbacks.md
 ```
--- a/docs/training/README.md
+++ b/docs/training/README.md
@@ -33,6 +33,7 @@ This directory contains comprehensive documentation for training and customizing
 | **[Configuration Container Overview](config-container-overview.md)** | Central configuration object for all training settings | First time setting up training |
 | **[Entry Points](entry-points.md)** | Training entry points and execution flow | Understanding how training starts |
 | **[Data Preparation](data-preparation.md)** | Dataset formats for pretraining, SFT, PEFT, and VLM fine-tuning | Preparing data or choosing dataset config fields |
+| **[Packed SFT Indexed Dataset](packed-sft-indexed-dataset.md)** | Prepare and consume `.sft.bin/.sft.idx` data | Moving packed SFT from Parquet to MCore indexed storage |
 | **[Training Loop Settings](training-loop-settings.md)** | Training loop parameters and configuration | Configuring batch sizes, iterations, validation |
 
 ### Optimization and Performance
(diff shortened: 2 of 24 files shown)


yaoyu-33:
Closing in favor of #5289, which adds weighted multi-source blending for the existing packed Parquet path.

The real NVIDIA Nemotron 3 Nano 30B-A3B SFT A/B on 8 H100s (sequence length 2048, GBS 128, four interleaved cold optimizer steps) did not show a bin/idx benefit: indexed averaged 104.679 s/step versus 103.810 s/step for Parquet. Parquet was nominally 0.83% faster, while same-format run-to-run variation was 1.1-2.6%, so the formats are effectively tied at this model scale. Batch generation took only about 3-4 ms of a roughly 104 s step.

The benchmark was limited to first complete optimizer steps because subsequent DeepEP steps hit the existing timeout; all four measured steps completed successfully with no skipped or NaN iterations. Given the lack of a material training-speed benefit and Parquet being about 51% smaller in the test dataset, the additional indexed storage format is not worth carrying.

PR #5231, opened 2026-08-01
fix(peft): sync grouped expert LoRA over expert DP

### LINKED ISSUE #5229: [bug] Fix grouped-expert LoRA gradient synchronization when EP=1 and ETP differs from TP
## Summary

Megatron-LM https://github.com/NVIDIA/Megatron-LM/pull/5916 fixes standard MCore expert parameters when `EP=1, ETP!=TP`.
Megatron-Bridge https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4998 fixes Bridge's independent `calc_params_l2_norm()`.

Bridge's `GroupedExpertLinearAdapter` still uses the old condition:

use_expert_process_groups = EP > 1

It must use:

use_expert_process_groups = EP > 1 or ETP != TP

## Impact

For grouped-expert LoRA with:

- `share_expert_adapters=False`
- `EP=1`
- `ETP!=TP`

Bridge sets `weight.allreduce=True`. MCore DDP consequently puts these adapter
weights in ordinary DP/CP buffers instead of expert-DP buffers, synchronizing
their gradients over the wrong process group.

MCore #5916 cannot fix this because Bridge creates and annotates these parameters.

## Tests

Cover:

- EP=1, TP=ETP -> ordinary DP
- EP>1 -> expert DP
- EP=1, TP>ETP -> expert DP
- EP=1, ETP>TP -> expert DP
- Verify MCore `group_params_for_buffers()` creates an expert BufferKey

References:
- https://github.com/NVIDIA/Megatron-LM/pull/5916
- https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4998

### Minimal repro

```shell
use the config above `EP=1, ETP!=TP`
```

### Expected behavior

use EDP for experts instead of DP 

### Affected area

area:peft

### Regression?

Yes

### Environment

_No response_

### Logs

```shell

```

yaoyu-33:
## What does this PR do?

Marks grouped-expert LoRA parameters for expert-DP synchronization whenever expert parallelism is enabled or expert TP differs from regular TP. This aligns Bridge-created adapter parameters with the MCore expert-parameter process-group contract and ensures `group_params_for_buffers()` produces expert `BufferKey` entries.

## Test design and evidence

Tests cover the complete topology matrix:

- EP=1 and TP=ETP: ordinary DP
- EP>1: expert DP
- EP=1 and TP>ETP: expert DP
- EP=1 and ETP>TP: expert DP
- Pinned MCore buffer grouping for an affected adapter parameter

Before the fix, the two TP/ETP mismatch cases and the expert `BufferKey` assertion failed: the adapter had `allreduce=True` and was grouped into ordinary DP. After the fix, all five focused cases pass.

A two-GPU distributed repro using real NCCL process groups confirmed the same red-to-green transition for EP=1, TP=2, ETP=1: before the fix `(allreduce, is_expert_parallel) == (True, False)`; after the fix the workload completed with the expected `(False, True)`.

## Validation

- `uv run python -m pytest tests/unit_tests/peft/test_utils.py tests/unit_tests/peft/test_lora.py tests/unit_tests/peft/test_canonical_lora.py tests/unit_tests/peft/test_multi_lora_layers.py -k "not MegatronIntegration" -q` — 212 passed, 8 skipped, 5 deselected
- `uv run python -m torch.distributed.run --standalone --nproc_per_node=2 ...` — passed
- `git diff --check` — passed

Fixes #5229

### DIFF
--- a/src/megatron/bridge/peft/utils.py
+++ b/src/megatron/bridge/peft/utils.py
@@ -1792,8 +1792,9 @@ class GroupedExpertLinearAdapter(nn.Module):
         self.pg_collection = _get_pg_collection(
             pg_collection,
             model_parallel_config,
-            required_pgs=["ep", "expt_tp", "expt_dp"],
+            required_pgs=["tp", "ep", "expt_tp", "expt_dp"],
         )
+        tensor_parallel_group = _get_tensor_parallel_group(self.pg_collection)
         self.expert_tp_group = _get_tensor_parallel_group(self.pg_collection, is_expert=True)
         self.ep_group = _get_process_group(self.pg_collection, "ep")
         self.expert_dp_group = _get_process_group(self.pg_collection, "expt_dp")
@@ -1803,6 +1804,10 @@ class GroupedExpertLinearAdapter(nn.Module):
         expert_tp_size = _process_group_size(
             self.expert_tp_group,
             model_parallel_config.expert_tensor_parallel_size or 1,
+        )
+        tensor_parallel_size = _process_group_size(
+            tensor_parallel_group,
+            getattr(model_parallel_config, "tensor_model_parallel_size", 1) or 1,
         )
         linear_in_tp_axis = 2 if input_is_parallel else 1
         linear_out_tp_axis = 1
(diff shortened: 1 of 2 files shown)


bvolpato (approved):
LGTM! Nice, focused fix. Condition now matches MCore expert-topology contract, and topology matrix plus real NCCL repro make this easy to trust.

I checked this against GLM 5.2 GB200 topology with TP=8, ETP=1, and EP=64. Both old and new paths select expert process groups through EP > 1, so this remains behavior-preserving there while fixing EP=1 with ETP != TP.

Nonblocking: future GLM 5.2 grouped-expert LoRA smoke coverage would be useful, especially on GB200. Not needed for this PR given focused tests and green CI. Thanks for catching and fixing this.

zhongbozhu (approved):
Thanks for the fix, LGTM

Issue #5230, opened 2026-08-01
[feature] Support DSpark speculative decoding draft training


khazic:
### User problem

Megatron-Bridge has no support for **DSpark** speculative-decoding draft training.
DSpark is a semi-autoregressive speculative-decoding draft (Markov / confidence
heads) that trains a lightweight draft against a frozen target. It is already
implemented in the NeMo AutoModel speculative-decoding stack
(`nemo_automodel/components/speculative/dspark/`), but there is no path today to
train DSpark drafts on Megatron-based targets or to round-trip them through the
bidirectional Hugging Face conversion this library provides, the way the
EAGLE-3 / MTP speculative paths are already tracked here.

### Desired outcome

DSpark speculative decoding supported in Megatron-Bridge:

- DSpark draft model + trainer recipe (draft class, Markov / confidence heads,
  training-time acceptance metrics), following the AutoModel implementation;
- frozen colocated target supervision with the draft-vocab (d2t / t2d) mapping;
- HF <-> Megatron conversion for the DSpark draft checkpoint so trained drafts
  export for serving.

### Alternatives considered

Training DSpark drafts only in NeMo AutoModel and converting externally. That
misses the Megatron-based target path and the native bidirectional HF conversion
this library is built around.

### Extra context

DSpark reference implementation in NeMo AutoModel:
https://github.com/NVIDIA-NeMo/Automodel/tree/main/nemo_automodel/components/speculative/dspark

I would like to take this. I do not have assign permissions on this repo, so could a maintainer assign me?

aroshanghias-nvd:
Thanks for volunteering to work on this. Before assigning the full scope, could you propose a staged first contribution? In particular, please describe:

1. the minimal first PR you would submit;
2. the Megatron-side DSpark model and checkpoint representation;
3. the Hugging Face artifact/schema that must round-trip for serving; and
4. which parts would remain in NeMo AutoModel versus Megatron-Bridge.

That will let the model and conversion maintainers confirm the repository boundary and acceptance criteria before implementation begins.

khazic:
Moving the staging plan and the open decisions out of the design page (#5233) and into this issue, since this is where they can be kept current. The design page now covers only the algorithm, where DSpark fits in this repo, and the proposed file surface.

### Staged plan

1. **Design (in review, #5233).** The algorithm, the mapping onto Megatron-Bridge, and the proposed file surface.
[...]

PR #5233, opened 2026-08-01
docs(training): add DSpark speculative decoding design

khazic:
## What

Adds a design proposal (RFC) page, `docs/training/dspark-speculative-decoding.md`,
for DSpark speculative-decoding **draft training** in Megatron-Bridge. Tracks #5230.

## Why a design PR first

Megatron-Bridge has no draft-training subsystem today (MTP is the only
draft-adjacent path, and it is a Megatron-Core native module). DSpark's draft (a
semi-autoregressive parallel backbone plus Markov and confidence heads) is a new
neural module that must live either in Megatron-Core (like MTP) or as a
ModelOpt-injected module, and its parallel/loss integration can only be validated
on multi-GPU. So the draft module and its trainer belong in later, separately
reviewed PRs; this PR establishes the algorithm spec, the Bridge integration
shape, and the plan.

## Contents

- The DSpark draft: a semi-autoregressive block (parallel backbone attending to
  the frozen target's hidden states as context K/V) plus a Markov/RNN sequential
  head and a confidence head; the draft shares and freezes the target embedding
  and LM head.
- Training objective: position-weighted CE (0.1) + total-variation (0.9) +
  confidence BCE (1.0); acceptance = `1 - TV`, `tau = sum cumprod(accept) + 1`.
- Bridge mapping: follows the ModelOpt distillation subsystem pattern (provider
  hook, `forward_step_dspark`, a `post_training/dspark.py` loss, a thin
  `training/dspark.py` entry, and `dspark.*` HF weight mappings).

Reference implementation: NeMo AutoModel
`nemo_automodel/components/speculative/dspark/`. Paper:
[arXiv:2607.05147](https://arxiv.org/abs/2607.05147).

Docs-only; the page is registered in the "Training and Customization" toctree.

### DIFF
--- a/docs/training/dspark-speculative-decoding.md
+++ b/docs/training/dspark-speculative-decoding.md
@@ -1,0 +1,212 @@
+# DSpark Speculative Decoding (Draft Training) Design
+
+## Overview
+
+This page is a design proposal (RFC) for adding **DSpark** speculative-decoding
+draft training to Megatron-Bridge. It tracks
+[issue #5230](https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/5230).
+
+DSpark ([arXiv:2607.05147](https://arxiv.org/abs/2607.05147)) is a
+speculative-decoding method that trains a lightweight **draft** against a frozen
+target. The draft proposes a block of tokens; the target verifies them. DSpark's
+draft is *semi-autoregressive*: a parallel backbone predicts a whole block in one
+pass, and a small sequential head restores the intra-block token dependency that a
+purely parallel drafter loses (the "suffix decay" problem). A confidence head
+predicts per-position acceptance so the serving engine can schedule its verify
+budget.
+
+A working draft-training reference already exists in NeMo AutoModel at
+`nemo_automodel/components/speculative/dspark/`. This document specifies how the
+same training recipe maps onto Megatron-Bridge and Megatron-Core, and proposes a
+phased plan. It intentionally does **not** ship the draft module itself: that is a
+Megatron-Core (or ModelOpt) neural module whose parallel/loss integration must be
+validated on multi-GPU runs, so it belongs in later, separately reviewed PRs.
+
+## Background: the DSpark draft
+
+### Semi-autoregressive draft
+
+For a block of `block_size` positions (`gamma`, default 7 offline / 5 in
+production), the draft factorizes the block distribution as
+`P(x | x0) = prod_k p_k(x_k | x0, x_1..x_{k-1})`:
+
+- **Parallel backbone.** A small transformer predicts every position of the block
+  in one forward pass. Crucially, the backbone does not re-encode the prompt: it
+  attends to the frozen target's hidden states as the attention **context K/V**,
+  with the block's own "noise" slots as the queries. Each block query attends to
+  the context strictly before its anchor and bidirectionally within its own block.
+  This is the block-parallel (DFlash-style) half of the draft.
+- **Sequential head.** A lightweight module adds intra-block dependency on top of
+  the parallel logits. DSpark provides a **Markov head** (a low-rank,
+  token-conditioned additive logit bias, rank 256) and an **RNN head** (a
+  GRU-like recurrence carrying prefix state across the block). At training time
+  the head is teacher-forced with the previous block tokens; only at inference is
+  it run serially.
+- **Confidence head.** A small projection `c_k = sigmoid(w^T [h_k ; embed(x_{k-1})])`
+  predicts each position's acceptance probability.
+
+The draft **shares and freezes the target's token embedding and LM head**; there
+is no separate (compressed) draft vocabulary or `d2t`/`t2d` remapping. The
+target's selected decoder-layer hidden states are fused by a learned linear
+projection (`fc`) into the draft hidden size and used as the attention context.
+
+### Training objective
+
+The draft is trained with three position-weighted terms (weights `exp(-k/gamma)`,
+emphasizing earlier positions):
+
+- `L_ce` (weight 0.1): cross-entropy of the draft logits against the target's
+  teacher-forced next tokens.
+- `L_tv` (weight 0.9): total-variation distance `0.5 * ||p_draft - p_target||_1`
+  to the target's next-token distribution. This is the dominant term and a direct
+  acceptance proxy.
+- `L_conf` (weight 1.0): binary cross-entropy training the confidence head against
+  the analytical acceptance label `c_k* = 1 - TV(p_draft, p_target)`.
+
+The target's teacher distribution is produced by passing the target's last hidden
+state through the (shared, frozen) LM head and is always detached, so no gradient
+flows into the target.
+
+### Acceptance and confidence
+
+Training measures acceptance analytically rather than by running rejection
+sampling. Per position, `accept_rate = clamp(1 - TV, 0, 1)`. The expected accepted
+prefix length of a block is `tau = sum_k cumprod(accept_rate)_k + 1` (a token
+survives only if every earlier token in its block is accepted; the `+1` counts the
+verified anchor token). `tau` is the training-time speedup proxy. The confidence
+head regresses to `accept_rate`; the paper applies a post-hoc Sequential
+Temperature Scaling calibration for serving.
+
+### Reference implementation (NeMo AutoModel)
+
+The AutoModel implementation separates cleanly into an architecture-agnostic core
+and per-target glue:
+
+- Architecture-agnostic core: the Markov/RNN heads, the three-term loss and
+  acceptance metrics, anchor sampling and block "noise" embedding, the eval mask,
+  the confidence-head module, and the block-attention mask builder.
+- Per-target glue: the draft backbone (attention that reads target hidden states
+  as context K/V, plus the decoder MLP), a draft config builder, a one-line
+  registry entry, and the frozen-target hidden-state capture.
+
+This split is what makes a Megatron-Bridge port tractable: the core is nearly
+verbatim, and only the backbone must be re-expressed in Megatron-Core layer specs.
+
+## Where DSpark fits in Megatron-Bridge
+
+### Current state
+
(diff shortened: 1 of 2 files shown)


docs/training/dspark-speculative-decoding.md
```diff
@@ -0,0 +1,212 @@
...
+
+Training measures acceptance analytically rather than by running rejection
+sampling. Per position, `accept_rate = clamp(1 - TV, 0, 1)`. The expected accepted
+prefix length of a block is `tau = sum_k cumprod(accept_rate)_k + 1` (a token
+survives only if every earlier token in its block is accepted; the `+1` counts the
+verified anchor token). `tau` is the training-time speedup proxy. The confidence
+head regresses to `accept_rate`; the paper applies a post-hoc Sequential
+Temperature Scaling calibration for serving.
+
+### Reference implementation (NeMo AutoModel)
```
yaoyu-33: not needed...

yaoyu-33 (dismissed):

docs/training/dspark-speculative-decoding.md
khazic: Removed that section. The pointer to the AutoModel reference in the overview stays.

docs/training/dspark-speculative-decoding.md
```diff
@@ -0,0 +1,167 @@
...
+projection (`fc`) into the draft hidden size and used as the attention context.
+
+### Training objective
+
+The draft is trained with three position-weighted terms (weights `exp(-k/gamma)`,
+emphasizing earlier positions):
+
+- `L_ce` (weight 0.1): cross-entropy of the draft logits against the target's
+  teacher-forced next tokens.
+- `L_tv` (weight 0.9): total-variation distance `0.5 * ||p_draft - p_target||_1`
```
yaoyu-33: **[High] This halves the dominant loss relative to the cited DSpark objective.** [Paper Eq. 10](https://arxiv.org/html/2607.05147#S3.E10) minimizes the weighted raw L1 distance `||p_draft - p_target||_1` with coefficient 0.9; the factor `1/2` appears in [Eq. 8](https://arxiv.org/html/2607.05147#S3.E8) only when forming the analytical acceptance/confidence target. The primary [DeepSpec implementation](https://github.com/deepseek-ai/DeepSpec/blob/005e03b81cec38b7da6399833d609ee89a2587f2/deepspec/modeling/dspark/loss.py#L73-L87) and current [NeMo AutoModel implementation](https://github.com/NVIDIA-NeMo/Automodel/blob/15da2681c6d827fa55c680e96dcaa8414db4cf88/nemo_automodel/components/speculative/dspark/loss.py#L59-L112) make the same distinction. With the documented coefficient, an implementation following this line gets half the intended training gradient (the proposed follow-up #5241 makes exactly that change). Please define the training term as weighted raw L1 and retain `0.5 * L1` only for the acceptance/confidence target, or explicitly document and validate a different coefficient contract.

docs/training/dspark-speculative-decoding.md
```diff
@@ -0,0 +1,167 @@
...
+- `training/post_training/distillation.py`: the custom loss (`loss_func_kd`).
+- `training/gpt_step.py`: `forward_step_modelopt`, which selects that loss when the
+  model is a distillation model.
+- `training/distill.py`: a thin entry point, `distill(config) = pretrain(config,
+  forward_step_modelopt)`.
+
+Plus an `examples/distillation/` recipe and a `docs/training/` page. This
+provider-hook plus custom-forward-step plus custom-loss plus thin-entry pattern is
+the established, low-friction way to add a training subsystem here, and it is the
+recommended shape for DSpark.
```
yaoyu-33: **[High] The recommended extension surface is already deprecated and conflates two different owners.** Current Bridge main warns that [`ModelProviderMixin`-based construction must migrate to `ModelConfig` + `ModelBuilder`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/2086ee62cf2433c132b78778d43a99ea0ac1d2bf/src/megatron/bridge/models/model_provider.py#L241-L245), so a new dynamic provider subclass cannot be the promised stable API. The two choices above also lead to different contracts: an MCore-native design makes MCore own the parallel module/loss/sharded checkpoint, while a ModelOpt-injected design makes ModelOpt own the transformation/module/loss/export and Bridge only configures it. Current ModelOpt main already has an [HF DSpark plugin](https://github.com/NVIDIA/Model-Optimizer/blob/a21173a4d52cd7376a9e1ae274944d5983e3897c/modelopt/torch/speculative/plugins/hf_dspark.py) and [`DSparkExporter`](https://github.com/NVIDIA/Model-Optimizer/blob/a21173a4d52cd7376a9e1ae274944d5983e3897c/modelopt/torch/export/plugins/hf_spec_export.py#L510-L522), while Bridge currently pins [ModelOpt 0.46.0rc1](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/2086ee62cf2433c132b78778d43a99ea0ac1d2bf/pyproject.toml#L133-L137), which predates that support. Please choose one end-to-end owner, or give each alternative its own complete module/loss/construction/checkpoint/export contract, and make the Bridge hook builder-backed rather than prescribing the legacy provider path.

docs/training/dspark-speculative-decoding.md
```diff
@@ -0,0 +1,167 @@
...
+  token IDs), keeping the target and draft in one model.
+- Offline precomputation of target features, streamed to a draft-only training
+  job (as AutoModel's `precompute_dspark.py` does), which sidesteps holding the
+  target in memory but adds a cache format.
+
+### HF to/from Megatron conversion
+
+Because the draft reuses the target's embedding and LM head and adds `fc`,
+`markov_head`, `confidence_head`, and the draft decoder layers, the bridge mapping
+registry must gain `dspark.*` entries (QKV and gated-MLP splits for the draft
```
yaoyu-33: **[High] `dspark.*` Bridge mappings are not implied until the frozen-weight and export schema are defined.** The references do not physically alias target parameters: AutoModel creates draft-owned `embed_tokens`/`lm_head`, [copies the target weights, and freezes the replicas](https://github.com/NVIDIA-NeMo/Automodel/blob/15da2681c6d827fa55c680e96dcaa8414db4cf88/nemo_automodel/components/speculative/dspark/draft_qwen3.py#L214-L230), then [controls their trainability separately](https://github.com/NVIDIA-NeMo/Automodel/blob/15da2681c6d827fa55c680e96dcaa8414db4cf88/nemo_automodel/components/speculative/dspark/draft_qwen3.py#L265-L282). That distinction determines PP placement, sharded save/resume, and whether serving export stores or reconstructs them. It also differs by owner: ModelOpt already owns its DSpark deployment schema/exporter, whereas an MCore-native checkpoint would need a separately specified HF schema before `AutoBridge` can map it. Please define alias-versus-replica storage, TP/vocab-sharded exact-L1 computation, selected/final-hidden PP transport, CP masking, online checkpoint versus offline-cache schema, and the repository that owns export; then list only the Bridge mappings that remain necessary.

yaoyu-33 (changes requested):
Architecture disposition: changes requested. This is a useful docs-only RFC, but it does not yet define an implementable stable contract. Before merge, it should correct the objective scaling, choose one end-to-end owner (or specify two complete alternative contracts), and define the frozen-weight, distributed-loss, checkpoint, and export boundaries. No DSpark core implementation is requested in this PR.

khazic:
Thanks @yaoyu-33, all three are addressed in 02156f95f. I verified each against the sources you cited before changing anything.

**Objective scaling.** You are right and the page was wrong. Eq. 10 minimizes the weighted raw L1 at coefficient 0.9, and the `1/2` belongs to the Eq. 8 acceptance label. AutoModel's `_compute_local_l1_term` accumulates `l1_dist_per_token * loss_weight_mask` with no `1/2`, and the `0.5` appears only in `_compute_accept_rate_3d`; DeepSpec splits it the same way. The page now defines `L_l1` as the raw weighted L1 and keeps `0.5 * L1` only in the acceptance label, and #5234 adds a regression asserting the pair together (disjoint distributions give `l1_loss == 2`, `accept_rate == 0`) so the two cannot drift again. I have closed #5241, which would have made exactly the change you flagged.

**Owner.** Settled on Megatron-Core native, with the reasoning written into a new "Decision" section rather than left implicit. The deciding fact is that ModelOpt's DSpark is HF-path only today: `plugins/` has `hf_dspark.py` and `modeling_dspark.py` and export has `DSparkExporter`, but the Megatron-side plugins are `megatron_eagle.py` and `megatron_medusa.py`, with no `megatron_dspark.py`. Going the ModelOpt route would mean landing a Megatron-native DSpark in Model-Optimizer and then waiting on the `nvidia-modelopt==0.46.0rc1` pin before any Bridge milestone could move. The consequence is spelled out too: Bridge owns the HF export schema, because `DSparkExporter` covers ModelOpt's own artifacts and not an MCore-native checkpoint.

**Deprecated extension surface.** Dropped. The construction path is now `ModelConfig` + `ModelBuilder` (a `DSparkModelConfig` naming a `DSparkModelBuilder`, mirroring `MambaModelConfig`/`MambaModelBuilder`), and the `dspark_provider.py` row is gone from the table along with the distillation-provider blueprint section that recommended it.

**Contracts that were missing.** The page now specifies replica-not-alias storage for the frozen embedding and LM head and why that choice drives PP placement, sharded save/resume, and export; shard-local exact L1 under TP (with gathered or top-k paths ruled out, since the acceptance label depends on its exactness); selected-hidden transport under PP; block-aligned anchor sampling under CP; the versioned offline feature-cache schema; and the checkpoint/export ownership split. It also documents the Megatron-Core loss return contract and its `calculate_per_token_loss=True` requirement, plus the measured memory contract for the probability L1. Both of those carry numbers from #5234's validation rather than estimates.

yaoyu-33 (approved):

kamran-nvidia:
/ok to test [ac09592](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5233/commits/ac0959283e21de5b05936903ea9bcc22e5a7e27d)

PR #5236, opened 2026-08-02
ci(cache): AUT-1188 reuse baseline MCore dependencies

svcnemo-autobot:
## Background

Downstream MCore dispatches invalidated expensive MBridge dependency layers. Follow-up CI exposed functional failures and an unsafe boundary for contributor-controlled MCore revisions.

## What changed

- Cache stable baseline MCore dependencies and test run→baseline→legacy donor fallback.
- Rebuild/assert dispatched `helpers_cpp` and pin compatible NVSkills.
- Allow only canonical main ancestry or explicitly mirrored PR revisions before executing MCore code.
- Isolate recipe outputs, reset reusable CUDA graphs, and adapt recipe-built distillation configs.

## Details

```mermaid
flowchart LR
    R[Dispatched revision] --> V[Provenance validation]
    V --> B[Cached dependency build]
    B --> H[Native helper rebuild]
    H --> T[Functional gates]
```

Revision validation precedes every fetch/build path; real Git fixtures cover accepted ancestry and mirrors plus rejected orphan, arbitrary, malformed, and fork inputs.

## Tested

- Full pre-commit, cache/donor, Docker-layer, provenance, and `git diff --check` gates
- `tests/unit_tests/training/test_train.py`: 58 passed
- H100 CUDA-graphs: 3/3 cases passed on both ranks
- H100 distillation: passed on both ranks


docker/Dockerfile.ci
```diff
@@ -105,15 +105,39 @@ RUN --mount=type=bind,source=docker/patches/deepep.patch,target=/opt/deepep.patc
 
 COPY pyproject.toml uv.lock /opt/Megatron-Bridge/
 COPY src/megatron/bridge/__init__.py src/megatron/bridge/package_info.py /opt/Megatron-Bridge/src/megatron/bridge/
-COPY 3rdparty/Megatron-LM/pyproject.toml 3rdparty/Megatron-LM/setup.py /opt/Megatron-Bridge/3rdparty/Megatron-LM/
-COPY 3rdparty/Megatron-LM/megatron/training/__init__.py /opt/Megatron-Bridge/3rdparty/Megatron-LM/megatron/training/
-COPY 3rdparty/Megatron-LM/megatron/core/__init__.py 3rdparty/Megatron-LM/megatron/core/package_info.py /opt/Megatron-Bridge/3rdparty/Megatron-LM/megatron/core/
-COPY 3rdparty/Megatron-LM/megatron/core/datasets/Makefile 3rdparty/Megatron-LM/megatron/core/datasets/helpers.cpp /opt/Megatron-Bridge/3rdparty/Megatron-LM/megatron/core/datasets/
```
ko3n1g, line 111: @svcnemo-autobot We say the layer-cache of COPY is too fragile, using a build-arg that clones in-container is more stable?

svcnemo-autobot: 💬 Yes—for this specific dependency layer, but not because `COPY` itself is unreliable. `COPY 3rdparty/Megatron-LM` hashes the entire dispatched checkout, so any MCore source change invalidates the expensive dependency install. PR #5236 instead passes an immutable baseline SHA, clones that baseline before the mutable `COPY`, installs reusable dependencies, then copies the dispatched checkout and syncs only its delta. The stability comes from the immutable SHA and layer ordering; cloning a mutable branch via build arg would still be fragile.

ko3n1g: @svcnemo-autobot even though we copy files and not whole dirs, it still hashes the enter dispatched checkout?

svcnemo-autobot: 💬 Not for the two narrow MBridge `COPY`s: they hash only `pyproject.toml`, `uv.lock`, `__init__.py`, and `package_info.py`. The dispatched MCore step is different: `COPY 3rdparty/Megatron-LM ...` names the directory, so BuildKit hashes every non-ignored file under that directory. Since `.dockerignore` does not exclude MCore source, any dispatched source change invalidates that COPY and all following layers. The new baseline dependency install is intentionally before this directory COPY, so it remains reusable.

ko3n1g: @svcnemo-autobot i'm still confused, in the original file upstream/main, while line exactly causes the invalidation?

svcnemo-autobot: 💬 In original upstream/main, the first invalidation is whichever narrow MCore `COPY` sees a changed file: `docker/Dockerfile.ci:108–111`. For example, a dispatched change to MCore `pyproject.toml` invalidates line 108; a change to `megatron/core/package_info.py` invalidates line 110. That invalidates the expensive `RUN` at lines 117–144. Ordinary MCore source files not listed there do **not** invalidate that dependency layer; they are first included by `COPY --chmod=644 .` at line 146, after it. The PR’s benefit is specifically making dependency-metadata changes reuse an immutable baseline install, not avoiding invalidation from every MCore source change.

ko3n1g: @svcnemo-autobot okay but this PR doesn't change that, it makes it worse: Each time `ARG=BASELINE_MCORE_REF` changes (which is literally each time MLM -> MBridge downstream is called), the build-arg is updated, mounted into the RUN layers and triggers a cache-invalidation. Prior to that the cache invalidation what limited to those specific files, now it spans over the whole MLM repo thru the `BASELINE_MCORE_REF` var

svcnemo-autobot: 💬 The key distinction is that `BASELINE_MCORE_REF` is **not** the dispatched MLM SHA. The workflow records it at `cicd-main.yml:373–374`, before replacing the submodule, so it is the MCore gitlink pinned by the MBridge checkout—currently `6513e3e…`. Repeated MLM→MBridge dispatches change `MCORE_REF`/the checkout, but keep `BASELINE_MCORE_REF` constant; therefore the baseline clone/install layers remain cacheable. They invalidate only when MBridge updates its pinned MCore baseline, which is appropriate because baseline dependencies may then change. My earlier wording conflated the baseline and dispatched refs; thanks for catching that.

ko3n1g (dismissed):

docker/Dockerfile.ci
ko3n1g: @svcnemo-autobot okay so put different: We first install MBridge:HEAD with MLM(MBridge:main), and then later run a reinstall of MLM(MBRidge:HEAD).

Why the three branches of:
```
  if [ "$MCORE_TRIGGERED_TESTING" = "true" ]; then \
        echo "⚙️ MCore testing mode: syncing the dispatched dependency delta"; \
        uv sync --only-group build && \
        uv sync --link-mode copy --all-extras --all-groups --no-group diffusion && \
        uv pip install --no-deps --reinstall -e 3rdparty/Megatron-LM; \
    elif [ -n "$BASELINE_MCORE_REF" ]; then \
        uv pip install --no-deps --reinstall -e 3rdparty/Megatron-LM; \
    else \
        uv sync --locked --only-group build && \
        uv sync --link-mode copy --locked --all-extras --all-groups --no-group diffusion; \
    fi
```

Doesn't uv enforce a re-install which makes `uv pip install --no-deps --reinstall -e 3rdparty/Megatron-LM; ` unnecessry?

svcnemo-autobot: 💬 That’s a good question. MCore is already declared as an editable path source in `pyproject.toml`, so `uv sync` may make the explicit `uv pip install --reinstall -e` redundant. I’m verifying the exact behavior in the CI environment before removing it or defending it.

ko3n1g (dismissed):

ko3n1g (approved):

yaoyu-33 (approved):

PR #5241, opened 2026-08-02
fix(training): train the DSpark TV term at 0.5 * L1 as documented

khazic:

## What

The trained TV term summed the raw probability L1 distance, which is twice the total-variation distance. With the default alphas the objective was effectively `0.1 * CE + 1.8 * TV` instead of the documented `0.1 * CE + 0.9 * TV`, and the `l1_loss` metric reported twice the documented TV value.

## Fix

Move the L1-vs-TV convention into the helper itself: `_l1_distance_per_token` becomes `_tv_distance_per_token` and returns `0.5 * ||p_draft - p_target||_1` directly. `accept_rate = 1 - tv` and `l1_num = (tv * weight).sum()` then fall out with no inline `0.5` constants, so no future call site can accidentally grab the un-halved distance.

Note: the `l1_loss` metric key is kept for continuity but now reports the TV value (half the raw L1), matching the module docstring and the design doc. Runs compared across this change will show the term drop by 2x; that is the fix, not a regression.

## Tests

New unit test `test_l1_loss_is_total_variation_not_full_l1` builds two disjoint saturated distributions (raw L1 = 2.0, TV = 1.0) and asserts both the metric and the loss equal 1.0. The test fails on the parent commit and passes here. All 17 dspark unit tests pass; ruff check and format are clean.

## Merge note

This branch and the masked-input-robustness fix PR (same stack) both touch the `l1_dist` block of `dspark_loss`, so merging both into the DSpark core branch (#5234) will conflict there. Resolve by scrubbing first, then using the TV helper: `tv_dist = _tv_distance_per_token(...)`, `tv_dist = torch.where(eval_bool, tv_dist, 0.0)`, then `accept_rate = (1.0 - tv_dist).clamp_(0.0, 1.0)`.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -82,6 +82,7 @@ training/moe-optimization.md
 training/peft.md
 training/packed-sequences.md
 training/multi-token-prediction.md
+training/dspark-speculative-decoding.md
 training/callbacks.md
 ```
 
--- a/docs/training/dspark-speculative-decoding.md
+++ b/docs/training/dspark-speculative-decoding.md
@@ -1,0 +1,212 @@
+# DSpark Speculative Decoding (Draft Training) Design
+
+## Overview
+
+This page is a design proposal (RFC) for adding **DSpark** speculative-decoding
+draft training to Megatron-Bridge. It tracks
+[issue #5230](https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/5230).
+
+DSpark ([arXiv:2607.05147](https://arxiv.org/abs/2607.05147)) is a
+speculative-decoding method that trains a lightweight **draft** against a frozen
+target. The draft proposes a block of tokens; the target verifies them. DSpark's
+draft is *semi-autoregressive*: a parallel backbone predicts a whole block in one
+pass, and a small sequential head restores the intra-block token dependency that a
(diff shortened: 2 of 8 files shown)


khazic:
Closing this: the change is in the wrong direction and @yaoyu-33 is right on the review in #5234.

I checked it independently. Paper [Eq. 10](https://arxiv.org/html/2607.05147#S3.E10) minimizes the weighted **raw** L1 distance at coefficient 0.9. The factor `1/2` appears in [Eq. 8](https://arxiv.org/html/2607.05147#S3.E8) only, where the analytical acceptance/confidence target is formed. NeMo AutoModel makes the same split: `_compute_local_l1_term` accumulates `l1_dist_per_token * loss_weight_mask` with no `1/2`, and the `0.5` shows up only in `_compute_accept_rate_3d`. DeepSpec does the same.

So the implementation was already correct, and what was actually wrong was the docstring in `loss.py` and the objective section of the design page, both of which described the training term as `0.5 * ||p_draft - p_target||_1`. Landing this PR would have halved the dominant term's gradient while leaving the published coefficient in place.

Both descriptions are corrected in #5233 and #5234 instead, and #5234 adds a regression test that asserts the pair together (disjoint distributions give `l1_loss == 2` with `accept_rate == 0`) so the objective and its documentation cannot drift apart again.

PR #5242, opened 2026-08-02
fix(training): exclude dropped anchor blocks from the DSpark loss and metrics

khazic:

## What

`block_keep_mask` was consumed only by the `tau` diagnostic. The loss weight was built from `eval_mask` alone, so a dropped anchor block (keep = 0) whose eval prefix was still set received full CE/TV/confidence gradient, and its positions were averaged into `accept_rate`, biasing both the objective and the diagnostics.

## Fix

Define the supervised-position mask once in `dspark_loss` (`supervised = eval_mask * block_keep_mask.unsqueeze(-1)`), derive the decayed loss weight from it, and pass it into `_acceptance_metrics` (now keyword-only per the repo style rule on same-type parameters). One mask drives the loss weight, `accept_rate`, and the `tau` block validity, so the three sites cannot drift.

## Tests

Two new unit tests: dropped blocks produce exactly zero gradient and the loss equals the eval-mask-zeroed equivalent; `accept_rate`/`tau` ignore a dropped block that deliberately mismatches the teacher. Both fail on the parent commit and pass here. All 18 dspark unit tests pass; ruff is clean.

## Merge note

the masked-input-robustness fix PR (same stack) sanitizes padded inputs keyed on `eval_mask` alone. Whichever of the two branches merges into the DSpark core branch (#5234) second should rebase that sanitization onto the combined `supervised` mask (one-line change: build `supervised` first and use `supervised.bool()` as the scrub mask), so that positions of dropped anchors are scrubbed as well as zero-weighted.


khazic:
Status note: this PR needs a rebase and is not reviewable as it stands.

It was stacked on #5241, which I have now closed. Paper Eq. 10 trains the weighted **raw** L1 and the factor `1/2` belongs only to the Eq. 8 acceptance label, matching DeepSpec and NeMo AutoModel, so #5241 would have halved the dominant term's gradient. Because the stack is linear (#5241 to #5242 to #5243 to #5244), this branch still carries that commit as its base and its diff still shows the `_tv_distance_per_token` rename.

The change this PR actually contributes is still valid and I am keeping it. #5234 has meanwhile been reworked fairly heavily (Megatron-Core 3-tuple return contract, no `dp_group`, and both full-vocabulary terms fused into one chunked, checkpointed pass), so the fix has to be reapplied on top of that rather than replayed. I will rebase once #5234 settles, to avoid redoing it if that review moves again.

khazic:
Rebased onto the reworked #5234 in f7fd089f2. The closed #5241 is out of this branch's history, so the diff no longer carries the `0.5 * L1` change or the `_tv_distance_per_token` rename.

Reapplied rather than replayed, because #5234 changed underneath it: `dp_group` is gone, the loss returns the Megatron-Core 3-tuple with reducible `[numerator, denominator]` pairs, and both full-vocabulary terms now run through one chunked, checkpointed pass. The fix itself is unchanged in intent. One combined supervision mask (`eval_mask` AND the kept-anchor flag) drives the loss weight, and it now also drives `num_tokens` and the acceptance denominators, which the previous version could not do because those did not exist yet. Dropped anchors leave the metric denominators entirely instead of sitting in them at weight zero.

Tests: the two original cases are kept, with the equivalence check extended to `num_tokens` and a new assertion that the dropped anchor is out of the `accept rate` and `tau` denominators. 26 unit tests pass, and the 2-GPU functional test passes under `torchrun --nproc_per_node=2`.

khazic:
Merged the current head of #5234 in, which is what this branch stacks on.

That also brings over the docs fix for the module docstring in `dspark/loss.py`: the `Return contract` section used an RST underline, which MyST parses as a setext H2, so the generated API page started at H2 and `sphinx-build --fail-on-warning` rejected it with `Document headings start at H2, not H1 [myst.header]`. #5234 replaced the underline with bold text; this branch was still carrying the older version and failing the docs build for that reason.

No change to the fix itself.

PR #5244, opened 2026-08-02
fix(training): sanitize masked sentinel and non-finite inputs in the DSpark loss

khazic:

## What

`dspark_loss` computed CE, softmax, and BCE over every position and only zeroed masked positions by multiplying with the zero weight afterwards. Two failure modes survive that scheme once real padded batches arrive:

1. `target_ids` padded with anything other than `-100` (e.g. `-1` or `vocab_size`) at masked positions crashes `F.cross_entropy` with a device-side assert.
2. Non-finite padded logits at masked positions poison the loss through `0 * NaN == NaN`, and a non-finite forward row also produces NaN gradients through cross_entropy's backward even when the upstream gradient is zero, so output-side masking alone cannot fix it.

## Fix

Sanitize the inputs at masked positions before any op: `draft_logits` rows are zeroed via `torch.where`, `target_ids` sentinels are replaced with 0 via `masked_fill`, the per-position TV distance is scrubbed on the reduced `[B, N, L]` tensor (cheaper than copying the full-vocab teacher, which is detached and only consumed through this reduction), and `confidence_pred` is zeroed at masked positions. The padded-input contract is documented on `DSparkForwardOutput.eval_mask`.

The draft-side `torch.where` costs one extra logits-sized allocation per step. This is a deliberate robustness trade: it is the only placement that guarantees finite gradients regardless of how the future producer pads, and the loss already materializes two fp32 softmax copies of the same size. If a future producer is guaranteed to pad the draft with finite values, it can be dropped.

`heads.get_prev_embeddings` intentionally stays a strict lookup. Clamping sentinels there would silently map genuinely corrupt negative ids at supervised positions to row 0 (and corrupt the RNN head's recurrent state for later positions in the block), so the docstring now states the producer contract instead: replace sentinel pads before calling, exactly as `dspark_loss` sanitizes `target_ids`.

## Tests

New unit test poisons masked positions with `-1` target ids, `-inf` draft rows, and NaN teacher/confidence values, then asserts the loss, all metrics, and the gradients are finite and exactly equal to the clean-input reference, with zero gradient at masked positions. It fails (crashes) on the parent commit and passes here. All 17 dspark unit tests pass; ruff is clean.

## Merge note

The sanitize mask is `eval_mask`, matching this branch's loss weighting. the block-keep-mask fix PR (same stack) redefines the weight as `eval_mask AND block_keep_mask`; whichever branch merges into the DSpark core branch (#5234) second should key the sanitization on that combined `supervised` mask so dropped-anchor padding is scrubbed too. This branch also conflicts textually with the TV-half-factor fix PR (same stack) in the `l1_dist` block; resolve by scrubbing first, then computing TV via the helper from that branch.


khazic:
Status note: this PR needs a rebase and is not reviewable as it stands.

It was stacked on #5241, which I have now closed. Paper Eq. 10 trains the weighted **raw** L1 and the factor `1/2` belongs only to the Eq. 8 acceptance label, matching DeepSpec and NeMo AutoModel, so #5241 would have halved the dominant term's gradient. Because the stack is linear (#5241 to #5242 to #5243 to #5244), this branch still carries that commit as its base and its diff still shows the `_tv_distance_per_token` rename.

The change this PR actually contributes is still valid and I am keeping it. #5234 has meanwhile been reworked fairly heavily (Megatron-Core 3-tuple return contract, no `dp_group`, and both full-vocabulary terms fused into one chunked, checkpointed pass), so the fix has to be reapplied on top of that rather than replayed. I will rebase once #5234 settles, to avoid redoing it if that review moves again.

khazic:
Rebased onto #5242 (itself now on the reworked #5234) in ba3eede3b. The closed #5241 is out of this branch's history, so the diff no longer carries the `0.5 * L1` change.

Reapplied rather than replayed. The sanitizing is the same idea and the reasoning is now written into the code, because it is the non-obvious part: masking the outputs is not enough, since `0 * NaN` is NaN in the forward and a non-finite row poisons the backward whatever its weight. Replacing the *inputs* with `torch.where` is what makes it safe, because `where`'s backward selects exact zeros into the masked slots instead of multiplying by zero, so no NaN reaches `draft_logits.grad`.

Two things changed with the new base. The scrub keys off #5242's combined supervision mask, so a poisoned position inside a dropped anchor is covered too, which a new test asserts. And the reduced-L1 scrub now sits ahead of the fused chunk output rather than the old `_tv_distance_per_token` call.

Tests: the original poisoned-versus-clean case is kept and strengthened to compare every reporting pair's numerator and denominator rather than a fixed metric list. 28 unit tests pass, and the 2-GPU functional test passes under `torchrun --nproc_per_node=2`.

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5244/commits/ba3eede3b4515d06fd6877447ac34cf34693b7a3

khazic:
Merged the updated #5242 in (which itself now carries the current #5234 head), so this branch is back on top of the stack it belongs to.

That clears the docs failure: `dspark/loss.py`'s module docstring used an RST underline for the `Return contract` section, MyST read it as a setext H2, and `sphinx-build --fail-on-warning` failed with `Document headings start at H2, not H1 [myst.header]`. #5234 fixed it by switching to bold; this branch was pinned to the version before that fix.

The other red checks on the previous run were not code failures: `Launch_Unit_Tests_Core` hit `Test timed out after 18 minutes`, and `release-summary` / `Nemo_CICD_Test` were just reporting that one failed job upstream. Could someone re-trigger CI on this head?

No change to the sanitization fix itself.

PR #5249, opened 2026-08-02
fix(diffusion): calculate WAN FLOPs from runtime geometry

cuichenx:
# What does this PR do?

Calculate WAN training FLOPs from the latent and text-context geometry actually executed by each batch. The theoretical MODEL_TFLOP/s/GPU numerator no longer depends on the compatibility-only `model.seq_length` field.

# Changelog

- Collect padded and unpadded WAN query and cross-attention key/value lengths from the existing `PackedSeqParams` batch metadata.
- Aggregate DP-global query-token, query-square, context-token, and query-context-product sums without duplicating CP or virtual-pipeline replicas.
- Add a WAN-specific matrix-multiplication estimate for self-attention, cross-attention, MLP, patch/text embedding, and output-head work.
- Preserve the existing three-value `resolve_global_flops_seqlen_stats` API and the original three-int packed-data collective for non-WAN models. The five-int collective is selected from a rank-uniform provider capability only for runtime cross-attention accounting.
- Add focused tests for runtime-estimator dispatch, packed/padded cross-attention statistics, distributed extrapolation and collective shape, exact WAN arithmetic, and invariance to `model.seq_length`.

There are no changes to the performance evaluator, evaluation tests, flat performance overrides, FLUX, or generic throughput logging.

# Root cause and impact

PR #4779 synchronized the CLI sequence length into `model.seq_length`. That is meaningful for language models because the field controls token workload. For WAN it is a compatibility placeholder: work is determined by packed latent video geometry and text-context length.

The target WAN 14B mock workload executes 37,440 query tokens and 512 context tokens per sample. Its previous generic estimates used sequence lengths 1024 and 8192, so both 7 and 63 MODEL_TFLOP/s/GPU were invalid undercounts; their 9x ratio was an estimator artifact.

| Measurement | Run/control | Old MODEL_TFLOP/s/GPU | New MODEL_TFLOP/s/GPU | Old step time | New step time | Step-time change |
|---|---|---:|---:|---:|---:|---:|
| NeMo CI result | Stored baseline vs. MR current | 6.9933 | 63.1000 | 39.1637 s | 39.1675 s | +0.010% |
| Controlled measurement with legacy estimator | Old/new/new/old on 16 GB200 GPUs, 30 pooled samples per variant | 7.0000 | 63.4833 | 38.904237 s | 38.929890 s | +0.0659% |
| Same controlled measurement with this corrected WAN estimator | Identical timings, both variants re-scored with 395.781682 PFLOP/step | 635.83 | 635.41 | 38.904237 s | 38.929890 s | +0.0659% |

The first two rows show the bug directly: reported TFLOP/s jumps by about 9x while physical step time is effectively unchanged. With one valid numerator applied to both variants, the rates move only with the measured time. The new point estimate is 0.06594% slower, within run-to-run variance, so there is no demonstrated performance improvement.

This issue is exposed on WAN because its generic sequence field does not describe runtime geometry. Text-model estimators remain valid. Other diffusion architectures may also need architecture-specific estimators, but their patch/context formulas must be derived separately rather than applying WAN logic or a broad domain opt-out.

# Common-path impact audit

- Only WAN defines `_get_num_floating_point_operations_with_runtime_stats` and only `WanForwardStep` passes cross-attention boundaries.
- Ten representative non-WAN formula paths were compared directly against `origin/main`—dense GPT, dynamic lengths, transformer MoE, hybrid Mamba/MoE, GDN, MLA, Kimi KDA, DeepSeek-V4, VLM/ViT, and LoRA—and were bit-identical. The legacy provider override was also identical.
- Dense models still perform no FLOP-metadata collective. Packed non-WAN models retain the same one three-int all-reduce; only WAN uses five ints.
- A rank-local width selector found during independent review was replaced before final publication with a rank-uniform model-capability selector. A zero-local-cross regression verifies that every WAN DP rank still enters the five-int collective.

The common training-loop plumbing is required because the exact latent and text lengths are batch-dependent and the loop owns per-step DP-global FLOP accumulation. A provider-only static value would remain wrong for different video resolutions, frame counts, or context lengths; a WAN-only reduction inside `forward_step` would duplicate the existing synchronization and step/VPP accounting path.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci) in the Contributing doc for how to trigger CI.

Validation in `nvcr.io/nvidian/nemo:26.08.rc3`, with the mounted checkout first on `PYTHONPATH`:

- 211 focused WAN, FLUX, FLOP utility, and common training-loop tests passed locally.
- HSG-OCI Slurm job `5839477` passed the same 211 tests on one full GB200 node (`nvl72110-T18`, 4 GPUs, explicit segment 1) at validated commit `fbc548b58816c6197bd46fce1a89f43238c91fbe`.
- Ruff formatting and checks passed on all changed Python files.
- `git diff --check` passed.
- Independent read-only final review approved with no findings after verifying formula, common-path isolation, and distributed collective safety.

# Before your PR is Ready for review

**Pre checks**:

- [x] Added focused unit tests for the changed contracts and arithmetic.
- [x] Evaluated documentation needs; no user-facing configuration workflow changed.
- [x] No dependency or optional-install component is affected.

# Additional Information

- Follow-up to #4779.
- Opened and retained as a draft per repository policy.
- WAN with PEFT is not enabled by current recipes; hypothetical WAN+LoRA accounting remains a separate follow-up rather than broadening this BF16 performance fix.

### DIFF
--- a/src/megatron/bridge/diffusion/models/wan/wan_provider.py
+++ b/src/megatron/bridge/diffusion/models/wan/wan_provider.py
@@ -72,6 +72,59 @@ class WanModelProvider(TransformerConfig, ModelProviderMixin[VisionModule]):  # 
     text_len: int = 512
     text_dim: int = 4096
 
+    def _get_num_floating_point_operations_with_runtime_stats(
+        self,
+        *,
+        batch_size: int,
+        seqlen_sum: int | None,
+        seqlen_squared_sum: int | None,
+        cross_seqlen_sum: int | None,
+        cross_seqlen_product_sum: int | None,
+    ) -> int:
+        """Estimate WAN training FLOPs from the executed latent and context lengths."""
+        if batch_size <= 0:
+            return 0
+        if any(
+            value is None for value in (seqlen_sum, seqlen_squared_sum, cross_seqlen_sum, cross_seqlen_product_sum)
+        ):
+            raise ValueError("WAN FLOP accounting requires runtime self- and cross-attention sequence metadata")
+
+        query_tokens = int(seqlen_sum)
+        query_tokens_squared = int(seqlen_squared_sum)
+        context_tokens = int(cross_seqlen_sum)
+        query_context_tokens = int(cross_seqlen_product_sum)
+
+        hidden_size = self.hidden_size
+        kv_channels = self.kv_channels or hidden_size // self.num_attention_heads
+        num_query_groups = self.num_attention_heads if self.num_query_groups is None else self.num_query_groups
+        query_projection_size = kv_channels * self.num_attention_heads
+        kv_projection_size = kv_channels * num_query_groups
+
+        self_attention_flops = (
+            2 * query_tokens * hidden_size * (query_projection_size + 2 * kv_projection_size)
+            + 4 * query_tokens_squared * query_projection_size
+            + 2 * query_tokens * query_projection_size * hidden_size
+        )
+        cross_attention_flops = (
+            2 * query_tokens * hidden_size * query_projection_size
+            + 4 * context_tokens * self.crossattn_emb_size * kv_projection_size
+            + 4 * query_context_tokens * query_projection_size
+            + 2 * query_tokens * query_projection_size * hidden_size
+        )
+        mlp_projection_factor = 6 if self.gated_linear_unit else 4
+        mlp_flops = mlp_projection_factor * query_tokens * hidden_size * self.ffn_hidden_size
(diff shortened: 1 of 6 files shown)


cuichenx:
waiting for nemo ci golden value update

balasaajay (approved):

PR #5253, opened 2026-08-02
fix(qwen3.5-vl): use top-level prefix for dense MTP mappings

### LINKED ISSUE #5252: [bug] Qwen35VLBridge MTP mappings use wrong Megatron prefix (language_model.mtp.* instead of mtp.*), crashes weight import
## Problem

`Qwen35VLBridge.mapping_registry()` (the dense Qwen3.5/3.6 VL bridge, e.g. for
[`Qwen/Qwen3.5-27B`](https://huggingface.co/Qwen/Qwen3.5-27B)) builds its MTP (Multi-Token
Prediction) parameter mappings with the wrong Megatron-side prefix, so every MTP parameter
silently fails to resolve during weight conversion.

In `src/megatron/bridge/models/qwen_vl/qwen35_vl_bridge.py`:

```python
mapping_list.extend(
    Qwen35Bridge._get_dense_lm_mappings(hf_prefix="model.language_model.", megatron_prefix="language_model.")
)
mapping_list.extend(Qwen35Bridge._get_dense_mtp_mappings(megatron_prefix="language_model."))
mapping_list.extend(_get_vision_mappings())
```

The main decoder mappings correctly use `megatron_prefix="language_model."` because in the
instantiated Megatron VL model, the text decoder lives inside a submodule literally named
`language_model` (a sibling of the vision tower). But the MTP submodule is **not** nested under
`language_model` — it's attached at the top level of the model, a sibling of `language_model`.
So its real Megatron parameter names are plain `mtp.layers.0.*`, not
`language_model.mtp.layers.0.*`.

Because `_get_dense_mtp_mappings` is called with `megatron_prefix="language_model."`, every
resulting mapping is keyed `language_model.mtp.layers.0.*`, which never matches the real
parameter name (`mtp.layers.0.*`) looked up at runtime. Every single MTP parameter fails
`megatron_to_hf_lookup` / `hf_to_megatron_lookup` and is treated as unmapped.

Downstream, in `src/megatron/bridge/models/conversion/model_bridge.py`,
`build_conversion_tasks()` leaves unmatched slots as a literal `None` in the `tasks` list (this
is different from the deliberate per-rank-skip case, where the task exists but
`task.megatron_module` is `None`). `load_weights_hf_to_megatron()` then does:

```python
    # None means megatron module not on current rank, skip if this task is not going to happen
        continue
```

When `task` itself is the literal `None` (a totally missing task, not a task with a `None`
module), this raises `AttributeError: 'NoneType' object has no attribute 'megatron_module'`,
crashing the weight import entirely rather than just warning about unmapped MTP weights.

## Minimal repro

```python

# Any dense Qwen3.5/3.6-architecture VL checkpoint, e.g. a Qwen3.5-27B-VL variant.
bridge = AutoBridge.from_hf_pretrained("<dense Qwen3.5/3.6 VL checkpoint>")

# Enable MTP.
model_cfg = bridge.to_megatron_provider()
model_cfg.mtp_num_layers = 1

model = model_cfg.provide()  # fresh Megatron model, MTP submodule attached at top level

# Import pretrained HF weights into the freshly constructed model.
bridge.load_weights_hf_to_megatron(model)
```

## Observed output

```
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.enorm.weight
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.hnorm.weight
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.eh_proj.weight
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.final_layernorm.weight
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.mtp_model_layer.self_attention.linear_proj.weight
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.mtp_model_layer.self_attention.linear_qkv.layer_norm_weight
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.mtp_model_layer.self_attention.linear_qkv.weight
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.mtp_model_layer.self_attention.q_layernorm.weight
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.mtp_model_layer.self_attention.k_layernorm.weight
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.mtp_model_layer.mlp.linear_fc1.layer_norm_weight
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.mtp_model_layer.mlp.linear_fc1.weight
WARNING | megatron.bridge.models.conversion.model_bridge:build_conversion_tasks:1602 - WARNING: No mapping found for megatron_param: mtp.layers.0.mtp_model_layer.mlp.linear_fc2.weight

Traceback (most recent call last):
  ...
  File ".../megatron/bridge/models/conversion/model_bridge.py", line 1046, in load_weights_hf_to_megatron
AttributeError: 'NoneType' object has no attribute 'megatron_module'
```

That specific traceback is from a downstream trainer calling into `AutoBridge.load_weights_hf_to_megatron()`; the root cause is entirely inside Megatron-Bridge's mapping construction.

## Expected behavior

MTP parameters for the dense Qwen3.5/3.6 VL bridge should resolve correctly (matching the
top-level `mtp.layers.*` Megatron parameter names), the same way they already do for the plain
(non-VL) `Qwen35Bridge.mapping_registry()`, which calls
`_get_dense_mtp_mappings(megatron_prefix="")` correctly.

## Root cause / fix

One-line fix in `Qwen35VLBridge.mapping_registry()`:

```diff
-        mapping_list.extend(Qwen35Bridge._get_dense_mtp_mappings(megatron_prefix="language_model."))
+        mapping_list.extend(Qwen35Bridge._get_dense_mtp_mappings(megatron_prefix=""))
```

The mapping *content* (which HF weight maps to which Megatron weight) was already correct —
only the prefix on the Megatron-side key was wrong.

I have a PR ready with this fix plus a regression test: it will follow shortly and reference this
issue.

**Note on a likely related bug (unverified):** `Qwen35VLMoEBridge` (the MoE VL variant, e.g. for
Qwen3.5-397B-A17B) has the exact same pattern at its own
`_get_moe_mtp_mappings(megatron_prefix="language_model.", ...)` call site. This looks like the
same class of bug with the same fix (empty prefix), but I have not verified it against real
MoE+VL+MTP training logs — my confirmed evidence below is for the dense variant only. Flagging it
here in case a maintainer wants to fix it in the same pass.

## Affected area

area:ckpt

## Regression?

Not sure — this looks like it may have been wrong since the dense VL bridge's MTP mapping was
added; the plain (non-VL) `Qwen35Bridge` has always used the correct empty prefix.

## Environment

- Branch/commit: `main` (as of this report)
- Model: dense Qwen3.5/3.6-architecture VL checkpoint (27B), `mtp_num_layers=1`
- This surfaced when a downstream RL-training framework called `AutoBridge.load_weights_hf_to_megatron()` to import pretrained HF weights into a freshly-constructed Megatron model at cold start.

## Related

- #3348 / the corresponding PR hit a very similar `No mapping found for megatron_param: mtp.layers.0.*` symptom for the plain (non-VL) `Qwen3Bridge`, for a different underlying reason (mappings were entirely missing there, rather than mis-prefixed here).

worktrial2-trajectory:
## Problem

`Qwen35VLBridge.mapping_registry()` (the dense Qwen3.5/3.6 VL bridge) builds its MTP mappings
with the wrong Megatron-side prefix:

```python
mapping_list.extend(Qwen35Bridge._get_dense_mtp_mappings(megatron_prefix="language_model."))
```

In the real instantiated Megatron VL model, the MTP submodule is attached at the top level (a
sibling of `language_model`, same as the vision tower), so its real parameter names are
`mtp.layers.0.*`, not `language_model.mtp.layers.0.*`. Every MTP parameter therefore fails to
resolve via `megatron_to_hf_lookup` / `hf_to_megatron_lookup`, and `load_weights_hf_to_megatron()`
then crashes with `AttributeError: 'NoneType' object has no attribute 'megatron_module'` when it
hits the resulting missing task entries in `build_conversion_tasks()`.

Confirmed against a dense Qwen3.5/3.6-27B VL checkpoint with `mtp_num_layers=1`, importing
pretrained HF weights into a freshly-constructed Megatron model — every MTP parameter (`enorm`,
`hnorm`, `eh_proj`, `final_layernorm`, and the MTP transformer layer's attention/MLP weights)
logs `No mapping found for megatron_param: mtp.layers.0.*`, followed by the `AttributeError`
above.

Fixes #5252

## Solution

One-line fix in `Qwen35VLBridge.mapping_registry()`:

```diff
-        mapping_list.extend(Qwen35Bridge._get_dense_mtp_mappings(megatron_prefix="language_model."))
+        mapping_list.extend(Qwen35Bridge._get_dense_mtp_mappings(megatron_prefix=""))
```

This mirrors the plain (non-VL) `Qwen35Bridge.mapping_registry()` in
`src/megatron/bridge/models/qwen/qwen35_bridge.py`, which already calls
`_get_dense_mtp_mappings(megatron_prefix="")` correctly. The mapping *content* was already right —
only the Megatron-side prefix was wrong.

Also corrected the docstring/comment on `_get_dense_mtp_mappings` in `qwen35_bridge.py`, which
documented the (incorrect) assumption that VL models should pass `"language_model."` here — that
comment was the source of the confusion that produced this bug.

## Tests

Added `test_mapping_registry_mtp_params_use_top_level_prefix` to
`TestQwen35VLBridgeMappingRegistry` in `test_qwen35_vl_bridge.py`, asserting:
- `mapping_registry().megatron_to_hf_lookup("mtp.layers.0.<param>")` resolves for the MTP norm/
  projection params, and
- the incorrect `"language_model.mtp.layers.0.enorm.weight"` does *not* resolve.

Ran the full `test_qwen35_vl_bridge.py` and `test_qwen35_bridge.py` suites locally (42 and 39
tests respectively, all passing) plus `ruff check`/`ruff format --check` on the touched files.
Verified the test fails on `main` before this change and passes after, reproducing the reported
symptom exactly (all MTP params `MISSING`, `language_model.mtp.*` present instead).

## Notes

- `Qwen35VLMoEBridge` (the MoE VL variant, e.g. Qwen3.5-397B-A17B) has the exact same pattern at
  its own `_get_moe_mtp_mappings(megatron_prefix="language_model.", ...)` call site — likely the
  same class of bug with the same fix (empty prefix). I have **not** verified this against real
  MoE+VL+MTP training logs, so I've left it out of this PR to keep the change minimal to the
  confirmed issue. Flagging it in case a maintainer wants to fold it in, or I'm happy to send a
  follow-up once/if it's confirmed.

### DIFF
--- a/src/megatron/bridge/models/qwen_vl/qwen35_vl_bridge.py
+++ b/src/megatron/bridge/models/qwen_vl/qwen35_vl_bridge.py
@@ -357,7 +357,12 @@ class Qwen35VLBridge(MegatronModelBridge):
         mapping_list.extend(
             Qwen35Bridge._get_dense_lm_mappings(hf_prefix="model.language_model.", megatron_prefix="language_model.")
         )
-        mapping_list.extend(Qwen35Bridge._get_dense_mtp_mappings(megatron_prefix="language_model."))
+        # MTP is attached at the top level of the Megatron VL model (a sibling of
+        # `language_model`, not nested under it), so its real parameter names are
+        # "mtp.layers.*" rather than "language_model.mtp.layers.*". Passing a
+        # "language_model." prefix here would produce mappings that never match the
+        # actual Megatron parameter names.
+        mapping_list.extend(Qwen35Bridge._get_dense_mtp_mappings(megatron_prefix=""))
         mapping_list.extend(_get_vision_mappings())
         return MegatronMappingRegistry(*mapping_list)
 
--- a/tests/unit_tests/models/qwen_vl/test_qwen35_vl_bridge.py
+++ b/tests/unit_tests/models/qwen_vl/test_qwen35_vl_bridge.py
@@ -282,6 +282,18 @@ class TestQwen35VLBridgeMappingRegistry:
         names = self._get_mapping_names(bridge.mapping_registry())
         assert any("patch_embed" in n for n in names)
 
+    def test_mapping_registry_mtp_params_use_top_level_prefix(self, bridge):
+        """MTP is a sibling of `language_model` in the real Megatron VL model, not
+        nested under it, so its Megatron param names must be "mtp.layers.*" rather
+        than "language_model.mtp.layers.*" (see regression this guards against:
+        MTP params silently failing to resolve during weight conversion)."""
+        registry = bridge.mapping_registry()
+        assert registry.megatron_to_hf_lookup("mtp.layers.0.enorm.weight") is not None
+        assert registry.megatron_to_hf_lookup("mtp.layers.0.hnorm.weight") is not None
+        assert registry.megatron_to_hf_lookup("mtp.layers.0.eh_proj.weight") is not None
+        assert registry.megatron_to_hf_lookup("mtp.layers.0.final_layernorm.weight") is not None
+        assert registry.megatron_to_hf_lookup("language_model.mtp.layers.0.enorm.weight") is None
+
 
 # =====================================================================
 # Tests for Qwen35TokenClassificationBridge
--- a/src/megatron/bridge/models/qwen/qwen35_bridge.py
+++ b/src/megatron/bridge/models/qwen/qwen35_bridge.py
@@ -583,8 +583,10 @@ class Qwen35Bridge(MegatronModelBridge):
         """Get MTP (Multi-Token Prediction) parameter mappings for dense Qwen3.5.
 
         Args:
-            megatron_prefix: Prefix for Megatron param names. Use "" for LM and
-                "language_model." for VL models.
+            megatron_prefix: Prefix for Megatron param names. In both the plain LM
+                and the VL model, MTP is attached at the top level of the Megatron
+                model (a sibling of "language_model", not nested under it), so this
+                should always be "".
 
         Returns:
             List of mapping objects for the MTP portion.
@@ -594,8 +596,8 @@ class Qwen35Bridge(MegatronModelBridge):
         # =================================================================
         # MTP (Multi-Token Prediction) mappings
         # MTP uses standard attention (not GDN) and dense MLP.
-        # Megatron VL prefix: language_model.mtp.*
-        # Megatron ML prefix: mtp.*
+        # Megatron prefix: mtp.* (top-level, not nested under "language_model.",
+        # even for the VL model).
         # HF prefix: mtp.* (top-level, not under model.language_model.)
         # =================================================================
         mtp_param_mappings = {


src/megatron/bridge/models/qwen_vl/qwen35_vl_bridge.py
```diff
@@ -357,7 +357,12 @@ def mapping_registry(self) -> MegatronMappingRegistry:
         mapping_list.extend(
             Qwen35Bridge._get_dense_lm_mappings(hf_prefix="model.language_model.", megatron_prefix="language_model.")
         )
-        mapping_list.extend(Qwen35Bridge._get_dense_mtp_mappings(megatron_prefix="language_model."))
+        # MTP is attached at the top level of the Megatron VL model (a sibling of
+        # `language_model`, not nested under it), so its real parameter names are
+        # "mtp.layers.*" rather than "language_model.mtp.layers.*". Passing a
+        # "language_model." prefix here would produce mappings that never match the
+        # actual Megatron parameter names.
+        mapping_list.extend(Qwen35Bridge._get_dense_mtp_mappings(megatron_prefix=""))
```
yaoyu-33, line 365: Blocking: this appears to reverse the prefix used by the model that this bridge itself provides. Qwen35VLModelProvider.provide() passes mtp_block_spec into Qwen3VLModel; Qwen3VLModel registers the resulting Qwen3VLGPTModel as self.language_model, and MCore GPTModel registers the MTP block as self.mtp. Standard named_parameters() on the outer VLM therefore exposes language_model.mtp.layers.*, not top-level mtp.layers.*. The existing MoE-VL path and its tests use that same hierarchy. Please show the exact type(model) and MTP names from provider.provide()/provide_distributed_model(), or keep megatron_prefix="language_model."; with the current implementation this change makes the official provider model unmappable.

tests/unit_tests/models/qwen_vl/test_qwen35_vl_bridge.py
```diff
@@ -282,6 +282,18 @@ def test_mapping_registry_has_vision_patch_embed(self, bridge):
         names = self._get_mapping_names(bridge.mapping_registry())
         assert any("patch_embed" in n for n in names)
 
+    def test_mapping_registry_mtp_params_use_top_level_prefix(self, bridge):
+        """MTP is a sibling of `language_model` in the real Megatron VL model, not
+        nested under it, so its Megatron param names must be "mtp.layers.*" rather
+        than "language_model.mtp.layers.*" (see regression this guards against:
+        MTP params silently failing to resolve during weight conversion)."""
+        registry = bridge.mapping_registry()
+        assert registry.megatron_to_hf_lookup("mtp.layers.0.enorm.weight") is not None
```
yaoyu-33, line 291: This test only proves that the registry contains the literal key introduced by the production change; it cannot detect disagreement with the instantiated Megatron model and therefore passes even if conversion is broken. Please construct a tiny Qwen35VL provider/model with mtp_num_layers=1 (or a faithful wrapper preserving the language_model -> GPTModel -> mtp hierarchy), collect its actual named_parameters(), and assert that every MTP name resolves or produces a non-None conversion task. That test should fail before the real fix and would settle whether the reported top-level name comes from the Bridge provider or from downstream unwrapping/reassembly.

PR #5254, opened 2026-08-03
sjayaram/k8s excalibur

sjayaram-nv:
This branch contains changes for adding the XCalibur executor backend to enable llm benchmarking runs in K8s.


malay-nagda (dismissed):

scripts/performance/setup_experiment.py
```diff
@@ -686,6 +711,36 @@ def main(
...
         )
+    elif xcalibur_namespace is not None:
+        try:
+            from utils.executors import xcalibur_executor
+        except ImportError:
+            from .utils.executors import xcalibur_executor
+        executor = xcalibur_executor(
+            namespace=xcalibur_namespace,
+            image=container_image,
+            num_nodes=num_gpus // gpus_per_node,
```
sudostock:
```suggestion
            num_nodes=-(num_gpus // -gpus_per_node),
```

sudostock: Switch to ceiling division to match other executors.

scripts/performance/setup_experiment.py
```diff
@@ -48,6 +48,11 @@
     )
     from .utils.utils import configure_slurm_gpu_tuning, select_config_variant_interactive
 
+try:
+    from nemo_run.core.execution.xcalibur import XCaliburExecutor as _XCaliburExecutor
+except ImportError:
+    _XCaliburExecutor = None  # type: ignore[assignment,misc]
+
```
sudostock: This block appears to be unused.

scripts/performance/utils/executors.py
```diff
@@ -21,6 +21,7 @@
 import nemo_run as run
 from nemo_run.config import get_nemorun_home, set_nemorun_home
 from nemo_run.core.execution.launcher import SlurmTemplate
+from nemo_run.core.execution.xcalibur import XCaliburExecutor
```
sudostock: This will break non-Xcal workloads unless https://github.com/NVIDIA-NeMo/Run/pull/579 is merged first or if user runs with an older nemo-run.

Might want to move this import into the executor definition. It'd make it lazy and not fire unless someone was trying to use it.

scripts/performance/setup_experiment.py
```diff
@@ -587,7 +611,8 @@ def main(
     if export_nsys_sqlite and not enable_nsys:
         logger.warning("--export_nsys_sqlite was set without --enable_nsys; no Nsys SQLite export will be generated.")
 
-    script_name = ENTRYPOINT_BOOTSTRAP
+    # XCalibur uses run_script.py directly; all other executors use bootstrap.py.
+    script_name = "run_script.py" if xcalibur_namespace else ENTRYPOINT_BOOTSTRAP
```
sudostock, line 610: Why this change? The other executors go through this ENTRYPOINT mechanism. I think doing so skips any recipe level env changes would mean jobs run on xcal and other executors are not apples to apples

scripts/performance/setup_experiment.py
sudostock: I think a check for xcal needs to get added to this line as well otherwise this would trigger if save_dir and xcal used.

scripts/performance/utils/executors.py
sjayaram-nv: Made it a lazy import.

scripts/performance/setup_experiment.py
sjayaram-nv: bootstrap.py is a new entrypoint; currently, the container image (nvcr.io/nvidia/nemo:26.04.01) has an older Megatron-Bridge baked into /opt/Megatron-Bridge - it only has run_script.py. This is a temp workaround; will test with nemo container v 26.06 to see if bootstrap.py is available.

scripts/performance/setup_experiment.py
sjayaram-nv: Thank you. Done.

scripts/performance/setup_experiment.py
bdubauski: nvcr.io/nvidia/nemo:26.08.00 mbridge container is available now. Can you update the code to make consistent with other executors?

sjayaram-nv: This is restored back as ENTRYPOINT_BOOTSTRAP for all executors. Resolving.

scripts/performance/utils/executors.py
```diff
@@ -361,3 +361,42 @@ def kubeflow_executor(
         packager=run.GitArchivePackager(include_submodules=True),
     )
     return executor
+
+
+def nvcre_executor(
```
sudostock, line 366: Add docstring and return type to match existing executor pattern

scripts/performance/perf_plugins.py
```diff
@@ -185,6 +189,9 @@ def setup(self, task: Union["run.Partial", "run.Script"], executor: "run.Executo
         if isinstance(executor, SlurmExecutor):
             # NOTE: DO NOT change to f-string, `%q{}` is Slurm placeholder
             launcher.nsys_filename = "profile_%p_%q{SLURM_JOB_ID}_node%q{SLURM_NODEID}_rank%q{SLURM_PROCID}"
+        elif NvcreExecutor is not None and isinstance(executor, NvcreExecutor):
+            # Nvcre pods use PET_* env vars for distributed rank info.
+            launcher.nsys_filename = "profile_node${PET_NODE_RANK}_rank%p"
```
sudostock: Depending on where this line is evaluated it might need to be switched to `%q{PET_NODE_RANK}` `%q{}` is an nsys pattern to interpolate values at runtime

sudostock:
There are no tests

scripts/performance/argument_parser.py
```diff
@@ -681,6 +681,99 @@ def parse_cli_args():
...
+        required=False,
+        default="24h",
+    )
+    nvcre_args.add_argument(
+        "--nvcre_test_scale",
+        type=str,
+        help="Nvcre test scale: 'intra-node', 'intra-rack', or 'full-scale'.",
+        required=False,
+        default=None,
+    )
```
sudostock: Not sure what this flag is or needed for. Can you verify all the flags added, we don't need to expose every NVCRE option. What we expose should be tailored to what we need for our benchmarks and anything a user is likely to need to change.

The NVCRE tool encompasses more things than we use.

scripts/performance/utils/executors.py
```diff
@@ -361,3 +361,42 @@ def kubeflow_executor(
...
+    namespace: str,
+    image: str,
+    num_nodes: int,
+    gpus_per_node: int,
```
sudostock: Align with the established naming in other executors:
`nodes`, `num_gpus_per_node`, `container_image`

scripts/performance/setup_experiment.py
sudostock: Add same nvcre_namespace guard as above

scripts/performance/setup_experiment.py
```diff
@@ -87,10 +87,14 @@ def _filter_run_script_args(argv: List[str]) -> List[str]:
...
+    * ``--offline`` — controls HF Hub access on the launcher node only; offline
+      behaviour in the rank-local script is governed by the ``HF_HUB_OFFLINE``
+      environment variable, not by this flag.
 
-    All of these take a value, passed either as ``--flag value`` (two tokens) or
-    ``--flag=value`` (one token).
+    Value-taking flags are passed as ``--flag value`` or ``--flag=value``.
+    Boolean flags (no following value) are listed in ``_LAUNCHER_ONLY_BOOL``.
     """
+    _LAUNCHER_ONLY_BOOL = {"--offline", "--dryrun"}
```
sudostock: `"-d"` is also the shortened version of the dryrun flag

sjayaram-nv:
> There are no tests

Added tests. But not able to see if it has run as part of the CI pipeline.

scripts/performance/argument_parser.py
sjayaram-nv: Removed just this one - had kept it for future use. All other flags seem relevent.

PR #5262, opened 2026-08-03
fix(model): stop building a DeepSeek query norm HF does not define

### LINKED ISSUE #5261: [bug] DeepSeek MLA with q_lora_rank=null builds a trainable Q-norm the HF architecture does not define
> Edited: earlier revisions of this issue misdiagnosed the parameter and its consequence. The text below is the verified account. Evidence gathered against `main` @ 573e088c9 with the pinned MCore submodule (`3rdparty/Megatron-LM` @ 6513e3e23).

### Problem

For a DeepSeek checkpoint with `q_lora_rank: null`, Bridge builds one parameter per layer that has no counterpart in the HF model:

```
decoder.layers.*.self_attention.linear_q_proj.layer_norm_weight
```

It cannot be loaded, because there is nothing to load it from. Dumping the converted model shows it sitting at its initialization while every neighbouring norm took its value from the checkpoint (norm weights in the source were randomized to `uniform(0.5, 1.5)` so that "loaded" and "left at init" are distinguishable):

```
input_layernorm.weight                              requires_grad=True  min=0.5000 max=1.5000
self_attention.linear_q_proj.layer_norm_weight      requires_grad=True  min=1.0000 max=1.0000
self_attention.linear_kv_up_proj.layer_norm_weight  requires_grad=True  min=0.5625 max=1.4219
mlp.linear_fc1.layer_norm_weight                    requires_grad=True  min=0.5000 max=1.4922
```

`requires_grad=True`. It participates in training.

That is the bug. A converted model should be the model, and this one has an extra per-layer normalization with its own learnable gain that the source architecture does not have.

### Why it happens

When `q_lora_rank is None`, MCore builds a single `linear_q_proj` rather than the query LoRA pair. With `qk_layernorm=True` — which `deepseek_v2_bridge.py:53` and `deepseek_v3_bridge.py:62` set unconditionally — MCore fuses the **Q normalization** into that projection:

```
[TE] q_lora_rank=None -> linear_q_proj=TELayerNormColumnParallelLinear, linear_q_up_proj=IdentityOp
[TE] q_lora_rank=1536 -> linear_q_proj=IdentityOp,                      linear_q_up_proj=TELayerNormColumnParallelLinear
```

(from calling `QKNormConfigResolver(config, submodules).resolve()` directly)

Both lines are the same role in the same resolver, `mla_qk_norm_config.py::_resolve_mla_qk_layernorm`; this repo maps the second to `self_attn.q_a_layernorm.weight` (`models/deepseek/common.py:57`). MCore's local-backend error message names it outright: *"Please use a fused norm+linear for `linear_q_proj` if you intend to have a Q-norm."*

HF defines no such norm. In `modeling_deepseek_v3.py:384-389`, `q_lora_rank is None` builds only `q_proj`, with nothing on the query path.

### Consequences

- **Training runs against the wrong architecture.** The optimizer has a degree of freedom the source model does not have, and the remaining weights co-adapt to it.
- **Export silently discards it.** The parameter has no mapping, so `stream_weights_megatron_to_hf` skips it with a warning. The exported checkpoint is therefore not the model that was trained, and the loss is not confined to that tensor — the weights that adapted around it are still there.
- **Loading crashes**, on the `None` conversion-task slot the missing mapping produces, with an `AttributeError` that names no parameter.

Inference immediately after conversion is *not* affected: with the Q-norm still at its initialization, two tiny `DeepseekV3ForCausalLM` checkpoints differing only in `q_lora_rank` both pass the repo parity gate at TP=1.

| variant | next token | cosine |
|---|---|---|
| `q_lora_rank=64` (control) | match | 0.999930 |
| `q_lora_rank=None` | match | 0.999903 |

That is a statement about conversion, not about training. (Method note: a first attempt at this comparison was invalid — HF `RMSNorm` weights initialize to ones, which makes a redundant normalization idempotent and undetectable. The numbers above are from checkpoints whose norm weights were randomized first. These are also 2-layer randomly-initialized toys and do not speak to drift across a deep trained stack.)

### Why the obvious fixes do not work

Three routes to not building the parameter are closed in this MCore:

1. **`qk_layernorm=False`** also removes the KV norm, and `kv_a_layernorm` exists in every DeepSeek checkpoint:
   ```
   qk_layernorm=True    q_lora_rank=None -> q_proj=TELayerNormColumnParallelLinear, kv_up=TELayerNormColumnParallelLinear
   qk_layernorm=False   q_lora_rank=None -> q_proj=TEColumnParallelLinear,          kv_up=TEColumnParallelLinear
   ```
   `common.py:40` maps `linear_kv_up_proj.layer_norm_weight` to a real HF weight, so turning the flag off trades a phantom parameter for a missing real one. The unconditional `qk_layernorm=True` is the better of the two.
2. **A custom spec with a fused KV projection while QK norm is disabled** is rejected by `_reject_disabled_norm`.
3. **A custom spec with an explicitly non-fused `linear_q_proj`** is overridden back to the fused class by `_mla_fused_linear_or_default` (`mla_qk_norm_config.py:234-238`).

`qk_layernorm` cannot express "KV norm yes, Q norm no", and the spec data is overridden on the way through. What does compose is the resolver call itself: `MLASelfAttention._resolve_qk_norm_config` is an ordinary overridable method returning a plain dict, so a subclass can keep the KV norm and substitute a non-fused `linear_q_proj`. That needs no Megatron-LM change, which matters because upstream `main` is byte-identical to the pinned submodule here — bumping MCore would not help.

### Expected behavior

The parameter should not be created. #5262 does this in the bridge, via the subclass hook above, and also stops the unmapped parameter from crashing the load.

Two follow-ups remain out of scope there:

- **Upstream.** MLA would be cleaner if `qk_layernorm` decoupled Q and KV, or if MCore did not fuse a Q norm into `linear_q_proj` when there is no query LoRA. Then the bridges could derive the flag from the HF config instead of hardcoding it, and the subclass could be dropped.
- **Existing checkpoints.** Anything already trained in Bridge under this configuration contains the extra parameter; loading those into the corrected model is not handled.

### Affected models

Both go through `models/deepseek/common.py`, so both bridges are affected:

| Model | Bridge | Layers | `input_layernorm` keys | `q_proj` keys | `q_a_proj` / `q_a_layernorm` |
|---|---|---|---|---|---|
| `deepseek-ai/DeepSeek-V2-Lite` | `deepseek_v2_bridge` | 27 | 27 | 27 | 0 / 0 |
| `kakaocorp/kanana-2-30b-a3b-thinking` | `deepseek_v3_bridge` | 48 | 48 | 48 | 0 / 0 |

Counts from each repo's `model.safetensors.index.json`. DeepSeek-V2 236B and DeepSeek-V3 both set `q_lora_rank`, so the headline models never exercise this path.

(`deepseek-ai/DeepSeek-V2-Lite` could not be used for the runs above: its `trust_remote_code` modeling file calls `DynamicCache.from_legacy_cache`, which no longer exists in the `transformers` version this repo pins.)

### Minimal repro

```shell
# Spec level, no GPU required.
python - <<'PY'
from types import SimpleNamespace
from megatron.core.models.gpt.gpt_layer_specs import get_gpt_layer_with_transformer_engine_spec
from megatron.core.transformer.mla_qk_norm_config import QKNormConfigResolver

spec = get_gpt_layer_with_transformer_engine_spec(
    num_experts=8, moe_grouped_gemm=False, qk_layernorm=True, multi_latent_attention=True
)
sub = spec.submodules.self_attention.submodules
for q_lora_rank in (None, 1536):
    cfg = SimpleNamespace(
        q_lora_rank=q_lora_rank, qk_layernorm=True, qk_l2_norm=False,
        multi_latent_attention=True, experimental_attention_variant=None,
        normalization="RMSNorm", transformer_impl="transformer_engine",
    )
    r = QKNormConfigResolver(cfg, sub).resolve()
    print(q_lora_rank, "->", r["linear_q_proj"].__name__, r["linear_q_up_proj"].__name__)
PY
```

End to end, building either affected model and calling `load_hf_weights` emits one warning per layer and then raises `AttributeError: 'NoneType' object has no attribute 'megatron_module'`.

### Affected area

area:model

### Regression?

No

### Environment

Branch / commit: `main` @ 573e088c9, MCore submodule @ 6513e3e23

bzantium:
# What does this PR do ?

Fixes #5261

Stop building a per-layer query normalization that the HF DeepSeek architecture does not define, and stop the resulting unmapped parameter from crashing the weight load.

# Bug and root cause

MCore derives Q and KV normalization from a single `qk_layernorm` flag. DeepSeek needs it enabled for `kv_a_layernorm`, which every checkpoint ships, so both bridges set `provider.qk_layernorm = True` unconditionally.

When `q_lora_rank is None` there is no query LoRA, and that same flag makes MCore fuse a **query** normalization into `linear_q_proj` instead:

```
[TE] q_lora_rank=None -> linear_q_proj=TELayerNormColumnParallelLinear, linear_q_up_proj=IdentityOp
[TE] q_lora_rank=1536 -> linear_q_proj=IdentityOp,                      linear_q_up_proj=TELayerNormColumnParallelLinear
```

Both lines come from the same resolver, `mla_qk_norm_config.py::_resolve_mla_qk_layernorm`, and this repo maps the second to `self_attn.q_a_layernorm.weight` (`models/deepseek/common.py:57`). HF defines no query-side norm when there is no query LoRA — `modeling_deepseek_v3.py:384-389` builds a bare `q_proj` — so nothing can be loaded into the first.

Dumping the converted model showed the parameter sitting at its init while every neighbouring norm took its checkpoint value (source norms randomized to `uniform(0.5, 1.5)` to tell the two apart):

```
input_layernorm.weight                              requires_grad=True  min=0.5000 max=1.5000
self_attention.linear_q_proj.layer_norm_weight      requires_grad=True  min=1.0000 max=1.0000
self_attention.linear_kv_up_proj.layer_norm_weight  requires_grad=True  min=0.5625 max=1.4219
mlp.linear_fc1.layer_norm_weight                    requires_grad=True  min=0.5000 max=1.4922
```

`requires_grad=True`, so training optimized a degree of freedom the source model does not have, and export dropped it silently. On top of that the missing mapping produced a `None` conversion-task slot, which the load path dereferenced.

# Changelog

- Add `MLASelfAttentionWithoutQueryNorm`, which keeps the KV norm and drops the query norm when `q_lora_rank is None`, plus `get_deepseek_decoder_block_spec` to install it.
- Point both DeepSeek bridges at that spec builder instead of `get_gpt_decoder_block_spec`.
- Mirror `get_gpt_decoder_block_spec` parameter for parameter, including `vp_stage` and `pp_rank`. `GPTModelProvider.provide()` inspects the callable's signature and only forwards `vp_stage` when it is declared, so omitting it routed interleaved pipeline parallelism into the layer-offset helper without a virtual stage.
- Share the attention substitution as `replace_mla_self_attention` and add a `mtp_layer_spec_transform` provider field, so a standalone MTP stage applies it on the fallback path where `mtp_block_spec` re-derives the layer spec from MCore.
- Require Transformer Engine when there is no query LoRA, with a message that names the setting at fault instead of surfacing an MCore internal error.
- Raise on a Megatron parameter with no mapping at all, naming the parameter, instead of warning and continuing.
- Add unit tests for each of the above, and a `q_lora_rank=None` state-dict round trip in the DeepSeek functional group. The existing toy fixture uses `q_lora_rank=512`, which is the unaffected control, so nothing in the suite reached the changed branch.

## Scope of the `None` task slot

`build_conversion_tasks` has two producers of a `None` slot, and they are not the same
condition. A Megatron parameter with no registry entry is a wrong model: the parameter
stays at its initial value on import and is dropped on export. That now raises by name and
points at `_HCAlphaSecondaryMapping` as the way to declare a deliberate no-op.

The second producer is an HF checkpoint that does not carry a weight the mapping names. It
is benign and pre-existing, so the consumer-side skips stay. Removing them would turn a
missing optional weight into a crash in shared conversion code for every model family.

## Backend support

The local backend cannot express this architecture, and could not before this PR.
`_resolve_mla_qk_layernorm` builds `linear_q_proj` from the backend's fused norm+linear
implementation whenever `q_lora_rank is None`, and `LocalSpecProvider` has none, so
`_require_linear` raises. Keeping the query norm instead trips `_raise_unused_q_norm`. Both
branches are closed, and DeepSeek needs `qk_layernorm` on for `kv_a_layernorm`.

Reproducing MCore's resolver inside the bridge to work around this would fork logic that
belongs upstream, so the bridge fails early with an explicit Transformer Engine
requirement.

`qk_layernorm=False` is not an alternative: it also removes `linear_kv_up_proj.layer_norm_weight`, which `common.py:40` maps to a real HF weight. Two spec-level workarounds are closed as well — `_reject_disabled_norm` rejects a fused KV projection while QK norm is disabled, and `_mla_fused_linear_or_default` overrides an explicitly non-fused `linear_q_proj` back to the fused class. Overriding `_resolve_qk_norm_config` on the attention module is the one hook that composes, and it needs no Megatron-LM change (upstream `main` is byte-identical to the pinned submodule here).

# Validation

Model construction, TP=1, tiny `DeepseekV3ForCausalLM` checkpoints:

| checkpoint | `linear_q_proj` norm | `linear_kv_up_proj` norm | MTP params | MTP `q_proj` norm |
|---|---|---|---|---|
| `q_lora_rank=None` | **0** (was 2) | 2 | – | – |
| `q_lora_rank=64` (control) | 0 | 2 | – | – |
| `q_lora_rank=None`, MTP=1 | **0** | 3 | 22 | **0** |

No `No mapping found for megatron_param` warnings remain. The MTP row above is the shared-stage layout, where `mtp_block_spec` reuses the corrected decoder spec. A standalone MTP stage owns no decoder layers and re-derives its spec straight from MCore, which is why the substitution is also registered as `mtp_layer_spec_transform` rather than relying on the shared spec.

Logit parity, `compare_hf_and_megatron/compare.py`, TP=1:

| variant | next token | cosine |
|---|---|---|
| `q_lora_rank=64` (control) | match | 0.999930 |
| `q_lora_rank=None`, before this PR | match | 0.999903 |
| `q_lora_rank=None`, after this PR | match | 0.999909 |

Removing the parameter does not move inference. The defect was never an inference-time divergence — it was a phantom trainable parameter — and an earlier revision of #5261 that claimed otherwise has been corrected.

Conversion-loop regression, against the parent commit:

```text
python -m pytest tests/unit_tests/models/test_model_bridge.py -q -k skips_unmapped_task_slots
```

- Before: `AttributeError: 'NoneType' object has no attribute 'megatron_module'` and `... 'param_weight'`, 2 failed.
- After: 2 passed.

Unit tests:

```text
python -m pytest tests/unit_tests/models/deepseek/ tests/unit_tests/models/test_model_bridge.py -q
94 passed
```

Run against the Megatron-Core commit pinned by this branch, not the submodule checkout,
so the QK-norm resolver under discussion is the one exercised.

The whole of `tests/unit_tests/models/` was run on this branch and on the parent commit in
the same environment, to check that raising on an unmapped parameter does not disturb other
model families:

| | failed | passed | skipped | errors |
|---|---|---|---|---|
| parent `bd25f3ed3` | 66 | 2800 | 36 | 206 |
| this branch | 66 | 2809 | 36 | 206 |

Same failures and errors, nine more passes, which are the tests added here. The pre-existing
failures are environment-related and unrelated to this change.

State-dict round trip, one GPU:

```text
python -m pytest tests/functional_tests/test_groups/models/deepseek/test_deepseek_conversion.py -k round_trip -q
1 passed
```

HF has `q_proj.weight` and `kv_a_layernorm.weight` and no query LoRA. Megatron has
`linear_q_proj.weight` and `linear_kv_up_proj.layer_norm_weight` and no
`linear_q_proj.layer_norm_weight`. Every original weight survives HF to Megatron to HF
bit-exactly, and no `q_a_proj` or `q_a_layernorm` key is invented. The structural
expectations come from the HF architecture, not from the conversion registry.

Coverage note: the unit tests reach the resolver and both affected bridges, but no test loads a real affected checkpoint. `kakaocorp/kanana-2-30b-a3b-thinking` is too large for the 2-GPU functional budget, and `deepseek-ai/DeepSeek-V2-Lite` cannot be loaded at all under the pinned `transformers` — its `trust_remote_code` modeling file calls `DynamicCache.from_legacy_cache`, which no longer exists. The model-construction and parity evidence above therefore comes from tiny `DeepseekV3ForCausalLM` checkpoints built to the same shape (`q_proj` present, `q_a_proj` / `q_a_layernorm` absent) rather than from CI. The round trip is now committed as `TestDeepSeekWithoutQueryLoRA` on a generated toy checkpoint, so the changed branch has end-to-end coverage even though no real affected checkpoint fits the budget.

Static checks: `ruff check src tests` passed, `ruff format --check` clean on changed files, `git diff --check` passed.

Environment caveat: these runs used a CUDA 12.9 / A100 container built from `uv.lock`, not the CI `megatron-bridge` image — `nvidia-resiliency-ext` publishes glibc 2.39 wheels only (this host is glibc 2.35), the CUDA source-build extras do not compile here, and Transformer Engine had to be built from the pinned commit with `NVTE_WITH_NCCL_EP=0` for `sm80`. `gradient_accumulation_fusion` was disabled for the parity runs because APEX is absent; it selects the weight-gradient kernel and does not affect the forward pass. The toy checkpoints are 2-layer and randomly initialized, so they do not speak to drift across a deep trained stack. CI on the real image remains the authoritative run.

# Scope

- No public API, CLI, config, dependency, lockfile, workflow, or MCore changes.
- Model construction changes only for MLA with `q_lora_rank is None`; the `q_lora_rank` path resolves identically to before.
- Conversion behaviour changes for one condition: a Megatron parameter with no registry entry now raises by name instead of warning and continuing. `_HCAlphaSecondaryMapping` in `deepseek_v4_bridge.py` is the existing way to declare a deliberate no-op. The separate `No mapping found for global_name` warning that `test_fp8_param_export.py` asserts is a different site and is untouched, as are the consumer-side skips for a weight the HF checkpoint does not carry.
- Checkpoints trained before this change still contain the extra parameter; loading them into the corrected model is not addressed here.

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Fixes #5261


src/megatron/bridge/models/deepseek/attention.py
```diff
@@ -0,0 +1,66 @@
...
+    def _resolve_qk_norm_config(self, submodules):
+        """Replace the fused query projection with a plain one when there is no query LoRA."""
+        layer_classes = super()._resolve_qk_norm_config(submodules)
+        if self.config.q_lora_rank is None:
+            backend = get_backend(self.config.transformer_impl)
+            layer_classes["linear_q_proj"] = backend.column_parallel_linear()
+        return layer_classes
+
+
+def get_deepseek_decoder_block_spec(config, *, use_transformer_engine: bool) -> ModuleSpec:
```
yaoyu-33: [P1] Preserve the virtual-pipeline arguments in this wrapper

This is not a drop-in replacement for `get_gpt_decoder_block_spec`: the MCore builder also accepts `vp_stage` and `pp_rank`. `GPTModelProvider.provide()` checks the callable signature and only forwards `vp_stage` when that parameter is present. Because this wrapper omits it, interleaved PP calls the pinned MCore `get_transformer_layer_offset(..., vp_stage=None)`, which asserts that `vp_stage` must be supplied when `virtual_pipeline_model_parallel_size` is set. Please preserve and forward `vp_stage`/`pp_rank` (and preferably the full builder contract), and add a focused test that builds distinct virtual stages or at least verifies their layer slices.

src/megatron/bridge/models/conversion/model_bridge.py
```diff
@@ -1194,6 +1194,13 @@ def load_weights_hf_to_megatron(
 
         _hf_import_cache: Dict[str, torch.Tensor] = {}
         for task in self._with_progress_tracking(hf_to_megatron_tasks, description):
+            # build_conversion_tasks returns List[None | WeightConversionTask]; a slot stays None
+            # when no mapping matched that global parameter, which is expected whenever the
+            # Megatron model owns a parameter the HF architecture has no counterpart for. It
+            # already warned by name, so skip instead of raising an AttributeError that names
+            # nothing.
+            if task is None:
```
yaoyu-33: [P1] Do not silently skip an unmapped real parameter

Here `None` is not the existing “parameter lives on another PP rank” task (that has a real task with `megatron_module=None`); it means `mapping_registry.megatron_to_hf_lookup(global_name)` failed for an actual global Megatron parameter. Continuing leaves that trainable parameter at initialization on HF import and omits it on HF export, changing a hard failure into an incomplete conversion for every model family. The DeepSeek architecture fix should remove the phantom parameter, so it should also remove the `None` slot instead of teaching shared conversion to tolerate it. Please retain a named hard failure (or an explicit, narrowly allowlisted skip contract) and test an incomplete registry through `build_conversion_tasks` to assert that conversion fails with the missing parameter name. The current tests manually inject `[None, mapped_task]`, so they assert the proposed loop behavior without independently proving state-dict/task completeness; other task consumers also still assume every entry is a `WeightConversionTask`. `_HCAlphaSecondaryMapping` demonstrates the existing explicit pattern: it registers a concrete no-op export mapping precisely to avoid an unmapped task.

tests/unit_tests/models/deepseek/test_deepseek_attention.py
```diff
@@ -0,0 +1,122 @@
...
+            n_shared_experts=1,
+            q_lora_rank=None,
+        )
+
+        bridge_cls().provider_bridge(hf_pretrained)
+
+        assert provider.qk_layernorm is True, "the KV norm still has to be requested"
+        assert provider.transformer_layer_spec.func is get_deepseek_decoder_block_spec
+
+    def test_spec_helper_is_a_drop_in_for_the_mcore_builder(self):
```
yaoyu-33: [P1] Exercise the affected state dictionaries, not only the resolver registry

These tests only use `SimpleNamespace`/spec objects and class-name assertions. They never instantiate an HF or Megatron model in the affected branch, and the existing DeepSeek functional fixture sets `q_lora_rank=512`, which is the unaffected control. Please add a tiny `q_lora_rank=None` conversion test with independently expected structure: HF has `q_proj.weight` and `kv_a_layernorm.weight` but no `q_a_proj`/`q_a_layernorm`; Megatron has `linear_q_proj.weight` and `linear_kv_up_proj.layer_norm_weight` but no `linear_q_proj.layer_norm_weight`. An exact HF→Megatron→HF state-dict round trip (plus MTP coverage if feasible) would catch both incorrect construction and missing task slots without making the conversion registry its own oracle.

src/megatron/bridge/models/deepseek/attention.py
```diff
@@ -0,0 +1,66 @@
...
+    builds a bare ``q_proj``.
+
+    The result is a trainable parameter with no HF counterpart, which cannot be loaded
+    and is silently dropped on export. This subclass keeps the KV norm and drops the
+    query norm so the converted model matches the source architecture.
+    """
+
+    def _resolve_qk_norm_config(self, submodules):
+        """Replace the fused query projection with a plain one when there is no query LoRA."""
+        layer_classes = super()._resolve_qk_norm_config(submodules)
```
yaoyu-33, line 75: [P1] Handle the optional non-TE backend before the parent resolver rejects it

Both DeepSeek bridges pass `use_transformer_engine=HAVE_TE`, and Transformer Engine is an optional extra. With `HAVE_TE=False`, pinned MCore `get_gpt_layer_local_spec` supplies standalone `q_layernorm` and `kv_layernorm` modules. The `super()` call here enters `QKNormConfigResolver._reject_common_spec_conflicts`, sees a non-trivial Q norm with `q_lora_rank=None`, and raises before line 43 can replace the projection. Thus the affected checkpoints still cannot be constructed with the supported local fallback. Please either resolve the KV-only local case explicitly (plain Q projection + identity Q norm while retaining the KV norm) or fail early with an intentional TE requirement, and add a local-backend test.

src/megatron/bridge/models/deepseek/attention.py
```diff
@@ -0,0 +1,66 @@
...
+
+    Returns:
+        The decoder block spec, with MLA self-attention replaced by
+        :class:`MLASelfAttentionWithoutQueryNorm` when there is no query LoRA.
+    """
+    spec = get_gpt_decoder_block_spec(config, use_transformer_engine=use_transformer_engine)
+    if getattr(config, "q_lora_rank", None) is not None:
+        return spec
+
+    for layer_spec in spec.layer_specs:
```
yaoyu-33: [P2] Keep the override when MTP is on a standalone pipeline stage

`gpt_provider.mtp_block_spec()` reuses this corrected block only while `spec.layer_specs` is non-empty. The repository explicitly supports standalone MTP stages via `set_deepseek_v3_pipeline_model_parallel_layout(..., mtp_standalone=True)`; on that stage the provider sees zero decoder layers and falls back to raw MCore `get_gpt_decoder_layer_specs()`, whose attention module is never replaced here. For `q_lora_rank=None`, the MTP layer therefore regains `linear_q_proj.layer_norm_weight`. Please preserve the DeepSeek attention override in that fallback (for example with a model-specific MTP spec) and cover a Q-LoRA-null standalone-MTP layout, rather than only TP=1 MTP where a decoder spec is available.

src/megatron/bridge/models/deepseek/attention.py
yaoyu-33: Confirmed in the PR checkout against the pinned MCore runtime: constructing this spec with PP=2 and VPP=2 reaches the exact `vp_stage must be provided if virtual pipeline model parallel size is set` assertion. The targeted DeepSeek/model-bridge tests pass because none exercise the provider call with virtual-stage routing.

tests/unit_tests/models/deepseek/test_deepseek_attention.py
yaoyu-33: Independent focused validation at this exact head now confirms the core TE/TP=1/no-MTP path: a tiny real `q_lora_rank=None` HF model had `q_proj.weight` and `kv_a_layernorm.weight` but no Q-LoRA parameters; the constructed MCore model had `linear_q_proj.weight` and `linear_kv_up_proj.layer_norm_weight` but no query `layer_norm_weight`; `build_conversion_tasks` contained no `None` slots; and those two weights survived HF→MCore→HF bit-exactly. This is positive evidence for the central substitution, but it is not committed coverage and does not exercise the separate VPP, local-backend, or standalone-MTP failures, so the requested focused regression test remains necessary.

yaoyu-33 (changes requested):
The central TE/TP=1/no-MTP query-norm substitution is independently validated, including a bit-exact affected-branch HF→MCore→HF round trip. I am requesting changes for the remaining P1 issues: preserve and test virtual-stage routing; retain strict failure for genuinely unmapped global parameters instead of silently skipping them; support or intentionally gate the optional local backend; and commit a real affected state-dict/round-trip regression test. The standalone-MTP fallback is also still open as P2. Please update the linked inline threads and request re-review on the exact new head; CI approval should wait for that revision.

src/megatron/bridge/models/deepseek/attention.py
yaoyu-33: Confirmed against the exact PR head and pinned MCore runtime: when the callable decoder block returns an empty `layer_specs` list (the standalone-MTP stage condition), `mtp_block_spec` takes its raw `get_gpt_decoder_layer_specs` fallback and passes an `MLASelfAttention` layer—not `MLASelfAttentionWithoutQueryNorm`—into the MTP builder. The DeepSeek override is concretely lost on this branch.

src/megatron/bridge/models/deepseek/attention.py
yaoyu-33: Confirmed against the exact PR head and pinned MCore runtime using the real local MLA spec: `_resolve_qk_norm_config` raises before the subclass replacement with the resolver error that `q_layernorm` is non-trivial while `q_lora_rank is None`. This is therefore a runtime construction failure, not only a static inference.

src/megatron/bridge/models/deepseek/attention.py
bzantium: Fixed. The wrapper now mirrors `get_gpt_decoder_block_spec` exactly and forwards both arguments.

Added a signature-parity check against MCore so a future parameter fails loudly here, and a test on the `partial` the bridges actually hand to the provider.

src/megatron/bridge/models/conversion/model_bridge.py
bzantium: Both sites now raise with the parameter name, and a test pushes an incomplete registry through `build_conversion_tasks`.

I kept the consumer-side skips though. A `None` slot also appears when the mapping exists but the HF checkpoint simply doesn't carry that weight (`model_bridge.py:1847-1861`), which is benign and pre-existing. Removing the skips would turn that into a crash for every family. So: no mapping raises, missing HF weight still skips.

Ran the whole of `tests/unit_tests/models/` against the parent commit to be sure nothing else leaned on the old behaviour. No new failures.

tests/unit_tests/models/deepseek/test_deepseek_attention.py
bzantium: Added. You were right that nothing reached this branch, since the existing fixture is `q_lora_rank=512`.

`TestDeepSeekWithoutQueryLoRA` generates a `q_lora_rank=None` toy model, checks the structure you listed, and asserts a bit-exact HF to Megatron to HF round trip with nothing dropped and no `q_a_*` invented. The expectations come from the HF architecture, not the registry.

It builds through `to_megatron_model`, since `get_model` rejects the MLA fields for this family today. MTP is still spec-level only.

src/megatron/bridge/models/deepseek/attention.py
bzantium: Went with your second option, an explicit TE requirement.

The KV-only local case can't be resolved without reimplementing the resolver. With no query LoRA it always wants the fused norm+linear for `linear_q_proj`, and the local backend has none, so clearing `q_layernorm` just swaps `_raise_unused_q_norm` for `_require_linear`. Both doors are shut, and `qk_layernorm` has to stay on for `kv_a_layernorm`.

The bridge now fails early with a message naming `transformer_impl`, with a test against the real local spec.

src/megatron/bridge/models/deepseek/attention.py
bzantium: Fixed. The substitution is now `replace_mla_self_attention`, shared between the block builder and MTP, and both bridges register it as `mtp_layer_spec_transform` so `mtp_block_spec` applies it on the empty-`layer_specs` fallback.

Tests cover the bare layer spec, the q-LoRA path staying untouched, and both bridges registering it.

bzantium:
All five threads are answered on the new head, b67b8d9, and the description is updated to match.

Four are implemented as asked. The local backend is the exception: MCore has no non-fused path for `linear_q_proj` when there is no query LoRA, so I took your second option and gated it with an explicit Transformer Engine requirement rather than reimplementing the resolver in the bridge. Details in that thread.

src/megatron/bridge/models/conversion/model_bridge.py
yaoyu-33: Implemented and pushed as `ee9f7d9f4`.

`build_conversion_tasks` now validates the complete gathered Megatron parameter-name list before rank-local task construction, so owning and non-owning PP ranks fail consistently when a mapping is missing. A mapped HF key that is absent now raises with both the Megatron and HF names unless the mapping explicitly opts into `allow_hf_name_mismatch`. The normal and FP8 builders return concrete `WeightConversionTask` objects only, and the downstream `None` filters (including MIMO) are removed.

Focused coverage now includes owning/non-owning PP missing mappings, missing HF checkpoint keys, the explicit mismatch opt-in, and remote-PP FP8 tasks. Validation: 270 focused unit tests passed, pre-commit passed, and a tiny real `q_lora_rank=None` HF→MCore→HF GPU round trip retained the independently expected state-dict structure and preserved the query projection and KV norm bit-exactly.

@bzantium, could you review this additive commit?

yaoyu-33:
@bzantium worked with agent to re-write a bit for this part, plz review.

bzantium:
Reviewed ee9f7d9, looks good. Validating the gathered list up front also fixes something my version got wrong: it only raised on ranks that owned the parameter. Left the rest in the inline thread.

Merging main then surfaced a conflict with the new invariant. #5374 made the modelopt export path tolerate `None` slots, and its test asserted `build_conversion_tasks` leaves one for an unmapped `*_amax` buffer, which now raises instead.

Resolved by keeping the guard and retargeting the test. The defect #5374 fixed was the plan builder truncating after a hole, and that guard still belongs there since the signature accepts `WeightConversionTask | None`. Only the half that asserted the builder produces the hole is gone; the sparse list is now constructed in the test. Renamed to `test_modelopt_plan_keeps_tasks_after_a_sparse_slot`.

Merge is pushed. The rest of it was the `scale_block_size` argument you'd expect, and the MCore pin moved with main.

src/megatron/bridge/models/conversion/model_bridge.py
bzantium: Pushed ff91efc7 for two nits here. The flag's comment in `param_mapping.py` only described the name-mismatch case, and the new error only pointed at synthesized names. Both now also mention gating the mapping on the config, which is the better option when the config settles it.

yaoyu-33 (dismissed):
Approved at exact head 96c66bd6df764774b7ff844ed72d3236bc824d32. The DeepSeek q_lora_rank=None structure, TE gate, VPP/MTP routing, strict mapping contract, focused unit coverage, and bit-exact GPU round trip have been validated; the author also reviewed the maintainer follow-up.

kamran-nvidia:
/ok to test [49bb737](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5262/commits/49bb737ec8ad73a9e6debb7de0ee8eabe439c0dc)

kamran-nvidia:
/ok to test [b6a6612](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5262/commits/b6a66123a148c01553597afa958671b5ef59277c)

bzantium:
Four failures on b6a6612, and three of them are ours. All four pass on main in the last scheduled run, so none of this is pre-existing.

Exaone and nemotronh hit the new validation from ee9f7d9. Exaone registers MTP mappings unconditionally and the toy checkpoint has no MTP weights, so it reports 15 missing `mtp.layers.0.*` entries. Nemotronh reports one, `embedding.word_embeddings.weight -> backbone.embeddings.weight`. Both look like the case your error message already describes, where the HF config should gate the registration.

qwen_audio raises the same ValueError but reads differently: the entire `language_model.*` set comes back missing, `final_layernorm`, `embed_tokens`, every MLP weight. That is too much to be unconditional registration, so I suspect the validation is not resolving the VLM name prefix. Probably worth its own look before we assume it is a bridge-side fix.

gemma_vl is unrelated. EADDRINUSE on port 52177, no mapping error anywhere in the log.

The check is finding something real, exaone genuinely maps weights that are not in the checkpoint. But it makes a previously tolerated condition fatal for at least three families at once, and each one needs its own bridge change. Would you rather split ee9f7d9 out? The DeepSeek fix here is independent and its jobs are green, so #5261 could land while the validation goes in with the conditional registrations it needs. Either way I am happy to do the work.

bzantium:
Correcting my triage above. I called exaone and nemotronh false positives of the new validation. They are not. Both are real defects the check exposed, and they are the same defect this PR is about: a Megatron parameter with nothing in the checkpoint to load into it.

Exaone is fixed in d3af42b. `ExaoneMoeForCausalLM` declares `_keys_to_ignore_on_load_unexpected = [r"mtp.*"]`, so released checkpoints carry MTP weights while the modelling class does not build them. The toy is constructed from that class with `num_nextn_predict_layers: 1`, so it saves no MTP weights at all, while the provider still builds MTP layers and the bridge still registers the 15 MTP mappings off the same config value. The roundtrip was comparing layers that had never been loaded. I dropped MTP from the toy config so it represents something the class can produce. Verified by running the roundtrip the test drives at TP=2 on two A100s: 89/89 weights matched, no validation error.

nemotronh I could not pin down, so I am not proposing a fix. The mapping is right: `backbone.embeddings.weight` is in the released `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16` index, alongside `lm_head.weight`, and `tie_word_embeddings` is false. What is reproducible is that the toy fixture saves a checkpoint without it. I built the toy exactly as the fixture does and checked the saved keys, and `backbone.embeddings.weight` is absent every time while `lm_head.weight` survives. It is not the `_fix_tied_weights_keys` shim: mapping the list to itself, to `backbone.embeddings.weight`, or to an empty dict all drop it, and the two parameters do not share storage after construction. So the toy is not representative of the model, and before this validation the embedding was simply never loaded and nobody noticed. Someone who knows that fixture will find it faster than I will.

qwen_audio still looks different from these two. The entire `language_model.*` set comes back missing, which is too much to be a registration gap, so I would look at whether the validation resolves that prefix before assuming it is a bridge fix. gemma_vl remains unrelated, EADDRINUSE on port 52177.

Given that, I withdraw the suggestion to split ee9f7d9 out. It is finding real problems and it belongs here. The open question is just whether nemotronh and qwen_audio get fixed in this PR or tracked separately, and I would rather not guess at fixes for model families I do not own.

bzantium:
The three functional failures turned out to be the same problem in different places, and ee9f7d9 was right to flag all of them. Fixed in abd5d21.

Each of those toys is built from a HF class whose current save layout no longer matches the released checkpoint the bridge maps, so the mapped weights were not in the file at all and the runs were comparing parameters that had never been loaded. That is what the new check exists to catch, so I fixed the toys rather than the check.

exaone was the clearest of them. `ExaoneMoeForCausalLM` declares `_keys_to_ignore_on_load_unexpected = [r"mtp.*"]`, so released checkpoints carry MTP weights but the class never builds them. The toy config set `num_nextn_predict_layers`, which made the provider build MTP layers and the bridge register 15 MTP mappings against weights the toy could not possibly save, so I dropped MTP from the toy.

The other two are naming. nemotronh saves `backbone.embedding.weight` while the released Nemotron-3-Nano index has `backbone.embeddings.weight`, and qwen_audio nests the language model one level deeper than `Qwen/Qwen2-Audio-7B-Instruct` does, `language_model.model.model.*` instead of `language_model.model.*`. Both are corrected right after save. The audio tower, projector and head were already fine.

I ran what I could locally on two A100s. qwen_audio passes outright, exaone's roundtrip matches 89/89 weights, and nemotronh clears the validation and then stops on the APEX fused wgrad extension, which my container lacks and CI has. gemma_vl only failed in that one run, with EADDRINUSE on port 52177 and no mapping error anywhere in the log, so I left it alone.

What I did not touch is that this mismatch is not only a test artifact. Anyone who re-saves a Nemotron-H or Qwen2-Audio checkpoint with the pinned transformers ends up with a file these bridges cannot map, and before ee9f7d9 it would have loaded quietly with those weights left at their initialization. Whether the bridges should accept both layouts felt bigger than this PR and not really mine to decide for those families, but I can pick it up if you would rather have it here.

adityavavreNVDA:
/ok to test [bdbe7eb](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5262/commits/bdbe7ebe47b026a76116501579edc1ba3ccf2c29)

bzantium:
Everything is green now, including the four jobs that were failing before. The exaone, nemotronh and qwen_audio fixes went in with the last branch update, and gemma_vl passed on the rerun, so that port collision was a one off.

@yaoyu-33 @kamran-nvidia @adityavavreNVDA this is only waiting on review at this point. The earlier review shows as dismissed, so it needs another pass from whoever has a moment. Thanks for keeping the branch current and running CI while I was chasing down the fixtures.

yaoyu-33 (approved):

PR #5250, opened 2026-08-02
chore(deps): AUT-1290 bump MCore main and validate dev

svcnemo-autobot:
## Background

Advance Bridge to MCore main while validating the same payload against MCore dev in #5251.

## What changed

- Merge current Bridge `main` into the shared compatibility payload.
- Apply dual-compatible checkpoint, distillation, and FSDP handling.
- Pin immutable dependency sources and independently refresh the main lock.

## Details

```mermaid
flowchart LR
  U[Bridge main] --> C[Compatibility 8b94495f]
  C --> M[Main lane]
  C --> D[Dev validation]
  M --> L[Only landing PR]
```

Both lanes carry `8b94495f94db86bb2e228313a2e3521540dff458`. Only this PR may merge after both lanes pass; #5251 must never merge.

## Tested

Exact lane deltas, signed DCO history, gitlinks, independent revision-2 locks, and immutable git sources verified. Independent convergence review passed. Corrected lane image build is running.

PR #5278, opened 2026-08-03
fix(fsdp): support MCore FullyShardedDataParallel factory in isinstance checks

cuichenx:
# What does this PR do ?

Restore Megatron-Bridge compatibility with newer Megatron-Core, where `mcore_fsdp_adapter.FullyShardedDataParallel` became a **factory function** (dispatching to V1/V2) instead of a class. Passing that symbol into `isinstance()` now raises `TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union`, which fails **every** mbridge-in-mcore CI run (both unit and functional suites) through `unwrap_model`.

Root cause landed in Megatron-Core via NVIDIA/Megatron-LM#5865 ("Add MFSDP v2 MCore integration with FullyShardedOptimizer"). The factory's own docstring states: *"Use the explicit V1 or V2 implementation classes for type checks."*

# Changelog

- `models/conversion/utils.py` (`unwrap_model`): build the `isinstance` tuple from the concrete FSDP wrapper **classes** — `FullyShardedDataParallelV1`/`V2` when present, or the legacy `FullyShardedDataParallel` class on older Core — via `getattr` + `isinstance(cls, type)` filtering. The factory function is excluded automatically (it is not a type). This unblocks all failing CI jobs, which route through here (e.g. `model_bridge.py:load_weights_hf_to_megatron → unwrap_model`).
- `training/train.py`: fix the latent sibling call site (`isinstance(model_chunk, megatron_FSDP)`, manual FSDP buffer registration) by moving its import to the same `try FullyShardedDataParallelV1 except FullyShardedDataParallel` guard already established in `training/setup.py`.

# GitHub Actions CI

See the CI section in the Contributing doc.

**Pre checks**:

- [ ] Did you write any new necessary tests? — existing unit + functional coverage (`test_train.py::TestFSDPRegistration`, conversion tests) already exercises `unwrap_model`; this restores them to green.
- [ ] Did you add or update any necessary documentation? — n/a
- [x] Does the PR affect components that are optional to install? — the newer-vs-older Core symbols are handled with import guards / `getattr` fallbacks.

# Additional Information

- Verified against the four failing mcore CI runs (PRs #6166, #6201, #6177, and the scheduled `main` run) — all share this single root cause.
- Follow-up (out of scope): `train.py` and `setup.py` type-check only against V1 (`megatron_FSDP`). If Core's factory ever dispatches `FullyShardedDataParallelV2` instances at runtime, those two sites would miss them. Consider a shared helper mirroring the `unwrap_model` tuple so all three sites stay aligned.

### DIFF
--- a/src/megatron/bridge/models/conversion/utils.py
+++ b/src/megatron/bridge/models/conversion/utils.py
@@ -46,13 +46,24 @@ def unwrap_model(model, module_instances=None):
     if module_instances is None:
         from megatron.core.distributed import DistributedDataParallel as DDP
         from megatron.core.distributed import TorchFullyShardedDataParallel as torch_FSDP
-        from megatron.core.distributed.fsdp.mcore_fsdp_adapter import (
-            FullyShardedDataParallel as megatron_FSDP,
-        )
+        from megatron.core.distributed.fsdp import mcore_fsdp_adapter
         from megatron.core.distributed.fsdp.src.megatron_fsdp.megatron_fsdp import MegatronFSDP
         from megatron.core.transformer.module import Float16Module
 
-        module_instances = (DDP, torch_FSDP, megatron_FSDP, Float16Module, MegatronFSDP)
+        # ``FullyShardedDataParallel`` is a class in older Megatron-Core but a factory function
+        # dispatching to V1/V2 in newer versions, so gather only the concrete wrapper *classes*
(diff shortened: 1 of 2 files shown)


cuichenx:
duplicate of https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5250

PR #5283, opened 2026-08-03
fix(training): surface distributed failures without cleanup hangs

yaoyu-33:
## Summary

- prove that the Gemma4-VL SFT stall is not caused by a pipeline boundary shape or dtype mismatch
- log the original rank-local exception and traceback before exceptional cleanup begins
- abort framework-owned process groups during exceptional cleanup so a rank-local failure is surfaced instead of hidden behind outstanding collectives
- keep Gemma4-VL SFT unverified because the full-dimension workload still exceeds H100 memory
- correct the inference card leaf to unverified because the recorded run stopped after 29 of 32 requested tokens

## Root cause

A tiny structural Gemma4-VL proxy completed three optimizer steps for PP1, PP1/EP2, PP2 multimodal, and PP2 image-free controls. The PP2 path sent and received matching BF16 boundary tensors, all ranks consumed compatible metadata, and selective-recompute replay used the same shapes as forward.

A full-dimension random-initialized TP4/PP2/EP1 diagnostic preserved sequence length 4096, image expansion, frozen vision, selective recompute, 128 MoE experts, and GBS32. It completed step 1, then all first-stage ranks failed during step-2 microbatch 34 with the same 352 MiB CUDA OOM in the MoE combine reduce-scatter as NVBug 6491103. Second-stage ranks were waiting in pipeline send/receive. The apparent hang came from failed ranks entering orderly process-group destruction while peers still had outstanding collective work.

The fix changes only ordinary exceptional cleanup. Bridge now logs the active exception with rank context and its full traceback before cleanup, then uses the PyTorch process-group abort path for framework-owned distributed resources. Cleanup errors remain diagnostic-only and cannot replace the original exception. The in-process restart wrapper retains cleanup ownership and does not use this fatal path.

The same full-dimension diagnostic now reports the original OOM and terminates all ranks promptly. It does not claim to solve the underlying H100 memory limit.

The exact-checkpoint real-data rerun remains pending because that immutable artifact was not accessible from the available validation environment. No SFT verification status or checkpoint is claimed.

## Verification

- focused pretrain unit tests: 10 passed, including log-before-cleanup/abort ordering, rank/type/message/traceback context, original-exception identity, cleanup-error logging, and in-process restart ownership
- model-card validator: passed
- `uv run --active --no-sync pre-commit run --all-files`: passed
- tiny PP1/PP2 structural matrix: four configurations completed 3/3 optimizer steps
- full-dimension before fix: step 1 completed, step 2 OOM was masked by distributed cleanup
- full-dimension after abort fix: the same OOM was printed and all ranks exited promptly
- small PP2 blocked-peer failure diagnostic after the logging follow-up: the injected `torch.OutOfMemoryError` was logged with rank 0 and its full traceback before abort; both ranks exited about four seconds after injection, inside the 150-second fail-fast timeout

Strict mypy was also attempted on the changed production file; the repository checkout reported existing package-stub and untyped-decorator errors, with no new diagnostic on the abort call.

### DIFF
--- a/examples/model_verification_cards/gemma-4-26b-a4b-it/card.yaml
+++ b/examples/model_verification_cards/gemma-4-26b-a4b-it/card.yaml
@@ -5,13 +5,16 @@ title: gemma_4_26b_a4b_it
 summary: >
   Performance disclaimer: this model has not been performance-tuned; reported
   timing and throughput metrics are sanity checks, not optimized performance
-  results. Gemma 4 26B-A4B-it BF16 CPU and GPU import, exact CPU and GPU
-  export, and deterministic multimodal inference are verified. Manual forward
-  comparison remains unverified because the recorded run compared a
+  results. Gemma 4 26B-A4B-it BF16 CPU and GPU import and exact CPU and GPU
+  export are verified. Deterministic multimodal inference remains unverified
+  because the recorded run stopped after 29 of the requested 32 tokens. Manual
+  forward comparison remains unverified because the recorded run compared a
   tensor-parallel padding position rather than the prompt's next token. Full
   SFT remains unverified: the current two-node EP8 recipe does not complete its
   first optimizer step, while the supported single-node TP4/PP2/EP1 fallback
-  completes one finite step but stalls during the second. Long-context SFT and
+  completes one finite step before running out of memory during the second.
+  The formerly reported stall was failure cleanup waiting on outstanding
+  collectives, not a pipeline-boundary shape mismatch. Long-context SFT and
   PEFT are pending. The published Bridge recipes do not currently provide VLM
   pretraining.
 verification_index:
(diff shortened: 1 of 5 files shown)


yaoyu-33:
Root cause and fix are now pushed in e5104e593d09b09a61a3393f60dbc493499b64ae.

The PP boundary was ruled out: tiny PP1/PP2 multimodal controls reached three steps with matching BF16 boundary and recompute shapes. A full-dimension TP4/PP2 reproduction then exposed the hidden failure: step-2 microbatch 34 OOMs on the first stage in the 352 MiB MoE combine reduce-scatter. The failed ranks were hanging in Bridge process-group destruction while the second stage waited in pipeline communication.

The fix aborts framework-owned process groups on exceptional cleanup. The same full-dimension workload now prints the original OOM and terminates promptly instead of appearing hung. Focused tests are 9/9 green, the card validator passes, and pre-commit passes. SFT remains unverified because this fix surfaces the memory failure; it does not make the workload fit. The exact-checkpoint rerun is still blocked by artifact availability in the current validation environment.

PR #5292, opened 2026-08-04
fix(quant): wrap post-quantization VLM generation in torch.no_grad()

Batool1227:
## Summary
- `.eval()` disables dropout/etc but not autograd. In both
  `ptq_generate_vlm.py` and `quantize_vlm.py`, the post-quantization
  sanity-check generation call ran with gradient tracking enabled, so every
  decode step retained a full backward-pass graph it never uses.
- On a 27B VLM with TP=8 this was enough to cause a CUDA OOM at
  ~79GB/79.18GB even after model loading and quantization validation had
  already succeeded.
- Also reorders `quantize_vlm.py` to save the quantized checkpoint before
  running the test generation, so the checkpoint lands on disk even if the
  sanity check fails for an unrelated reason (previously it ran test-then-save,
  meaning any failure in the test step lost the quantized checkpoint entirely).

## Test plan
- [x] Reproduced the OOM on Qwen3.5-27B-VL, TP=8, single node 8xH100-80GB,
      fp8 PTQ checkpoint, in both `ptq_generate_vlm.py` and `quantize_vlm.py`.
- [x] Applied the fix, reran both previously-OOMing jobs, confirmed both now
      complete successfully (checkpoint save + test generation).
- [x] `ruff check` / `ruff format --check` pass on both changed files.
- Not adding a functional test: this is a runtime memory-tracking bug only
  reproducible with a large model under TP -- the project's functional test
  budget is capped at 2 GPUs, which can't reproduce the failure.


Batool1227:
Suggested labels for triage: `bug`, `area:quant`. (I don't have write access to self-label as an external contributor.)

yaoyu-33 (approved):

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5292/commits/0698727a008c16cca40a8caa66a4968d90abc077

PR #5295, opened 2026-08-04
fix(model): map the fused input norm for MLA with q_lora_rank=None

### LINKED ISSUE #5294: DeepSeek bridge leaves every input layernorm unmapped when q_lora_rank is None
**Describe the bug**

Affects `deepseek-ai/DeepSeek-V2-Lite` and
`kakaocorp/kanana-2-30b-a3b-thinking`, both public.

## Summary

A DeepSeek checkpoint with `q_lora_rank: null` loads with **every layer's input
layernorm left at its random initialization**. Nothing raises, and no weight
mismatch is reported — the parameter simply has no mapping, so the conversion
never visits it.

## Why the configuration is reachable

MLA normally compresses the query with a LoRA pair, so MCore builds
`linear_q_down_proj` + `linear_q_up_proj` and the fused input norm attaches to
the down projection. With `q_lora_rank=None` there is nothing to compress:
MCore builds a single `linear_q_proj`, and Transformer Engine fuses the input
norm into *that* module instead. The parameter is therefore

```
decoder.layers.*.self_attention.linear_q_proj.layer_norm_weight
```

and no standalone `input_layernorm` module exists to satisfy the registry's
existing `decoder.layers.*.input_layernorm.weight` entry.

`src/megatron/bridge/models/deepseek/common.py` already maps the *weight* of
that single-matrix projection:

```python
# For models without MLA
"decoder.layers.*.self_attention.linear_q_proj.weight": "model.layers.*.self_attn.q_proj.weight",
```

but has no entry for the norm that comes with it. DeepSeek-V2 236B and
DeepSeek-V3 both set `q_lora_rank`, so the path is never exercised by the
headline models.

The comment above that line ("For models without MLA") is also slightly
misleading: `linear_q_proj` is used both by non-MLA models *and* by MLA with
`q_lora_rank=None`. The affected models below have `multi_latent_attention=True`.

## Affected models

Both go through `models/deepseek/common.py`, so a single registry gap hits both
the V2 and V3 bridges.

| Model | Bridge | Layers | `input_layernorm` keys | `q_proj` keys | `q_a_proj` / `q_a_layernorm` |
|---|---|---|---|---|---|
| `deepseek-ai/DeepSeek-V2-Lite` | `deepseek_v2_bridge` | 27 | 27 | 27 | 0 / 0 |
| `kakaocorp/kanana-2-30b-a3b-thinking` | `deepseek_v3_bridge` | 48 | 48 | 48 | 0 / 0 |

Counts are from each repo's `model.safetensors.index.json`. The zero
`q_a_proj` / `q_a_layernorm` columns confirm there is no query LoRA to map, and
the `input_layernorm` count equals the layer count — every one of them is
currently dropped.

## Reproduction

No GPU required; the mapping lookup is enough to show the gap.

```python

for repo in ["deepseek-ai/DeepSeek-V2-Lite", "kakaocorp/kanana-2-30b-a3b-thinking"]:
    bridge = AutoBridge.from_hf_pretrained(repo)
    registry = bridge._model_bridge.mapping_registry()
    for name in (
        "decoder.layers.0.self_attention.linear_q_proj.weight",
        "decoder.layers.0.self_attention.linear_q_proj.layer_norm_weight",
    ):
        print(f"{repo}\n  {name}\n    -> {registry.megatron_to_hf_lookup(name)}")
```

The weight resolves; the norm returns `None`.

End to end, building either model and calling `load_hf_weights` emits one
warning per layer:

```
WARNING: No mapping found for megatron_param: decoder.layers.0.self_attention.linear_q_proj.layer_norm_weight
...
```

## Fix

One registry entry in `src/megatron/bridge/models/deepseek/common.py`:

```diff
-        # For models without MLA
+        # Single-matrix q projection: used both without MLA and by MLA with
+        # q_lora_rank=None, which has no q_a_proj/q_a_layernorm to map.
         "decoder.layers.*.self_attention.linear_q_proj.weight": "model.layers.*.self_attn.q_proj.weight",
+        # In that q_lora_rank=None case TE fuses the input norm into the q projection,
+        # so there is no standalone input_layernorm module and the entry above maps
+        # nothing. Both spellings target the same HF tensor; only one exists per model.
+        "decoder.layers.*.self_attention.linear_q_proj.layer_norm_weight": "model.layers.*.input_layernorm.weight",
```

Both spellings point at the same HF tensor and only one of them exists for a
given model, so the pre-existing `decoder.layers.*.input_layernorm.weight`
entry is unaffected.

MTP mappings are generated by rewriting this same `param_mappings` dict, so
models that also set `num_nextn_predict_layers` pick up the new entry for
`mtp.layers.*.mtp_model_layer.self_attention.linear_q_proj.layer_norm_weight`
without further changes. Neither public model above ships MTP layers; that half
was verified separately on a non-public MTP-enabled checkpoint of the same
architecture (48 decoder + 3 MTP layers): 51 unmapped parameters before the fix,
0 after, with `mtp.layers.0.eh_proj.weight` then matching the checkpoint exactly.

## Secondary: the failure mode hides its own cause

`build_conversion_tasks` is declared

```python
def build_conversion_tasks(...) -> List[None | WeightConversionTask]:
```

and does leave `None` slots — the backfill loop skips any global parameter whose
mapping lookup returns `None`. `load_weights_hf_to_megatron` then reads
`task.megatron_module` without checking:

```
AttributeError: 'NoneType' object has no attribute 'megatron_module'
```

The traceback names no parameter, despite the builder having already warned
about each one. Suggested alongside the mapping fix:

```diff
         for task in self._with_progress_tracking(hf_to_megatron_tasks, description):
+            # build_conversion_tasks returns List[None | WeightConversionTask]: a slot stays
+            # None when no mapping matched that global parameter. It already warned; crashing
+            # here with AttributeError buries which parameter was unmapped.
+            if task is None:
+                continue
             # None means megatron module not on current rank, skip if this task is not going to happen
             if task.megatron_module is None:
                 continue
```

Worth noting the more dangerous variant this class of gap allows: had the
mapping been *wrong* rather than absent, there would be no `None` slot and no
error at all — just a model quietly training from a partly random
initialization. A missing mapping is currently only loud by accident.

## Suggested test

A unit test over the registry needs no weights and covers both bridges:

```python
@pytest.mark.parametrize("bridge_cls", [DeepSeekV2Bridge, DeepSeekV3Bridge])
def test_single_matrix_q_projection_maps_its_fused_input_norm(bridge_cls):
    """q_lora_rank=None fuses the input norm into linear_q_proj; it must still map."""
    registry = bridge_cls().mapping_registry()

    mapping = registry.megatron_to_hf_lookup(
        "decoder.layers.0.self_attention.linear_q_proj.layer_norm_weight"
    )

    assert mapping is not None
    assert mapping.hf_param == "model.layers.0.input_layernorm.weight"
```

## Verification performed

8x B300, TP1/PP1/EP8, DeepSeek-V3-shaped 48-layer / 128-expert / MTP=3
checkpoint with `q_lora_rank=None`:

| | before | after |
|---|---|---|
| unmapped parameters | 51 | 0 |
| model construction | ok | ok |
| weight load | `AttributeError` | completes |
| `mtp.layers.0.eh_proj.weight` vs checkpoint | — | exact match, std 0.015242 both sides |

bzantium:
# What does this PR do?

Maps the fused input norm that MLA uses when `q_lora_rank` is None, so those checkpoints stop loading with every layer's input layernorm left at its random initialization.

Fixes #5294.

## Why

With `q_lora_rank=None` there is no query LoRA to build, so Megatron-Core creates a single `linear_q_proj` and Transformer Engine fuses the input norm into it. The parameter is then

```
decoder.layers.*.self_attention.linear_q_proj.layer_norm_weight
```

and no standalone `input_layernorm` module exists to satisfy the registry's existing `decoder.layers.*.input_layernorm.weight` entry. `models/deepseek/common.py` already maps the weight of that single-matrix projection but not the norm that comes with it.

Nothing raises. The parameter has no mapping, so the conversion never visits it and the model trains from a partly random initialization. DeepSeek-V2 236B and V3 both set `q_lora_rank`, which is why the headline models never exercise this; `deepseek-ai/DeepSeek-V2-Lite` (27 layers) and `kakaocorp/kanana-2-30b-a3b-thinking` (48 layers) do not set it, and both go through this shared mapping list.

## What changes

One registry entry in `models/deepseek/common.py`. Both spellings target the same HF tensor and only one exists for a given model, so the pre-existing `input_layernorm` entry is unaffected. MTP mappings are rewritten from the same dict, so MTP-enabled models pick the entry up without further changes.

The PR also guards the two loops in `models/conversion/model_bridge.py` that consume conversion tasks. `build_conversion_tasks` is declared `List[None | WeightConversionTask]` and does leave `None` slots for unmapped global parameters, but both loops read `task.megatron_module` directly, so an unmapped parameter surfaces as

```
AttributeError: 'NoneType' object has no attribute 'megatron_module'
```

naming no parameter, despite the builder having already warned about each one. That is what made this bug hard to place. Worth noting the more dangerous variant the same gap allows: had the mapping been wrong rather than absent, there would be no `None` slot and no error at all.

## Tests

Three cases in `TestCommonMappingSingleMatrixQProjection`, no weights needed:

- the fused `linear_q_proj.layer_norm_weight` resolves to `input_layernorm.weight`
- the standalone `input_layernorm.weight` entry still resolves (regression guard for the `q_lora_rank` set case)
- MTP layers inherit the new entry

Removing just the new mapping line fails the first and third and leaves the second passing, so the tests track the fix rather than the file.

`tests/unit_tests/models/deepseek/test_deepseek_bridges.py` passes 25/25. `ruff check` and `ruff format` clean.

Verified end to end on 8x B300, TP1/PP1/EP8, with a DeepSeek-V3-shaped 48-layer / 128-expert / MTP=3 checkpoint that sets `q_lora_rank=None`: 51 unmapped parameters before, 0 after, weight load completing where it previously raised.

### DIFF
--- a/src/megatron/bridge/models/conversion/model_bridge.py
+++ b/src/megatron/bridge/models/conversion/model_bridge.py
@@ -1194,6 +1194,11 @@ class MegatronModelBridge(
 
         _hf_import_cache: Dict[str, torch.Tensor] = {}
         for task in self._with_progress_tracking(hf_to_megatron_tasks, description):
+            # build_conversion_tasks returns List[None | WeightConversionTask]; a slot stays None
+            # when no mapping matched that global parameter. It already warned, and crashing here
+            # with AttributeError buries which parameter was unmapped.
+            if task is None:
+                continue
             # None means megatron module not on current rank, skip if this task is not going to happen
             if task.megatron_module is None:
                 continue
@@ -1322,6 +1327,11 @@ class MegatronModelBridge(
             conversion_tasks = self.build_conversion_tasks(hf_pretrained, megatron_model)
 
         for task in conversion_tasks:
+            # build_conversion_tasks returns List[None | WeightConversionTask]; a slot stays None
+            # when no mapping matched that global parameter. It already warned, and crashing here
+            # with AttributeError buries which parameter was unmapped.
+            if task is None:
+                continue
             # None means megatron module not on current rank, skip if this task is not going to happen
             if task.megatron_module is None:
                 continue
--- a/src/megatron/bridge/models/deepseek/common.py
+++ b/src/megatron/bridge/models/deepseek/common.py
@@ -57,8 +57,12 @@ def get_common_mapping_list(hf_config=None) -> list:
         "decoder.layers.*.self_attention.linear_q_up_proj.layer_norm_weight": "model.layers.*.self_attn.q_a_layernorm.weight",
         # Mcore local spec
         "decoder.layers.*.self_attention.q_layernorm.weight": "model.layers.*.self_attn.q_a_layernorm.weight",
-        # For models without MLA
+        # Single-matrix q projection: used both without MLA and by MLA with q_lora_rank=None,
+        # which has no q_a_proj/q_a_layernorm to map.
         "decoder.layers.*.self_attention.linear_q_proj.weight": "model.layers.*.self_attn.q_proj.weight",
+        # TE fuses the input norm into that projection, so no standalone input_layernorm module
+        # exists. Both spellings target the same HF tensor; only one exists per model.
+        "decoder.layers.*.self_attention.linear_q_proj.layer_norm_weight": "model.layers.*.input_layernorm.weight",
     }
 
     mapping_list = []
--- a/tests/unit_tests/models/deepseek/test_deepseek_bridges.py
+++ b/tests/unit_tests/models/deepseek/test_deepseek_bridges.py
@@ -332,6 +332,40 @@ class TestDeepSeekV3MaybeModifyLoadedHFWeight:
         assert result["up"] is w2
 
 
+class TestCommonMappingSingleMatrixQProjection:
+    """Input-norm coverage when q_lora_rank is None, so MLA uses a single q projection."""
+
+    def test_fused_input_norm_is_mapped(self):
+        """TE fuses the input norm into linear_q_proj, and that spelling must map."""
+        registry = MegatronMappingRegistry(*get_common_mapping_list())
+
+        mapping = registry.megatron_to_hf_lookup("decoder.layers.0.self_attention.linear_q_proj.layer_norm_weight")
+
+        assert mapping is not None
+        assert mapping.hf_param == "model.layers.0.input_layernorm.weight"
+
+    def test_standalone_input_norm_still_maps(self):
+        """The q_lora_rank set case keeps its own input_layernorm module."""
+        registry = MegatronMappingRegistry(*get_common_mapping_list())
+
+        mapping = registry.megatron_to_hf_lookup("decoder.layers.0.input_layernorm.weight")
+
+        assert mapping is not None
+        assert mapping.hf_param == "model.layers.0.input_layernorm.weight"
+
+    def test_mtp_layers_pick_up_the_fused_input_norm(self):
+        """MTP mappings are rewritten from the same dict, so they inherit the new entry."""
+        hf_config = SimpleNamespace(num_nextn_predict_layers=1, num_hidden_layers=61)
+        registry = MegatronMappingRegistry(*get_common_mapping_list(hf_config=hf_config))
+
+        mapping = registry.megatron_to_hf_lookup(
+            "mtp.layers.0.mtp_model_layer.self_attention.linear_q_proj.layer_norm_weight"
+        )
+
+        assert mapping is not None
+        assert mapping.hf_param == "model.layers.61.input_layernorm.weight"
+
+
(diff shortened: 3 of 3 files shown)


bzantium:
Duplicate of #5262, which I opened earlier and had lost track of. #5262 is the more complete change — it guards three task loops rather than two, including `stream_weights_megatron_to_hf`, and adds coverage in `test_model_bridge.py` as well. Closing this in favour of it. Apologies for the noise.

PR #5313, opened 2026-08-04
fix(conversion): initialize CPU imports with Gloo

cuichenx:
## What changed

- run CPU Hugging Face to Megatron import construction and checkpoint save inside a temporary single-rank Gloo context
- preserve a caller-managed distributed context when one already exists
- add focused regression coverage for both paths
- update the model-verification-card skill to reject local patches, unpublished commits, source overlays, monkeypatches, and copied trees as verification evidence

## Root cause

`AutoBridge.import_ckpt()` requested CPU model initialization without first initializing a CPU process group. `provide_distributed_model()` therefore entered its standalone fallback, unconditionally selected a CUDA device, and initialized NCCL. A genuine CPU-only import failed before weight mapping.

## User impact

CPU checkpoint import can construct and save through the existing temporary Gloo helper without requiring the CUDA/NCCL fallback. Verification failures must now be fixed through a pushed upstream PR and rerun from that exact clean commit before a card item can be marked verified.

## Validation

- `uv run --no-project python -m pytest tests/unit_tests/models/test_auto_bridge.py -k "import_ckpt" tests/unit_tests/training/test_model_load_save.py -k "import_ckpt or temporary_distributed_context" -q` — 7 passed
- `uv run --no-sync pre-commit run --all-files` — passed
- skill `quick_validate.py` — passed
- `git diff --check` — passed

## Pending model validation

The full GPT-OSS CPU import/export rerun will use the exact pushed PR head. It remains pending and is not claimed as verified here. The newer Transformer Engine load-time CUDA-driver dependency is a separate upstream blocker for driverless 26.08 containers.

### DIFF
--- a/skills/create-model-verification-card/SKILL.md
+++ b/skills/create-model-verification-card/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: create-model-verification-card
-description: Create or update concise, agent-readable Megatron Bridge model verification cards. Use when adding a model support card, auditing cross-model convergence comparability or verification coverage, recording conversion, deterministic inference, training, checkpoint resume, post-SFT export, or performance results, or preparing a model-support PR. Enforce the required core inventory, convergence-versus-performance contracts, optional canonical performance item, public Slurm launcher commands, training metrics, important-feature allowlist, and a strict privacy boundary that excludes private runtime wiring, internal paths, credentials, and job metadata.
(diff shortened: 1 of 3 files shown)


cuichenx:
Full `openai/gpt-oss-20b` CPU import from this exact pushed PR head was run against the pinned MCore commit. This PR cleared the original unconditional CUDA-device/NCCL fallback and reached model construction. The next failure is an independent MCore bug: YARN constructor-time CPU initialization calls `torch.cuda.current_device()`.

That owning fix is now public as NVIDIA/Megatron-LM#6270. Its focused eight-rank tests pass from the exact pushed MCore PR head. End-to-end GPT-OSS CPU import remains pending and is **not verified**; it must be rerun from a clean pushed Bridge commit that consumes the public MCore fix.

cuichenx:
Closing as superseded. Current main already contains equivalent temporary Gloo scoping for CPU import through #5007. The broader CPU-only conversion gaps are being tracked separately.

PR #3853, opened 2026-05-15
[models] fix: support CPU-only checkpoint conversion

yaoyu-33:
## Summary

Fixes #3295.

- Use Gloo and CPU initialization when CUDA is not visible for provider-based model construction.
- Disable MCore CPU-offload hooks/flags and TE-only `persist_layer_norm` behavior during CPU-only initialization.
- Fall back to the local GPT layer spec for CPU-init/no-CUDA builds so TransformerEngine SelfAttention is not instantiated.
- Add focused unit coverage for no-visible-CUDA provider initialization and GPT layer-spec selection.

## Validation

- cw job `11784762` completed `0:0`
  - log: `/lustre/fsw/portfolios/coreai/projects/coreai_dlalgo_llm/users/yuya/logs/issue3295/issue3295-cpu7_11784762.log`
  - `uv sync --frozen --no-install-project --no-install-package megatron-core`
  - focused pytest: `6 passed, 30 deselected`
  - no-CUDA tiny Llama `AutoBridge.to_megatron_model(wrap_with_ddp=False, use_cpu_initialization=True)` completed with `REPRO_DONE`
- cw job `11784769` completed `0:0`
  - log: `/lustre/fsw/portfolios/coreai/projects/coreai_dlalgo_llm/users/yuya/logs/issue3295/issue3295-pc2_11784769.log`
  - `uv run --no-sync pre-commit run --all-files`

PR #5317, opened 2026-08-04
[model] fix: use local GPT spec for CPU-only initialization

cuichenx:
<details><summary>Summary</summary>

## What

- Select the MCore-local GPT layer spec when `use_cpu_initialization=True`
  and CUDA is unavailable, avoiding CUDA-only Transformer Engine module
  construction on CPU-only hosts.
- Keep the existing TE spec when CUDA is available.
- Propagate the same local-versus-TE choice to MTP block construction.

This revisits the narrow spec-selection portion of #3853 in response to a real
full `openai/gpt-oss-20b` CPU-import failure. It does not revive that closed PR's
broader provider monkeypatches.

## Compatibility gate

The local-spec versus later GPU/TE checkpoint compatibility concern that closed
#3853 is treated as an acceptance gate, not assumed away. Current MCore provides
canonical sharded-state mappings and direct local-GPT-spec ↔ TE-GPT-spec
save/load coverage, including SequentialMLP ↔ TEGroupedMLP interchange.

Before this PR is called verified, a full GPT-OSS checkpoint produced by the
CPU-only local-spec path must load into the GPU/TE model from exact clean pushed
commits. If it does not, this approach is not acceptable.

## Validation

- [x] `tests/unit_tests/models/test_gpt_provider.py`: 29 passed in a
  driverless 26.06 container.
- [x] `uv run --no-sync pre-commit run --all-files`.
- [ ] Full `openai/gpt-oss-20b` CPU import from a clean public integration
  commit.
- [ ] Load the CPU-produced checkpoint into the GPU/TE GPT-OSS model.

No tests will be quarantined without Chen's explicit approval. No local source
patch or unpublished overlay may be counted as verification evidence.

</details>

### DIFF
--- a/src/megatron/bridge/models/gpt_provider.py
+++ b/src/megatron/bridge/models/gpt_provider.py
@@ -97,6 +97,11 @@ def local_layer_spec(config: "GPTModelProvider") -> ModuleSpec:
     )
 
 
+def _use_local_spec_for_cpu_only_initialization(config: "GPTModelProvider") -> bool:
+    """Return whether CPU initialization must avoid Transformer Engine modules."""
+    return bool(config.use_cpu_initialization) and not torch.cuda.is_available()
+
+
 def modelopt_transformer_layer_spec(config: "GPTModelProvider") -> ModuleSpec:
     """Layer specification for quantization with ModelOpt."""
     # arbitrary attention mask is used for speculative decoding training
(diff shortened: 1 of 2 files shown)


cuichenx:
Closing this standalone approach. Selecting local GPT/MTP specs does not by itself solve driverless Transformer Engine import, and local-to-TE checkpoint schema compatibility is not guaranteed across model families. Any future CPU-only conversion fix should be validated as an end-to-end design rather than landing this isolated selector.

PR #5322, opened 2026-08-04
[model] fix: support MCore FSDP factory wrappers

cuichenx:
## Summary

- keep conversion `unwrap_model()` compatible with both the Bridge-pinned legacy MCore FSDP class API and the newer MCore factory API
- unwrap both concrete `FullyShardedDataParallelV1` and `FullyShardedDataParallelV2` wrappers without passing the factory function to `isinstance()`
- add regression coverage for both API generations, including the legacy case where the V1/V2 symbols do not exist

## Root cause

Newer MCore versions expose `mcore_fsdp_adapter.FullyShardedDataParallel` as a factory function and provide V1/V2 as the concrete wrapper classes. Bridge conversion code included the factory in an `isinstance()` tuple, which raises before MFSDP conversion reaches weight loading.

This is separate from the TP=2 vocabulary-scatter defect fixed by #5353. That fix is present on current main; this PR removes the newer MCore compatibility blocker that prevented QA from re-verifying the three MFSDP conversion examples.

## Compatibility

- legacy API: use `FullyShardedDataParallel` when it is a class; V1/V2 symbols may be absent
- factory API: use the explicit, independent V1 and V2 wrapper classes
- no training-path, dependency, CI, public API, or MCore submodule changes

## Validation

- Red on clean `origin/main` `c62be0886a`:
  - `uv run python -m pytest tests/unit_tests/models/test_conversion_utils.py -k unwrap_model -vv`
  - factory case failed with `TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union`; legacy case passed
- Green on `ea19acb7f`:
  - `uv run python -m pytest tests/unit_tests/models/test_conversion_utils.py -vv` — 7 passed

### DIFF
--- a/src/megatron/bridge/models/conversion/utils.py
+++ b/src/megatron/bridge/models/conversion/utils.py
@@ -46,13 +46,20 @@ def unwrap_model(model, module_instances=None):
     if module_instances is None:
         from megatron.core.distributed import DistributedDataParallel as DDP
         from megatron.core.distributed import TorchFullyShardedDataParallel as torch_FSDP
-        from megatron.core.distributed.fsdp.mcore_fsdp_adapter import (
-            FullyShardedDataParallel as megatron_FSDP,
-        )
+        from megatron.core.distributed.fsdp import mcore_fsdp_adapter
         from megatron.core.distributed.fsdp.src.megatron_fsdp.megatron_fsdp import MegatronFSDP
         from megatron.core.transformer.module import Float16Module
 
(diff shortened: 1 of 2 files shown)


yaoyu-33 (approved):

yaoyu-33:
Superseded by #5431, which has now merged the MCore FSDP factory-wrapper compatibility fix into `main` together with focused coverage in `tests/unit_tests/models/test_conversion_unwrap_utils.py`. Rebasing this PR now conflicts on the same implementation and would only duplicate the code/tests, so closing this PR rather than resolving it as a duplicate.

PR #5331, opened 2026-08-05
build: bump nvidia-modelopt to 0.46.0rc1

kevalmorabia97:
<details><summary>Claude summary</summary>

## What
- Bump `nvidia-modelopt` to `0.46.0rc1` (from `0.44.0rc5`) in `override-dependencies`.
- Regenerate `uv.lock`.

## Lockfile delta
```
Resolved 327 packages
Updated nvidia-modelopt v0.44.0rc5 -> v0.46.0rc1
```

No transitive movement — `nvidia-modelopt`'s own dependency list is unchanged
between the two versions, so the only lockfile edits are the override
specifier, the package version, and the wheel URL + sha256.

Relocked with `uv 0.7.2` to match the `UV_VERSION` pin in
`docker/Dockerfile.ci`; the lockfile `revision` header is untouched.

## Test plan
- [ ] L0 CI green
- [ ] L1 CI green (label `needs-more-tests` applied)
- [ ] L2 CI green (label `full-test-suite` applied — modelopt drives the
      quantization paths that only L2 exercises)

## Quarantined tests (this bump)
_None yet — will be appended as flakes are identified during CI iteration._

</details>

PR #5337, opened 2026-08-05
refactor(quant)!: simplify quantization examples and fix calibration dataset load

kevalmorabia97:
### 1. Replace the hand-rolled calibration loop (`quantize.py`)

The old `get_calib_dataloader` / `_hf_dataset_forward_loop_func` pair had three defects:

- **`load_dataset("cnn_dailymail", ...)`** — the bare legacy id. Fails hard on the pinned `huggingface-hub` 1.26.0 with `HfUriError: Repository id must be 'namespace/name'`. Reproduced standalone with no modelopt imported, so it is not version-specific to the bump. ModelOpt's dataset registry maps this to `abisee/cnn_dailymail` internally.
- **`text[:max_sequence_length]` sliced characters, not tokens.** The parameter said sequence length and defaulted to 512, but the sample is a `str`, so calibration saw roughly a quarter of the intended context.
- **`megatron_generate(..., osl=1)` per sample at batch size 1** — full generate machinery to produce one token, when calibration only needs prefill.

All three go away by delegating to `get_megatron_calibration_forward_loop`, which handles tokenization, token-accurate truncation, batching, DP sharding, and a logits-free prefill. Adds `--calib-seq-length`, `--calib-batch-size`, and `--calib-dataset`.

`--calib-dataset` accepts a ModelOpt registered dataset name, a HuggingFace path, a local directory, or a `.jsonl` file, so calibration data can be supplied without Hub access. `apply_chat_template` is left at ModelOpt's default (`True`) rather than forced off, so chat-format calibration datasets are formatted correctly; it is a no-op for plain-text datasets like the `cnn_dailymail` default.

**CI note:** ModelOpt's registry resolves `cnn_dailymail` to the namespaced `abisee/cnn_dailymail`. CI functional tests run with `HF_HUB_OFFLINE=1` against a pre-seeded `HF_HOME` that currently holds only the legacy `cnn_dailymail` cache key, so `L2_Launch_quantization_aware_training` fails with `ConnectionError: Couldn't reach 'abisee/cnn_dailymail' on the Hub (OfflineModeIsEnabled)` until the test-data cache is seeded with the namespaced id. That seeding has been requested from the Megatron-Bridge admins.

### 2. Correctness fixes

- **No explicit `.eval()` is added.** ModelOpt already handles it: `mtq.calibrate` wraps calibration in `model.eval()` and restores the prior mode (`model_quant.py:106-124`), and `megatron_prefill` / `megatron_generate` each call `model.eval()` themselves.
- **Skip calibration when it cannot help** — `weight_only or not mtq.need_calibration(cfg)` avoids the dataset download and forward pass for weight-only and dynamic configs.
- **`ptq_generate.py` pointed users at `examples/models/quantize.py`**, a path that does not exist, and omitted `--hf-model-id`.

### 3. Share common logic

`quantize_utils.py` already existed, but only 2 of 5 scripts imported it; the rest re-declared their own `--tp/--pp/--ep/--etp` and their own `console = Console()`. New helpers:

| Helper | Replaces | Copies removed |
|---|---|---|
| `require_torchrun()` | WORLD_SIZE guard | 5 |
| `require_checkpoint(path, *, quantize_command)` | "checkpoint missing → run this" block | 3 |
| `build_bridge_and_provider(...)` | `AutoBridge.from_hf_pretrained` + `is_safe_repo` + 5 provider assignments + `finalize()` + `initialize_model_parallel()` | 5 |
| `load_quantized_megatron_model(...)` | `load_megatron_model` + `mp_overrides` dict | 3 |
| `print_parallelism_summary(provider)` | 4 `console.print` lines | 5 |

`add_common_quantization_args` was split: `add_common_model_args` (model + parallelism + `--trust-remote-code`) is used by all five scripts, and the quantization-only flags stay in the original function so `ptq_generate.py` / `export.py` do not inherit `--export-quant-cfg` or `--megatron-save-path`.

`build_bridge_and_provider` keeps `AutoBridge` and the provider knobs visible and named — these are Megatron-Bridge examples, so the bridge → provider → model flow is the thing being taught. It is just no longer taught five times. ModelOpt's `load_mbridge_model_from_hf` would collapse it further but was deliberately not used, for the same reason.

### 4. Reuse `modelopt.torch.utils.distributed`

| Was | Now | Sites |
|---|---|---|
| `torch.distributed.get_rank() == 0` | `dist.is_master()` | 5 |
| `if is_initialized(): destroy_process_group()` | `dist.cleanup()` | 6 |
| `torch.distributed.barrier()` | `dist.barrier()` | 3 |
| `disable=torch.distributed.get_rank()` (tqdm) | `disable=not dist.is_master()` | 1 |
| `broadcast_object_list` + `is_initialized()`/`else` fork | `dist.broadcast(obj, src=dist.size() - 1)` | 1 |

Not purely cosmetic: `dist.broadcast` short-circuits at `size() == 1`, so the 9-line fork in `export.py` collapses to 2 lines; `dist.rank()` is safe before init (returns 0) where `torch.distributed.get_rank()` raises; and `dist.cleanup()` adds a suppressed barrier before destroy.

## Testing

All runs in `nvcr.io/nvidian/nemo:26.08.rc4` with `nvidia-modelopt==0.46.0rc1`, 2× RTX 6000 Ada.

**Functional tests — 10 passed, 0 failed:**

| Test | Result | Time | Scripts exercised |
|---|---|---|---|
| `test_quantization_workflow` | 3 passed | 21:34 | `quantize.py`, `ptq_generate.py` (real Llama-3.2-1B + real calibration) |
| `test_export_workflow` | 2 passed | 13:47 | `quantize.py`, `export.py` |
| `test_qwen3_moe_quantization_workflow` | 2 passed | 12:15 | `quantize.py`, `ptq_generate.py` (MoE TP=2/ETP=2, PP re-shard) |
| `test_qwen3_vl_quantization_workflow` | 2 passed | 11:18 | `quantize_vlm.py`, `ptq_generate_vlm.py` |
| `test_qat_workflow` | 1 passed | 08:34 | `quantize.py`, `pretrain_quantized_llama3_8b.py` |

`test_quantization_workflow` is the direct regression check: it is the test that failed with `HfUriError` before this change, and it passes here against the real `cnn_dailymail` calibration path.

**Scope note:** the table above was run just before the final `modelopt.torch.utils.distributed` swap and comment cleanup. Those were re-verified afterwards: `test_quantization_workflow` re-run on the final code gives the same **3 passed** (20:34), and all six scripts pass CLI smoke tests (import, `--help`, and exit 2 with `required: --hf-model-id`). Re-runs of the export and VLM suites were still in flight at the time of writing; CI is the authority.

**Lint:** `ruff check`, `ruff format --check`, and `ruff check --select I` clean at the pinned v0.9.9.

### DIFF
--- a/examples/quantization/quantize.py
+++ b/examples/quantization/quantize.py
@@ -17,78 +17,53 @@ This example demonstrates how to use the AutoBridge to perform quantization
 from a Hugging Face model to a quantized Megatron-LM model on multiple GPUs.
 
 The process is as follows:
-1. An AutoBridge is initialized from a pretrained Hugging Face model
-    (e.g., "meta-llama/Llama-3.2-1B"). This downloads the model from the Hub and loads it.
+1. An AutoBridge is initialized from the pretrained Hugging Face model given by `--hf-model-id`.
+    This downloads the model from the Hub (or loads it from a local path) and loads it.
 2. ModelOpt quantization is applied to the Megatron-LM model using the specified configuration.
 3. The quantized Megatron-LM model is saved in Megatron's native checkpoint format
     using the `--megatron-save-path` argument.
 
 Usage:
-torchrun --nproc_per_node 2 examples/models/quantize.py --export-quant-cfg fp8
-torchrun --nproc_per_node 2 examples/models/quantize.py --export-quant-cfg fp8 --megatron-save-path ./megatron_checkpoint
+torchrun --nproc_per_node 2 examples/quantization/quantize.py \
+    --hf-model-id meta-llama/Llama-3.2-1B --export-quant-cfg fp8
+torchrun --nproc_per_node 2 examples/quantization/quantize.py \
+    --hf-model-id meta-llama/Llama-3.2-1B --export-quant-cfg fp8 --megatron-save-path ./megatron_checkpoint
 """
 
 import argparse
-import os
-import sys
 import warnings
 
 import modelopt.torch.quantization as mtq
+import modelopt.torch.utils.distributed as dist
 import torch
-from datasets import load_dataset
 from megatron.core.utils import unwrap_model
+from modelopt.torch.utils.plugins.megatron_calibration import get_megatron_calibration_forward_loop
 from modelopt.torch.utils.plugins.megatron_generate import megatron_generate
 from quantize_utils import (
     QUANT_CFG_CHOICES,
     add_common_quantization_args,
+    build_bridge_and_provider,
     console,
     create_quantization_stats_table,
     get_modelopt_torch_quantization_config,
+    print_parallelism_summary,
+    require_torchrun,
 )
-from tqdm import tqdm
 
-from megatron.bridge import AutoBridge
 from megatron.bridge.models.decorators import torchrun_main
-from megatron.bridge.models.hf_pretrained.utils import is_safe_repo
 
 
 warnings.filterwarnings("ignore")
 
-HF_MODEL_ID = "meta-llama/Llama-3.2-1B"
-
-
-def get_calib_dataloader(calib_size=512, max_sequence_length=512):
-    """Return a dataloader for calibration."""
-    # Only rank 0 downloads/prepares the dataset to avoid race conditions when multiple
-    # ranks simultaneously write to the same HuggingFace cache directory.
-    rank = torch.distributed.get_rank()
-    if rank == 0:
-        dataset = load_dataset("cnn_dailymail", name="3.0.0", split="train")
-    torch.distributed.barrier()
-    if rank != 0:
-        dataset = load_dataset("cnn_dailymail", name="3.0.0", split="train")
-
-    text_column = "article"
-
-    calib_size = min(len(dataset), calib_size)
-    for i in range(calib_size):
-        yield dataset[i][text_column][:max_sequence_length]
-
-
-def _hf_dataset_forward_loop_func(model, tokenizer, calib_size):
-    """Forward loop function for calibration using HuggingFace dataset."""
-    dataloader = get_calib_dataloader(calib_size)
-
-    for prompt in tqdm(dataloader, total=calib_size, disable=torch.distributed.get_rank()):
-        tokens = tokenizer(prompt, return_tensors="pt")
-        megatron_generate(model, tokens.input_ids.cuda(), osl=1)
+# Resolved by ModelOpt's dataset registry, which pins the concrete namespaced Hub id.
+DEFAULT_CALIB_DATASET = "cnn_dailymail"
 
 
 def _custom_prompt_forward_loop_func(model, prompts, tokenizer, is_rank_0, osl=32):
     """Forward loop function for testing quantized model with custom prompts."""
     all_prompts = prompts.split("|")
 
-    for idx, prompt in tqdm(enumerate(all_prompts), disable=torch.distributed.get_rank()):
+    for idx, prompt in enumerate(all_prompts):
         tokens = tokenizer(prompt, return_tensors="pt")
         generated_ids = megatron_generate(model, tokens.input_ids.cuda(), osl=osl, enable_kv_cache=False)
         generated_texts = tokenizer.batch_decode(generated_ids)
@@ -99,7 +74,7 @@ def _custom_prompt_forward_loop_func(model, prompts, tokenizer, is_rank_0, osl=3
 
 @torchrun_main
 def main(
-    hf_model_id: str = HF_MODEL_ID,
+    hf_model_id: str,
     tp: int = 1,
     pp: int = 1,
     ep: int = 1,
@@ -107,6 +82,9 @@ def main(
     megatron_save_path: str | None = None,
     export_quant_cfg: str = "fp8",
     calib_size: int = 512,
+    calib_seq_length: int = 512,
+    calib_batch_size: int = 1,
+    calib_dataset: str = DEFAULT_CALIB_DATASET,
     compress: bool = False,
     weight_only: bool = False,
     export_kv_cache_quant: bool = False,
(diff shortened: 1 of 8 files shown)


examples/quantization/quantize.py
```diff
@@ -164,26 +125,29 @@ def main(
...
+                console.print("[yellow]Weight-only or dynamic quantization: skipping calibration.[/yellow]")
+            ptq_forward_loop_func = None
+        else:
+            ptq_forward_loop_func = get_megatron_calibration_forward_loop(
+                bridge.hf_pretrained.tokenizer,
+                dataset_name=calib_dataset,
+                num_samples=calib_size,
+                seq_length=calib_seq_length,
+                batch_size=calib_batch_size,
+                pack=True,
```
yueshen2016, line 141: `L2_Launch_quantization_aware_training` fails here (h100 + gb200):

```
dataset_utils.py:630  load_dataset(streaming=True, **config, split=s)
ConnectionError: Couldn't reach 'abisee/cnn_dailymail' on the Hub (OfflineModeIsEnabled)
```

The old in-script `load_dataset` was non-streaming, so it was cache-satisfiable under `HF_HUB_OFFLINE=1`; ModelOpt streams, so it isn't. Cache seeding may get past module resolution, but streaming still has to fetch data files afterwards.

Suggest using the flag this PR adds instead — a local `.jsonl` goes through HF's `json` builder (`dataset_utils.py:507`) and never hits the Hub:

```python
# test_qat_workflow.py
+"--calib-dataset", "/home/TestData/<...>/cnn_dailymail_calib.jsonl",
```

Also covers `--calib-dataset`'s local-path branch, which is otherwise untested.

yueshen2016 (commented):
One note on the failing `L2_Launch_quantization_aware_training` job (fails on both h100 and gb200) — inline.

examples/quantization/quantize_utils.py
```diff
@@ -135,11 +289,6 @@ def add_common_quantization_args(parser: argparse.ArgumentParser) -> None:
         default=512,
         help="Samples to use for PTQ calibration.",
     )
```
yueshen2016, line 291: Why is `--compress` dropped here? `quantize_vlm.py` keeps both the flag and the `mtq.compress()` call, so the LLM and VLM scripts have diverged on it. Intentional, or fallout from the argparse split?

yueshen2016 (commented):
One question on `--compress` — inline.

kevalmorabia97:
Thanks — this is a good catch and the diagnosis is right. I ran the experiments to close the two open questions.

**Your open question: does `streaming=True` resolve and fetch data files under `HF_HUB_OFFLINE=1`?**

Yes. Tested with the hub cache **deleted** so only the processed arrow cache remained, under `HF_HUB_OFFLINE=1` *and* `HF_DATASETS_OFFLINE=1` (modelopt 0.46.0rc1, `nemo:26.08.rc4`):

| Call | Result |
|---|---|
| `load_dataset("abisee/cnn_dailymail", "3.0.0", split="train", streaming=True)` | OK — real article text |
| `get_dataset_samples("cnn_dailymail", ...)` (ModelOpt, streams internally) | OK — 4 samples |

So streaming gets past both module resolution *and* the data read from a local cache; the second CI round-trip you were worried about would most likely have gone green. Supporting datum: on the `hub/datasets--…` cache form, **non-streaming fails identically** to streaming — the failure is purely module resolution, not streaming.

One correction that matters if we ever do seed: the working cache form is `HF_HOME/datasets/<org>___<name>/` (processed arrow), **not** `HF_HOME/hub/datasets--<org>--<name>/`. A `hf download --repo-type dataset` populates the latter and fails offline. The Internal-Developer guide's `hub/` path is written for *models*, which do resolve from the hub cache. I also confirmed a processed cache copied to a different `<ns>___<name>` key loads fine under the new name, so a bucket-to-bucket copy would suffice — no re-upload.

**That said, I agree with your recommendation and will take it.** Even though seeding would probably work, the `.jsonl` route is in-tree, has no admin dependency, no CI round-trip, and — as you note — is the only thing that gives `--calib-dataset`'s local-path branch any coverage. I verified a 1024-row / 3.5 MB `.jsonl` loads fully offline through the json builder.

**Your directory-form caveat is exactly right, with numbers.** The staged layout has to satisfy `name="3.0.0"` from the registry config:

| Staged layout | Size | Offline result |
|---|---|---|
| Full snapshot dir named `cnn_dailymail/` (README + `1.0.0`/`2.0.0`/`3.0.0`) | 2.4 GB | OK |
| `3.0.0` only, README kept | 799 MB | `DataFilesNotFoundError` |
| `3.0.0` only, no README | 799 MB | `BuilderConfig '3.0.0' not found. Available: ['default']` |

The README's YAML declares all three configs, so the directory form effectively requires the full 2.4 GB snapshot. The `.jsonl` at 3.5 MB avoids that coupling entirely — further reason to prefer it.

**On `--disable-hf-datasets-file-lock`: not dead, please keep it.** I tested by replacing `datasets.builder.FileLock` with a tripwire that raises on instantiation:

```
jsonl:            FileLock USED
registry(cached): FileLock USED
```

Both the `.jsonl` path and the registry/streaming path still go through `datasets.builder.FileLock`, so the flag is still load-bearing for read-only test-data directories.

I'll push the `--calib-dataset <local .jsonl>` change to the QAT test (and the PTQ/export tests, which hit the same calibration path). That still needs the small `.jsonl` staged into the CI test-data path — much lighter than the dataset seeding, and I'll follow the Internal-Developer upload procedure for it.

---

Unrelated, correcting the automated review above: it asks to confirm the ModelOpt APIs exist in `nvidia-modelopt==0.44.0rc5`. That pin is stale — #5331 merged and `main` is on `0.46.0rc1`, which this branch is rebased onto. `megatron_calibration` does **not** exist in `0.44.0rc5` (hence the dependency), and `get_megatron_calibration_forward_loop`, `mtq.need_calibration`, and `modelopt.torch.utils.distributed.{is_master,broadcast,size,barrier,cleanup}` are all verified present in `0.46.0rc1`.

examples/quantization/quantize_utils.py
kevalmorabia97: Intentional — I was asked to drop `compress` from the LLM script to simplify it, and moving the flag into `quantize_vlm.py`'s own parser was the follow-through so no script would accept a flag it silently ignores. Not argparse fallout.

That said, your divergence point is fair, so I've restored it rather than defend the split. `3e2091b89`:

- `--compress` moved back into `add_common_quantization_args`, so both scripts get it from the shared helper exactly as on `main`
- `compress` param, the `mtq.compress()` block, and the call-site arg restored in `quantize.py`
- the duplicate VLM-local `--compress` removed, since it now comes from the shared helper again

Verified via `--help`: `--compress` present on `quantize.py` and `quantize_vlm.py`, absent from `ptq_generate.py` / `ptq_generate_vlm.py` / `export.py`.

One thing I did *not* restore: while `compress` was gone I had added two guards that never existed on `main` — rejecting `--compress` combined with `--export-kv-cache-quant`, and skipping the post-quantization generation sanity check when compressed (since `megatron_generate` doesn't support compressed forward for every format). Both are real hazards, but they'd be new behaviour rather than a revert, and `--compress` has no test coverage on either script. Happy to add them here or in a follow-up if you'd like them.

examples/quantization/quantize.py
kevalmorabia97: Not true. It passes after cached HF_HOME in CI machine is fixed

yueshen2016 (approved):

PR #5361, opened 2026-08-06
fix(data): preserve chat template EOS semantics

yaoyu-33:
# What does this PR do?

Follow-up to #5344, which is now merged. This PR isolates the GLM-5.2 chat-template, assistant-mask, EOS, and packing-cache semantics needed by the SFT recipe.

## Summary

- use the pinned official GLM-5.2 Hugging Face chat template instead of a recipe-local override
- recover assistant spans from structural role markers when a template has no `{% generation %}` blocks or per-turn end marker
- honor `add_eos=True` in the Hugging Face chat path and avoid duplicate EOS tokens
- supervise the appended EOS only for a complete terminal assistant turn; confirmed right truncation inside an answer keeps the synthesized EOS out of loss
- preserve complete assistant spans under confirmed left/right truncation, while exact `max_length` alone is not treated as proof of truncation
- fingerprint packed artifacts from the effective tokenizer revision, chat template, token IDs, preprocessing, and EOS policy instead of the synthetic `tokenizer_model_name = "glm5"`

## Why GLM-5.2 needs this fallback

Qwen3 has explicit `<|im_start|>` / `<|im_end|>` turn framing. GLM-5.2 starts roles with `<|system|>`, `<|user|>`, `<|assistant|>`, or `<|observation|>`, but does not emit a per-turn end marker. Its assistant span therefore ends at the next structural role marker or the logical sequence boundary.

GLM-5.2 also uses `<|endoftext|>` as EOS, while its official chat template does not emit that token. Dataset EOS handling must remain separate from assistant-boundary inference.

## Validation

- `uv run --active --no-sync pre-commit run --all-files`
- focused ruff and format checks for the changed source and tests
- Python compile check for all changed Python files
- focused pytest collection is currently blocked in the local environment because the installed `nvidia-resiliency-ext` is older than the minimum required by current main; Linux CI is running on this head

## Relationship to other PRs

- #5344 is already merged
- #5535 should land after this PR because it relies on the official-template and EOS behavior established here

# Before your PR is ready for review

- [x] Read and followed the contributor guidelines
- [x] Added focused regression tests
- [x] Kept the change scoped to data behavior, recipe cleanup, model-card configuration, and tests

# Additional Information

- Related to MB-1140


src/megatron/bridge/data/datasets/utils.py
```diff
@@ -971,6 +977,26 @@ def _chat_preprocess(
...
+    if add_eos:
+        eos_id = getattr(tokenizer, "eos_id", None)
+        if eos_id is None:
+            hf_tokenizer = get_processor_tokenizer(tokenizer)
+            eos_id = getattr(hf_tokenizer, "eos_token_id", None)
+        if eos_id is None:
+            raise ValueError("Chat preprocessing with add_eos=True requires a tokenizer EOS token ID.")
+        eos_id = int(eos_id)
+        eos_in_loss = tokenized.truncation_side != "right" and (loss_mode == "full" or terminal_assistant)
+        if not input_ids or input_ids[-1] != eos_id:
```
yaoyu-33, line 991: [P1] Avoid duplicating template-emitted EOS followed by formatting

This checks only `input_ids[-1]`. With Transformers 5.12.1 and the official Qwen2.5-Instruct tokenizer, a complete assistant response ends `A<|im_end|>\n`, where `<|im_end|>` is the EOS token. Because the newline is last, default `add_eos=True` produces and supervises `A<|im_end|>\n<|im_end|>` for every ordinary Qwen chat row. Please recognize terminal EOS before trailing template formatting and add a real-tokenizer non-GLM regression.

PR #5362, opened 2026-08-06
fix(data): preserve tokenizer fork safety

### LINKED ISSUE #5328: [bug] gpt_sft force-enables TOKENIZERS_PARALLELISM at import, disabling the tokenizers fork-deadlock guard for the whole SFT path
## Problem

Importing `megatron.bridge.data.datasets.gpt_sft` force-enables tokenizer parallelism for the whole process when the user has not set it:

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/src/megatron/bridge/data/datasets/gpt_sft.py#L53-L54

```python
if os.getenv("TOKENIZERS_PARALLELISM") is None:
    os.putenv("TOKENIZERS_PARALLELISM", "True")
```

`TOKENIZERS_PARALLELISM` is not only a throughput switch. When a fast tokenizer has already used its rayon pool and the process then forks, `tokenizers` detects the fork and disables parallelism, emitting `The current process just got forked, after parallelism has already been used. Disabling parallelism to avoid deadlocks`. Setting the variable explicitly suppresses that automatic protection and keeps the pool enabled across the fork. A forked child inherits the pool's bookkeeping but none of its worker threads, so the first batch encode in the child blocks forever.

The SFT data path forks: `datasets.map(num_proc=...)` uses fork, and so does any `multiprocessing.Pool` built around the same dataset object. Enabling parallelism by default in the module that defines that path removes the guard exactly where it is needed.

Two secondary problems with the same two lines:

1. **`os.putenv` does not update `os.environ`.** The C environment changes, so the Rust tokenizer sees `True` through `getenv`, but Python-side inspection of `os.environ` shows nothing. A user debugging a hang cannot see the setting that causes it from inside their own process.
2. **The default is inconsistent with the rest of the repository.** Every other location sets it to `False`: `examples/models/stepfun/step35/slurm_pretrain.sh`, `examples/models/stepfun/step37/slurm_pretrain.sh`, `scripts/performance/rubin_launch_interactive.sh`, `scripts/performance/utils/executors.py`, `examples/evaluation/utils/executors.py`, and `src/megatron/bridge/diffusion/models/wan/inference/__init__.py`.

## Evidence

Observed while adding batched tokenization to a forked worker pool over the offline packed-sequence path, on a 192-core node:

| Fork pool size | Result |
|---|---|
| 4 workers | completes |
| 8 workers | deadlock |
| 64 workers | deadlock |

In the deadlocked runs every process is asleep on a futex, memory is not exhausted, and `loadavg` decays to ~0, which reads as a stalled job rather than a hang:

```
$ awk '/^State:/{print $2}' /proc/<pid>/status ; cat /proc/<pid>/wchan   # all 68 processes
     68 S futex_wait_queue
$ free -g
Mem:  total 4011   used 340   free 3535
```

`py-spy dump` on a stuck worker:

```
Thread (idle): "MainThread"
    _encode_plus (transformers/tokenization_utils_tokenizers.py:924)
    __call__ (transformers/tokenization_utils_base.py:2490)
    apply_chat_template (transformers/tokenization_utils_base.py:3078)
    ...
    worker (multiprocessing/pool.py:125)
    _launch (multiprocessing/popen_fork.py:71)
```

That the failure rate rises with pool size is consistent with the fork hazard rather than with a data-dependent bug.

## Scope

The currently shipped offline packing path encodes one conversation per call, which does not dispatch to the parallel batch path, so this does not reproduce on an unmodified checkout today. The report is about the guard being disabled process-wide by default: any SFT code path that batch-encodes inside a forked worker inherits a deadlock, including a user's own `map(num_proc=...)` callable.

## Proposed direction

- Do not set `TOKENIZERS_PARALLELISM` to `True` from library import. Either leave it unset so `tokenizers` can apply its own fork protection, or set it to `False` in the data path that forks.
- If the intent of the line was to silence the fork warning, `False` achieves that and is what the rest of the repository already does.
- Use `os.environ[...] = ...` instead of `os.putenv` so the effective value is visible to Python.
- Setting a process-wide environment variable as an import side effect is surprising in a library; if a parallelism default is wanted it would be better expressed where the tokenizer is constructed.

## Acceptance criteria

- [ ] Importing `gpt_sft` does not enable tokenizer parallelism across a subsequent fork.
- [ ] The chosen default is consistent with the other call sites in the repository.
- [ ] Any environment variable the library sets is visible through `os.environ`.
- [ ] A regression test covers a fork pool whose workers batch-encode, and completes rather than hanging.

## Affected area

`area:data`

## Regression?

Not a regression. The lines date from the initial import of the NeMo LM data path.

## Environment

- `main` at `bcd2d134c`
- 192-core node, fork-based `multiprocessing.Pool`, Hugging Face fast tokenizer

yaoyu-33:
# What does this PR do ?

Stop the GPT SFT dataset module from force-enabling Hugging Face tokenizer parallelism at import time, preserving the tokenizer library's fork-deadlock protection.

# Changelog

- Remove the process-wide TOKENIZERS_PARALLELISM=True import side effect from gpt_sft.
- Add a regression test that warms a fast tokenizer, forks a worker pool, and verifies batch encoding completes with bounded timeouts.

# GitHub Actions CI

Local validation:

- `uv run --no-sync pre-commit run --all-files` — passed.
- Minimal fork-pool harness — fixed code exits 0; the removed behavior reproducibly times out and exits 124.
- Targeted pytest collection is blocked on macOS arm64 because the pinned Megatron-Core imports Triton, which has no macOS wheel; the new test is intended to run in Linux CI.

**Pre checks**:

- [x] Did you add or update any necessary documentation? (Not applicable; behavior-only fix.)
- [x] Does the PR affect components that are optional to install? (No.)
  - [x] Reviewer: Does the PR have correct import guards for all optional libraries? (Not applicable.)

# Additional Information

Fixes #5328

### DIFF
--- a/src/megatron/bridge/data/datasets/gpt_sft.py
+++ b/src/megatron/bridge/data/datasets/gpt_sft.py
@@ -50,9 +50,6 @@ NEMO_DATASETS_CACHE = Path(os.getenv("NEMO_DATASETS_CACHE", DEFAULT_NEMO_DATASET
 DEFAULT_NEMO_MODELS_CACHE = NEMO_CACHE_HOME / "models"
 NEMO_MODELS_CACHE = Path(os.getenv("NEMO_MODELS_CACHE", DEFAULT_NEMO_MODELS_CACHE))
 
-if os.getenv("TOKENIZERS_PARALLELISM") is None:
-    os.putenv("TOKENIZERS_PARALLELISM", "True")
-
 logger = logging.getLogger(__name__)
 
 # hack to avoid the "not enough disk space" error in some slurm cluster
--- a/tests/unit_tests/data/datasets/test_gpt_sft.py
+++ b/tests/unit_tests/data/datasets/test_gpt_sft.py
@@ -33,6 +33,7 @@ from megatron.bridge.data.packing.gpt_sft import GPTSFTPackedDataset
 from megatron.bridge.data.samplers import build_pretraining_data_loader
 
 
+@pytest.mark.unit
 def test_import_preserves_tokenizers_fork_safety():
     """Importing the SFT dataset must not enable tokenizer parallelism across a fork."""
     code = textwrap.dedent(
@@ -56,7 +57,7 @@ def test_import_preserves_tokenizers_fork_safety():
 
         with multiprocessing.get_context("fork").Pool(1) as pool:
             result = pool.apply_async(encode_batch)
-            assert result.get(timeout=5) == len(batch)
+            assert result.get(timeout=10) == len(batch)
         """
     )
     environment = os.environ.copy()
@@ -67,7 +68,7 @@ def test_import_preserves_tokenizers_fork_safety():
         capture_output=True,
         env=environment,
         text=True,
-        timeout=15,
+        timeout=30,
         check=False,
     )
     assert result.returncode == 0, result.stderr
--- a/tests/unit_tests/data/datasets/test_gpt_sft.py
+++ b/tests/unit_tests/data/datasets/test_gpt_sft.py
@@ -13,6 +13,10 @@
 # limitations under the License.
 
 import json
+import os
+import subprocess
+import sys
+import textwrap
 from pathlib import Path
 from unittest.mock import MagicMock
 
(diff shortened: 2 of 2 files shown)


janbernloehr:
Can we cherry pick this into `r0.6.0` because this regressed over the `v0.6.0` release

PR #5364, opened 2026-08-06
[step35] [work around] skip roundtrip weight verification for MTP

shifangx:
# What does this PR do ?
  # Background
MTP `shared_head.output.weight` is tied to the main `lm_head` in Megatron-Core, but for Step-3.5-Flash,  model.layers.45.transformer.shared_head.output.weight has different value comapred to lm_head.weight.
As shown in the following:
```
Model dir: stepfun-ai/Step-3.5-Flash

[A] model.layers.45.transformer.shared_head.output.weight
    shard=model-00002.safetensors  shape=(128896, 4096)  dtype=torch.bfloat16
[B] lm_head.weight
    shard=model-00002.safetensors  shape=(128896, 4096)  dtype=torch.bfloat16

=== comparison ===
torch.equal (same dtype)      : False
value-exact (fp32 upcast)     : False
max |A-B|                     : 0.2905464172363281
mean |A-B|                    : 0.02121051587164402
differing elements            : 527488999 / 527958016 (99.9112%)
allclose(atol=1e-1,rtol=1e-5) : False
```
So we need to skip roundtrip weight verification for MTP and align the Step-3.5-Flash example scripts/docs with the validated 4-GPU (EP=4) convert + inference workflow.

  # Changelog

  - **Roundtrip conversion:** MTP `shared_head.output.weight` is tied to the main `lm_head` in Megatron-Core, so comparing it during HF↔Megatron roundtrip verification is meaningless. Skip that comparison and emit a warning instead of aborting the conversion.
  - **`examples/models/stepfun/step35/conversion.sh`:** switch from the single-process `convert.sh import` to the multi-GPU roundtrip converter (`hf_megatron_roundtrip_multi_gpu.py`) run under `torch.distributed.run --nproc_per_node=4` with `TP=1 / PP=1 / EP=4`; add a skip-if-exists guard so an existing checkpoint is not re-converted.
  - **`examples/models/stepfun/step35/inference.sh`:** default to 4 GPUs with `EP=4` (was 8 / `EP=8`); update the default prompt to `"Explain hyper-connections in transformer models."`.
  - **`examples/models/stepfun/step35/README.md`:** update the import command to the 4-GPU roundtrip invocation, change the inference section to 4 GPUs / `EP=4`, and refresh the "Expected output" block with the current generation.

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)

### DIFF
--- a/examples/conversion/hf_megatron_roundtrip_multi_gpu.py
+++ b/examples/conversion/hf_megatron_roundtrip_multi_gpu.py
@@ -83,6 +83,18 @@ IGNORE_PRECISION_PARAMS = [
 # FP8 dtypes whose dequantisation is inherently lossy — allclose is meaningless.
 _FP8_DTYPES = {torch.float8_e4m3fn, torch.float8_e5m2}
 
+# MTP (multi-token-prediction) tied output heads — skip allclose.
+# Megatron-Core defines MTP with a SINGLE shared output layer that is tied to the
+# main model output layer (HF ``lm_head.weight``). On export every MTP head is
+# therefore materialized as a copy of ``lm_head``. The published HF checkpoint
+# instead stores a *separate* per-layer tensor
+# (``model.layers.<N>.transformer.shared_head.output.weight``) whose values
+# legitimately differ from ``lm_head``. Comparing the exported (tied) copy against
+# the HF per-layer tensor is meaningless, so skip it like the FP8 case. The main
+# LM head is ``lm_head.weight`` (not ``shared_head.output.weight``) and is still
+# verified normally.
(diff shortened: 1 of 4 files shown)


shifangx:
/ok to test [80cf0d0](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5364/commits/80cf0d054062d4bb1ea0844e5fb6da03e2b0e240)

shifangx:
This pr is only a work around to convert Step35 HF weights to megatron format without MTP layers.
But what we need to support untied head weight for MTP layers.

PR #5374, opened 2026-08-06
fix(modelopt): handle sparse export task lists

yaoyu-33:
## Summary

- preserve the existing sparse `build_conversion_tasks` contract and leave non-ModelOpt conversion paths unchanged
- restore the ModelOpt metadata guard for `None` task slots
- materialize concrete tasks once at the ModelOpt export-plan boundary, retaining every real task after a sparse hole
- add a real producer-to-consumer regression covering `mapped -> unmapped -> mapped`

This fixes NVBug 6565708.

## Root cause

`build_conversion_tasks` preserves the shared pipeline-parallel name order and may leave `None` for a global name without a conversion mapping. Existing callers such as the MIMO orchestrator explicitly recognize that sentinel.

#4566 narrowed the ModelOpt helper types and removed the existing `task is None` guard without changing the producer. The new export-plan prepass therefore dereferenced a sparse slot before normal export processing.

The fix restores the sentinel handling removed by #4566 and converts the sparse list to concrete ModelOpt tasks once before indexing, mapping, and hook construction. It does not catch `AttributeError`, alter the producer, or change other conversion paths.

## Tests

- `uv run python -m pytest tests/unit_tests/models/test_model_bridge.py tests/unit_tests/models/test_modelopt_utils.py` (85 passed)
- focused red check reproduces `AttributeError` at `task.megatron_module` without the sentinel fix

### DIFF
--- a/src/megatron/bridge/models/conversion/modelopt_utils.py
+++ b/src/megatron/bridge/models/conversion/modelopt_utils.py
@@ -477,7 +477,7 @@ def _build_hf_modelopt_pre_ep_quant_metadata(
 
 
 def collect_modelopt_quant_metadata(
-    conversion_tasks: list[WeightConversionTask],
+    conversion_tasks: list[WeightConversionTask | None],
 ) -> dict[str, QuantMeta]:
     """Collect ModelOpt quantization metadata from conversion task modules."""
     from modelopt.torch.export import quant_utils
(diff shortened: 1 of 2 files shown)


aroshanghias-nvd:
@cuichenx can you review this?

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5374/commits/f61f4ac37e268e6e5e2fa93c44d691904f0839b2

PR #5376, opened 2026-08-06
fix(peft): preserve grouped expert SwiGLU LoRA order

yaoyu-33:
# What does this PR do?

Preserves exact per-expert LoRA tensor ordering when gated grouped-expert `linear_fc1` adapters are loaded from distributed checkpoints. Related to NVBug 6565837.

# Changelog

- Remove a second gate/up de-interleave from the PEFT grouped-expert SwiGLU merge callback. Its build callback already emits pure gate and pure up shards in merge order.
- Replace the synthetic fused-shard merge test with an exact build-to-merge identity regression for two experts at ETP=1 and ETP=2.
- Keep the change scoped to `megatron.bridge.peft`; generic checkpoint code and checkpoint formats are unchanged.

# Root cause and PR #5366

The factory build callback splits each local fused `[gate|up]` tensor into pure gate and up `ShardedTensor` values. The merge callback incorrectly split those pure tensors again, moving the middle two quarters of each expert's fused axis.

Merged PR #5366 does not affect this path. It changes low-memory GLU normalization in `training/checkpointing.py`; this defect is in the separate PEFT factory in `peft/utils.py`.

# Validation

The unchanged regression failed on current `main` before the fix with 16 of 32 elements moved (50%) and passed after the fix. The deterministic tensor labels encode expert, gate/up projection, row, and column independently.

```text
uv run python -m pytest tests/unit_tests/peft/test_utils.py -k grouped_expert_linear_fc1 -q --tb=short
3 passed, 99 deselected

Two-rank EP=2 distributed checkpoint save/load with exact tensor equality
Passed

uv run pre-commit run --all-files
Passed
```

# GitHub Actions CI

Draft PR; CI can be triggered after review.

- [x] Added focused red-to-green regression coverage.
- [x] Documentation changes are not needed for this internal checkpoint mapping correction.
- [x] No optional dependency behavior is changed.

# Additional Information

- NVBug 6565837

### DIFF
--- a/src/megatron/bridge/peft/utils.py
+++ b/src/megatron/bridge/peft/utils.py
@@ -1621,18 +1621,6 @@ def _apply_grouped_expert_swiglu_sharded_factory(
         ]
 
     def sh_ten_merge_fn(sub_state_dict):
-        if not singleton_local_shards and len(sub_state_dict) > 1:
-            # Dist checkpoint load reconstructs one local fused shard per expert-TP
-            # rank, so the incoming tensors look like [gate_0|up_0, gate_1|up_1, ...].
-            # Restore the fused [gate_0, gate_1, ..., up_0, up_1, ...] layout before
-            # concatenating back along the SwiGLU axis.
-            gate_parts = []
-            up_parts = []
-            for tensor in sub_state_dict:
-                gate_part, up_part = torch.chunk(tensor, 2, dim=swiglu_shard_axis)
-                gate_parts.append(gate_part)
-                up_parts.append(up_part)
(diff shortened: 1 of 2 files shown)


yaoyu-33:
The only failed check is unrelated to this PEFT change: `L0_Launch_models_gemma4_vl` passed 5 conversion cases, then `test_gemma4_dense_vl_autoconfig_roundtrip` hit a transient Gloo bind failure (`EADDRINUSE`, port 56485). Core unit tests and all PEFT-relevant checks passed. I reran the failed job: https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/31122540378/job/92919103863

PR #5379, opened 2026-08-06
fix(peft): preserve shared expert LoRA values across EP reshards

yaoyu-33:
# What does this PR do ?

Preserve every shared grouped-expert LoRA checkpoint shard during load so EP resharding computes the intended expert-parallel mean instead of selecting one rank value.

# Changelog

- Mark Bridge training, direct adapter load/export, and standalone adapter load workflows with load-time sharding metadata.
- Give every expert slot an independent CPU checkpoint destination during reshard loads, avoiding aliasing and extra GPU staging pressure.
- Reuse the original buffer for save and known unchanged-topology loads, keeping ordinary resumes bounded in host memory.
- Average normal and SwiGLU expert shards with a bounded-memory FP32-or-higher accumulator and cast once, preventing low-precision drift.
- Read source EP topology from caller metadata, the run configuration, or legacy checkpoint args; reduce local means across destination EP ranks only when topology changes.
- Preserve local-checkpoint resumes and detect legacy 2D shared adapters before validating source EP topology.
- Reject non-legacy EP>1 adapter loads when source topology is genuinely unavailable instead of silently producing rank-dependent weights.
- Preserve MSC-backed checkpoint paths and tolerate malformed run configs while locating fallback metadata.
- Add focused regression coverage for FC1 and FC2 shared-expert adapters across both `linear_in.weight` and `linear_out.weight`, same-topology loads, cross-EP reduction, BF16 stability, legacy detection, and adapter load/export entry points.
- Confirm the direct merge remains an arithmetic mean; the original defect was buffer aliasing, not merge ordering or reduction semantics.

# GitHub Actions CI

Validated after independent review:

- Focused `uv run python -m pytest` selection — 45 passed.
- Tiny-model distributed checkpoint check — EP=2 save → EP=2 load retained each rank value; EP=2 save → EP=1 load produced exact means `2`, `9`, `24`, and `38` for all four shared-expert adapter tensors.

**Pre checks**:

- [x] Make sure you read and followed the Contributor guidelines.
- [x] Did you add or update any necessary documentation? No documentation change is needed because this restores existing checkpoint semantics without changing user-facing APIs or configuration.
- [x] Does the PR affect components that are optional to install? No.
  - [x] Reviewer: Does the PR have correct import guards for all optional libraries? Not applicable.

# Additional Information

- NVBug 6565838
- Fail-before used distinct EP-rank values across every affected tensor: FC1 `linear_in` expected `2`, observed `3`; FC1 `linear_out` expected `9`, observed `7`; FC2 `linear_in` expected `24`, observed `27`; FC2 `linear_out` expected `38`, observed `35`. This confirms last-writer buffer aliasing.

### DIFF
--- a/examples/conversion/adapter/export_adapter.py
+++ b/examples/conversion/adapter/export_adapter.py
@@ -54,6 +54,7 @@ from megatron.bridge import AutoBridge
 from megatron.bridge.peft.lora import LoRA, VLMLoRA
 from megatron.bridge.peft.utils import enable_legacy_shared_expert_adapter_loading
 from megatron.bridge.training.checkpointing import (
+    _checkpoint_expert_parallel_size,
     _generate_model_state_dict,
     _model_sharded_state_dict_load_metadata,
     apply_peft_adapter_filter_to_state_dict,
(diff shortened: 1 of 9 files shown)


yaoyu-33:
Superseded by #5396, which fixes the shared expert LoRA initialization invariant directly. Closing this checkpoint-mean approach in favor of the root-cause fix.

PR #5389, opened 2026-08-06
feat(data): infer assistant loss mask for gpt-oss Harmony chat template

cuichenx:
## What

Adds an OpenAI Harmony (gpt-oss) candidate to `infer_assistant_mask_boundary_config` so chat SFT with `loss_mode="assistant"` produces a correct assistant-only loss mask for gpt-oss tokenizers.

## Why

gpt-oss ships a chat template **without** `{% generation %}` blocks, so HF's `return_assistant_tokens_mask` path yields nothing. Both text chat entry points — `text_chat_collate_fn` (direct HF SFT) and `GPTSFTChatDataset` — then fall through to `build_assistant_loss_mask` with an inferred boundary config. Since the inferer only recognized ChatML / Kimi / Gemma / Llama markers, gpt-oss matched none and raised `"Unable to build assistant loss mask..."`. This makes gpt-oss text SFT work out of the box with no call-site changes.

## How

The Harmony assistant header (`<|start|>` + role) is emitted separately in the template, so the concatenated `<|start|>assistant` is not a literal substring. Detection therefore keys off the standalone structural specials `<|start|>`, `<|channel|>`, `<|message|>` (Harmony-unique — no collision with the existing ChatML/Llama candidates).

- `assistant_start = "<|start|>assistant"` → the loss span begins at `<|channel|>`, so **every channel of the turn is trained, including the `analysis` (reasoning) channel**.
- End variants: `<|return|>` (primary, final generation), with `<|end|>` (message) and `<|call|>` (tool call) fallbacks.

The per-turn scanner already handles one assistant turn rendering as multiple channel segments (`analysis` then `final`): it bounds the turn to a single window and masks from the first `<|start|>assistant` content through the **last** end token, so the two blocks are masked contiguously.

## Tests

`tests/unit_tests/data/test_conversation_processing.py`:

1. `test_infer_assistant_mask_boundary_config_from_gpt_oss_harmony_template` — asserts the derived start/end tokens and that detection works even though `<|start|>assistant` is not a literal template substring.
2. `test_gpt_oss_harmony_assistant_mask_covers_analysis_and_final_channels` — runs `tokenize_chat_example` end-to-end over a two-channel (analysis + final) assistant turn and asserts the analysis chain-of-thought **and** the final answer are both in the loss, while the user turn and the assistant priming header (`<|start|>assistant`) are not. The fake tokenizer renders the Harmony byte structure independently of the masking logic, so the assertions exercise the real mask builder rather than restating it.

## Notes / follow-ups

- I could not run the tokenizer against a real gpt-oss checkpoint here. Two things worth confirming against the actual `openai/gpt-oss-*` tokenizer: (a) `"<|start|>assistant"` tokenizes as a clean prefix of the rendered stream, and (b) the terminal token per turn is covered by `<|return|>` / `<|end|>` / `<|call|>`.
- Includes the `analysis` channel in loss by design (matches gpt-oss training). If a future caller wants final-only, that's a separate template-`{% generation %}` route.


cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5389/commits/c0e62c615627aa9253cfafd15571dbc1f809c2d3

yaoyu-33 (approved):

maanug-nv (approved):

PR #5395, opened 2026-08-06
feat(training): enable MFSDP V2

wujingyue:
## What

- Bump Megatron-LM to `59b72fa57` and regenerate `uv.lock` for MFSDP V2 support.
- Keep MFSDP V1 as the default while making `ddp.megatron_fsdp_version=2` opt into the experimental V2 path.
- Enforce V2's supported DP-only BF16 configuration and reject checkpointing, gradient clipping, and other unsupported features early.
- Add a two-GPU, ten-step dense `HybridModelProvider` smoke test using `hybrid_layer_pattern="**"` and `clip_grad=0.0`.
- Remove unused temporary-directory fixtures from the non-checkpoint MFSDP tests; retain the checkpoint-resume test's real temporary directory.

## Why

MFSDP V2 has a distinct wrapper and optimizer contract from V1. Bridge must select its safe configuration explicitly and avoid V1-only assumptions.

## Validation

- Two-GPU functional smoke: existing MFSDP V1 basic training and the new MFSDP V2 dense-Hybrid test.

Related to NVIDIA/Megatron-LM#6325.

### DIFF
--- a/tests/functional_tests/test_groups/megatron_fsdp/test_megatron_fsdp.py
+++ b/tests/functional_tests/test_groups/megatron_fsdp/test_megatron_fsdp.py
@@ -82,8 +82,8 @@ class Llama3FSDPTestModelProvider(GPTModelProvider):
 
 
 @dataclass
-class DenseHybridFSDPV2TestModelProvider(HybridModelProvider):
-    """Small dense HybridModel configuration for the MFSDP V2 smoke test."""
+class DenseHybridSmokeModelProvider(HybridModelProvider):
+    """Small dense HybridModel configuration for the training smoke test."""
 
     normalization: str = "RMSNorm"
     activation_func: Callable = F.silu
@@ -133,8 +133,8 @@ def create_fsdp_model_config(seq_length: int, bf16: bool = True, **kwargs) -> Ll
     return Llama3FSDPTestModelProvider(**base_config)
 
 
-def create_dense_hybrid_fsdp_v2_model_config(**kwargs) -> DenseHybridFSDPV2TestModelProvider:
-    """Create the dense two-layer HybridModel configuration used by MFSDP V2."""
+def create_dense_hybrid_smoke_model_config(**kwargs) -> DenseHybridSmokeModelProvider:
+    """Create the dense two-layer HybridModel configuration used by the smoke test."""
     base_config = {
         "tensor_model_parallel_size": 1,
         "pipeline_model_parallel_size": 1,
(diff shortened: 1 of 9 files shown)


3rdparty/Megatron-LM
wujingyue, line 1: This PR requires to sync past https://github.com/NVIDIA/Megatron-LM/pull/5865. Should I wait for the next time you bump up the Megatron-LM version?

uv.lock
wujingyue, line 1: Not sure where this change came from. Probably an artifact from updating 3rdparty/Megatron-LM.

wujingyue:
cc @rapatel

3rdparty/Megatron-LM
cuichenx: can this wait until #5431 is merged? sorry this bump is taking a while

wujingyue: No problem at all!

tests/functional_tests/test_groups/megatron_fsdp/test_megatron_fsdp.py
```diff
@@ -479,7 +540,7 @@ def test_fsdp_pretrain_save_resume(self, tmp_path):
             clear_directories(shared_base_dir)
 
     @pytest.mark.run_only_on("GPU")
-    def test_fsdp_precision_aware_optimizer_decoupled_grad_consistency(self, tmp_path):
```
wujingyue, line 482: I removed tmp_path because it's not used by the test. As a result, I also removed the try-finally.

3rdparty/Megatron-LM
wujingyue: Hey @cuichenx and @yaoyu-33, I rebased this PR to main. Will you review this?

cuichenx (commented):
i would suggest introducing a fsdp_compat.py or mcore_compat.py file, and then do the try except FSDP import and MEGATRON_FSDP_TYPES in that one place instead of everywhere. Do you think that makes sense? Otherwise LGTM

wujingyue:
> i would suggest introducing a fsdp_compat.py or mcore_compat.py file, and then do the try except FSDP import and MEGATRON_FSDP_TYPES in that one place instead of everywhere. Do you think that makes sense? Otherwise LGTM

Makes sense -- I've made that change in the latest commit. I'm not sure about the new dependency from src/megatron/bridge/models/conversion/utils.py to src/megatron/bridge/training/; let me know if you see a better approach. cc @cuichenx

cuichenx (approved):
LGTM

PR #5396, opened 2026-08-06
fix(peft): synchronize shared expert LoRA across EP

yaoyu-33:
# What does this PR do ?

Keep shared expert LoRA parameters bitwise identical across expert-parallel ranks from initialization through backward.

# Changelog

- Broadcast CUDA-initialized shared grouped-expert LoRA weights from expert-parallel group rank zero.
- Stage CPU-initialized weights through the existing NCCL expert-parallel group, then copy the synchronized values back to CPU.
- Reject unmaterialized meta parameters instead of silently leaving rank-local values.
- Retain the existing SUM gradient synchronization across expert-parallel ranks.
- Add normal unit-CI coverage plus two-rank exact-parity coverage for FC1 and FC2 adapters, including every linear_in and linear_out parameter and nonzero gradient.

# Root cause

Expert LoRA layers use the expert RNG tracker, whose seed differs by expert-parallel rank. CPU initialization can also use rank-dependent seeds. The existing gradient all-reduce applies identical updates but cannot remove an initialization offset.

Before the fix, the regression test found all 32 LoRA-A elements mismatched across two expert-parallel ranks. After the fix, CUDA initialization, deliberately rank-divergent CPU initialization, and nonzero gradients match exactly for both shared expert adapter layouts.

# GitHub Actions CI

Targeted validation:

- `uv run --no-sync --with pytest python -m torch.distributed.run --standalone --nproc_per_node=2 -m pytest tests/unit_tests/peft/test_shared_expert_lora_distributed.py -vv` (3 passed)
- `uv run --no-sync --with pytest python -m pytest` on the focused normal-CI init/CPU-init/gradient tests (3 passed)
- Existing shared-expert sharded-state, SwiGLU, extra-state, and gradient-sync tests (4 passed)
- `uv run --no-sync --active pre-commit run --all-files` (passed)

- [x] Added focused regression tests.
- [x] No documentation update is needed for this correctness fix.
- [x] No optional-install component behavior is changed.

# Additional Information

Related to NVBug 6565838.

### DIFF
--- a/src/megatron/bridge/peft/utils.py
+++ b/src/megatron/bridge/peft/utils.py
@@ -1293,7 +1293,7 @@ class ParallelLinearAdapter(nn.Module):
         weights = [
             weight
             for module in (self.linear_in, self.linear_out)
-            if isinstance(weight := getattr(module, "weight", None), torch.Tensor) and weight.is_cuda
+            if isinstance(weight := getattr(module, "weight", None), torch.Tensor)
         ]
         if not weights:
             return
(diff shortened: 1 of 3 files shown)


jstjohn (approved):

src/megatron/bridge/peft/utils.py
```diff
@@ -1281,6 +1282,37 @@ def _allreduce_shared_expert_grad(self, grad: torch.Tensor) -> torch.Tensor:
         torch.distributed.all_reduce(grad, group=self.ep_group)
         return grad
 
+    def _synchronize_shared_expert_parameters(self) -> None:
```
jstjohn, line 1285: A lot of other megatron layers that support parallelism get around this kind of pattern using seeded initializers that make the same weights on every rank (where appropriate), right?

jstjohn: Curious why that doesn't automatically happen here.

cuichenx (approved):

PR #5397, opened 2026-08-06
GPT-OSS 20B Verification Card

maanug-nv:
# What does this PR do ?

Adds model verification card for GPT-OSS.

# Changelog

# GitHub Actions CI

**Pre checks**:

# Additional Information

- Related to # (issue)


maanug-nv:
For GPT-OSS 20B long-context SFT, the base `gpt_oss_20b_sft_8gpu_h100_bf16_config` uses GBS 128. At 32K sequence length with offline-packed Tulu3 `train[:10000]`, the packed dataset produced 120 packed rows, so the base GBS 128 configuration failed before training with:

`train dataset size (120) < global batch size (128)`

GBS 128 at 32K sequence length is also a very large single-node validation workload: 4,194,304 token slots per optimizer step. On H100 this would take several hours for a bounded verification run.

I could not directly add `--global-batch-size 32` to the command in the card, because the model verification card validator intentionally rejects training-card commands that override global or micro batch size. Batch size is treated as part of the recipe-owned convergence contract, not as a card-only launch override.

Therefore, I added a new 32k sequence length, GBS=32 recipe. 

The verified long-context run completed successfully with GBS 32. To keep the card command compliant and reproducible, this PR adds `gpt_oss_20b_sft_8gpu_h100_bf16_32k_config`, inheriting from the existing 20B SFT recipe and only moving the sequence/packing length to 32768 and recipe-owned GBS to 32.

The remaining long-context execution settings, such as TP4/CP2, sequence parallelism, non-fused CE, and full recompute, are still explicit in the verification card command because they were execution/memory settings required for the H100 validation run.

src/megatron/bridge/models/gpt_oss/gpt_oss_bridge.py
```diff
@@ -60,7 +60,7 @@ def provider_bridge(self, hf_pretrained: PreTrainedCausalLM) -> GPTModelProvider
         provider.normalization = "RMSNorm"
         provider.gated_linear_unit = True
         provider.add_bias_linear = True
-        provider.add_qkv_bias = False
+        provider.add_qkv_bias = getattr(hf_pretrained.config, "attention_bias", True)
```
cuichenx, line 63: i think this is correct but unnecessary since we already we add_bias_linear = True?

cuichenx (approved):

PR #5398, opened 2026-08-07
perf: let FORCE_PCT_BINDING override the B300 PCT binding auto-disable

rsalagame-nvidia:
By Default, on Bia B300, all workloads will have PCT-enabled with the changes below.


bdubauski (approved):
LGTM. Simpler approach with PCT binding disabled by default

PR #5400, opened 2026-08-07
fix(data): collate ragged packed sequence metadata

yaoyu-33:
## What does this PR do?

Fixes NVBug 6570726, a regression introduced by #5344 where offline-packed SFT fails in `GPTSFTPackedDataset.collate_fn` before training step 1 when packed rows contain different numbers of logical sequences.

The finetuning batch sampler intentionally sends the full global batch assigned to each data-parallel rank through `collate_fn`, then splits the collated tensors into MBS1 microbatches. Therefore MBS1 does not imply that cumulative-sequence metadata is rectangular at collate time.

This change extends each logical and physical cumulative-offset row to the batch-local maximum boundary count by repeating its terminal offset. The repeated offsets represent zero-length sequences, matching the existing static `pad_cu_seqlens` convention while preserving the direct MCore/TE `cu_seqlens_q` / `cu_seqlens_kv`, optional `*_padded`, and per-row `max_seqlen_q` / `max_seqlen_kv` contract.

## Testing

- Added a focused regression through the real packed SFT `collate_fn` with two rows containing different logical sequence counts.
- Confirmed the test fails on unmodified main with `ValueError: expected sequence of length 4 at dim 1 (got 2)`.
- `uv run python -m pytest` focused packed-dataset, logical/physical boundary, batch-splitting, and packed-sequence utility coverage: 34 passed.

An optional small 2-GPU packed-SFT smoke was attempted but the allocation was canceled before Python execution, so it is not counted as validation evidence.

## Tracking

- NVBug 6570726
- Regression from #5344 / e4cb868

### DIFF
--- a/src/megatron/bridge/data/packing/gpt_sft.py
+++ b/src/megatron/bridge/data/packing/gpt_sft.py
@@ -275,8 +275,18 @@ class GPTSFTPackedDataset(GPTSFTDataset):
         }
 
         if self.return_cu_seqlen:
-            # Offline packed SFT requires micro-batch size 1, so no sentinel padding is
-            # needed to batch variable-length metadata. Emit MCore/TE field names directly.
+            # The DataLoader collates every row assigned to this data-parallel rank before
+            # the training loop splits that global batch into MBS1 microbatches. Extend the
+            # boundary arrays with zero-length sequences so ragged rows form tensors while
+            # retaining the direct MCore/TE cumulative-offset contract.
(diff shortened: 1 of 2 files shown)


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5400/commits/43dc2b20ca22b2b1cfa06e734432671a148a8e64

PR #5406, opened 2026-08-07
fix(training): preserve CLI mixed precision overrides

### LINKED ISSUE #5365: [bug] ddp.grad_reduce_in_fp32 CLI override is silently overwritten by mixed_precision.setup()
### Problem

The following `grad_reduce_in_fp32` CLI override has no effect:

```bash
torchrun ... /opt/Megatron-Bridge/scripts/training/run_recipe.py \
    --recipe nemotron_3_nano_sft_config \
    mixed_precision=bf16_mixed \
    ddp.grad_reduce_in_fp32=False
```

The final training run still uses `grad_reduce_in_fp32=True`. No warning is emitted.

My workaround was to register a new named recipe via monkey-patch appended to `mixed_precision.py` at container startup:

```python
@register
def bf16_mixed_without_grad_reduce_in_fp32() -> MixedPrecisionConfig:
    return MixedPrecisionConfig(
        bf16=True,
        params_dtype=torch.bfloat16,
        pipeline_dtype=torch.bfloat16,
        autocast_enabled=False,
        grad_reduce_in_fp32=False,
    )
```

Then, 

```bash
torchrun ... recipe_runner.py ... mixed_precision=bf16_mixed_without_grad_reduce_in_fp32
```

## Root cause analysis (Claude)

Repo at commit #8dbc2977d

The execution order defeats CLI overrides on ddp fields:

1. `apply_cli_overrides()` (`recipe_runner.py`:479) sets `ddp.grad_reduce_in_fp32=False`.
2. Later, `runtime_config_update()` (`config.py`:1839–1843) resolves the `mixed_precision` string → `MixedPrecisionConfig` and calls `cfg.mixed_precision.setup(cfg.model, cfg.optimizer, cfg.ddp)`.
3. `setup()` → `update_config_with_precision_overrides()` (`mixed_precision.py`:148–155) unconditionally overwrites all matching ddp fields — including `grad_reduce_in_fp32`. `bf16_mixed` has `grad_reduce_in_fp32=True`, so the CLI value is silently clobbered.

Pure CLI alternatives don't help: `mixed_precision.grad_reduce_in_fp32=False` fails because `mixed_precision` is a scalar string node in Hydra. The `_target_` instantiation path breaks on `torch.dtype` fields (`params_dtype`, `pipeline_dtype`) that arrive as strings in the constructor.

### Proposed Fixes

Option 1 (preferred): Add a `skip_fields` set to `update_config_with_precision_overrides()`, populated from fields explicitly set via CLI, so user intent always wins.

Option 2: Apply `mixed_precision.setup()` before `apply_cli_overrides()` so CLI overrides always have the last word. Smaller diff, but changes semantics for all `runtime_config_update()` interactions.

Option 3 (short-term): Upgrade the log in `update_config_with_precision_overrides()` from DEBUG to WARNING when overwriting a field. Doesn't fix the bug but makes it visible.

### Minimal repro

```shell
srun -l ... torchrun ... \
    /opt/Megatron-Bridge/scripts/training/run_recipe.py --recipe nemotron_3_nano_sft_config \
        mixed_precision=bf16_mixed \
        ddp.grad_reduce_in_fp32=False
        # Tried too and not working: mixed_precision.grad_reduce_in_fp32=False
```

### Expected behavior

grad_reduce_in_fp32 is really set to False.

### Affected area

area:recipe

### Regression?

Not sure

### Environment

Tested with 26.06 GA container (and an ealier rc container).
Claude analysis on main branch (commit #ab2a992)

### Logs

```shell

```

aroshanghias-nvd:
## Summary

- retain leaf paths explicitly assigned through Hydra CLI overrides
- preserve matching model, optimizer, and DDP values across mixed-precision runtime setup
- cover both direct and mapping-style Hydra override syntax

## Root cause

CLI overrides are applied before `runtime_config_update()`. The later `MixedPrecisionConfig.setup()` call copied preset values into matching target fields and silently replaced explicit CLI values such as `ddp.grad_reduce_in_fp32=False`.

The fix snapshots only direct CLI-assigned fields on `model`, `optimizer`, or `ddp` that overlap `MixedPrecisionConfig`, runs the existing setup unchanged, and restores those explicit values before validation.

## Validation

- `srun` job `15273935`: 16 targeted unit tests passed
- exact `ddp.grad_reduce_in_fp32=False` regression passed
- mapping-style `ddp={grad_reduce_in_fp32:false}` regression passed
- full pre-commit hooks passed
- Ruff format and checks passed
- `git diff --check` passed

Closes #5365


adityavavreNVDA:
Verified as correct: the `_`-prefixed attribute can't leak into serialization (`config_utils.py:104` filters `_`-prefixed names); restore ordering is right, before `set_data_parallel_size()` and `validate()`; sweeps can't reach `override.value()` since Hydra raises inside the wrapped `try`. The `"." not in field_name` is correct. `update_config_with_precision_overrides` only touches direct attributes and never recurses, so the preserve surface exactly matches the clobber surface. Worth a comment saying so.

**1. The same bug is unfixed 25 lines below.** `comm_overlap.setup(cfg.model, cfg.optimizer, cfg.ddp)` runs after the restore, and `_apply_cfgs` (`comm_overlap.py:588-592`) is stricter than the function fixed here. Unconditional `setattr`, no value comparison, no logging at all. It writes `overlap_grad_reduce`, `overlap_param_gather`, `align_param_gather`, and `bucket_size` onto `ddp`, so `ddp.overlap_grad_reduce=False` is discarded exactly as `grad_reduce_in_fp32` was.

**2. The reported symptom is only half fixed.** The override now wins, but `mixed_precision.py:156` is still DEBUG, so you end up with `mixed_precision.grad_reduce_in_fp32=True` alongside `ddp.grad_reduce_in_fp32=False` which is  inconsistent, saved that way, with no signal. That was Option 3 in the issue; a WARNING when a preserved value diverges from the preset is a few lines.

**3. YAML config-file values are still clobbered** Provenance is only recorded under `if cli_overrides:`. Defensible since the issue was CLI-specific, just flagging that it should be a stated decision rather than an accident.

**4. Tests only cover one of three branches.** The fix guards `{"model", "optimizer", "ddp"}`, but both parametrize entries override `ddp`. Two more entries would close it, e.g. `model.bf16=False` and an optimizer field.

PR #5438, opened 2026-08-08
feat(data): blend several sources on the direct-HF SFT path

bzantium:
# What does this PR do ?

Let the direct Hugging Face SFT path draw its training rows from several sources at once, so a training mixture no longer has to be prepared outside Bridge.

# Changelog

- Let `DirectHFSFTDatasetConfig.source` take a list of sources as well as a single one, following the same shape as `GPTDatasetConfig.data_path`. A list blends the training rows.
- Add `source_weights` and `blend_seed`. Each weight is an epoch count for its source and defaults to one pass each.
- Add `resolve_blend_weights`, `blend_sft_rows`, and `load_and_blend_hf_datasets` to `data/sources/hf.py`, alongside the existing source loading helpers.
- Mix sources after they are adapted to canonical SFT rows, so sources whose raw column layouts differ can be blended when their schema adapters agree.
- Require explicit `validation_source` and `test_source` when several sources are given, since a blend has no single source to derive those splits from. A single source is unchanged.
- Add unit coverage for the blend arithmetic and the config surface, and a single-GPU functional training step over a blend.
- Document the field and its weight semantics in the data preparation guide.

## Weight semantics

Weights are epoch counts over each source: `1.0` keeps every row once, `2.5` keeps every row twice plus half of them again, and `0.5` keeps half of them, drawn without replacement.

They count rows, not tokens. A Megatron pretraining blend can read its weights as token shares because every sample it draws is `sequence_length` tokens, but SFT rows vary in length, so a row share is not a token share. A token-share blend needs fixed-size units and would have to follow per-source packing, which the direct Hugging Face path does not have yet.

# GitHub Actions CI

Local validation in the 26.04 container:

- `python -m pytest tests/unit_tests/data/builders tests/unit_tests/data/sources` gives 175 passed, covering the new cases plus the existing builder and source suites.
- `torchrun --nproc_per_node=1 -m pytest tests/functional_tests/test_groups/training/test_sft.py -k blended_sources` gives 1 passed, running one optimizer step over a two-source blend on a single GPU.
- `ruff check` and `ruff format --check` are clean.

The new cases cover the weight arithmetic, seed determinism, source interleaving, validation errors, and the config surface through serialization and CLI overrides.

**Pre checks**:

# Additional Information

- No new dependencies. The blend uses `random` from the standard library.
- Existing configs are unaffected: a single `source` keeps its current behavior, including validation and test derivation.
- Possible follow-up: once the direct Hugging Face path can pack per source, a token-share blend becomes expressible and would match the Megatron blend semantics more closely.

### DIFF
--- a/docs/training/data-preparation.md
+++ b/docs/training/data-preparation.md
@@ -189,6 +189,28 @@ dataset.preprocessing = PromptCompletionSFTPreprocessingConfig(
 
 Structured multi-turn rows require chat preprocessing; Bridge does not silently flatten them into prompt-completion text.
 
+To train on several sources at once, pass a list to `source`. Each entry in `source_weights` is an epoch count for its source: `1.0` keeps every row once, `2.5` keeps every row twice plus half of them again, and `0.5` keeps half of them. Fractional passes draw without replacement. Omitting the weights gives every source one pass.
+
+```python
+from megatron.bridge.data.builders import DirectHFSFTDatasetConfig, HFDatasetSourceConfig
+
+dataset = DirectHFSFTDatasetConfig(
+    seq_length=4096,
+    source=[
+        HFDatasetSourceConfig(dataset_name="squad"),
+        HFDatasetSourceConfig(dataset_name="gsm8k"),
+    ],
+    source_weights=[1.0, 2.0],
+    blend_seed=1234,
+    validation_source=HFDatasetSourceConfig(dataset_name="squad", split="validation"),
+    do_test=False,
+)
+```
+
+Sources are adapted to canonical SFT rows before mixing, so their raw column layouts do not have to match. A single source still derives its own validation and test splits; a list has no single split to derive from, so set `validation_source` and `test_source` explicitly or disable those splits.
+
+Weights count rows, not tokens. SFT rows vary in length, so sources with equal row counts can still contribute very different token counts.
+
 Known semantic datasets should use their preset name, for example `squad`, `gsm8k`, `openmathinstruct2`, `cord_v2`, `raven`, `rdr`, `medpix`, `cv17`, or `llava_video_178k`. Do not combine `dataset_name` with `path_or_dataset`, `subset`, or `schema_adapter`; a preset owns those coupled fields. `split`, `load_kwargs`, and `adapter_kwargs` remain available for split selection and declarative runtime options such as a video root. Presets validate published split support: `raven`, `rdr`, and `llava_video_178k` are train-only; `medpix` and `squad` have no test split; `gsm8k` has no validation split; and OpenMathInstruct-2 exposes training variants only. Disable unsupported derived validation/test splits or supply explicit compatible sources.
 
 For text chat, `hf_processor_path=None` reuses the training tokenizer only when that tokenizer already defines the intended chat template. Otherwise select a vocabulary-compatible instruction processor explicitly, as above.
--- a/docs/training/data-preparation.md
+++ b/docs/training/data-preparation.md
@@ -189,7 +189,7 @@ dataset.preprocessing = PromptCompletionSFTPreprocessingConfig(
 
 Structured multi-turn rows require chat preprocessing; Bridge does not silently flatten them into prompt-completion text.
 
-To train on several sources at once, pass a list to `source`. Each entry in `source_weights` is an epoch count for its source: `1.0` keeps every row once, `2.5` keeps every row twice plus half of them again, and `0.5` keeps half of them. Fractional passes draw without replacement. Omitting the weights gives every source one pass.
+To train on several sources at once, pass a list to `source`. Each entry in `source_weights` is an epoch count for its source: `1.0` keeps every row once, `2.5` keeps every row twice plus half of them again, and `0.5` keeps half of them. Fractional passes draw without replacement and round up to a whole row. Omitting the weights gives every source one pass.
 
 ```python
 from megatron.bridge.data.builders import DirectHFSFTDatasetConfig, HFDatasetSourceConfig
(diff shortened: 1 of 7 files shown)


yaoyu-33 (dismissed):

src/megatron/bridge/data/sources/hf.py
```diff
@@ -324,3 +325,98 @@ def load_and_adapt_hf_dataset(source: HFDatasetSourceConfig) -> list[dict[str, A
...
+            is not positive.
+    """
+    if not sources:
+        raise ValueError("A blend must name at least one source.")
+    if weights is None:
+        return [1.0] * len(sources)
+    if len(weights) != len(sources):
+        raise ValueError(f"Blend sources and weights must be equal in number, got {len(sources)} and {len(weights)}.")
+    for weight in weights:
+        if weight <= 0:
```
yaoyu-33: Please reject non-finite weights here (for example, by converting once and checking `math.isfinite`). `NaN <= 0` and `inf <= 0` are both false, so `config.validate()` accepts these values; dataset construction then fails later at `int(weight)` with a low-level `ValueError`/`OverflowError`. Add focused `NaN` and infinity cases so invalid values fail at the configuration boundary with the documented blend-weight error.

src/megatron/bridge/data/sources/hf.py
```diff
@@ -324,3 +325,98 @@ def load_and_adapt_hf_dataset(source: HFDatasetSourceConfig) -> list[dict[str, A
...
+    """
+    rng = random.Random(seed)
+    blended: list[dict[str, Any]] = []
+    for index, (rows, weight) in enumerate(zip(rows_per_source, weights)):
+        if not rows:
+            raise ValueError(f"Blend source {index} produced no rows.")
+        full_passes = int(weight)
+        for _ in range(full_passes):
+            blended.extend(rows)
+        remainder = round(len(rows) * (weight - full_passes))
```
yaoyu-33: Please make the fractional-row rounding contract explicit and prevent a positive source weight from silently contributing zero rows. For a one-row source, `weight=0.5` reaches `round(0.5) == 0` (banker's rounding); in a multi-source blend the call succeeds but that source disappears entirely despite passing the positive-weight validation. Either retain at least one row for every positive source or reject a weight/source-size combination that rounds to zero, and cover the one-row case.

yaoyu-33 (commented):
Focused validation found two blend-arithmetic edge cases that should be fixed and covered before merge. The distributed sampler/resume path and the one-step packed GPU smoke otherwise behaved as expected.

src/megatron/bridge/data/builders/direct_hf_sft.py
```diff
@@ -51,17 +53,29 @@
...
+
+    ``source`` takes one source or a list of them. A list blends the training
+    rows across the sources, weighted by ``source_weights``.
     """
 
     seq_length: int
-    source: HFDatasetSourceConfig
+    source: HFDatasetSourceConfig | list[HFDatasetSourceConfig]
+    source_weights: list[float] | None = None
+    blend_seed: int = 1234
```
yaoyu-33, line 78: Please validate `blend_seed` at the configuration boundary (at minimum, reject `None` and other non-integers). Direct Python construction currently accepts `blend_seed=None`, `config.validate()` succeeds, and `random.Random(None)` then reseeds from system entropy independently on each build/process. In a focused probe, eight builds from the same validated config produced eight distinct blend orders; separate DP workers can therefore disagree on the base-example order. Add a config-level regression case proving an invalid seed fails with an actionable error before dataset loading.

src/megatron/bridge/data/builders/direct_hf_sft.py
```diff
@@ -101,12 +115,50 @@ def validate(self) -> None:
...
+    def split_source(self) -> HFDatasetSourceConfig | None:
+        """The source that derives validation and test splits, if one is unambiguous."""
+        sources = self.training_sources
+        return sources[0] if len(sources) == 1 else None
+
+    def _validate_sources(self) -> None:
+        """Validate the training sources and any blend weights."""
+        # Any override re-serializes this config through OmegaConf, which returns
+        # dataclasses as plain mappings even when the override targeted another field.
+        if isinstance(self.source, list):
```
yaoyu-33, line 137: Please normalize override-provided `validation_source` and `test_source` mappings with `_as_source` as well. Starting from the ordinary scalar-source config, a blend can be introduced through Hydra with `~source` plus `+source=[...]`; setting the newly required split sources in that same override leaves both fields as plain `dict` values because their prior values were `None`. `config.validate()` then fails with `AttributeError: dict object has no attribute validate`. This means an override-configured blend cannot keep validation/test enabled even when both required sources were supplied. Please rebuild non-`None` split mappings before using them and add a regression test that starts with optional split fields unset, applies the full blend/split override, and validates successfully.

docs/training/data-preparation.md
```diff
@@ -189,6 +189,28 @@ dataset.preprocessing = PromptCompletionSFTPreprocessingConfig(
 
 Structured multi-turn rows require chat preprocessing; Bridge does not silently flatten them into prompt-completion text.
 
+To train on several sources at once, pass a list to `source`. Each entry in `source_weights` is an epoch count for its source: `1.0` keeps every row once, `2.5` keeps every row twice plus half of them again, and `0.5` keeps half of them. Fractional passes draw without replacement. Omitting the weights gives every source one pass.
```
yaoyu-33: Please mirror this new blend section into `docs/fern/versions/nightly/pages/training/data-preparation.mdx`. The Fern nightly page is checked in separately and currently jumps directly from prompt-completion guidance to named datasets, so the published guide omits `source_weights`, `blend_seed`, the required split-source behavior, and the example added here. This is not generated automatically in the PR: the preceding changes to this guide updated both files together, while the current Fern checks still pass despite the content drift.

yaoyu-33 (approved):
Verified the follow-up commit addresses all five review findings. Focused validation on the exact head passed 69 direct-HF source/builder tests, changed-file lint/format checks, and independent seeded-distribution plus two-rank sampler-resume probes. The prior GPU smoke covered the unchanged training path; fresh exact-SHA CI is still required.

PR #5448, opened 2026-08-08
fix(models): use HybridEP for MiniMax-M3 expert parallelism

yaoyu-33:
# What does this PR do?

Uses HybridEP as the MiniMax-M3 model-specific expert-parallel dispatcher default for inference and the public H100 pretrain/SFT recipes. This avoids the affected BF16 grouped all-to-all transport path and replaces the earlier inference-only FP32 transport workaround.

# Root cause

On the affected release runtime, finite MiniMax-M3 BF16 tensors become corrupted at variable-split EP all-to-all dispatch/combine and produce step-0 NaN logits. Isolation below MCore found stale or uninitialized `ncclIbRequest.send.onlyWriteImm` state in the SPCX grouped all-to-all path; requests can be misclassified, losing or corrupting payload data. The same model, checkpoint, prompt, and EP32 topology remain finite with NCCL internal IB or HybridEP.

This is not a prompt-formatting, checkpoint-conversion, router-math, or weight-roundtrip issue. It is also separate from the older cuBLAS router crash.

# Changes

- Configure the MiniMax-M3 Bridge mapping to use `flex` / `hybridep`, 16 dispatcher SMs, with HybridEP permute fusion disabled.
- Apply the same settings to the 256-H100 pretrain and 128-H100 packed-SFT recipes, including an 8-rank NVLink domain for H100 nodes.
- Remove the MiniMax-specific inference-only BF16-to-FP32 all-to-all dispatch/combine workaround and return to the standard MCore MoE layer.
- Keep MiniMax-M3's FP32 router projection and parameter behavior unchanged.
- Keep the generated-suffix-only VLM decoding fix from the earlier revision of this PR.
- Add provider/recipe assertions and update the MiniMax-M3 verification card. Training entries remain explicitly unverified.

Bridge-created MiniMax-M3 providers use the new configuration. Existing native checkpoints may still contain a serialized `alltoall` setting, so inference callers loading those checkpoints must also override the deserialized model config.

# Reproduction and controlled validation

The public reproduction shape is:

```bash
./scripts/inference/infer.sh --nodes 4 --gpus-per-node 8 \
  --task vlm-generation \
  --hf_model_path MiniMaxAI/MiniMax-M3 \
  --hf-revision 50942730318c7943fe83db7ec8e9f9177ecb1cf8 \
  --megatron_model_path <verified-imported-checkpoint> \
  --tp 1 --pp 1 --ep 32 --etp 1 \
  --image_path docs/images/tp1.png \
  --prompt "Read the green caption near the top of the image. Answer immediately with only the exact quoted term, including capitalization and hyphen; do not explain or reason." \
  --max_new_tokens 64 \
  --trust_remote_code
```

With the affected all-to-all path, step 0 produces NaN logits and empty/token-0 degeneration. With the four HybridEP model settings above and an 8-rank H100 NVLink domain, a matched 32-H100 TP1/PP1/EP32/ETP1 run observed 50/50 finite full-vocabulary logit vectors, no NaNs or infinities, no token 0, natural EOS, and the exact historical token sequence ending in `Pre-LayerNorm`.

This was a controlled configuration-level validation on clean Bridge/MCore source, not a training run and not a performance comparison.

# Validation and known dependency

- Model verification card schema: passed; its focused validator suite passed (18 tests).
- Focused MiniMax-M3 provider/recipe tests passed (37 tests).
- Focused VLM generation utility tests passed (13 tests).
- Full repository pre-commit hooks passed.
- The local GPU driver is older than the repository container runtime, so these local tests ran on CPU and do not substitute for GPU training validation.
- The earlier revision's generated-suffix decoding tests and full CI passed, but that prior CI result does not validate this new commit.

No backward pass, optimizer step, convergence run, or training performance run is claimed. The public recipes require 128 or 256 H100s, beyond the 32-GPU investigation ceiling.

Packed-SFT verification also requires the padding-safe dropless HybridEP routing changes merged to MCore dev in [NVIDIA/Megatron-LM#5542](https://github.com/NVIDIA/Megatron-LM/pull/5542). The current Bridge MCore main pin does not yet contain that complete change, so this PR should not merge until the applicable MCore main/backport pin is available and the bounded training checks pass.

# Before this PR is ready for review

- [x] Root cause isolated below MCore.
- [x] Deterministic real-model HybridEP inference validated on the target topology.
- [x] Bridge mapping, H100 training recipes, focused assertions, and verification card updated.
- [x] No MCore source, dependency, CI, or public API changed.
- [ ] Pin an MCore main revision with padding-safe HybridEP routing for packed SFT.
- [ ] Pass updated Bridge CI.
- [ ] Run bounded finite-loss pretrain/SFT validation before marking training verified.

### DIFF
--- a/examples/model_verification_cards/minimax-m3/card.yaml
+++ b/examples/model_verification_cards/minimax-m3/card.yaml
@@ -170,13 +170,31 @@ items:
       with only the exact quoted term, including capitalization and hyphen."
       The bounded completion remained coherent but did not reach the requested
       OCR target before the maximum; this result does not claim task accuracy
-      or natural EOS.
+      or natural EOS. MiniMax-M3 providers and H100 training recipes default
+      to the following model configuration when the runtime provides HybridEP:
+      moe_token_dispatcher_type="flex",
+      moe_flex_dispatcher_backend="hybridep",
+      moe_flex_dispatcher_num_sms=16, and
+      moe_permute_fusion_into_hybridep=false, with 8 HybridEP ranks per NVLink
+      domain on 8-GPU H100 nodes. Apply the same dispatcher overrides to both
+      the initial provider and the deserialized checkpoint model config;
+      changing only the provider does not replace the dispatcher saved in the
+      checkpoint. The public VLM command above does not yet expose checkpoint
+      dispatcher overrides, so the HybridEP result remains separate from its
+      command-level verification. A matched controlled validation of this
+      configuration kept every observed generation-logit vector finite with
+      zero NaNs and infinities, selected no token 0, reached EOS, and reproduced
+      the exact historical completion ending in "Pre-LayerNorm". The training
+      entries below remain unverified: selecting this model-specific default
+      does not establish training correctness, a performance result, or a
+      global dispatcher default.
 
   pretrain:
     H100:
       status: unverified
       precision: bf16
-      enabled_features: {}
+      enabled_features:
+        moe_dispatcher: hybridep
       command: null
       last_verified: null
       metrics:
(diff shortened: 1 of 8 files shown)


yaoyu-33:
Thanks—the asymmetry was not intentional. In both pinned MCore revisions, `MoELayer.combine()` currently delegates directly to `token_combine()`, so this was not a present correctness gap. I updated the FP32 path in `8ed1059be` to call `super().combine(output.float())` and preserve the base-layer contract. The focused inference transport test and all-file pre-commit checks pass.

PR #5482, opened 2026-08-10
fix(model): use public Gemma4 block spec target

yaoyu-33:
## Summary

- Define the Gemma 4 MoE block-spec function under the public `gemma4_block_spec` name so new `run_config.yaml` files serialize a reloadable target.
- Keep `_gemma4_block_spec` as a legacy alias and allowlist only that exact private target so existing checkpoints remain reloadable.
- Cover both the real provider serialization round trip and a legacy private-target reload.

## Problem

`Gemma4ModelProvider.transformer_layer_spec` used a partial of `_gemma4_block_spec`. YAML serialization derives callable targets from the function `__qualname__`, so the pre-existing public alias did not help: new checkpoints still recorded the private name, and strict checkpoint reload rejected it.

## Scope

This is limited to the Gemma 4 block-spec target and its exact legacy checkpoint compatibility entry. It does not change inference precision handling, model behavior, public conversion APIs, dependencies, CI, or Megatron-LM.

## Validation

- Red on current `main`: the provider default serialized `megatron.bridge.models.gemma.modeling_gemma4._gemma4_block_spec`, and strict reload raised `InstantiationException` at `model.transformer_layer_spec`.
- Focused unit tests after the fix: 82 passed.
- Real checkpoint-backed `google/gemma-4-26B-A4B-it` inference on 8x H100 with BF16 weights, TP4/PP2/EP1:
  - reloaded the legacy `run_config.yaml` and checkpoint;
  - crossed the PP boundary without a private-target or Transformer Engine dtype exception;
  - produced finite full-vocabulary logits;
  - stopped naturally after 10 generated tokens with `The animal on the candy is a turtle.`;
  - all eight ranks exited cleanly.

No model verification card status was changed.

## Reference

NVBug 6491102.

### DIFF
--- a/src/megatron/bridge/models/gemma/gemma4_provider.py
+++ b/src/megatron/bridge/models/gemma/gemma4_provider.py
@@ -36,9 +36,9 @@ from megatron.bridge.models.gemma.modeling_gemma4 import (
     Gemma4OutputLayer,
     Gemma4RotaryEmbedding,
     _attach_ple_modules,
-    _gemma4_block_spec,
     _install_ple_forward,
     _install_tied_kv,
+    gemma4_block_spec,
     get_gemma4_layer_spec,
     wire_gemma4_kv_sharing,
 )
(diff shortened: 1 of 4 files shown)


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5482/commits/8e480b48b906bf8740a434cb52356d98f0a5d059

PR #5484, opened 2026-08-10
[peft] feat: default MoE expert LoRA to per-expert adapters

cuichenx:
# What does this PR do ?

Flip the default of `share_expert_adapters` from `True` to `False` for both `LoRA` and `CanonicalLoRA`, so grouped MoE expert linears get **one adapter per local expert** by default instead of one adapter shared across the experts co-located on an EP rank.

## Motivation: the shared default has EP-dependent semantics

A shared expert adapter is shared only across the experts on a **single EP rank** (the plain shared `ParallelLinearAdapter` has no cross-EP replication — only the newer shared-outer path added one). So the effective sharing granularity, and therefore the **trained numerics**, silently changes with the expert-parallel degree:

- EP=1 → 1 adapter shared across all experts
- EP=8 → 8 independent adapters, each over 1/8 of the experts

The same recipe produces different capacity/numerics at different EP. Per-expert adapters are **EP-invariant**, which is the more defensible default.

## Ecosystem alignment

| Framework | MoE-expert LoRA default | Shared offered? |
|-----------|------------------------|-----------------|
| NeMo Automodel | per-expert | no — per-expert only |
| SkyRL (`skyrl-tx`) | per-expert | no — per-expert only |
| Miles (Bridge LoRA consumer) | per-expert | shared-outer opt-in only |
| Megatron-Bridge (before) | **shared** | yes |

Two first-party implementations offer per-expert only, and Miles — a downstream Megatron-Bridge LoRA consumer — already defaults to per-expert.

## Changelog

- `peft/lora.py`, `peft/canonical_lora.py`: default `share_expert_adapters=False`; docstrings explain the EP-invariance rationale.
- `docs/training/peft.md`: default tables + prose updated.
- Unit tests: two default-value asserts flipped; nine tests that specifically exercise the shared path pin `share_expert_adapters=True` explicitly (so shared mode stays covered).

## Breaking change

Marked with the `breaking-change` label (kept out of the title per request):
- Recipes on the default now produce larger (per-expert) adapters with different numerics.
- Previously-trained shared adapters require `share_expert_adapters=True` to load.
- Pin `share_expert_adapters=True` to recover the prior behavior (e.g. to save parameters).

The shared layout is unchanged and fully supported via the flag.

### DIFF
--- a/docs/training/peft.md
+++ b/docs/training/peft.md
@@ -76,7 +76,7 @@ The following table lists key hyperparameters for configuring LoRA, which contro
 | `alpha` | `float` | `32` | Scaling parameter for LoRA |
 | `dropout` | `float` | `0.0` | Dropout rate for LoRA layers |
 | `normalize_moe_lora` | `bool` | `False` | Reduce expert-layer rank to `dim // moe_router_topk` while keeping dense layers at the full rank |
-| `share_expert_adapters` | `bool` | `True` | Share one adapter across all local experts on an EP rank instead of creating one adapter per local expert |
+| `share_expert_adapters` | `bool` | `False` | Create one adapter per local expert (EP-invariant); set `True` to instead share one adapter across all local experts on an EP rank |
 
 #### Target Modules
 The following table lists specific submodules within transformer architectures that are commonly targeted for LoRA, enabling efficient fine-tuning of attention and feedforward components:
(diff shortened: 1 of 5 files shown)


cuichenx:
pending perf optimization

PR #5497, opened 2026-08-11
[ckpt] fix: restore frozenset Energon group keys

yaoyu-33:
## What changed

Allow Bridge's restricted Energon checkpoint loader to restore `frozenset` group keys written by the default `torch.save` protocol.

## User impact and root cause

Energon task encoders may use any hashable value as a batch-group key. When a supported task encoder leaves a partial bucket keyed by a `frozenset`, Bridge saves the dataloader state with `torch.save`. The default pickle protocol records that key as `__builtin__.frozenset`, while the restricted loader only allowed `builtins.frozenset`. Resume therefore failed with an `UnpicklingError` on a checkpoint produced by Bridge itself.

The fix canonicalizes only this legacy module spelling before applying the existing safe-type allowlist. Other globals remain subject to the same restrictions.

## Regression evidence

The new focused test uses a real `SavableDataLoaderState` and `torch.save` round trip.

- Before the fix: `uv run --active --no-sync python -m pytest --confcutdir=tests/unit_tests/utils tests/unit_tests/utils/test_safe_pickle.py::test_energon_group_bucket_frozenset_key_round_trip -q` — failed with `Restricted unpickler refused to load '__builtin__.frozenset'`.
- After the fix: the same command — 1 passed.
- Adjacent tests: `uv run --active --no-sync python -m pytest --confcutdir=tests/unit_tests/utils tests/unit_tests/utils/test_safe_pickle.py -q` — 53 passed.
- `git diff --check` — passed.
- `uv run --active --no-sync pre-commit run --all-files` — all hooks passed.

## Scope

This change only handles the legacy pickle spelling of the already-allowed `frozenset` type. It does not broaden the unpickler's class allowlist or change other Energon checkpoint semantics.


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5497/commits/f6b1cfacccb6f7581b64b11d37b02597019ca428

PR #5519, opened 2026-08-11
[1/4] feat(data): add DeepSeek chat collation

yaoyu-33:
## Summary

- add a model-owned DeepSeek V4 renderer/collator based on the official encoding contract, including chat/thinking modes, `truncate_history_thinking`, tools, and assistant/last-turn/full loss masks
- reuse the existing `_MODEL_COLLATE_SPECS` registry and shared text batching through one optional `tokenize_impl` hook
- select the DSV4 collator from the tokenizer `name_or_path` in the Direct HF SFT builder while leaving shared conversation processing and `DirectSFTDataset` unchanged

## Validation

- `uv run --active --no-sync pre-commit run --all-files`
- all 4 official DeepSeek V4 encoding fixtures match exactly
- real-tokenizer assistant masking and left/right truncation pass
- isolated shared-collator padding, label shifting, and loss-mask integration pass

Standard local pytest collection is blocked on this host by the installed CUDA/Megatron environment; CI is the authoritative pytest run.

### DIFF
--- a/src/megatron/bridge/data/builders/direct_hf_sft.py
+++ b/src/megatron/bridge/data/builders/direct_hf_sft.py
@@ -197,6 +197,14 @@ def select_direct_hf_sft_collate(
     raise ValueError("Prompt-completion preprocessing supports text-only examples.")
 
 
+def _model_collate_key(processor: Any) -> str:
+    tokenizer = get_processor_tokenizer(processor)
+    name_or_path = getattr(tokenizer, "name_or_path", "")
+    if isinstance(name_or_path, str) and "deepseek-v4" in name_or_path.lower().replace("_", "-"):
+        return "deepseek-v4"
+    return type(processor).__name__
+
+
 def build_direct_hf_sft_split(
     config: DirectHFSFTDatasetConfig,
     source: HFDatasetSourceConfig,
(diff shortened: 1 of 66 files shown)


src/megatron/bridge/data/builders/direct_hf_sft.py
```diff
@@ -197,6 +197,14 @@ def select_direct_hf_sft_collate(
     raise ValueError("Prompt-completion preprocessing supports text-only examples.")
 
 
+def _model_collate_key(processor: Any) -> str:
+    tokenizer = get_processor_tokenizer(processor)
+    name_or_path = getattr(tokenizer, "name_or_path", "")
+    if isinstance(name_or_path, str) and "deepseek-v4" in name_or_path.lower().replace("_", "-"):
```
yaoyu-33: can we check a finger print as well maybe vocab size or a special token id?

yaoyu-33: Addressed in `d04ec8ad8`. The fallback now requires the normalized DeepSeek-V4 model name plus the official tokenizer fingerprint: total tokenizer size `129280` and `<｜Assistant｜>` token ID `128804`. Added positive coverage for the official fingerprint and negative coverage showing that a matching name with the wrong fingerprint stays on the generic collator path.

PR #5525, opened 2026-08-12
feat(data): scalable canonical-grid reads for MegatronMIMO data loading

going-song:
## What

Adds opt-in **scalable data-parallel reads** to MegatronMIMO. By default every data-loading
rank reads the full global micro-batch and `forward_step` slices out its module-DP shard, so
per-rank read and preprocessing cost scales with the DP degree. With
`dataset.megatron_mimo_scalable_dp` each rank's sampler reads only its shard and the slice is
skipped:

```text
default            scalable
read    [4, S]     read    [2, S]      (per language rank, dp=2)
slice   [4]->[2]   slice   skipped
forward [2, S]     forward [2, S]      (same shard geometry, identical routing)
```

Second of three PRs splitting #4608 (original work by **@sailor1493**, credited
as co-author and signer), per the maintainers' [split request in #4609](https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/4609#issuecomment-4926364154).
Part 1 merged as #5131.

## Why a canonical grid (mechanism deviation from #4608)

#4608 sharded each module's loader by its own module-DP `(rank, size)`. A sampler's shard
assignment is a function of `(micro_batch_size, data_parallel_size)`, so under `"cyclic"` —
the direct-HF-SFT default since #5048 — modules with different DP sizes read **different
sample sets**, while the `BridgeCommunicator` routes modality embeddings positionally along
the batch dim. We hit this live as an embedding-count crash in
`align_embeddings_by_token_positions`; with uniform image sizes it would silently train on
mispaired image-text.

Here all loaders shard on one **canonical grid** — the LCM of the module DP sizes. Every
group sampler uses the identical `(micro_batch // grid, grid)` geometry and global
`consumed_samples`; a rank whose module has DP size `d` concatenates the windows of its
`grid // d` consecutive groups (`CanonicalGroupBatchSampler`). Group streams are
module-independent, so every module materializes the identical ordered micro-batch under
`"single"` **and** `"cyclic"`; for `"single"` the stream is bit-identical to the default
path's slice. Geometries are truncated to whole micro-batches so ragged dataset sizes cannot
desynchronize the groups. ("Canonical group" is the vocabulary of #4608's reordering
machinery, which ③ will port.)

## Configuration

```python
cfg.dataset.megatron_mimo_scalable_dp = True   # DirectHFSFTDatasetConfig; example: --scalable-dp
```

`"single"` and `"cyclic"` are supported; `"batch"`/`"external"` and `drop_last=False` are
rejected in `validate()` and again at loader build. Off by default — with the switch unset, the
sampler parameters, loader arguments and slicing behavior are identical to main.

## Notes for reviewers

- Scalable reads reject `expert_tensor_parallel_size > 1` (unvalidated combination). The
  default slicing path shares the same assumption — happy to hoist the guard into
  `MegatronMIMOParallelismConfig` validation if preferred.
- No train-loop changes: the provider forces `variable_seq_lengths=True`, so the schedule's
  `micro_batch_size` is shape-inert and the LR/consumed accounting already uses the full
  micro-batch in both modes (#4608's schedule/LR adjustment is unnecessary here).
- Under `"cyclic"`, scalable on/off are two different valid epoch shuffles (per-iteration
  losses not comparable); `"single"` is order-identical.

## Testing

258 unit tests passing locally, including the cross-geometry alignment invariant (a rank
covering groups {0,1} reads exactly the concatenation of the single-group ranks' windows, for
`"cyclic"` in both `data_sharding` modes) and a library-path end-to-end test with DP {2,3}
(grid = 6). Live PP=2 runs (5 GPUs of an 8xA100 node, `language=tp1,pp2,dp2` + `images=dp1`, Qwen3.5-0.8B random
init, cord_v2, mbs 4, 10 iters): `cyclic`+scalable completes 10/10 where module-DP sharding
crashes; `single` scalable matches the default losses within bf16 noise (max |diff| 0.0043);
the default path matches prior losses within run-to-run noise (max |diff| 0.0024); packing on/off parity under scalable ~0.001.
(One drive-by repair: the example-config test's arg namespace was already failing on main —
#5131 added `pack_sequences_in_batch` to `_build_dataset_config` without updating it — so this
PR adds both missing kwargs there.)

## Known limitations

ETP > 1 and `"batch"`/`"external"` loaders are rejected for scalable reads; MegatronMIMO still
cannot run CP (pre-existing); checkpoint resume is unit-tested but not yet exercised in a live
run, and multi-node validation is pending, as with #4608.

Signed-off-by: kayeon.song <kayeon.song@navercorp.com>

### DIFF
--- a/src/megatron/bridge/data/megatron_mimo/canonical_sampler.py
+++ b/src/megatron/bridge/data/megatron_mimo/canonical_sampler.py
@@ -1,0 +1,213 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Canonical-grid batch sampling for MegatronMIMO scalable data-parallel reads.
+
+With ``megatron_mimo_scalable_dp`` every module's loaders must materialize the same
+ordered global micro-batch, because the ``BridgeCommunicator`` routes modality
+embeddings to language ranks by contiguous position along the batch dim. A sampler's
+shard assignment depends on its ``(data_parallel_rank, data_parallel_size,
+micro_batch_size)``, so modules with different DP sizes cannot shard by their own DP:
+a shuffling sampler would hand them different sample sets. Instead, all loaders shard
+on one shared **canonical grid** — the least common multiple of the module DP sizes.
+A rank whose module has DP size ``d`` covers ``grid // d`` consecutive canonical
+groups and concatenates their windows, so concatenating any module's rank batches
+reproduces the identical ordered global micro-batch under any deterministic sampler
+(``single`` and ``cyclic`` included; the shuffle seed derives from the epoch alone).
+"""
+
+from __future__ import annotations
+
+import math
+from typing import Callable, Iterator
+
+from torch.utils.data import DataLoader, Dataset
+
+from megatron.bridge.data.samplers import MegatronPretrainingRandomSampler, MegatronPretrainingSampler
+
+
+def canonical_grid_size(module_dp_sizes: list[int]) -> int:
+    """Return the shared sampler grid size: the LCM of every module's DP size."""
+    if not module_dp_sizes or any(dp is None or dp < 1 for dp in module_dp_sizes):
+        raise ValueError(f"module DP sizes must be positive integers (got {module_dp_sizes}).")
+    return math.lcm(*module_dp_sizes)
+
+
+def covered_canonical_groups(dp_rank: int, dp_size: int, grid_size: int) -> list[int]:
+    """Return the canonical groups this rank reads.
+
+    The groups are the ``grid_size // dp_size`` consecutive slots matching the
+    contiguous batch-dim chunk the ``BridgeCommunicator`` routes to this DP rank.
+    """
+    if grid_size % dp_size != 0:
+        raise ValueError(
+            f"canonical grid size ({grid_size}) is not divisible by the module DP size ({dp_size}); "
+            "module DP sizes must divide their least common multiple."
+        )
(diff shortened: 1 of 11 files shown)


src/megatron/bridge/data/megatron_mimo/dp_utils.py
```diff
@@ -197,9 +204,24 @@ def get_megatron_mimo_sampling_info(
         return 0, 1, False
 
     needs_data = _needs_data_for_module(my_grid, my_module)
-    # All data-loading ranks use the same sampler settings so they load
-    # identical global micro-batches.  Module-local DP slicing happens later
-    # in forward_step via slice_batch_for_megatron_mimo.
+    if scalable_dp:
+        module_parallelism = megatron_mimo_cfg.module_parallelisms[my_module]
+        if module_parallelism.expert_tensor_parallel_size != 1:
+            raise NotImplementedError(
```
going-song, line 210: The default slicing path relies on the same assumption — should this live in `MegatronMIMOParallelismConfig` validation instead?

src/megatron/bridge/data/megatron_mimo/canonical_sampler.py
```diff
@@ -0,0 +1,213 @@
...
+        raise ValueError("megatron_mimo_scalable_dp requires drop_last=True (partial windows misalign modules).")
+    if micro_batch_size % grid_size != 0:
+        raise ValueError(
+            f"micro_batch_size ({micro_batch_size}) must be divisible by the canonical grid size ({grid_size})."
+        )
+    group_micro_batch_size = micro_batch_size // grid_size
+    # Truncate to whole global micro-batches: the flat samplers round their active range at
+    # per-group granularity, which diverges per group for a non-multiple dataset size (the
+    # cyclic data_sharding=False stride would give groups unequal window counts).
+    total_samples = (len(dataset) // micro_batch_size) * micro_batch_size
```
going-song, line 134: Prevents unequal per-group window counts on ragged dataset sizes (cyclic + `data_sharding=False`); regression-tested.

tests/unit_tests/examples/test_qwen35_vl_mimo_finetune.py
```diff
@@ -158,6 +158,8 @@ def test_dataset_config_enables_only_requested_or_known_validation_splits(
                 dataloader_type="single",
                 trust_remote_code=False,
                 do_validation=do_validation,
+                pack_sequences_in_batch=False,
```
going-song, line 161: This test has been failing on main since #5131 (missing `pack_sequences_in_batch`); adding both kwargs.

liding-nv (approved):

PR #5573, opened 2026-08-13
feat(model): let callers supply a media-token validity mask

DanialTaheri:
## What

`NemotronOmniModel.forward` gains a keyword-only `media_token_validity_mask`
that takes precedence over the masks it currently derives. Behavior is
unchanged when the argument is omitted.

## Why

`_merge_projected_media` enforces a strict 1:1 contract — one projected feature
per valid placeholder. That contract is right; the question is where the
placeholder mask comes from. Today it is derived from the padding mask, or a
2-D attention mask:

```python
media_token_validity_mask = None
if padding_mask is not None:
    media_token_validity_mask = ~padding_mask
elif attention_mask is not None and attention_mask.dim() == input_ids.dim():
    media_token_validity_mask = attention_mask
```

Both answer **"is this a real token?"**. The merge needs **"is this a media
anchor?"**. Those coincide only while every media token in a non-padding
position anchors an image.

A media placeholder is an ordinary vocabulary entry, so text can legitimately
contain it. Such a row attaches no image, but the derived mask still marks the
position valid, so the merge demands a feature that was never meant to exist
and raises.

This is not hypothetical. The text post-training data can include `<image>` with zero attached images —
competitive-programming statements where `<image>` replaced inline math, e.g.
`"for any character <image> there is exactly one character"`. They carry between
1 and 12 placeholders each.

Callers know how many media items each row carries, so they can answer the
question the model cannot. This lets them do so.

Rows that do carry media keep every position valid, so a genuine
placeholder/feature disagreement is still reported rather than masked away — the
argument cannot be used to silence real misalignment.

## Tests

Two tests via the existing `_BoundaryModel` harness, which exercises the real
forward on CPU:

- `test_text_containing_the_placeholder_trains_with_a_caller_mask` — a row that
  spells the placeholder raises without a mask, and trains with one.
- `test_caller_mask_takes_precedence_over_the_derived_one` — an explicit mask
  wins over a padding mask that would otherwise treat the placeholder as an
  anchor.

Verified against this branch: 9 passed, 0 failed for
`tests/unit_tests/models/nemotron_omni/test_nemotron_omni_model.py`
(`-k "caller_mask or validity_mask or placeholder or text_only or media_alignment or collapsed"`).

`ruff check` and `ruff format --check` clean under the repo config.

### DIFF
--- a/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py
+++ b/src/megatron/bridge/models/nemotron_omni/modeling_nemotron_omni.py
@@ -642,6 +642,7 @@ class NemotronOmniModel(MegatronModule):
         sound_clips: Optional[torch.Tensor] = None,
         sound_length: Optional[torch.Tensor] = None,
         *,
+        media_token_validity_mask: torch.Tensor | None = None,
         inference_params=None,
         **kwargs,
     ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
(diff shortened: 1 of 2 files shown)


cuichenx (approved):

DanialTaheri:
Both failures are unrelated to this change.

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5573/commits/dbb8d4acac42e9c711b012494a139b0ec6d984d4

kamran-nvidia:
/ok to test [7759634](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5573/commits/77596340c832d111dd5150ecfdd06733d1f52a15)

PR #5575, opened 2026-08-14
Generalize ModelOpt quantized weight export

mxinO:
## Summary

- convert distributed Megatron parameters into ModelOpt canonical deployment tensors through the public non-mutating export API
- preserve normal HF-name mapping while supporting quantized weights, scales, input scales, grouped experts, TP/PP/EP, and tied/shared parameters
- omit training-only quantizer state from deployment export
- document the deployment-tensor contract and pin the required ModelOpt revision

Depends on NVIDIA/Model-Optimizer#2190.

## Testing

- 40 focused model bridge and ModelOpt utility unit tests
- two-rank distributed TP2/PP2/EP2 parity test against ModelOpt terminal export
- generated lock-file validation in container


mxinO:
Closing this draft while the real-quant refit design is reworked and validated locally. The branch is retained for reference.

PR #5579, opened 2026-08-14
Add canonical ModelOpt quantized export bridge

mxinO:
## Summary
- capture opaque ModelOpt export state from Megatron weights
- preserve TP, PP, and EP ownership while mapping state to canonical HF names
- pack each expert before grouped export and emit ModelOpt canonical quantization metadata

Depends on NVIDIA/Model-Optimizer#2193.

## Verification
- 39 focused CPU tests passed
- two-rank TP2, PP2, EP2, and grouped-EP export parity passed on GPU
- Ruff format/check, compileall, and git diff check passed

This PR is intentionally draft while the dependent NeMo-RL integration is validated.


mxinO:
Closing this draft while the real-quant refit design is reworked and validated locally. The branch is retained for reference.

PR #5608, opened 2026-08-16
fix(recipe): add library aliases for the colliding GLM-5.2 SFT recipes

pruprakash:
# What does this PR do ?

Fixes the GLM-5.2 verification card's two SFT reproduction commands launching a 131,072-token benchmark job instead of the library configuration their published metrics were measured on, by giving those library recipes a distinct name.

# Changelog

- `recipes/glm/h100/glm5.py`: add `glm52_h100_sft_config` alias.
- `recipes/glm/gb200/glm5.py`: add `glm52_gb200_sft_config` alias.
- `recipes/glm/{__init__,h100/__init__,gb200/__init__}.py`: export the aliases.
- `examples/model_verification_cards/glm5-2/card.yaml`: both SFT commands name the alias.
- `tests/unit_tests/recipes/test_glm5_recipes.py`: 4 parametrized cases over the aliases and the shipped card commands.

# Additional Information

- **Root cause:** both SFT names are exported by `megatron.bridge.recipes` and `megatron.bridge.perf_recipes`; `load_recipe` binds the benchmark definition and the library workloads had no other name.
- **Blast radius:** card only. Purely additive — the colliding names still resolve to the benchmark recipes.
- **Regression?** Yes. The SFT leaves were verified 2026-07-22 / 2026-08-04; the collision registry that changed what those names mean landed later. Found while testing #5583, not caused by it.
- **Verification:** red-green in `nvcr.io/nvidian/nemo:nightly` — RED is a collection `AttributeError` without the aliases and `2 failed` with the card unchanged; GREEN is `20 passed`. On the patched tree `load_recipe` returns gbs=32/seq=2048 (H100) and gbs=8/seq=8192 (GB200), matching the card.
- **Not included:** the other 7 `BENCHMARK_RECIPE_PRECEDENCE_COLLISIONS` entries still lack a library alias, and `validate_card.py` still does not check `--recipe`.

### DIFF
--- a/examples/model_verification_cards/glm5-2/card.yaml
+++ b/examples/model_verification_cards/glm5-2/card.yaml
@@ -240,7 +240,7 @@ items:
         moe_dispatcher: deepep
       command: >
         ./scripts/training/train.sh --nodes 52 --gpus-per-node 8
-        --recipe glm52_sft_416gpu_h100_bf16_config --mode sft
+        --recipe glm52_h100_sft_config --mode sft
         --pretrained_checkpoint work/model-verification/glm5-2/gpu-megatron/iter_0000000
         --max_steps 100 --seq_length 2048
         dataset.hf_output_root=work/data/glm5-2/tulu3-full-sft-pad32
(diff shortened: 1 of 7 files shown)


linmuchuiyang:
Independent reproduction on GB200 with nemo:26.08 (Bridge v0.6.0 c932511): importing `megatron.bridge.perf_recipes` before `megatron.bridge.recipes` binds `glm52_sft_192gpu_gb200_bf16_config` to the perf variant (NullTokenizer, benchmark data), so the card's SFT command trains on synthetic tokens without any error. We work around it by importing `megatron.bridge.recipes.glm.gb200.glm5` explicitly in a wrapper. Aliases help, but the shadowing itself remains as long as both modules export the same name — could the perf recipe be renamed (e.g. `..._perf_config`) or `load_recipe` error on ambiguous names?

PR #5493, opened 2026-08-11
feat(data): add in-batch packing for GPT SFT

yaoyu-33:
# What does this PR do ?

Adds opt-in runtime sequence packing for GPTSFTDataset and GPTSFTChatDataset while preserving lazy mmap-backed JSONL loading and existing offline packing support.

# Changelog

- Add typed in-batch packing configuration and validate that runtime and offline packing are mutually exclusive.
- Build THD packed batches with exact shifted tokens, labels, loss masks, position IDs, and packed-sequence metadata.
- Preserve logical microbatch boundaries when a global-batch sampler is used, and derive required CP/SP padding from training configuration.
- Add unit coverage for dataset, builder, sampler, batch, and configuration behavior.
- Document when to use offline versus in-batch GPT SFT packing.

# GitHub Actions CI

Local validation completed:

- uv run pre-commit run --all-files: passed.
- Focused GPT SFT, builder, sampler, batch, and configuration unit tests: 98 passed, 1 skipped.
- Linux validation against the exact pinned Megatron-Core revision: 49 passed, 1 existing GPU-only test skipped because the remote host driver is older than the container requirement.
- Functional sampler tests: 5 passed.

# Before your PR is "Ready for review"

- [x] Read and followed the contributor guidelines.
- [x] Added necessary tests.
- [x] Updated documentation.
- [x] No optional dependency components are affected.

# Additional Information

This is an opt-in data-path feature. Existing GPT SFT behavior is unchanged unless enable_in_batch_packing is enabled.

PR #5610, opened 2026-08-17
feat(data): add CoderForge SFT preset

yaoyu-33:
## Summary

- add a named CoderForge Preview `SWE_Rebench` SFT preset and JSON schema adapter
- restore enum types after Hydra overrides so typed selectors such as the attention backend remain valid
- propagate physical-gap metadata through packed batches and preserve supervised terminal EOS tokens during offline packing

Stacked on #5493.

## Validation

- focused unit suite: 438 passed, 5 deselected
- Qwen3 0.6B 128K SFT smoke tests on 8 H100 GPUs (TP1/CP8), three optimizer steps each:
  - no packing: finite losses, zero skipped or NaN iterations
  - in-batch packing: 128K bins, finite losses, zero skipped or NaN iterations
  - offline packing: 128K bins with no single-sequence cap, finite losses, zero skipped or NaN iterations

### DIFF
--- a/src/megatron/bridge/data/packing/gpt_sft.py
+++ b/src/megatron/bridge/data/packing/gpt_sft.py
@@ -226,10 +226,19 @@ class GPTSFTPackedDataset(GPTSFTDataset):
                 for i in range(len(item["seq_boundaries"]) - 1):
                     current_seq = item["input_ids"][item["seq_boundaries"][i] : item["seq_boundaries"][i + 1] - 1]
 
-                    # Stop logical lengths at the last non-EOS token so alignment padding is excluded.
+                    # Alignment padding uses zero-loss EOS tokens. A supervised terminal EOS is part of
+                    # the logical sequence and must not be mistaken for a physical alignment gap.
                     current_seq_arr = np.array(current_seq)
-                    non_eos_positions = np.where(current_seq_arr != self.tokenizer.eos_id)[0]
-                    logical_seqlen = non_eos_positions[-1] + 1 if non_eos_positions.size > 0 else 0
+                    current_loss_mask = np.array(
+                        item["loss_mask"][item["seq_boundaries"][i] : item["seq_boundaries"][i + 1] - 1]
+                    )
+                    logical_seqlen = len(current_seq_arr)
+                    while (
+                        logical_seqlen > 0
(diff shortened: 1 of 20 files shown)


src/megatron/bridge/recipes/utils/dataset_utils.py
```diff
@@ -179,6 +179,37 @@ def default_tulu3_config(
     )
 
 
+def default_coderforge_config(
+    seq_length: int = 4096,
```
cuichenx, line 183: should we default to a higher value?

src/megatron/bridge/data/packing/gpt_sft.py
```diff
@@ -226,10 +226,19 @@ def collate_fn(self, batch):
                 for i in range(len(item["seq_boundaries"]) - 1):
                     current_seq = item["input_ids"][item["seq_boundaries"][i] : item["seq_boundaries"][i + 1] - 1]
 
-                    # Stop logical lengths at the last non-EOS token so alignment padding is excluded.
+                    # Alignment padding uses zero-loss EOS tokens. A supervised terminal EOS is part of
```
cuichenx, line 229: terminal EOS should not be supervised. loss mask is aligned to the target not input, so EOS token should always have loss mask = 0

tests/unit_tests/data/datasets/test_gpt_sft.py
```diff
@@ -524,9 +524,30 @@ def test_collate_fn_marks_offline_alignment_padding(self, tmp_path):
         assert processed["tokens"].tolist() == [[10, 11, 12, 2, 20, 21, 2, 2]]
         assert processed["cu_seqlens_q"].tolist() == [[0, 3, 5]]
         assert processed["cu_seqlens_q_padded"].tolist() == [[0, 4, 8]]
+        assert processed["pad_between_seqs"] is True
         assert processed["padding_mask"].dtype == torch.bool
         assert processed["padding_mask"].tolist() == [[False, False, False, True, False, False, True, True]]
 
+    def test_collate_fn_preserves_supervised_terminal_eos(self, tmp_path):
```
cuichenx, line 531: same as above, this test should be removed

PR #5613, opened 2026-08-17
[model] fix: Finalize providers before distributed model construction

yaoyu-33:
## What changed

Finalize legacy model providers in the training setup caller immediately before entering `provide_distributed_model()`, and add a focused regression test for heterogeneous Llama-Nemotron providers.

## Bug and impact

Heterogeneous providers defer Megatron-Core post-initialization, so derived fields such as `per_block_parameters` do not exist until `finalize()` runs. The training construction path previously entered `provide_distributed_model()` before establishing that lifecycle boundary, causing an `AttributeError` when model construction consumed the field.

`_build_distributed_model()` now finalizes a legacy provider after distributed initialization and immediately before invoking `provide_distributed_model()`. Finalization remains an explicit caller responsibility rather than an implicit side effect of the provider method.

## Validation

The regression test replaces the distributed construction method with an entry callback and verifies that the heterogeneous block configuration has already been parsed before that callback is entered.

```text
python -m pytest tests/unit_tests/training/test_setup.py::TestBuildDistributedModel tests/unit_tests/models/test_model_provider_mixin.py
10 passed

passed
```

## Scope

This does not change conversion mappings, provider configuration fields, public model-builder APIs, or Megatron-Core. It only makes the training setup lifecycle ordering explicit for deprecated `ModelProviderMixin` providers.

### DIFF
--- a/src/megatron/bridge/models/model_provider.py
+++ b/src/megatron/bridge/models/model_provider.py
@@ -247,6 +247,8 @@ class ModelProviderMixin(abc.ABC, Generic[ModelT]):
         if wrap_with_ddp and not ddp_config:
             raise ValueError("ddp_config is required when wrap_with_ddp is True")
 
+        self.finalize()
+
         if not torch.distributed.is_initialized():
             os.environ["RANK"] = os.environ.get("RANK", "0")
             os.environ["WORLD_SIZE"] = os.environ.get("WORLD_SIZE", "1")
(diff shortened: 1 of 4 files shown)


yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5613/commits/26d2fb10846cdab066e807d3bac2df3d772cf093

PR #5455, opened 2026-08-09
feat(data): add Nemotron Image v3 Energon tutorial

yaoyu-33:
# What does this PR do?

Adds a pinned, small-subset preparation and training tutorial for
`nvidia/Nemotron-Image-Training-v3` through the Energon path introduced by
#5285.

This is intentionally stacked on #5285. It does not download or materialize
the complete 6.9M-sample release.

## Contents

- Add `prepare_nemotron_image_v3.py` to join selected conversation JSONL files
  with separately stored media and write Bridge `ChatMLWebdataset` shards.
- Pin the dataset revision and verify the self-contained `turing` JSONL and
  media tar by size and SHA-256 before conversion.
- Restrict verified conversion to the pinned archive allowlist, reject
  unpinned subsets and loose/unverified media shadows, and provide an explicit
  `--skip-source-integrity-check` escape hatch.
- Normalize the mixed source message representation to typed ChatML, create a
  deterministic train/validation split, run Energon indexing, and persist a
  preparation manifest.
- Fail closed on invalid/remote/ambiguous media, unsupported video/audio,
  archive path escapes, stale output directories, and constructor-time archive
  failures without leaking open handles.
- Document a complete small-subset flow: pinned Hugging Face download,
  conversion/indexing, Qwen3-VL checkpoint import, and a public
  `scripts/training/train.sh` LoRA launch using `--dataset energon`, `vlm_step`,
  16K sequence length, and native Energon packing.
- Link the tutorial from the data, Energon, training, and Qwen3-VL entry points.
- Add focused happy-path and failure-path unit coverage, and restore collection
  of the existing `tests/unit_tests/tutorials/` suite.

## Validation

- Exact-head Core Unit job: `9044 passed, 14 skipped, 2 deselected`; both
  tutorial modules were collected, including all 13 Nemotron Image v3 cases.
- Real pinned `turing` conversion: 193 source conversations converted into 181
  train and 12 validation samples; source sizes and SHA-256 values matched,
  image bytes decoded, and Energon metadata/indexes were created.
- Real Qwen3-VL Energon loader: produced a finite multimodal batch with labels,
  loss mask, visual inputs, THD metadata, and a positive token count.
- Exact `train.sh` smoke: one 16K Qwen3-VL LoRA optimizer step plus validation
  completed with finite losses, zero skipped/NaN iterations, a persisted
  resolved config, and model plus Energon dataloader checkpoints.
- `uv run pre-commit run --all-files` passed in Linux.
[...]

PR #5630, opened 2026-08-18
fix(models): use physical boundaries for Qwen packed attention

yaoyu-33:
## What

- Pass physical cumulative sequence boundaries to Qwen3-VL core attention when Energon provides logical and padded THD metadata.
- Keep the original logical/padded metadata for Qwen MRoPE, hybrid GDN, and other decoder consumers.
- Add focused coverage for tail-only padding and the unpadded identity path.

## Why

Energon online packing can produce a pack where only the final cumulative offset differs between logical and physical layouts. Transformer Engine's padded-layout inference does not detect that tail-only difference, so attention backward can interpret a longer physical THD stream with inconsistent logical boundaries and produce exploding gradients or NaNs.

The physical-boundary view is limited to causal core attention. It avoids unsupported explicit padding selection while preserving the richer metadata everywhere else.

Follow-up to #5455.

## Validation

- 54 focused Qwen attention, packed-sequence utility, in-batch packing, and VLM-step unit tests passed.
- `uv run pre-commit run --all-files` passed.
- A matched 5-step 32-GPU BF16 Energon online-packing run completed with finite loss/gradients and zero skipped/NaN iterations.
- A matched 100-step run then completed all 100 iterations with 100 finite metric records, zero skipped/NaN iterations, and successful model plus dataloader checkpointing.
- The non-packing control also remained finite.

The runtime validation used 16K sequences, TP1/CP1/PP1/EP32, MBS1/GBS32, and `packing_buffer_size=16`.


lkhphuc:
Hi, why is this closed without merged? Is the problem not exist anymore?

PR #5635, opened 2026-08-18
fix(peft)!: make expert LoRA exact under suppressed tensor-parallel communication

### LINKED ISSUE #5634: [bug] Expert LoRA computes wrong deltas and gradients at expert_tensor_parallel_size > 1 (and at TP > 1 under moe_shared_expert_overlap)
### Problem

LoRA adapters on MoE expert layers train on wrong numbers whenever `expert_tensor_parallel_size > 1`, and one variant is wrong at plain `TP > 1` when `moe_shared_expert_overlap` is enabled. Nothing errors and nothing warns: training runs to completion and the adapter is wrong. In one path the adapter trains nothing at all, because its output is detached from autograd. All shipped recipes pin `expert_tensor_parallel_size=1`, so stock configurations are unaffected today; custom MoE topologies are not.

**The single rule all four defects violate:** an adapter's output must be in the same *parallel state* as the base linear's output — replicated, a shard, or a partial awaiting a sum — because the two are added together. At `expert_tensor_parallel_size > 1` (or EP>1) Megatron-Core sets `explicit_expert_comm` on expert linears and deletes their own collectives, because the MoE token dispatcher performs that sum later for the *base* weights. The base's output state therefore changes from *replicated* to *partial*. LoRA adapters are not routed through the dispatcher, so they must complete their own arithmetic; they did not follow the change.

Four defect families follow, none of which raises an error:

1. **`ParallelLinearAdapter(is_expert=True)`, fc2 case** (the default for grouped experts via `share_expert_adapters=True`, the only class for `.local_experts.`/sequential layouts, and CanonicalLoRA's expert route): the adapter's row-parallel `linear_in` never sums `A_r @ h_r` across ETP (the reduce is suppressed), and its `gather_output=True` delta is then summed once per rank by the dispatcher — the delta is built from unsummed partials AND over-counted ×ETP.
2. **Same class, fc1 case**: the forward is exact — so forward-only checks pass — but `explicit_expert_comm` forces `allreduce_dgrad=False`, so `dL/dA` on each rank keeps only its own-rank term. A forward-only parity check reports this path as passing while every cross-rank gradient term is lost.
3. **`GroupedExpertLinearAdapter`** (`share_expert_adapters=False`), all three backends (per-expert, `grouped_mm`, TE-fp8): `_gather_along_last_dim` is a bare `torch.distributed.all_gather` into `torch.empty_like` buffers — invisible to autograd. On the fc2 case the adapter output carries no graph at all (`out.requires_grad == False`; `dL/dA = dL/dB = 0`; the adapter silently trains nothing) on top of the same full-delta-vs-partial-term forward error; on fc1, `dL/dA is None`.
4. **Shared-expert fc2 under `moe_shared_expert_overlap`** — dense TP>1, no ETP required: the base linear's collectives are suppressed and `SharedExpertMLP.post_forward_comm` sums full-width partials across the TP group, so the adapter's gathered delta is counted exactly TP times. `MultiLoRALinear` has the same defect on the same case.

All four are invisible at ETP=1/TP=1 and masked at step 0 by the default `lora_B_init_method="zero"` (the delta starts identically zero, so smoke tests pass). All shipped recipes pin `expert_tensor_parallel_size=1`, so stock configs are unaffected today; any custom ETP>1 topology — and, for family 4, any TP>1 run with `moe_shared_expert_overlap` — trains silently wrong.

### Minimal repro

```shell
# 2 GPUs. Distills defect family 1; families 2-4 have analogous forms.
cat > /tmp/etp_lora_repro.py <<'EOF'
import os, torch, torch.distributed as dist
import megatron.core.parallel_state as ps
from megatron.core.model_parallel_config import ModelParallelConfig
from megatron.core.process_groups_config import ProcessGroupCollection
from megatron.bridge.peft.utils import ParallelLinearAdapter

dist.init_process_group("nccl")
torch.cuda.set_device(int(os.environ["LOCAL_RANK"]))
ps.initialize_model_parallel(tensor_model_parallel_size=2, expert_tensor_parallel_size=2)
pg = ProcessGroupCollection.use_mpu_process_groups()
dev = torch.device("cuda", torch.cuda.current_device())
torch.backends.cuda.matmul.allow_tf32 = False

TOK, IN, DIM, OUT = 8, 8, 4, 8
rank = ps.get_expert_tensor_parallel_rank()
g = torch.Generator().manual_seed(2026)
x = torch.randn(TOK, IN, generator=g).to(dev)
a = torch.randn(DIM, IN, generator=g).to(dev)
b = torch.randn(OUT, DIM, generator=g).to(dev)

cfg = ModelParallelConfig(tensor_model_parallel_size=2, expert_tensor_parallel_size=2,
                          sequence_parallel=False, params_dtype=torch.float32,
                          gradient_accumulation_fusion=False)
adapter = ParallelLinearAdapter(IN, OUT, DIM,
    base_linear_name="decoder.layers.0.mlp.experts.linear_fc2",
    activation="identity", input_is_parallel=True, is_expert=True, alpha=DIM,
    disable_tensor_parallel_comm=True, model_parallel_config=cfg,
    pg_collection=pg).to(dev)
with torch.no_grad():
    adapter.linear_in.weight.copy_(a[:, rank * IN // 2 : (rank + 1) * IN // 2])
    adapter.linear_out.weight.copy_(b[rank * OUT // 2 : (rank + 1) * OUT // 2, :])

delta = adapter(x[:, rank * IN // 2 : (rank + 1) * IN // 2]).detach()
dist.all_reduce(delta, group=pg.expt_tp)   # the dispatcher's expert-TP sum
ref = x @ (b @ a).t()                      # merged-weight reference (alpha/dim == 1)
if rank == 0:
    print("max relative error vs merged weight:",
          ((delta - ref).abs().max() / ref.abs().max()).item())
EOF
torchrun --nproc_per_node=2 /tmp/etp_lora_repro.py
# Observed: max relative error ~1.0 (delta is 2x a concatenation of unsummed partials).
# Grouped-adapter one-liner (family 3): GroupedExpertLinearAdapter(...)(x, [4, 4])
#   .requires_grad  ->  False at ETP=2.
torchrun --nproc_per_node=2 /tmp/etp_lora_repro.py
```

A full three-layer regression suite (closed-form algebra, CPU harness driving the real adapters through real `explicit_expert_comm` suppression, and a two-rank real-NCCL suite) accompanies the companion PR; every defect above is demonstrated there as a strict-xfail with its measured signature.

### Expected behavior

An expert-LoRA delta must be indistinguishable from folding `(alpha/dim) * B @ A` into the base weight — forward and gradients — under any parallelism layout, exactly as it is at ETP=1. Concretely for the repro: max relative error at fp32 (TF32 off) should be ~1e-6, not ~1.0; grouped-adapter outputs must carry `requires_grad=True` with nonzero `dL/dA`/`dL/dB`; fc1 `dL/dA` must include the cross-rank terms. Where a correct completion is not implementable for a path, it should refuse loudly rather than train wrong.

### Affected area

area:peft

### Regression?

No

### Environment

```
Branch / commit:  NVIDIA-NeMo/Megatron-Bridge main @ 0c565c9a0 (Megatron-LM pin d12f6c8c9)
Container:        nvcr.io/nvidia/pytorch:25.09-py3 -based CI image (docker/Dockerfile.ci)
GPU / machine:    2x NVIDIA H100 80GB HBM3, single node
Python / CUDA / PyTorch: image defaults (Python 3.12)
Relevant code:    explicit_expert_comm suppression: megatron/core/tensor_parallel/layers.py
                  (row-parallel reduce skip; allreduce_dgrad=False); dispatcher expert-TP sum:
                  moe/token_dispatcher.py combine_preprocess; shared-expert TP sum:
                  moe/shared_experts.py post_forward_comm; adapter paths:
                  src/megatron/bridge/peft/{utils,dora,lora_layers,multi_lora_layers}.py
```

### Logs

Verbatim pytest output from the five defect tests run pre-fix on 2x H100 (fix commit reverted; `torch.distributed.run --nproc_per_node=2`, `--runxfail` so the fail-first markers assert for real):

```shell
___________ test_default_fc2_adapter_delta_matches_merged_reference ____________
E   AssertionError: Tensor-likes are not close!
E
E   Mismatched elements: 64 / 64 (100.0%)
E   Greatest absolute difference: 9.130354881286621 at index (2, 3) (up to 0.0001 allowed)
E   Greatest relative difference: 22.013450622558594 at index (3, 6) (up to 0.0001 allowed)
_______________ test_default_fc1_dL_dA_matches_merged_reference ________________
E   AssertionError: Tensor-likes are not close!
E
E   Mismatched elements: 16 / 16 (100.0%)
E   Greatest absolute difference: 10.64639949798584 at index (1, 7) (up to 0.0001 allowed)
E   Greatest relative difference: 13.507515907287598 at index (1, 0) (up to 0.0001 allowed)
____________________ test_grouped_fc2_output_requires_grad _____________________
E   AssertionError: assert False
E    +  where False = tensor([[0., 0., 0., 0., 0., 0., 0., 0.],\n ... ]], device='cuda:1').requires_grad
_____________ test_shared_expert_fc2_delta_counted_once_by_tp_sum ______________
E   AssertionError: Tensor-likes are not close!
E
E   Mismatched elements: 64 / 64 (100.0%)
E   Greatest absolute difference: 8.840452194213867 at index (2, 2) (up to 0.0001 allowed)
E   Greatest relative difference: 1.000002384185791 at index (0, 0) (up to 0.0001 allowed)
_____________ test_multi_lora_shared_expert_fc2_delta_counted_once _____________
E   AssertionError: Tensor-likes are not close!
E
E   Mismatched elements: 119 / 128 (93.0%)
E   Greatest absolute difference: 31.396705627441406 at index (2, 0) (up to 0.5 allowed)
E   Greatest relative difference: 1.2800720930099487 at index (2, 4) (up to 0.15 allowed)
======================== 5 failed, 27 warnings in 4.34s ========================
```

Reading the signatures: the shared-expert case is the pure double-count (relative difference exactly 1.0, i.e. delta == 2x reference); the expert-fc2 case is structurally wrong (a concatenation of unsummed partials, so the error is unbounded, here up to 22x); the grouped case's output tensor is detached from autograd entirely (`requires_grad` False; all-zero values are just the default zero-init of the B weight in that test). Post-fix, the same five tests pass on the same hardware.

tigist-far:
# What does this PR do ?

LoRA adapters on MoE expert layers silently compute the wrong numbers whenever `expert_tensor_parallel_size > 1`, and one variant is wrong at plain `TP > 1` when `moe_shared_expert_overlap` is enabled. Nothing errors and nothing warns: training runs to completion and the adapter is wrong. In one path the adapter trains nothing at all, because its output is detached from autograd. Every shipped recipe pins `expert_tensor_parallel_size=1`, so stock configurations are unaffected; custom MoE topologies are not.

This PR makes all four affected paths exact, and makes the four configurations for which no correct implementation exists yet fail loudly instead of silently.

## How it breaks

A LoRA adapter is a branch added to a base linear's output, so it must deliver its result in the same *parallel state* as the base: replicated, a shard, or a partial awaiting a sum. At `expert_tensor_parallel_size > 1` Megatron-Core sets `explicit_expert_comm` and **deletes the expert linear's own collectives**, because the MoE token dispatcher performs that sum later in `combine`. The base's output state therefore changes from *replicated* to *partial*. The adapters were never told: they kept gathering a full-width delta that the dispatcher then counted once per rank, and internally they consumed their own un-summed partial as though it were complete.

| Base linear | Base's output state | What the adapter must emit |
|---|---|---|
| column-parallel | shard | shard (already correct) |
| row-parallel, ordinary TP | replicated | full width (already correct) |
| row-parallel, collectives suppressed | **partial** | **a partial term — this is what was wrong** |

Four defect families follow from this, spanning `ParallelLinearAdapter` (both fc1 and fc2 cases), `GroupedExpertLinearAdapter` (all three backends), and the shared-expert/`MultiLoRALinear` path under `moe_shared_expert_overlap`. Per-path detail, measured failure signatures, and a two-GPU repro: **#5634**.

## What the fix does

Two small, autograd-visible completions, applied uniformly:

- **Complete the low-rank activation** after `linear_in`: for the fc2 case, `reduce_from(copy_to(x))` over the expert-TP group — an all-reduce in *both* directions, the exact adjoint pair for a sum whose result every rank consumes differently. For the fc1 case, `copy_to(x)` — forward identity whose backward all-reduce restores the suppressed `dL/dz` sum. The grouped class's fc1 gather becomes `all_gather_last_dim_from_tensor_parallel_region` (backward = reduce-scatter-**sum**, the true adjoint).
- **Zero-embed the output shard** instead of gathering: `linear_out` keeps `gather_output=False` and the local shard is placed at this rank's offset in a full-width buffer of zeros. Each output element then has exactly one non-zero contributor, so the downstream sum (dispatcher over ETP; `post_forward_comm` over TP) reconstructs `B @ z` exactly once.

All groups are passed explicitly (`group=`) — the MCore default resolves the *dense* TP group, which is silently wrong for expert adapters.

## Why this design

**Constraints.** A candidate fix is a choice of shard axis for the adapter's `A` and `B`, plus whatever collectives make it correct. Three constraints bind:

1. **Exactness.** The delta must be *exactly* `(alpha/r)·B@A`, because serving merges it into the base weight. "Close" is not a weaker form of correct — it is a delta no merged weight can represent.
2. **State-matching.** The adapter's output must arrive in the base's parallel state, per the table above.
3. **Checkpoint compatibility.** Each of `A` and `B` must keep the shard axis the stored format already records for it. That axis is checkpoint metadata (`ShardedTensor.axis_fragmentations`) and is what lets a TP=4 checkpoint load at TP=2; the export path pairs each tensor with the matching `RowParallelMapping` / `ColumnParallelMapping` so its per-rank pieces reassemble into the right full tensor. This is a *function of the base flavor*, not a single axis per matrix: `A` is sharded on `r` under a column-parallel base and on `d_in` under a row-parallel one, while `B` is always sharded on `d_out`. The constraint forbids changing the axis for a given flavor, not the flavor-dependence itself, which is already part of the format. Moving an axis would break existing adapter checkpoints, mis-assemble exported `lora_A`/`lora_B`, and silently change the merge.

**Alternatives considered**, for the row-parallel case where the base's collectives are suppressed:

| Option | Exact? | Communication | Why not |
|---|---|---|---|
| **Zero-embed the local shard** (chosen) | Yes | one all-reduce on `[tokens, rank]`; output side is a local pad, no traffic | — |
| Keep gathering (status quo) | **No** | all-gather on `[tokens, d_out]` | this *is* the bug. Correct under ordinary TP where the base's output is replicated; under suppression the downstream sum counts the delta `P` times. Identical code, opposite correctness, which is why it went unnoticed |
| Gather, then scale by `1/P` | Forward only | same as status quo | bakes a topology constant into the numerics: `(alpha/r)·B@A` no longer merges unless export applies the same factor, and the constant is silently wrong if the summing group ever changes |
| Reduce-scatter the low-rank activation, shard `B` on the rank axis | Yes | half the chosen option's boundary traffic | genuinely cheaper, but moves `B`'s shard axis from `d_out` to `r`, violating constraint 3. Worth revisiting only alongside a checkpoint-format change |
| All-gather the input, then use the base's own sharding pattern | Yes | `d_in / r` times more traffic (roughly 100-700x) | the base's pattern needs a replicated input; re-replicating one costs far more than completing the rank-axis sum |

**Recommendation.** Zero-embedding is the only option that is exact, state-matched, and checkpoint-compatible. It also needs no compensating factor anywhere: each output element has exactly one non-zero contributor, so the sum the dispatcher already performs reconstructs `B @ z` once, and the embed's adjoint is a plain slice, so gradients need no compensation either. That is what keeps stored weights and the merge path untouched. The one alternative that beats it on cost (reduce-scatter) does so by roughly one small collective per adapted expert linear and requires a checkpoint migration to adopt; at `r=32` that is not a trade worth making, but it is the right thing to reconsider if the adapter checkpoint format is ever revised.

Where a correct completion does not exist without new machinery, the PR refuses loudly instead of computing silently wrong results.

## What this breaks (the reason for the `!`)

Four configurations that previously ran while **silently training wrong** now raise `NotImplementedError`, each naming the offending config key and its workaround. This is the intended blast radius of the `!` in the title.

Configs that previously ran while **silently training wrong** now raise `NotImplementedError` with messages naming the offending config:

1. DoRA on MoE expert linears, and on suppressed-comm TP-sharded bases (shared experts under `moe_shared_expert_overlap`).
2. `experts_shared_outer_loras=True` at ETP>1 (previously crashed with an unrelated shape error at first forward; export path has independent ETP>1 defects).
3. Expert-adapter `dropout > 0` at ETP>1.
4. `LoRALinear.weight` on grouped adapter classes (previously `AttributeError`).

Each refusal has a practical workaround today (ETP=1, the default `share_expert_adapters=True`, `dropout=0.0`, the export-side merge path); lifting them properly is tracked in #5636.

No checkpoint format, weight layout, or export-path change: shard axes and `sharded_state_dict` factories are untouched, and merges still compute exactly `α/r·B@A` with no parallelism factor — the zero-embed design is what keeps stored weights compensation-free.

# Changelog

<details>
<summary>Per-file changes</summary>

- `peft/utils.py` — `ParallelLinearAdapter`: add `_reduce_expert_low_rank_activation` and `_embed_row_parallel_shard`; constructor predicates (`_expert_row_parallel`, `_shared_expert_row_parallel`) force `lin_out_gather_output=False` on the two row-parallel-base cases.
- `peft/utils.py` — `GroupedExpertLinearAdapter`: replace the autograd-invisible `_gather_along_last_dim` with `_complete_low_rank_activation` / `_embed_expert_row_parallel_shard` across all three backends (per-expert, `grouped_mm`, TE-fp8); `.contiguous()` normalization before the activation-offloading tag; error text for the missing-group case kept byte-identical.
- `peft/utils.py` — construction-time `NotImplementedError` for `SharedOuterGroupedExpertAdapter` at ETP>1, and for expert-adapter `dropout > 0` at ETP>1 on both adapter classes (dropout draws per-rank RNG on tokens the dispatcher replicates across ETP — not expressible by any merged weight).
- `peft/dora.py` — `DoRA.transform` refuses MoE expert linears and suppressed-comm TP-sharded bases (its dense-group adapter and per-rank magnitude norm do not compose); replicated bases still transform.
- `peft/lora_layers.py` — `LoRALinear.weight` refuses grouped adapter classes (previously an accidental `AttributeError`; the merge path for grouped adapters is the export-side materializer).
- `peft/multi_lora_layers.py` — `MultiLoRALinear`: same zero-embed fix for the suppressed-comm row-parallel case (its mid-GEMM reduce was already correct).
- Tests (commit 1, fail-first): `test_expert_row_parallel_adapter_algebra.py` and `test_expert_column_parallel_adapter_backward_algebra.py` (torch-only closed-form ground truth with negative controls); `test_expert_lora_etp.py` (CPU suite driving the real adapters through real MCore `explicit_expert_comm` suppression via a one-rank gloo init + mock group sizes, only the collectives faked — each fake asserts it received exactly the tensor the real collective would see); `test_expert_lora_etp_distributed.py` (two-rank TP=2×ETP=2 real-NCCL suite, TF32 disabled for exact tolerances) + `L0_Launch_peft_expert_etp_distributed.sh`; DoRA refusal tests; `LoRALinear.weight` refusal tests; MultiLoRA behavior-pinning tests; one existing test updated to pass a `pg_collection` (the completion needs a resolvable group).

</details>

# GitHub Actions CI

**Pre checks**:

- [x] Did you write any new necessary tests? — three layers; see verification below
- [x] Did you add or update any necessary documentation? — docstrings state the invariants (partial-sum form, adjoints, group selection); no user-facing doc page applies to a bug fix; the new error messages name the config knob to change
- [x] Does the PR affect components that are optional to install? — **No.** The fix imports only `megatron.core.tensor_parallel.mappings` (a hard dependency); the existing TE `safe_import_from` guards in `peft/utils.py` are untouched

# Additional Information

<details>
<summary>Verification (A/B on 2×H100, real NCCL)</summary>

## Verification (A/B on 2×H100, real NCCL)

Commit 1 checked out alone reproduces every defect: each demonstration is `xfail(strict=True)` and was verified to fail **on its intended assertion** (not a harness error) against the pre-fix tree; commit 2 deletes the markers, so a strict XPASS gate mechanically proves each demonstrated defect is fixed.

| | Pre-fix (fail-first) | Post-fix (same tests as live assertions) |
|---|---|---|
| CPU suite | 5 passed, 8 xfailed | **13/13 passed** |
| Distributed 2-GPU (TP=2×ETP=2) | 4 passed, 7 xfailed | **11/11 passed, both ranks** |
| DoRA / weight-property / MultiLoRA regression tests | 1 passed + 2 xfailed | 26 passed |
| Closed-form algebra | 6 passed | 6 passed |
| Full `tests/unit_tests/peft/` regression | — | 419 passed |

Measured defect signatures pre-fix (matching the closed-form predictions): fc2 combined delta = 2× a concatenation of unsummed partials; fc1 `dL/dA` missing all cross-rank terms under an exact forward; grouped fc2 output `requires_grad == False`; shared overlap and MultiLoRA deltas exactly 2× at TP=2 (MultiLoRA greatest rel. diff 1.27).

</details>

<details>
<summary>Performance and memory — no added communication at ETP=1; two output-side paths got faster</summary>

## Performance and memory

Measured on the same 2x H100 setup, adapter in isolation per the A/B protocol: 50 CUDA-event-timed iterations after 10 warmup, even token counts so the pad path never fires, bf16 (the per-expert-backend row is fp32, that backend's trigger), 4096 tokens, hidden 2048, ffn 2688, rank 32, 4 local experts, TP=2 x ETP=2:

| Case | fwd fixed (ms) | fwd broken (ms) | bwd fixed (ms) | bwd broken (ms) | peak fixed / broken (MiB) |
|---|---|---|---|---|---|
| Default adapter, expert fc2 | 0.251 ± 0.033 | 0.234 ± 0.058 | 0.361 | 0.259 | 147.1 / 147.1 |
| Default adapter, expert fc1 | 0.216 ± 0.015 | 0.195 ± 0.014 | 0.369 | 0.370 | 138.8 / 138.8 |
| Grouped adapter fc2, grouped_mm | 0.370 ± 0.016 | 0.368 ± 0.113 | 0.543 | n/a (no graph) | 195.9 / 153.2 |
| Grouped adapter fc1, grouped_mm | 0.351 ± 0.033 | 0.334 ± 0.013 | 0.531 | 0.231 | 194.3 / 177.5 |
| Grouped adapter fc2, per-expert backend | 0.658 ± 0.035 | 0.559 ± 0.018 | 0.949 | n/a (no graph) | 275.9 / 205.7 |
| Shared-expert-overlap fc2 | 0.210 ± 0.056 | 0.263 ± 0.019 | 0.336 | 0.427 | 224.8 / 224.8 |
| MultiLoRA fc2 (wrapper incl. base linear) | 0.527 ± 0.067 | 0.626 ± 0.222 | 0.372 | 0.360 | 249.5 / 257.5 |

Reading the table — with the essential caveat that the broken arm is **not same-work** (it computes wrong values, partial gradients, or no gradients at all, which is the bug):

- **Peak memory is byte-identical wherever the two arms do comparable work** (both default-adapter rows and shared-overlap), measuring the by-construction claim: the zero-embed buffer replaces the removed gather buffer one-for-one and no new collective saves tensors for backward. The grouped rows' higher fixed-arm peaks exist because the broken arm builds no (fc2) or a truncated (fc1) autograd graph — the delta is the cost of gradients existing.
- **The two output-side-only cases got faster** (shared-overlap −20% fwd, MultiLoRA −16% fwd): the fix removes a gather there and replaces it with a local zero-pad, while the embedded delta rides the reduce that `post_forward_comm`/the dispatcher already performs — zero added communication.
- Elsewhere forward overhead is +5–11%, worst +18% on the per-expert fallback backend where the completion collectives scale with local expert count; all collectives run on `[tokens, rank]` tensors and are latency-bound, and the whole adapter remains a sub-millisecond fraction of an MoE layer.
- Backward deltas where the broken arm has a graph reflect the restored communication plus gradient terms the broken arm never computed (e.g. grouped fc1's `dL/dA`).

At ETP=1 — every shipped recipe — all new branches short-circuit at group size 1, leaving one predicate check per adapter forward; dense LoRA paths are otherwise untouched. Not measured: end-to-end step-time at ETP>1 (the adapter is a small fraction of an MoE layer, but no full-model A/B was run), and a bitwise pre/post A/B at ETP=1 (asserted structurally; backed by the 419-test regression run rather than a bit comparison).

</details>

## Known limits / residual risk

- The claim that the dispatcher sums partial terms across expert-TP is proven from code for the `alltoall` dispatcher (`combine_preprocess` reduce-scatter over `expt_tp`). For the **flex/DeepEP** dispatcher the combine's sum semantics are inferred from the DeepEP contract, not verified — worth a maintainer's confirmation before relying on ETP>1 with `flex`.
- TE-fp8 grouped path and EP>1 × ETP>1 jointly are not covered by the 2-GPU suite (the latter needs ≥4 GPUs, above the functional-test cap); the collectives sit at bf16 boundaries outside TE autograd kernels by construction.
- All shipped recipes pin `expert_tensor_parallel_size=1`, so these defects are latent for stock configs; they fire on custom ETP>1 topologies (where we found them) and — for the shared-expert/MultiLoRA cases — at plain TP>1 with `moe_shared_expert_overlap`.

Fixes #5634. Related to #5636.

### DIFF
--- a/src/megatron/bridge/peft/dora.py
+++ b/src/megatron/bridge/peft/dora.py
@@ -21,7 +21,7 @@ from torch import nn
 from megatron.bridge.peft.base import PEFT
 from megatron.bridge.peft.dora_layers import DoRALinear, ParallelLinearDoRAAdapter
 from megatron.bridge.peft.module_matcher import ModuleMatcher
-from megatron.bridge.peft.utils import get_adapter_attributes_from_linear
+from megatron.bridge.peft.utils import get_adapter_attributes_from_linear, is_expert_linear
 
 
 logger = logging.getLogger(__name__)
@@ -90,7 +90,22 @@ class DoRA(PEFT, ModuleMatcher):
 
         if (ans := self.match(m, name, prefix)) is not None:
             _, full_name = ans
+            if is_expert_linear(full_name):
+                raise NotImplementedError(
+                    f"DoRA does not support MoE expert linears (matched {full_name}): the DoRA "
+                    "adapter is built without is_expert, so its sharding and collectives do not "
+                    "compose with expert parallelism (TP, ETP, or EP > 1). Exclude expert "
+                    "modules from target_modules to proceed."
+                )
             attrs = get_adapter_attributes_from_linear(m)
+            if attrs.disable_tensor_parallel_comm and attrs.base_linear_is_parallel:
+                raise NotImplementedError(
+                    f"DoRA does not support TP-sharded linears whose own tensor-parallel "
+                    f"communication is suppressed (matched {full_name}; e.g. shared experts "
+                    "under moe_shared_expert_overlap): DoRA's per-rank magnitude norm does not "
+                    "compose with the downstream TP reduction. Exclude the module from "
+                    "target_modules. A replicated base (e.g. a duplicated TELinear) is fine."
+                )
             logger.info(f"Adding DoRA to: {full_name}")
             adapter = ParallelLinearDoRAAdapter(
                 attrs.in_features,
--- a/src/megatron/bridge/peft/lora_layers.py
+++ b/src/megatron/bridge/peft/lora_layers.py
@@ -35,6 +35,19 @@ class LoRALinear(AdapterWrapper):
     @property
     def weight(self) -> torch.Tensor:
         """Return the effective base weight, including the LoRA delta when enabled."""
+        from megatron.bridge.peft.utils import GroupedExpertLinearAdapter, SharedOuterGroupedExpertAdapter
+
+        if isinstance(self.adapter, (GroupedExpertLinearAdapter, SharedOuterGroupedExpertAdapter)):
+            # Grouped bases expose weight0..N (no single 2D to_wrap.weight), and these
+            # adapters carry 3D per-expert weights with no `tp_group` attribute — the
+            # group-less merge below would silently merge only the local shard. The
+            # merge path for grouped adapters is the export-side materializer
+            # (models/conversion/peft_bridge.py), not this property.
+            raise NotImplementedError(
+                f"LoRALinear.weight is not supported for grouped expert adapters "
+                f"({type(self.adapter).__name__} on {self.adapter.base_linear_name}); "
+                "use the checkpoint/export merge path instead."
+            )
         base_weight = self.to_wrap.weight
         if not self._adapter_enabled:
             return base_weight
--- a/src/megatron/bridge/peft/multi_lora_layers.py
+++ b/src/megatron/bridge/peft/multi_lora_layers.py
@@ -115,6 +115,18 @@ class MultiLoRALinear(AdapterWrapper):
         # column-sharded adapter output must be gathered before the residual
         # add; a column-parallel base keeps the [tokens, out/tp] shard.
         self._gather_output = attrs.input_is_parallel or not attrs.base_linear_is_parallel
+        # Shared-expert fc2 under moe_shared_expert_overlap: the base's own TP
+        # collectives are suppressed (attrs.disable_tensor_parallel_comm) and
+        # SharedExpertMLP.post_forward_comm sums the full-width partials across
+        # the dense TP group. A gathered delta would be counted once per rank;
+        # zero-embedding this rank's shard instead gives each output element
+        # exactly one non-zero contributor, so the downstream sum counts it once.
+        # Mirrors ParallelLinearAdapter._shared_expert_row_parallel.
+        self._suppressed_comm_row_parallel = bool(
+            attrs.input_is_parallel and attrs.disable_tensor_parallel_comm and attrs.base_linear_is_parallel
+        )
+        if self._suppressed_comm_row_parallel:
+            self._gather_output = False
 
         # ModuleList of ParallelLinearAdapters gives per-adapter optimizer state
         # isolation, clean checkpoint serialization, and bridge export compatibility.
@@ -154,6 +166,21 @@ class MultiLoRALinear(AdapterWrapper):
         self.register_buffer(
             "rank_values", torch.full((n_adapters,), dim, dtype=dtype, device=device), persistent=False
         )
+
+    def _embed_row_parallel_shard(self, out: torch.Tensor) -> torch.Tensor:
+        """Zero-embed this rank's output shard at its dense-TP offset.
+
+        Padding is differentiable and its adjoint is the matching slice, so the
+        downstream ``post_forward_comm`` sum reproduces the delta exactly once with
+        no ``1/tp`` factor and no gradient compensation.
+        """
+        tp_size = parallel_state.get_tensor_model_parallel_world_size()
+        if tp_size <= 1:
+            return out
+        rank = parallel_state.get_tensor_model_parallel_rank()
+        shard_width = out.shape[-1]
+        left = rank * shard_width
+        return nn.functional.pad(out, (left, (tp_size - 1) * shard_width - left))
 
     def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
         linear_output, bias, layernorm_output = self.base_linear_forward(x, *args, **kwargs)
@@ -226,8 +253,13 @@ class MultiLoRALinear(AdapterWrapper):
         # Match the wrapped base linear's output layout: row-parallel base
         # produces a fully-summed [tokens, h_out] tensor (which we then SP
         # scatter); column-parallel base keeps the [tokens, h_out/tp] shard.
+        # A suppressed-comm row-parallel base (shared experts under overlap)
+        # emits a full-width partial that post_forward_comm sums across TP, so
+        # the delta must be contributed as a zero-embedded local shard instead.
         if self._gather_output:
             out = gather_from_tensor_model_parallel_region(out)
+        elif self._suppressed_comm_row_parallel:
+            out = self._embed_row_parallel_shard(out)
 
         if not self.disable_sequence_parallel_comm and self.input_is_parallel:
             if self.use_a2a:
--- a/src/megatron/bridge/peft/utils.py
+++ b/src/megatron/bridge/peft/utils.py
@@ -34,7 +34,10 @@ from megatron.core.dist_checkpointing.mapping import ShardedStateDict, ShardedTe
 from megatron.core.process_groups_config import ProcessGroupCollection
 from megatron.core.tensor_parallel import ColumnParallelLinear, RowParallelLinear, set_tensor_model_parallel_attributes
 from megatron.core.tensor_parallel.mappings import (
+    all_gather_last_dim_from_tensor_parallel_region,
+    copy_to_tensor_model_parallel_region,
     gather_from_sequence_parallel_region,
+    reduce_from_tensor_model_parallel_region,
     scatter_to_sequence_parallel_region,
 )
 from megatron.core.tensor_parallel.random import is_checkpointing
@@ -1022,6 +1025,18 @@ class ParallelLinearAdapter(nn.Module):
         self.tp_group = _get_tensor_parallel_group(self.pg_collection, is_expert=is_expert)
         self.ep_group = _get_process_group(self.pg_collection, "ep")
         self.expert_dp_group = _get_process_group(self.pg_collection, "expt_dp")
+        if (
+            is_expert
+            and dropout > 0.0
+            and _process_group_size(self.tp_group, model_parallel_config.expert_tensor_parallel_size or 1) > 1
+        ):
+            raise NotImplementedError(
+                f"Expert-adapter dropout is not supported at expert_tensor_parallel_size > 1 "
+                f"(got dropout={dropout} for {base_linear_name}): dropout draws per-rank RNG on "
+                "tokens the MoE dispatcher replicates across the expert tensor parallel group, so "
+                "the resulting delta is not expressible by any merged weight. Route dropout "
+                "through the expert-parallel RNG tracker or set dropout=0.0."
+            )
         _sequence_parallel = model_parallel_config.sequence_parallel
         model_parallel_config.sequence_parallel = False  # SP is irrelevant for the lora linear layer
         self.config = model_parallel_config
@@ -1070,6 +1085,24 @@ class ParallelLinearAdapter(nn.Module):
 
         if not base_linear_is_parallel:
             lin_out_gather_output = True
+
+        # An expert linear whose input is already sharded (experts.linear_fc2) emits a
+        # full-width partial that the MoE token dispatcher sums across the expert-TP
+        # group. The adapter must contribute a partial term of that sum, not the full delta: keep linear_out's
+        # local shard and zero-embed it into full width in forward(), so each output
+        # element has exactly one non-zero contributor and the dispatcher's sum
+        # reconstructs B @ z exactly once. Gathering here instead would hand every
+        # rank the same full-width delta, and the sum would count it once per rank.
+        self._expert_row_parallel = bool(is_expert and input_is_parallel and base_linear_is_parallel)
+        # Shared-expert fc2 under moe_shared_expert_overlap is the same disease over
+        # the DENSE TP group: the base's own collectives are suppressed (which is what
+        # disable_tensor_parallel_comm reports) and SharedExpertMLP.post_forward_comm
+        # sums the full-width partials across TP.
+        self._shared_expert_row_parallel = bool(
+            not is_expert and disable_tensor_parallel_comm and input_is_parallel and base_linear_is_parallel
+        )
+        if self._expert_row_parallel or self._shared_expert_row_parallel:
+            lin_out_gather_output = False
 
         self.linear_out = ColumnParallelLinear(
             dim,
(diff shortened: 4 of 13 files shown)


tigist-far:
Restructured the description for readability, top-down: what breaks and for whom, then the mechanism, then a new **Why this design** section covering the constraints and the alternatives considered, then the breaking changes. Verification, performance, and the per-file changelog moved into collapsed sections; known limits stay visible.

The per-path defect enumeration now lives only in #5634 rather than being duplicated here.

No code, test, or technical-claim changes — description only.

PR #5642, opened 2026-08-18
perf(training): add NVTX ranges around training loop iterations

dshirron3:
# What does this PR do ?
Wrap each training iteration with named NVTX push/pop ranges so nsys timelines clearly show per-step boundaries, including skip and early-exit paths. This is mainly to enable analysis tools to easily identify step boundries.

# Changelog

1) Wrap each training loop iteration in train() with an NVTX range named training_step_<step> via nvtx_range_push / nvtx_range_pop
2) Push the range at the start of each iteration (after handle_profiling_step)
3) Pop the range on skip-iteration (continue), early exit from train_step (should_exit), and before handle_profiling_stop on the normal path
4) Improves nsys timeline readability by marking full per-iteration boundaries (imports / @nvtx_decorator() on train_step were already present)
image

### DIFF
--- a/src/megatron/bridge/training/train.py
+++ b/src/megatron/bridge/training/train.py
@@ -385,6 +385,9 @@ def train(
         if nvtx_ctx is not None:
             nsys_nvtx_context = nvtx_ctx
 
+        nvtx_step = global_state.train_state.step
+        nvtx_range_push(suffix=f"training_step_{nvtx_step}")
+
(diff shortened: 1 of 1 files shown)


yaoyu-33 (approved):

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5642/commits/d304161ea8cd9c80c4187d0656fdd497c6be3a67

PR #5643, opened 2026-08-18
fix(data): keep object-store schemes out of memmap index paths

Josephasafg:
# What does this PR do ?

`_index_fn` builds the index sidecar path by joining `index_mapping_dir` with the data file's name, stripping only leading slashes. An object-store source therefore keeps its scheme inside a local path:

```
/tmp/index  +  msc://bucket/prefix/train_0.jsonl 
->  /tmp/index/msc://bucket/prefix/train_0.jsonl.idx
```
MSC reads the embedded `://` back as a nested URL and resolves the bucket profile out of it, so the parent `mkdir` and the sidecar write disagree about where the file lives and the write fails:

```
[Errno 2] No such file or directory:
  '/tmp/index/msc:/bucket/prefix/.train_0.jsonl.idx.npy.tmp'
  -> '/tmp/index/msc:/bucket/prefix/train_0.jsonl.idx.npy'
```

`build_index_files` then hides the cause: `safe_map` returns `None` for the failed file and `sum(build_status)` raises `TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'`, so the traceback points at a log-formatting line and the real error survives only as a `WARNING`.

Hit with `_JSONLMemMapDataset` over `msc://` JSONL, but the same path serves `GPTSFTDataset`: any remote JSONL source combined with `index_mapping_dir` fails this way.
`index_mapping_dir` is exactly what you need for a read-only data bucket, so the two features are unusable together.

## Testing

Three cases added to `test_index_fn` covering `msc://`, `gs://` and `s3://` data files with a local `index_mapping_dir`, plus an explicit assertion that no scheme survives.


Josephasafg:
@yaoyu-33 Can I get this reviewed please? Thanks!

Josephasafg:
@chtruong814 @yaoyu-33 I added one more test which should make the `codecov` CI check to pass

yaoyu-33 (approved):

PR #5645, opened 2026-08-18
feat(model): DP of Nemotron Omni vision encoder across CP ranks

huvunvidia:
## Summary

Every context-parallel rank currently encodes the **full** image set of a microbatch, so the vision tower's activations are replicated `CP` times while the language model's are sharded. This PR adds a `vision_dp_over_cp` flag that gives each CP rank an integer share of the patched images (or tubelets, when temporal compression is on).

- Uses MCore's existing `split_to_context_parallel_ranks_dynamic_res` / `gather_from_context_parallel_ranks_dynamic_res`, then gathers projected features back so every rank's media merge still sees the global feature set.
- RADIO already attends per image via `cu_seqlens`, so this is **data parallelism borrowing the CP process group**, not sequence sharding. No cross-rank attention communication is introduced; the only added collective is the post-projector gather. This mirrors what `llava_model.py` already does upstream for the temporal path.
- Two Omni-specific details:
  - Project **before** gathering, so the projector stays sharded too.
  - Zero-pad odd patch grids before pixel shuffle. When a microbatch has fewer images than CP ranks, the MCore splitter injects `1x1` placeholder images to keep every rank non-empty, and Omni's factor-2 pixel shuffle rejects an odd grid outright. Those rows are dropped by the gather.
- Guards that `pg_collection.cp.size()` matches `context_parallel_size`, so a divergent process-group collection fails loudly instead of sharding against the wrong ranks.

## Breaking changes

Two, hence `feat(model)!`:

1. **`vision_dp_over_cp` defaults to `True`.** It ANDs with `context_parallel_lm > 1`, so a CP=1 run — including both checked-in Nemotron Omni recipes — behaves exactly as before. Raising CP now silently enables image sharding too. Set `model.vision_dp_over_cp=False` to opt out.
2. **The field was named `vision_context_parallel` in earlier revisions of this PR.** Renamed at review request: the old name read as sequence-dimension context parallelism, which this is not, and it collided visually with the pre-existing and completely unrelated `dist_train.vision_context_parallel_size` (Qwen3-VL dist-train, gives the vision tower its *own* ranks). That knob is untouched.

## Measurements

Nemotron-3 Omni 30B-A3B SFT on 16 GPUs. Each pair of rows is an A/B from the same launcher with only `vision_dp_over_cp` flipped. `PP=1`, `DP=2`, `EP=8`, `ETP=1`, `MBS=2`, `GBS=64`, `GA=16` throughout.

| Dataset | TP | CP | Vision DP over CP | Step time | Peak memory |
|---|---|---|---|---|---|
| InfoVQA (high resolution) | 4 | 2 | false | 13.2 s | 81 GB |
| | 4 | 2 | **true** | **12.5 s** (-5.3%) | **73 GB** (-9.9%) |
| Mantis (multiple images) | 2 | 4 | false | 12.2 s | 81 GB |
| | 2 | 4 | **true** | **11.6 s** (-4.9%) | **77 GB** (-4.9%) |
| llava_video (video) | 4 | 2 | false | 13.6 s | 84 GB |
| | 4 | 2 | **true** | **13.0 s** (-4.4%) | **75 GB** (-10.7%) |

Peak memory drops 5-11% and step time 4-5% across all three media shapes. The step-time win is real work removed rather than just memory pressure relieved: each rank now encodes `1/CP` of the images instead of every one of them.

Mantis is the weakest case, and predictably so. The split is by image **count**, not patch count, and the remainder lands on the last rank — so a `CP=4` run over samples carrying 1-5 images leaves the tail rank doing more than its share. InfoVQA (one large image per sample, `CP=2`) and llava_video (fixed canvas, proportional tubelet split) both divide cleanly. A patch-count-balanced split on image boundaries would close that gap; that change belongs upstream in MCore.

## Test plan

- [x] `ruff check` / `ruff format --check` clean on all changed files
- [x] New unit tests for `_pad_patch_grid_to_even`: even grid passes through unchanged, odd grid is zero-padded so pixel shuffle accepts it, and padding keeps every real patch differentiable
- [x] Two-rank equivalence test (`test_vision_dp_over_cp_distributed.py`) asserts the sharded tower matches the unsharded one for fewer images than ranks, an exact split, and a remainder
- [x] Unit tests for the process-group mismatch guard and for the silent self-disable at CP=1
- [x] Multi-GPU A/B on InfoVQA, Mantis, and llava_video (table above) — covers high-resolution single-image, multi-image, and video paths
- [ ] `uv run python -m pytest tests/unit_tests/models/nemotron_omni/` (needs a GPU container; lint was run on the login node)
- [ ] `CP=8` placeholder-padding path (fewer images than ranks) is exercised by unit tests but not yet end-to-end

Docs and the accompanying perf skill are in #5646, which should merge **after** this PR.


huvunvidia:
Test coverage note addressed in `ab975bf9c`:

| Review ask | Added |
|---|---|
| Guard raises `ValueError` on `pg_collection.cp.size()` divergence | `test_vision_context_parallel_rejects_process_group_mismatch` |
| CP=2 sharded encoding reproduces CP=1 projected features | `test_vision_context_parallel_matches_unsharded_encoder` |
| — (not flagged) | `test_vision_context_parallel_is_disabled_when_cp_is_one` |

Two deliberate deviations:

**Unit, not functional.** The CP=2 test self-launches `torch.distributed.run --nproc_per_node=2` via `subprocess.run` from a normal pytest test. `Launch_Unit_Tests_Core.sh` already runs with `CUDA_VISIBLE_DEVICES="0,1"`, so it runs on *every* PR — an L1 functional test would only fire on main/schedule/label and consume a 2-GPU functional slot. CONTRIBUTING.md prefers unit over functional and endorses the `subprocess.run` pattern.

**Broader than one shape.** Parameterized over `num_images ∈ {1, 2, 3}`, which at CP=2 covers placeholder padding (fewer images than ranks), exact split, and remainder-to-last-rank.

Verified on 2 GPUs: 87 passed, 1 skipped for the `nemotron_omni` directory; ruff clean. Mutation check — changing `gather_from_context_parallel_ranks_dynamic_res(projected, num_padded_ranks)` to `..., 0)` makes the CP=2 test fail (`sharded (2, 128) != reference (1, 128)`), confirming it is not vacuous.

cuichenx (approved):
LGTM

PR #5693, opened 2026-08-19
perf(nemotron): add Nemotron 3.5 Lightning GB300 recipes

malay-nagda:
## What does this PR do?

Adds canonical Nemotron 3.5 Lightning throughput-benchmark recipes for an 8-GPU GB300 system.

- Adds BF16, MXFP8, NVFP4, and MXFP8 Megatron FSDP variants.
- Exports the recipes through `megatron.bridge.perf_recipes.nemotronh` and registers them in the performance-config inventory.
- Extends the existing Nemotron 3.5 recipe tests to cover GB300 model finalization, MTP settings, topology, precision-specific tuning, and FSDP defaults.
- Corrects the copied FSDP comment: the GB300 FSDP microbatch is 3, while the non-FSDP GB300 recipes use 4.

These are benchmark recipes for upper-bound throughput measurement, not production convergence recipes.

## Testing

- `uv run --with pre-commit --no-project pre-commit run --all-files` — passed.
- Targeted pytest was attempted but could not run in the local checkout: project sync fails while building `fast-hadamard-transform` because its isolated build environment does not declare `torch`, and the existing `.venv` does not contain `pytest` or `torch`.
- Static inventory validation confirmed that all four GB300 inventory entries resolve to defined recipe factories.

### DIFF
--- a/src/megatron/bridge/perf_recipes/nemotronh/gb300/nemotronh.py
+++ b/src/megatron/bridge/perf_recipes/nemotronh/gb300/nemotronh.py
@@ -581,7 +581,6 @@ def nemotron_3_5_lightning_pretrain_8gpu_gb300_fp8mx_config() -> ConfigContainer
     """Nemotron 3.5 Lightning pretrain: 8× GB300, MXFP8."""
     cfg = _build_nemotron_3_5_lightning_gb300_mxfp8()
     cfg.model.use_transformer_engine_op_fuser = True
-    cfg.model.moe_mlp_glu_interleave_size = 32
     cfg.mixed_precision.fp8_dot_product_attention = True
     cfg.env_vars = {
         **COMMON_PERF_ENV_VARS,
--- a/src/megatron/bridge/perf_recipes/nemotronh/gb300/nemotronh.py
+++ b/src/megatron/bridge/perf_recipes/nemotronh/gb300/nemotronh.py
@@ -34,6 +34,11 @@ from megatron.bridge.perf_recipes.nemotronh.common import (
 from megatron.bridge.utils.cuda_graph import set_cuda_graph_modules
 
 
+# Public Nemotron 3.5 Lightning checkpoint used by the Lightning recipe API.
+_NEMOTRON_3_5_LIGHTNING_MODEL_ID = "nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16"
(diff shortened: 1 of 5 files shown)


src/megatron/bridge/perf_recipes/nemotronh/gb300/nemotronh.py
```diff
@@ -524,3 +529,152 @@ def nemotronh_56b_pretrain_256gpu_gb300_fp8cs_config() -> ConfigContainer:
...
+    cfg.tokenizer.tokenizer_model = _NEMOTRON_3_5_LIGHTNING_MODEL_ID
+    cfg.tokenizer.hf_tokenizer_kwargs = {"revision": _NEMOTRON_3_5_LIGHTNING_MODEL_REVISION}
+    return cfg
+
+
+def nemotron_3_5_lightning_pretrain_8gpu_gb300_fp8mx_config() -> ConfigContainer:
+    """Nemotron 3.5 Lightning pretrain: 8× GB300, MXFP8."""
+    cfg = _build_nemotron_3_5_lightning_gb300_mxfp8()
+    cfg.model.use_transformer_engine_op_fuser = True
+    cfg.model.moe_mlp_glu_interleave_size = 32
```
cuichenx: this shouldn't be turned on? is this required for perf

malay-nagda: this is for perf (enabled w/ cutedsl fusion)

cuichenx: i don't think it's getting used here since nemotron 3.5 lightning uses squared relu and no gated linear unit. This flag does change the checkpoint layout so we don't want it accidentally picked up by users

malay-nagda: ok. will test this w/o `moe_mlp_glu_interleave_size` and then update here

malay-nagda: @cuichenx updated to remove `moe_mlp_glu_interleave_size ` for 3.5 MXFP8

cuichenx (approved):
LGTM

dingqingy-nv (approved):

PR #5727, opened 2026-08-21
fix(data): use Energon validation recipe for val split

### LINKED ISSUE #5626: [bug] Energon validation set is shuffled and never resets — validation loss is not reproducible
## What happens

The Energon val split is built with the **training** recipe —
`get_train_dataset(split_part="val")` in
`src/megatron/bridge/data/energon/base_energon_datamodule.py:178`.

So the validation stream is shuffled and infinitely repeating. The iterator is also created once
and never rebuilt, so each validation pass continues where the previous one stopped.

## Example

With `eval_interval=50`, `eval_iters=10`:

- validation at step 50 evaluates samples 1-320
- validation at step 100 evaluates samples 321-640
- ...

Every validation scores a **different, shuffled subset**. Val loss can move between steps purely
because the subset changed, and two runs of the same config produce different val curves.

Side effect of the infinite repeat: a val dataset smaller than `dp_size × num_workers` fails at
startup with `"Cannot repeat empty dataset indefinitely"` (`wrappers/repeat_dataset.py:57`).

## Expected

Validation evaluates the same samples every time it runs, so the curve reflects only the model.

## Suggested fix

Use Energon's val recipe for the val split — `get_val_dataset()` instead of
`get_train_dataset(split_part="val")`. It is unshuffled and finite, and its optional `limit`
wraps the data in `LimitDataset(reset_after_epoch=True)`, which rewinds after each pass — so
repeated validations replay the same samples.

This also removes the empty-dataset failure above, since nothing needs to repeat indefinitely.

Happy to open a PR if this looks right.

caohch-1:
# What does this PR do ?

Use Energon's validation dataset recipe for the validation split instead of constructing validation data through the training recipe.

# Changelog

- Route the training split through `get_train_dataset()` and the validation split through `get_val_dataset()`.
- Keep training-only shuffle configuration out of the validation path while preserving existing dataset-loading options.
- Add regression coverage for validation dataset recipe selection and update existing validation dataloader tests to use the validation builder.

# Validation

- A focused regression harness confirms that the pre-fix implementation selects `get_train_dataset()` for validation and the updated implementation selects `get_val_dataset()`.
- The same harness confirms that the training split continues to use `get_train_dataset()`.
- `git diff --check` passes.
- Python compilation checks pass for both changed files.

# GitHub Actions CI

This is an external contribution, so GitHub Actions CI will require approval from an NVIDIA developer.

**Pre checks**:

- [x] Did you add or update any necessary documentation? No documentation changes are needed for this fix.
- [ ] Does the PR affect components that are optional to install? No.

# Additional Information

Fixes #5626

### DIFF
--- a/src/megatron/bridge/data/energon/base_energon_datamodule.py
+++ b/src/megatron/bridge/data/energon/base_energon_datamodule.py
@@ -16,7 +16,7 @@ import logging
 from typing import Any, Literal, Optional
 
 from megatron.core.process_groups_config import ProcessGroupCollection
-from megatron.energon import WorkerConfig, get_savable_loader, get_train_dataset
+from megatron.energon import WorkerConfig, get_savable_loader, get_train_dataset, get_val_dataset
 
 
 logger = logging.getLogger(__name__)
@@ -88,7 +88,7 @@ class EnergonMultiModalDataModule:
         and batching samples for validation. Defaults to None and will be the same as task_encoder.
         pg_collection (ProcessGroupCollection, optional): Process group collection for distributed training.
         If provided, used instead of the global parallel_state. Defaults to None.
-        **kwargs: Additional keyword arguments. Will be passed to get_train_dataset() of Energon
+        **kwargs: Additional keyword arguments passed to the selected Energon dataset builder.
         """
 
         super().__init__()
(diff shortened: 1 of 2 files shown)


src/megatron/bridge/data/energon/base_energon_datamodule.py
```diff
@@ -171,24 +171,29 @@ def datasets_provider(self, worker_config, split: Literal["train", "val"] = "val
...
+                max_samples_per_sequence=self.max_samples_per_sequence,
+                **self.kwargs,
+            )
 
-        _dataset = get_train_dataset(
+        return get_val_dataset(
             self.path,
             batch_size=self.micro_batch_size,
-            task_encoder=task_encoder,
+            task_encoder=self.validation_task_encoder,
```
yaoyu-33, line 330: validation_task_encoder will this always exist? or required

caohch-1: Yes. `validation_task_encoder` is initialized in `__init__` as `validation_task_encoder or self.task_encoder`, so it falls back to the training task encoder when a separate validation encoder is not provided. The existing val path already used `self.validation_task_encoder`; this PR only changes the dataset builder from `get_train_dataset()` to `get_val_dataset()`.

If both encoders are `None` in direct construction, the value can still be None, but that is pre-existing behavior and is not introduced by this change.

psycherun-kaiko:
@yaoyu-33 gentle ping :) 

This would unblock reliable val curves 📉

PR #5541, opened 2026-08-12
[4/4] fix(data): align thinking-model chat templates and masks

yaoyu-33:
## Summary

- retain the existing offline Qwen 3.5/3.6 tools × thinking coverage
- translate canonical thinking controls to the native GLM-5.2 and Kimi-K3 tokenizer arguments
- infer assistant-only loss boundaries for GLM-5.2, Kimi-K3, and MiniMax-M3 templates that do not expose HF generation masks
- preserve OpenAI JSON-string tool arguments, tools, right truncation, padding, and shifted-label alignment
- keep MiniMax `thinking_mode="adaptive"` unchanged

This PR is stacked on #5538.

## Fixes

- GLM-5.2: map `truncate_history_thinking` to native `clear_thinking`; use conversation-prefix renders because completed assistant turns have no explicit end token
- Kimi-K3: map `enable_thinking` to native `thinking`; detect the custom XTML assistant structure without depending on the remote-code class name
- MiniMax-M3: recognize its assistant/tool role markers and include the assistant turn terminator in loss
- shared preprocessing: preserve `chat_template_kwargs` and top-level `tools` when re-rendering a conversation to derive its mask

The review also removed unconditional signature/vocabulary inspection from the per-row path. Vocabulary probing now occurs only for template-less tokenizers that expose the Kimi-style `thinking` parameter.

## Official template audit

Pinned snapshots checked on 2026-08-13:

| Model | Revision | Cases | MBridge ↔ native serving text/IDs |
|---|---|---:|---|
| `zai-org/GLM-5.2` | `b4734de4facf877f85769a911abafc5283eab3d9` | 4 | exact / exact |
| `moonshotai/Kimi-K3` | `9f62e4e9fffbd0a83ddd60e1c209d828994b3569` | 4 | exact / exact |
| `MiniMaxAI/MiniMax-M3` | `f0e1c1e04d40177e4673a22097036854f536e9c0` | 4 | exact / exact |

The four cases for each model are with/without tools × with/without thinking. Raw OpenAI JSON-string `function.arguments` is used in tool cases.

The existing Qwen audit remains exact in all 8 Qwen 3.5/3.6 cases.

## Loss-mask audit

MBridge masks system/user/tool-response content and assistant role headers. It supervises retained thinking, assistant responses, assistant tool calls, and model-defined assistant turn terminators where present.

Against ms-swift `ca937fbaf8e0c3dc4ea34358889430e36475463b`:

[...]

PR #5742, opened 2026-08-22
fix(data): apply the null-content contract to legacy chat turns

pruprakash:
# What does this PR do ?

Fixes ShareGPT chat SFT rows with a null `value` training the model on the literal string `None`, by applying the documented null-content contract to the legacy `from`/`value` branch of `normalize_chat_conversation`.

# Changelog

- `src/megatron/bridge/data/conversation_processing.py`: legacy `from`/`value` branch converts a null `value` to `""`; docstring widened to both schemas.
- `tests/unit_tests/data/test_conversation_processing.py`: 2 tests over the patched branch and null/empty equivalence.

# Additional Information

- **Root cause:** `conversation_processing.py:606-611` applies `None` -> `""` then `continue`s, so the legacy branch at `:612-615` copied `value` into `content` unconverted. Jinja stringifies `None` instead of raising, so it reaches the training text silently.
- **Blast radius:** any chat SFT run over a ShareGPT-format corpus, every model. Not DeepSeek-specific.
- **Regression?** No — broken since null handling landed in `46c42552` (#5541); the legacy branch (`0845a7dd`, #5078) never had it.
- **Verification:** red-green in `nvcr.io/nvidian/nemo:nightly` on EOS at base `4a3f2c2e`. RED (job 5901189): 2 failed on `content=None` vs `''`. GREEN (job 5901199): 2 passed, file 110 passed. `tests/unit_tests/data/` 796 -> 798 passed, same 2 pre-existing unrelated failures.
- **Not included:** the `raise` at `:616` also rejects an assistant turn with `tool_calls` and no `content` key, making DeepSeek-V4 tool-call-only support unreachable from `build_direct_hf_sft_split`. Separate root cause, and the remedy is a maintainer call — reported separately, not fixed here.

### DIFF
--- a/src/megatron/bridge/data/conversation_processing.py
+++ b/src/megatron/bridge/data/conversation_processing.py
@@ -578,8 +578,9 @@ def normalize_chat_conversation(
     """Normalize supported text-chat schemas to OpenAI-style messages.
 
     Rows may use ``messages``, singular ``conversation``, or legacy plural
-    ``conversations`` with ``from``/``value`` keys. Null OpenAI message content
-    is converted to an empty string before chat-template rendering.
+    ``conversations`` with ``from``/``value`` keys. Null message content is
+    converted to an empty string before chat-template rendering, in either
+    supported schema.
     """
     if isinstance(example_or_conversation, Mapping):
         source = example_or_conversation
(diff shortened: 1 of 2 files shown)


pruprakash:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5742/commits/82a8ee111e9251cf1255d6b88ce82ffc5c7049b0

PR #5753, opened 2026-08-23
fix(data): scope GLM assistant boundary scan to the current turn

pruprakash:
# What does this PR do ?

Fixes GLM-shaped SFT rows aborting the data pipeline with `ValueError: AssistantMaskBoundaryConfig did not match any loss-contributing spans in input_ids` whenever any turn quotes a role marker such as `<|observation|>`.

# Changelog

- `src/megatron/bridge/data/conversation_processing.py`: `_conversation_contains_boundary_tokens` gains a keyword-only `turns`; the non-terminal assistant guard passes `turns=[turn]`.
- `tests/unit_tests/models/test_glm52_kimi_k3_minimax_m3_chat.py`: 2 tests covering both branches of the guard.

# GitHub Actions CI

**Pre checks**:

# Additional Information

- **Root cause:** the guard at `conversation_processing.py:1449` scanned every turn's payload, but only the current turn renders between `content_start` and the next role header. Both `True` and the unknown `None` abort, so a marker quoted anywhere vetoed the boundary.
- **Blast radius:** GLM-shaped templates expose no `{% generation %}` mask to fall back to and `data/collators/sft.py:230` does not catch, so the row killed the training job instead of degrading.
- **Regression?** Yes — introduced by #5633 (`2d3733eb`, 2026-08-19), which added the guard and widened the GLM marker map to `<|system|>`/`<|user|>`/`<|observation|>`. Before it the row was mis-masked, not unloadable. Fail-closed is preserved when the marker is inside the assistant's own payload.
- **Verification:** red-green in `nvcr.io/nvidian/nemo:nightly` — RED (test patch only) 1 failed; GREEN 19 passed in the file, plus 129 passed across `test_conversation_processing.py` and `test_sft_processing.py`.
- **Not included:** a separate pre-existing defect (the `turn_search_start >= turn_limit` window from #5166 collapses for a *second* non-terminal assistant turn under a history-dependent template) is unfixed here — it needs a different window derivation for all templates, not just GLM.

### DIFF
--- a/src/megatron/bridge/data/conversation_processing.py
+++ b/src/megatron/bridge/data/conversation_processing.py
@@ -1446,8 +1446,10 @@ def _assistant_mask_from_conversation_turns(
         if role in rendered_turn_end_roles:
             content_end = after_end = turn_limit
             if turn_index + 1 < len(conversation):
+                # Only this turn's payload renders between ``content_start`` and the next role
+                # header, so a marker quoted in any other turn cannot shift that boundary.
                 boundary_tokens_in_payload = _conversation_contains_boundary_tokens(
-                    example_or_conversation, tokenizer, boundary_config
+                    example_or_conversation, tokenizer, boundary_config, turns=[turn]
                 )
                 if boundary_tokens_in_payload is not False:
                     return None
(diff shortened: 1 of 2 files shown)


pruprakash:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5753/commits/c18b8de30e96919e8fb9e9d7f4be825e9060bbf3

PR #5813, opened 2026-08-26
fix(model): pin fp32 MoE routing for Qwen3.5 text bridge

Josephasafg:
# What does this PR do ?

`Qwen35MoEBridge` was the only Qwen GDN+MoE path still routing in bf16, which made expert selection unstable between identical runs. Sets `moe_router_dtype="fp32"` to match the `Qwen3.5 VL` provider and the `Qwen3-Next` bridge.


Josephasafg:
@yaoyu-33 Can you please take a look? Thanks!

PR #5814, opened 2026-08-26
docs(model): correct DeepSeek V4 conversion status

yaoyu-33:
## Summary
- mark DeepSeek V4 CPU and GPU conversion unverified
- remove historical GPU claims that cannot be reproduced from the recorded public dependency graph
- record the exact upstream dependency prerequisite for renewed verification

## Evidence
- the card commit pins Megatron-Core `6513e3e23`, which does not implement `dsv4_hybrid`
- current main pins `731b791469`, which also lacks that attention variant
- both public revisions fail before tensor conversion with `Invalid experimental attention variant: dsv4_hybrid`

A dependency bump must be handled separately; after that lands, CPU and GPU import/export need complete reruns and tensor/reload audits.

## Validation
- model-card validator and YAML safe-load
- `git diff --check`
- repository pre-commit hooks via the no-project fallback (the exact project command cannot resolve the uninitialized Megatron-LM submodule in a fresh worktree)

### DIFF
--- a/examples/model_verification_cards/deepseek-v4-flash/card.yaml
+++ b/examples/model_verification_cards/deepseek-v4-flash/card.yaml
@@ -4,17 +4,19 @@
 title: deepseek_v4_flash
 summary: >
   Performance scope: pretrain_performance.GB200 and pretrain_performance.GB300
-  use their tuned canonical 128-GPU MXFP8 recipes.
-  Timing and throughput metrics from functional training items are sanity
-  checks rather than optimized performance results.
+  use tuned canonical 128-GPU MXFP8 recipes, but both evidence leaves are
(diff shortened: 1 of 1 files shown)


yaoyu-33:
Corrected the verification claims in 854f648284ecf1d6e75d821b19ecbaa4c586cdfc. I checked both recorded dependency graphs: the top-level Bridge commit pins Megatron-Core 6513e3e23, and the GB300 override pins cd4afffa6; both reject experimental_attention_variant=dsv4_hybrid. The affected GB200 training, SFT/export, long-context, resume, and performance leaves plus GB300 performance are now unverified with null metrics and explicit rerun gates. The verification index and summary are synchronized, and the card validator passes.

PR #5817, opened 2026-08-26
feat(model): text-only conversion mode for the Qwen3.5/3.6 VL MoE bridge

Josephasafg:
# What does this PR do ?

Qwen3.5/3.6 MoE checkpoints are `Qwen3_5MoeForConditionalGeneration`, so AutoBridge always builds the VL model, vision tower included, even for text-only fine-tuning.

`QWEN35_CONVERSION_MODE=text` makes the VL MoE bridge return a plain `GPTModelProvider` and map only the language model - the same switch `gemma4` already has with `GEMMA4_CONVERSION_MODE`. Default behavior is unchanged, the shared LM overrides move into a helper used by both bridges.

On export, the checkpoint's `config.json` still declares the VL architecture, so vLLM and transformers expect the `model.visual.*` tensors even though text mode never converts them. The bridge copies them from the source checkpoint on export - same passthrough pattern as `KimiK3Bridge` and `NemotronOmniBridge`, only here it's gated on text mode.


Josephasafg:
@yaoyu-33 Can I please get a review? Thanks!

Josephasafg:
Updated PR to also support `stream_weights_megatron_to_hf` for VL in text mode qwen

Josephasafg:
@yaoyu-33 can I get a review please? Thanks!

PR #5869, opened 2026-08-28
fix(data): chunked materialization with fork-based parallelism

kenkenpa2126:
# What does this PR do ?

Chunked materialization of dataset items with fork-based parallelism to avoid OOM when packing large datasets.

# Changelog

- Replaced `np.array([dataset[i] for i in range(len(dataset))])` with chunked processing: items are processed in fixed-size chunks (5000), each written to a temporary `.npy` file and released from memory immediately.
- Replaced `torch.multiprocessing.Pool` with fork-based `multiprocessing.Pool` to avoid torch's shared memory reduction, which leaks across chunks and causes OOM on large datasets (e.g. 700k+ samples, 32k seq_len).
- Pool is created fresh per chunk so worker processes are terminated and their memory freed after each chunk, allowing num_workers to scale without accumulating shared memory leaks.
- Updated existing tests to work with the new chunked implementation; added tests for parallel correctness and dict item preservation.

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci) in the Contributing doc for how to trigger the CI.

- [x] Did you add or update any necessary documentation? (No)
- [x] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc) (No)

# Additional Information

- Verified on NVIDIA H100 node (Intel Xeon Platinum 8480C, 2TB CPU RAM): 700k+ sample dataset (7GB JSONL, 32k seq_len) — torch.multiprocessing.Pool caused OOM/SIGBUS due to shared memory leaks despite 2TB RAM; serial (1 worker) processing was too slow (~4h for 86% completion); fork-based multiprocessing with 4 workers completes in ~47 minutes with no OOM.


kenkenpa2126:
Adding some additional benchmark data to help with the review.

This PR includes two independent changes to `_materialize_dataset_items`:
1. **Chunked processing**: items are processed in fixed-size chunks (5000) instead of all at once, with each chunk written to a temporary `.npy` file and released from memory immediately.
2. **Fork-based multiprocessing**: replaces `torch.multiprocessing.Pool` with `multiprocessing.Pool(get_context("fork"))` to avoid torch's `file_system` sharing strategy, which creates shared memory files in `/dev/shm` for each tensor passed between workers.

## Code comparison

The original implementation uses a single `torch.multiprocessing.Pool` to process all items, accumulating results in `list(pool.imap(...))` before converting to `np.array`:

```python
# Original implementation (simplified)
torch.multiprocessing.set_sharing_strategy("file_system")
with Pool(num_workers, initializer=_init_shared_dataset_worker, initargs=(dataset,)) as pool:
    items = list(tqdm(pool.imap(_get_shared_dataset_item, range(len(dataset))), total=len(dataset)))
return np.array(items)
```

This PR replaces it with chunked processing + fork-based Pool:

```python
# This PR (simplified)
ctx = mp.get_context("fork")
for start in tqdm(range(0, total, _MATERIALIZE_CHUNK_SIZE), desc="Chunked materialize"):
    end = min(start + _MATERIALIZE_CHUNK_SIZE, total)
    indices = list(range(start, end))
    with ctx.Pool(num_workers, initializer=_init_shared_dataset_worker, initargs=(dataset,)) as pool:
        items = pool.map(_get_shared_dataset_item, indices)
    chunk = np.array(items)
    np.save(chunk_file, chunk, allow_pickle=True)
    chunk_files.append(chunk_file)
    del chunk, items
    gc.collect()
arrays = [np.load(f, allow_pickle=True) for f in chunk_files]
return np.concatenate(arrays, axis=0)
```

A fresh Pool is created per chunk so worker processes are terminated (and their memory freed) after each chunk. Fork context avoids torch's shared memory reduction machinery — child processes inherit the dataset via copy-on-write instead of serializing tensors to `/dev/shm` files.

## Benchmark

To isolate the effect of each change, I tested 4 configurations:

| config | chunk | pool | description |
|---|---|---|---|
| original | no | torch.multiprocessing.Pool | MB v0.6.0 implementation |
| chunk_only | yes | torch.multiprocessing.Pool | chunking only |
| fork_only | no | multiprocessing.Pool(fork) | fork only |
| chunk_fork | yes | multiprocessing.Pool(fork) | this PR (both) |

**Environment:** H100 node (2TB CPU RAM), CPU-only (packing is a pure CPU workload — the offline packing step tokenizes samples and builds packed sequences, so no GPU is required), NeMo 26.08.00 container (MB v0.6.0), 700k+ samples at seq_len=32,768. Each run was executed with exclusive access to a node to avoid interference from other jobs, as they share `/dev/shm` and CPU resources.

### Code for each config (simplified)

**chunk_only** — same chunking logic as this PR, but keeps `torch.multiprocessing.Pool`:

```python
# chunk_only (simplified)
torch.multiprocessing.set_sharing_strategy("file_system")
for start in range(0, total, _CHUNK_SIZE):
    indices = list(range(start, min(start + _CHUNK_SIZE, total)))
    with torch.multiprocessing.Pool(num_workers, ...) as pool:
        items = pool.map(_get_shared_dataset_item, indices)
    chunk = np.array(items)
    np.save(chunk_file, chunk, allow_pickle=True)
    del chunk, items
    gc.collect()
return np.concatenate([np.load(f, allow_pickle=True) for f in chunk_files], axis=0)
```

**fork_only** — replaces `torch.multiprocessing.Pool` with fork-based `multiprocessing.Pool`, but no chunking (accumulates all items at once):

```python
# fork_only (simplified)
ctx = mp.get_context("fork")
with ctx.Pool(num_workers, ...) as pool:
    items = list(pool.imap(_get_shared_dataset_item, range(len(dataset))))
return np.array(items)
```

**Benchmark setup** (all configs use the same MB v0.6.0 pack pipeline):

```python
cfg_dataset = GPTSFTDatasetConfig(
    dataset_root=<data_dir>,
    seq_length=32768,
    do_validation=False,
    do_test=False,
    enable_offline_packing=True,
    offline_packing_specs=PackedSequenceSpecs(
        packed_sequence_size=32768,
        pad_seq_to_mult=1,
        num_tokenizer_workers=num_workers,
    ),
    dataset_kwargs={"chat": True, "use_hf_tokenizer_chat_template": True},
)
builder = GPTSFTDatasetBuilder(config=cfg_dataset, tokenizer=tokenizer)
builder.prepare_packed_data()
```

The only difference between configs is `_materialize_dataset_items` (monkey-patched via `PYTHONPATH` import). The packed output is identical across all configs.

### Results

**workers=1 (serial, no parallel Pool):**

| config | result | elapsed |
|---|---|---|
| original | ✅ completed | 4h 10m |
| chunk_only | ✅ completed | 128 min |
| fork_only | ✅ completed | 4h 15m |
| chunk_fork | ✅ completed | 127 min |

With workers=1, no Pool is created (serial path), so no shared memory is used. All configs complete without OOM and show nearly identical timings (~4h for `original`/`fork_only`, ~128 min for `chunk_only`/`chunk_fork` — note the chunked configs are actually faster even in serial mode). The serial path is CPU-bound (~50-70 it/s, slowing over time), so 700k+ items take 2-4 hours depending on the config.

**workers=4 (parallel Pool):**

| config | result | elapsed |
|---|---|---|
| original | **OOM at ~2%** | — |
| chunk_only | ✅ completed | ~60 min |
| fork_only | **OOM at ~2%** | — |
| chunk_fork | ✅ completed | ~60 min |

With workers=4, the parallel path uses a Pool with shared memory for tensor transfer. `original` and `fork_only` (no chunking) both OOM at ~2% of the dataset with `RuntimeError: unable to mmap ... Cannot allocate memory`, despite the process tree RSS being only ~9 GB at the time of OOM (see Key finding 4). `chunk_only` and `chunk_fork` complete in ~60 min (~4x faster than serial), because chunking creates a fresh Pool per chunk that releases shared memory on exit.

### Key findings

1. **Chunking is the primary fix for OOM.** Without chunking, all items are accumulated in memory via `list(pool.imap(...))` before `np.array()`, and both Pool implementations OOM at ~2% of the dataset (workers=4).

2. **Fork alone does not prevent OOM** — the OOM is caused by accumulating all items in memory, not by the Pool implementation.

3. **With chunking, fork vs torch.multiprocessing shows no measurable difference** in this benchmark (same ~60 min, similar peak memory). The shm delta is ~0MB for both, because each chunk creates a fresh Pool that releases shm on exit. However, fork avoids torch's `file_system` sharing strategy entirely, so it should be more robust in environments with limited `/dev/shm`, file descriptor limits, or `overcommit_memory=2` — these are common constraints in containerized environments. This benchmark ran with 1TB `/dev/shm` and no restrictive limits, so the difference was not observable.

4. **OOM cause is unclear.** At OOM, the process tree RSS was only ~9 GB (0.5% of 2TB RAM). `/dev/shm` (1TB), `overcommit_memory=0`, `ulimit -n=131,072`, cgroup `memory.max` unlimited, and inodes were all plentiful. The `mmap ... Cannot allocate memory` occurs despite ample resources — likely related to torch.multiprocessing's shared memory file accumulation within a single long-lived Pool, but the exact system limit hit was not identified.

**Note on measurement:** Memory values are from an external monitor process polling the process tree (`psutil` RSS/PSS/USS). The monitor failed to capture fine-grained memory growth in most runs (0MB readings), so peak values may be under-reported. The OOM-vs-completed results were consistent across multiple runs of each config.

## Takeaway

Chunking alone already fixes the OOM in our environment. The fork-based multiprocessing change avoids torch's `file_system` sharing strategy entirely (no shared memory files in `/dev/shm`), which may be more robust in constrained environments, but its effectiveness was not directly verified in this benchmark. Further investigation is needed to pinpoint the exact OOM mechanism; I will follow up if I find more details.

yaoyu-33 (dismissed):

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5869/commits/48820225c4ecd38300b03549acd27e682dc3660f

liding-nv:
@kenkenpa2126 looks like some tests were failed and might be related to the changes. could you fix them so we can re-run the tests

kenkenpa2126:
Thanks for the pointer @liding-nv — pushed a fix in d513353f.

**Root cause:** The chunked `_materialize_dataset_items` writes each chunk to a temp file and concatenates the chunk files at the end. For an empty dataset (`len(dataset) == 0`) zero chunk files are produced, so `np.concatenate([])` raises `ValueError: need at least one array to concatenate`. The three failing tests in `TestPackedSequenceDatasetKwargs` mock the split builder with an empty dataset and hit exactly this path. The pre-chunking implementation returned `np.array([])` for empty input, which is why this regressed.

**Fix:** Return `np.array([])` early when no chunk files exist, preserving the original behavior. Also added a unit regression test (`test_materialize_dataset_items_empty_dataset`).

**Verification:** the 3 failing functional tests pass locally, and `pytest tests/unit_tests/data/packing/` is green (106 passed). Lint (ruff check/format) clean.

Could you re-run CI with `/ok to test d513353f` when you get a chance?

kenkenpa2126:
@liding-nv @yaoyu-33 The failing tests have been fixed in d513353f, and all CI checks are now passing. Since the push dismissed the previous approval, could you re-approve when you get a chance so this can be merged?

huvunvidia:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5869/commits/aa4281f8141c26278490e705df2aa1e83771e9ce

kenkenpa2126:
@huvunvidia Thanks for merging main and triggering CI! The CI run on aa4281f failed with a runner infra error (failed to create job pod: json: cannot unmarshal number into Go struct field Toleration), not a code issue. Could you re-trigger /ok to test aa4281f when you get a chance?

huvunvidia:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5869/commits/e44ad4468c1964f3de9f8583fa543eb35d5dc9b0

huvunvidia:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5869/commits/b104b803f903b383780ee7559d728d51bf26d601

huvunvidia:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5869/commits/4199669379b7f56dd9d7aa65a7a7764942057b1b

sajadn (approved):

PR #5904, opened 2026-08-31
fix(model): handle singleton packed Qwen3-VL batches

leooop-al:
# What does this PR do ?

Fix singleton legacy packed-BSHD detection for Qwen3-VL.

A packed input with shape `[1, T]` is ambiguous: it can be either a singleton
legacy BSHD batch or an already packed THD input.

The legacy Qwen step supplies an attention mask and lets the model construct
MRoPE position IDs, while the existing THD path supplies explicit 3D MRoPE
position IDs. This change uses that distinction to convert singleton legacy
BSHD inputs to THD exactly once without repacking valid THD inputs.

# Changelog

- Handle `B=1` packed BSHD inputs when an attention mask is present and explicit
  MRoPE position IDs are absent.
- Preserve the existing packed THD path with explicit 3D MRoPE position IDs.
- Extend the regression test to cover singleton and multi-sample legacy packed
  BSHD inputs.
- Verify that legacy preprocessing is invoked for both input IDs and generated
  MRoPE position IDs.

# Testing

Passed:

- `pre-commit run --all-files`
- `python3 -m py_compile` for the modified model and test files
- `git diff --check`

The targeted pytest was not run locally because the pinned Megatron-LM
submodule was unavailable. The PR is therefore initially submitted as a draft.

# Duplicate-work check

No open PR was found for the `legacy_packed_bshd` or singleton packed-BSHD
fix at the time this change was prepared.

# Additional Information

AI assistance was used to analyze and prepare this change. The submitter has
reviewed the resulting diff.

### DIFF
--- a/src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
+++ b/src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/model.py
@@ -613,8 +613,14 @@ class Qwen3VLModel(MegatronModule):
 
         cp_rank = self.pg_collection.cp.rank()
         cp_size = self.pg_collection.cp.size()
+        # A singleton BSHD row is shape-identical to THD. The legacy Qwen step
+        # supplies a keep mask and lets this model construct MRoPE, while THD
+        # producers supply explicit 3D MRoPE positions.
         legacy_packed_bshd = (
-            packed_seq_params is not None and input_ids is not None and input_ids.dim() == 2 and input_ids.size(0) > 1
+            packed_seq_params is not None
+            and input_ids is not None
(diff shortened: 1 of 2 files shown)


yaoyu-33 (approved):

yaoyu-33:
CI has been started for the current head. One remaining blocker is DCO: commits `5b47d44f403a5dc88effb2079b17a60d8978acfe` and merge commit `75e21f760f2bbe01afa5716bfdd7a3b246691bbe` do not contain a `Signed-off-by:` trailer. Could you please sign off all commits and force-push the updated branch? We can retrigger CI for the new exact head afterward.

Issue #5807, opened 2026-08-26
[bug] _resolve_hf_tokenizer_revision calls snapshot_download without local_files_only, breaking revision-pinned tokenizers on pre-populated offline caches


janbernloehr:
### Problem

`_resolve_hf_tokenizer_revision()` in `src/megatron/bridge/training/tokenizers/tokenizer.py` resolves a revision-pinned HF tokenizer with:

```python
from huggingface_hub import snapshot_download

snapshot_path = snapshot_download(
    repo_id=tokenizer_model,
    revision=revision,
    allow_patterns=list(_HF_TOKENIZER_SNAPSHOT_ALLOW_PATTERNS),
    ignore_patterns=list(_HF_MODEL_WEIGHT_IGNORE_PATTERNS),
)
```

There is no `local_files_only` argument and no fallback. In an offline / air-gapped environment (`HF_HUB_OFFLINE=1`) whose HF cache was pre-populated by `transformers` (i.e. `AutoTokenizer`/`AutoProcessor.from_pretrained`, which fetch per file via `hf_hub_download`), this call fails and training crashes before step 1, even though every tokenizer file the patterns select is already in the cache.

Observed traceback tail:

```
huggingface_hub.errors.OfflineModeIsEnabled: Cannot reach https://huggingface.co/api/models/Qwen/Qwen3-8B/tree/b968826d9c46dd6066d109eabc6255188de91218?recursive=true&expand=false
```

This is a regression: before the revision-resolution path existed, tokenizer construction went through `transformers` (`hf_hub_download` under the hood), which degrades to the cache when offline, so the same pre-baked cache worked.

### Minimal repro

huggingface_hub 1.28.0.

Step 1 — populate a cache the way `transformers` does, with network access:

```python
# equivalent to AutoTokenizer.from_pretrained("Qwen/Qwen3-8B") — per-file fetches
from huggingface_hub import hf_hub_download
for f in ["config.json", "tokenizer.json", "tokenizer_config.json", "vocab.json", "merges.txt"]:
    hf_hub_download(repo_id="Qwen/Qwen3-8B", filename=f)
```

Step 2 — resolve the same repo at a pinned revision with no network:

```bash
HF_HUB_OFFLINE=1 python -c '
from huggingface_hub import snapshot_download
snapshot_download(
    repo_id="Qwen/Qwen3-8B",
    revision="b968826d9c46dd6066d109eabc6255188de91218",
[...]

PR #5908, opened 2026-08-31
fix(tokenizer): honor HF_HUB_OFFLINE in tokenizer revision resolution

MohammedAlkindi:
# What does this PR do ?

Makes `_resolve_hf_tokenizer_revision` honor `HF_HUB_OFFLINE`, so a revision-pinned HF tokenizer resolves against a cache that `transformers` populated.

# Changelog

- `src/megatron/bridge/training/tokenizers/tokenizer.py`: pass `local_files_only=HF_HUB_OFFLINE` to `snapshot_download`.
- `tests/unit_tests/training/test_tokenizer.py`: add `test_build_hf_tokenizer_revision_resolution_honors_offline_mode`.

### Why

`snapshot_download` needs the repo tree listing to expand `allow_patterns` / `ignore_patterns`, and it only caches that listing when it fetches it itself. A cache populated by per-file `hf_hub_download`, which is what `transformers.AutoTokenizer.from_pretrained` uses underneath, has every requested file on disk but no cached tree listing. Without `local_files_only`, the call then falls through to a network request for that listing and raises `OfflineModeIsEnabled`, even though nothing actually needs downloading.

### Evidence

Measured against `huggingface_hub` 1.29.0 (the version `uv.lock` pins here), same repo id, same allow/ignore patterns and same pinned-SHA revision shape the function uses, cache pre-populated via per-file `hf_hub_download`, `HF_HUB_OFFLINE=1`:

| call | result |
| --- | --- |
| `snapshot_download(...)` without `local_files_only` (current) | `OfflineModeIsEnabled: Cannot reach .../tree/<sha>?recursive=true...`, exit 1 |
| `hf_hub_download(..., revision=...)` | succeeds, exit 0 |
| `snapshot_download(..., local_files_only=True)` | succeeds, exit 0 |

Running the real function before and after the change against that cache: before, it raises `OfflineModeIsEnabled`; after, it returns the snapshot path with all four tokenizer files present.

Regression checks on the fix itself: offline against a genuinely empty cache still fails cleanly with `LocalEntryNotFoundError`, and with `HF_HUB_OFFLINE` unset it still downloads from the network as before. The paths that already worked are unchanged.

### What I could not verify

**I could not run this repo's test suite, so the new test has not been through pytest's collector.** `uv sync` fails at `megatron-core @ editable+3rdparty/Megatron-LM` (the submodule is not checked out), and the pinned build dependencies (`transformer-engine` built from git, `flashinfer-cubin`, `mamba-ssm`, `causal-conv1d`) require CUDA and `nvcc`, which this machine does not have. Please treat CI as the first real run of the test.

What I did run: `ruff check` and `ruff format --check` on both touched files (pass), and the before/after harness above against real `huggingface_hub` rather than a mock of the failure.

I reproduced with `hf-internal-testing/llama-tokenizer` rather than the `Qwen3-8B` from the issue, for speed. The code path and call shape are identical, and I confirmed it with both a pinned-SHA and a branch revision.

# Additional Information

- Related to #5807

### DIFF
--- a/src/megatron/bridge/training/tokenizers/tokenizer.py
+++ b/src/megatron/bridge/training/tokenizers/tokenizer.py
@@ -101,12 +101,22 @@ def _resolve_hf_tokenizer_revision(config: TokenizerConfig) -> TokenizerConfig:
         return config
 
     from huggingface_hub import snapshot_download
+    from huggingface_hub.constants import HF_HUB_OFFLINE
 
+    # `snapshot_download` needs the repo tree listing to expand `allow_patterns` /
+    # `ignore_patterns`, and only caches that listing itself. A cache populated via
+    # per-file downloads (e.g. `transformers.AutoTokenizer.from_pretrained`, which
+    # uses `hf_hub_download`) has the snapshot files but no cached tree listing, so
+    # without `local_files_only` this call falls through to a network request for
+    # the tree and raises `OfflineModeIsEnabled` even though every requested file is
+    # already on disk. Pass `local_files_only` explicitly so offline mode is honored
+    # consistently with `hf_hub_download`.
     snapshot_path = snapshot_download(
         repo_id=tokenizer_model,
         revision=revision,
         allow_patterns=list(_HF_TOKENIZER_SNAPSHOT_ALLOW_PATTERNS),
         ignore_patterns=list(_HF_MODEL_WEIGHT_IGNORE_PATTERNS),
+        local_files_only=HF_HUB_OFFLINE,
     )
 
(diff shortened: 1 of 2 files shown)


yaoyu-33:
Thank you for the detailed investigation and for taking the time to prepare this fix. We are going to close this PR for now, as we are not planning to move forward with this change in its current form. We will keep the underlying offline-tokenizer behavior tracked in #5807 and can revisit it as part of the broader tokenizer/cache handling work. We appreciate the contribution.

PR #5925, opened 2026-09-01
fix(model): preserve CPU initialization when unspecified

### LINKED ISSUE #5919: [bug] provide_distributed_model() resets provider.use_cpu_initialization when omitted
### Problem

`use_cpu_initialization` can be configured directly on a model provider and remains set after `finalize()`. However, calling `provide_distributed_model()` without passing the same option resets the field to `False` before model construction.

The relevant code on current `main` is:

```python
def provide_distributed_model(
    ...,
    use_cpu_initialization: bool | None = False,
    ...
):
    ...
    get_model(
        ...,
        use_cpu_initialization=use_cpu_initialization,
        ...
    )
```

```python
def get_model(
    ...,
    use_cpu_initialization: bool | None = False,
    ...
):
    model_provider.use_cpu_initialization = (
        use_cpu_initialization if use_cpu_initialization else False
    )
```

As a result, an omitted argument, `None`, and an explicit `False` all overwrite the existing provider value with `False`.

Omission should preserve the provider value because the provider is already the finalized model configuration, while the construction argument is an optional override. This is also consistent with:

- the `bool | None` parameter type, where `None` should have a meaning distinct from `False`;
- the neighboring `fp16` and `bf16` construction arguments, where `None` preserves the provider value;
- the MIMO provider, which already falls back to `self.use_cpu_initialization` when the argument is `None`.

The current workaround is to repeat the provider setting at every construction call:

```python
provider.provide_distributed_model(
    wrap_with_ddp=False,
    use_cpu_initialization=True,
)
```

### Minimal repro

```python
from unittest.mock import patch

from megatron.bridge.models.model_provider import ModelProviderMixin


class StopAtConstruction(Exception):
    pass


class Provider(ModelProviderMixin):
    def __init__(self):
        self.use_cpu_initialization = True

    def finalize(self):
        pass

    def provide(self, pre_process=None, post_process=None, vp_stage=None):
        raise AssertionError("model construction should have been intercepted")


provider = Provider()
provider.finalize()

print("after finalize:", provider.use_cpu_initialization)


def observe(model_provider, *args, **kwargs):
    print("at model construction:", model_provider.use_cpu_initialization)
    raise StopAtConstruction


with (
    patch("torch.distributed.is_initialized", return_value=True),
    patch(
        "megatron.bridge.models.model_provider._create_model",
        side_effect=observe,
    ),
):
    try:
        provider.provide_distributed_model(
            wrap_with_ddp=False,
            pg_collection=object(),
        )
    except StopAtConstruction:
        pass
```

### Expected behavior

`use_cpu_initialization` should use tri-state override semantics:

- omitted or `None`: preserve `provider.use_cpu_initialization`;
- `True`: explicitly enable CPU initialization;
- `False`: explicitly disable CPU initialization.

For the minimal reproduction above, the value observed at model construction should remain `True`.

### Affected area

area:model

### Regression?

Not sure

### Environment

Megatron-Bridge: current main
Commit: b0afb83c98ab3afb7e3587375925e3fbbabaface

The minimal reproduction is CPU-only and does not require model weights or a GPU.

ZenAlexa:
## What does this PR do?

Preserves a finalized model provider's `use_cpu_initialization` value when distributed model construction receives no override. This keeps provider configuration intact through model creation and retains explicit `True` and `False` overrides.

## Changelog

- Use `None` as the absence marker in both construction entry points.
- Assign `use_cpu_initialization` for explicit boolean values.
- Exercise omitted, `None`, `False`, and `True` through `provide_distributed_model`.
- Exercise the omitted override through the direct `get_model` entry point before model construction.

## Verification

- `uv run pre-commit run --files tests/unit_tests/models/test_model_instantiation.py`
- `python -m pytest --confcutdir=tests/unit_tests/models tests/unit_tests/models/test_model_instantiation.py::TestGetModel::test_provide_distributed_model_cpu_initialization_override -q` — 4 passed
- The direct `get_model` regression requires the Transformer Engine CUDA runtime; the NGC Python 3.12 validation job covers this path.

## Before your PR is "Ready for review"

- [x] Read and followed the contributor guidelines
- [x] Added the focused regression tests
- [x] Updated the affected API documentation
- [x] Confirmed that the change adds no optional-component dependency

## Additional Information

- Fixes #5919

### DIFF
--- a/src/megatron/bridge/diffusion/models/flux/flux_provider.py
+++ b/src/megatron/bridge/diffusion/models/flux/flux_provider.py
@@ -89,7 +89,7 @@ class FluxProvider(TransformerConfig, ModelProviderMixin[VisionModule]):
     apply_rope_fusion: bool = False
 
     # Initialization and performance settings
-    use_cpu_initialization: bool = True
+    use_cpu_initialization: bool = False
     gradient_accumulation_fusion: bool = False
     enable_cuda_graph: bool = False
     cuda_graph_scope: Optional[str] = None  # full, full_iteration
--- a/src/megatron/bridge/models/model_provider.py
+++ b/src/megatron/bridge/models/model_provider.py
@@ -509,7 +509,7 @@ class GetModelKwargs(TypedDict, total=False):
         use_torch_fsdp2: Use PyTorch FSDP2 instead of custom DDP.
         wrap_with_ddp: Whether to wrap model with DDP.
         data_parallel_random_init: Initialize parameters randomly across data parallel ranks.
-        use_cpu_initialization: Initialize model on CPU.
+        use_cpu_initialization: Override CPU initialization. None preserves the provider setting.
         init_model_with_meta_device: Initialize model on meta device.
         pre_wrap_hook: A single callable or list of callables that overrides all registered pre-wrap hooks.
         post_wrap_hook: A single callable that overrides all registered post-wrap hooks.
--- a/src/megatron/bridge/models/model_provider.py
+++ b/src/megatron/bridge/models/model_provider.py
@@ -194,7 +194,7 @@ class ModelProviderMixin(abc.ABC, Generic[ModelT]):
         use_torch_fsdp2: bool = False,
         wrap_with_ddp: bool = True,
         data_parallel_random_init: bool = False,
-        use_cpu_initialization: None | bool = False,
+        use_cpu_initialization: bool | None = None,
         init_model_with_meta_device: bool | None = None,
         pre_wrap_hook: Union[
             Callable[[list[MegatronModule]], list[MegatronModule]],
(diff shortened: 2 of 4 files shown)


ZenAlexa:
`051d28e9c` adds the direct `get_model` regression. NGC validation for this head requires a renewed `/ok to test 051d28e9c`.

yaoyu-33:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5925/commits/051d28e9ca92484e11c4a0b309b19e8e167a53d5

yaoyu-33 (approved):

ZenAlexa:
The approved head `051d28e9c` has been waiting on `cicd-container-build` since 00:01 UTC after routing and the build matrix completed. Could the ephemeral runner assignment for job `100072413262` be checked? Thanks! :)

ZenAlexa:
Thanks @yaoyu-33 for approving the head and enabling the NVIDIA validation. The container build, full L0 matrix, coverage jobs, lint, and DCO have all completed successfully; `051d28e9c` is ready for merge.

PR #1119, opened 2025-10-28
complex ckpt deprecation

dimapihtar:

PR #5927, opened 2026-09-01
fix(training): validate Gloo for memory-efficient checkpoints

nayopu:
## Problem

`CheckpointConfig.finalize()` rejects every combination of `dist_ckpt_optim_fully_reshardable=True` and
`distrib_optim_fully_reshardable_mem_efficient=True`, even when Gloo process groups are enabled. 
The error message says that the memory-efficient option requires Gloo, but the current assertion does not inspect the Gloo setting.

## Root cause and fix

The check was introduced in [complex checkpoint deprecation (#1119)](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1119).

Megatron Core's own argument validation rejects this combination when `use_gloo_process_groups` is disabled ([argument validation](https://github.com/NVIDIA/Megatron-LM/blob/d7288711ba278d160d2a5a22c099915c9fe1395c/megatron/training/arguments.py#L1380-L1389)).
The optimizer implementation documents and implements the memory-efficient path by selecting the Gloo DP group and returning the optimizer state only on DP rank 0 ([optimizer implementation](https://github.com/NVIDIA/Megatron-LM/blob/d7288711ba278d160d2a5a22c099915c9fe1395c/megatron/core/optimizer/distrib_optimizer.py#L1439-L1461)).
The same behavior is also stated in the Megatron Core configuration documentation ([configuration docstring](https://github.com/NVIDIA/Megatron-LM/blob/d7288711ba278d160d2a5a22c099915c9fe1395c/megatron/training/config/training_config.py#L509-L515)).

Move the cross-configuration check to `ConfigContainer`, where both `CheckpointConfig` and `DistributedInitConfig` are available. This preserves the rejection for the incompatible configuration and allows the supported one.

## Validation

- `uvx pre-commit run --all-files` — passed.
- `python3 -m compileall -q src/megatron/bridge/training/config.py tests/unit_tests/training/test_config.py` — passed.
- Added a focused unit test covering Gloo enabled and disabled configurations.
- The focused pytest could not run on macOS arm64 because the locked `nvidia-resiliency-ext` package has no compatible wheel.

## Scope

This changes validation only. It does not change defaults, enable memory-efficient checkpointing automatically, or modify checkpoint formats.

### DIFF
--- a/src/megatron/bridge/training/config.py
+++ b/src/megatron/bridge/training/config.py
@@ -645,11 +645,6 @@ class CheckpointConfig(MTrainCheckpointConfig):
                     f"Please set checkpoint.load to the base checkpoint directory."
                 )
 
-        if self.dist_ckpt_optim_fully_reshardable:
-            assert not self.distrib_optim_fully_reshardable_mem_efficient, (
-                "distrib_optim_fully_reshardable_mem_efficient requires use_gloo_process_groups"
-            )
-
 
 @dataclass(kw_only=True)
 class LoggerConfig(MTrainLoggerConfig):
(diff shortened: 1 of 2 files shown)


yaoyu-33 (approved):

nayopu:
@yaoyu-33 
Sorry, I've just force-pushed signed commits to pass the DCO test

PR #5289, opened 2026-08-04
feat(data): blend packed Parquet SFT sources

yaoyu-33:
## Summary

- add `packed_train_data_blend=([sources], [weights])` for offline-packed GPT SFT Parquet data
- treat each file, directory, or glob as one logical source while preserving shard support within a source
- build deterministic low-discrepancy weighted mappings with seeded source-local shuffling and optional over/under-sampling through `max_train_samples`
- allow a packed blend to be the standalone training source when validation and test are disabled, or pair it with a raw-data source for validation and test splits
- validate paths, positive finite weights, mutually exclusive single/blended paths, and reject `pad_cu_seqlens` because each source requires its own packing metadata
- document the supported JSONL-to-packed-Parquet-to-blend workflow and remove stale packed-storage guidance

## Motivation

A real NVIDIA Nemotron 3 Nano 30B-A3B SFT comparison on 8 H100s at sequence length 2048 and GBS 128 found bin/idx and Parquet statistically indistinguishable: indexed averaged 104.679 s per cold optimizer step and Parquet averaged 103.810 s. Since storage-format overhead is negligible at this model scale, weighted multi-source Parquet support is more useful than adding another packed storage format.

## Tests

Validated on the current head in a Megatron Bridge Linux container:

- `python -m pytest -q tests/unit_tests/data/packing/test_parquet.py tests/unit_tests/data/builders/test_gpt_sft_config.py tests/unit_tests/data/builders/test_gpt_sft_builder.py` — 101 passed
- `python -m pytest -q tests/unit_tests/training/test_config.py -k "packed or parquet or packing"` — 36 passed, 243 deselected
- `python -m pytest -q tests/unit_tests/doc_consistency/test_readme_consistency.py tests/unit_tests/tutorials/test_text_dataset_tutorials.py` — 26 passed
- end-to-end smoke: independently packed two JSONL sources to Parquet, then built a 3:1 blend through `GPTSFTDatasetBuilder`; 8 requested samples produced deterministic 6/2 source counts and collated successfully
- `pre-commit run --all-files` — passed

PR #5951, opened 2026-09-03
feat(data): blend JSONL SFT sources

yaoyu-33:
## Summary

- add GPTSFTDatasetConfig.per_split_data_source_manifest_path for MLM-style per-split JSONL source lists and alternating weight/path blends
- preserve the existing direct GPTSFTDataset path when a split contains only one source
- deterministically blend raw mmap rows by relative weight, with source-local without-replacement shuffling and repeat-on-oversampling
- feed the resolved raw blend into offline packing once, producing one packed Parquet cache per enabled packed split
- fingerprint builder-managed caches with source paths, weights, file size, seed, preprocessing, and packing semantics
- document configuration, sampling semantics, offline packing, and cache ownership

## Semantics

Weights are positive relative row ratios and do not need to sum to one. The default blended length is the sum of source lengths; max_train_samples overrides the runtime raw-blend length when packing is disabled. Unweighted path lists consume each source row once. A one-path split uses the existing single-source construction unchanged.

This complements #5289 but does not depend on it: this PR blends raw JSONL before creating one offline-packed artifact, whereas #5289 blends multiple already-packed Parquet sources at runtime.

## Test plan

- uv run python -m pytest -q tests/unit_tests/data/datasets/test_gpt_sft.py tests/unit_tests/data/builders/test_gpt_sft_config.py tests/unit_tests/data/builders/test_gpt_sft_builder.py tests/unit_tests/data/packing/test_offline.py tests/unit_tests/doc_consistency/test_readme_consistency.py tests/unit_tests/tutorials/test_text_dataset_tutorials.py (139 passed)
- end-to-end two-JSONL 3:1 blend smoke verifies a 6:2 source ratio in one emitted packed Parquet artifact
- uv run pre-commit run --all-files


yaoyu-33:
Cluster runtime validation at commit `358e2859b3a6e6da0e12afaeba66eedd82bdd111` passed on 1x H100.

- Three-source prompt/completion manifest: Dolly/SQuAD/GSM8K, 1500/900/600 rows with 0.5/0.3/0.2 weights; resolved blend counts matched exactly.
- Qwen3-0.6B, seq=256, MBS=1, GBS=8, 100 optimizer steps, identical seed/checkpoint/scheduler; validation/test/checkpoint saves disabled.
- Unpacked: 100/100 finite-loss steps, first-20 median 2.355924, last-20 median 1.794558 (-23.83%), skipped=0, NaN=0.
- Offline packed: one 1,972-row train Parquet, 94.83% packing efficiency, no sample repetition; 100/100 finite-loss steps, first-20 median 2.138398, last-20 median 1.701103 (-20.45%), skipped=0, NaN=0.
- Raw source hashes were unchanged; resolved configs differed only in offline-packing fields and the config-output filename.

This is a functional/loss-health smoke, not a convergence-parity or throughput claim, because packing changes token composition per physical sample.

PR #5952, opened 2026-09-03
fix(model): honor explicit meta-device disable

ZenAlexa:
# What does this PR do?

Honor the tri-state `init_model_with_meta_device` override through `get_model` and `provide_distributed_model`.

# Changelog

- Apply explicit `True` and `False` overrides to the provider.
- Preserve the provider setting when the override is omitted or `None`.
- Use the effective provider setting for the construction device context.
- Document the contract and test both public entry points.

# Validation

The source-loaded CPU regression passes all override cases at `6585f50ab`, checking real CPU and meta tensor construction and restoration of the outer device context. The project pre-commit checks pass.

# GitHub Actions CI

The project test matrix awaits authorization for `6585f50abb7bde7caa7a8fee1881d2b6ad91ea77`.

# Before review

- [x] Contributor guidelines followed.
- [x] Regression tests added.
- [x] Public API documentation updated.

# Additional Information

Follow-up to #5925.

### DIFF
--- a/src/megatron/bridge/models/model_provider.py
+++ b/src/megatron/bridge/models/model_provider.py
@@ -223,7 +223,7 @@ class ModelProviderMixin(abc.ABC, Generic[ModelT]):
             wrap_with_ddp: Whether to wrap model with DDP.
             data_parallel_random_init: Initialize parameters randomly across data parallel ranks.
             use_cpu_initialization: Override CPU initialization. None preserves the provider setting.
-            init_model_with_meta_device: Initialize model on meta device.
+            init_model_with_meta_device: Override meta-device initialization. None preserves the provider setting.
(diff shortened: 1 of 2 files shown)


ZenAlexa:
@yaoyu-33, could you review this meta-device follow-up to #5925 and authorize `/ok to test 050c5a306537df423701b810048a12a892ed274f`?

I've added the direct `get_model` regression for an explicit `False` override.

ZenAlexa:
I expanded the regression to both public entry points, covering omitted and explicit overrides together with the tensor construction device and context restoration.

@yaoyu-33, could you authorize `/ok to test 6585f50abb7bde7caa7a8fee1881d2b6ad91ea77` when convenient? The project pre-commit checks pass; the expanded matrix needs the project’s CI environment.

PR #5965, opened 2026-09-04
perf(qwen): enable Qwen3 235B NVFP4 full-iteration optimizations

rhmukundan:
## Summary

Enable Qwen3 235B-A22B NVFP4 optimizations on B200, B300, GB200, and GB300:

- Full-iteration CUDA graphs
- HybridEP MoE A2A overlap and delayed weight-gradient computation
- CUTEDSL grouped MLP and Transformer Engine op fusion
- FP4 parameter gather with TP communication overlap disabled
- Megatron-Core update for FP4 with 1F1B A2A overlap

## Testing

- Added targeted recipe unit coverage
- Pre-commit checks passed


malay-nagda (changes requested):
the PR has checked in M-LM in 3rdparty dir. Please remove that.

rhmukundan:
>the PR has checked in M-LM in 3rdparty dir. Please remove that.

@malay-nagda : I have made the change

malay-nagda (dismissed):

rhmukundan:
This MR cannot be merged until https://github.com/NVIDIA/Megatron-LM/pull/6135 is merged as it is dependent on changes from #6135. 
cc: @malay-nagda

src/megatron/bridge/perf_recipes/qwen/b200/qwen3_moe.py
```diff
@@ -555,30 +556,43 @@ def qwen3_235b_a22b_pretrain_64gpu_b200_nvfp4_config() -> ConfigContainer:
...
         "NVTE_BWD_LAYERNORM_SM_MARGIN": 16,
+        "NVTE_CUTEDSL_FUSED_GROUPED_MLP": 1,
         "NVTE_FWD_LAYERNORM_SM_MARGIN": 16,
         # NVFP4 fast-math path.
         "NVTE_USE_FAST_MATH": 1,
     }
     return cfg
 
 
+qwen3_235b_a22b_pretrain_256gpu_b200_nvfp4_full_iteration_config = qwen3_235b_a22b_pretrain_256gpu_b200_nvfp4_config
```
malay-nagda: @rhmukundan why use alias instead of a method?

rhmukundan: This is just a compatibility name for the same recipe. I originally used a wrapper, but CI treats that as a separate recipe and expects its environment config to be duplicated. Using an alias keeps the behavior identical and avoids that duplication.

malay-nagda: - If in future we do not use full-iter CG in `qwen3_235b_a22b_pretrain_256gpu_b200_nvfp4_config` this will be incorrect. 
- We should either remove this or have a separate method like before. 
    - In case of separate methods, `qwen3_235b_a22b_pretrain_256gpu_b200_nvfp4_config` can inherit from the full-iter CG method- both methods should have separate env vars section.

malay-nagda (approved):

src/megatron/bridge/perf_recipes/qwen/b200/qwen3_moe.py
rhmukundan: Addressed it in the latest commit

PR #5989, opened 2026-09-07
fix: use in-process HashStore for temporary_distributed_context rendezvous

### LINKED ISSUE #5967: [bug] temporary_distributed_context has a singleton TCP rendezvous race
### Problem

`temporary_distributed_context()` currently discovers a TCP rendezvous port by binding `localhost:0`, closes that socket, and then asks `torch.distributed.init_process_group()` to bind the same address:

https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8a8360510adcd43ad1154d54bb25928468bc7288/src/megatron/bridge/training/model_load_save.py#L131-L140

That creates a time-of-check/time-of-use race: another process can claim the port before Gloo binds it, causing `EADDRINUSE` / `address already in use`.

Standalone `AutoBridge.import_ckpt()` and `AutoBridge.export_ckpt()` use this temporary context when the caller has no initialized process group:

- https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8a8360510adcd43ad1154d54bb25928468bc7288/src/megatron/bridge/models/conversion/auto_bridge.py#L1494-L1497
- https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/8a8360510adcd43ad1154d54bb25928468bc7288/src/megatron/bridge/models/conversion/auto_bridge.py#L1577-L1583

PR #5696 correctly preserves a caller-owned process group, but standalone callers still enter this path.

### Minimal repro

```shell
# Run on Linux with Megatron-Bridge and the Gloo backend available.
# This is a race, so increase the count/parallelism if needed.
unset MASTER_ADDR MASTER_PORT
seq 1 200 | xargs -n1 -P32 sh -c '
python - <<'"'"'PY'"'"'
from megatron.bridge.training.model_load_save import temporary_distributed_context

with temporary_distributed_context(backend="gloo"):
    pass
PY
'
```

The same class of race produced an `EADDRINUSE` failure in the project's public CI:

https://github.com/NVIDIA-NeMo/Megatron-Bridge/actions/runs/31846107309/job/94940794151

### Expected behavior

A world-size-one temporary process group should initialize and clean up without racing on a host TCP port or depending on inherited rendezvous environment variables.

### Affected area

`area:ckpt`

### Regression?

Not sure. The current helper has used this rendezvous pattern since it was introduced.

### Environment

- Current `main`: `8a8360510adcd43ad1154d54bb25928468bc7288`
- Latest checked release: `v0.6.0` (`51885cf132b2814188b6855c25a8588254274c2a`)
- Backend: Gloo
- World size: 1
- Reproduced through standalone checkpoint conversion/export; the race itself is GPU-independent

### Suggested direction

Because this context is explicitly single-process, an in-process `torch.distributed.HashStore()` can be passed through `store=` to `init_process_group()`, avoiding TCP rendezvous entirely.

Closed PR #5595 proposed that implementation and focused Gloo/NCCL coverage:

https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5595

I could not find a merged replacement for #5595, and current `main` still contains the TCP bind-close-rebind implementation. Was #5595 intentionally superseded, or would a new focused PR for this change be welcome?

ManoharPaturi:
Fixes #5967.

**What's wrong**

`temporary_distributed_context()` finds a free TCP port, closes the socket, then asks `init_process_group()` to bind the same address. Anything that grabs the port in between makes the init fail with `EADDRINUSE`. It's a small window but real, and it showed up in public CI (link in the issue). Also, when `MASTER_ADDR`/`MASTER_PORT` happen to be set, the context silently inherits them, which is not great for a standalone helper.

**Fix**

The context is always `world_size=1, rank=0`, so it doesn't need TCP at all. Pass an in-process `dist.HashStore()` as the store and drop the port scan + env lookup. No host port is touched now, and rendezvous env vars are ignored.

Same approach as yaoyu-33's #5595, which got a positive bot review but was closed before a human looked at it. Happy to fold in any changes if that direction gets picked up instead.

**Tests**

- updated the 3 mocked tests to assert the new contract (`store=` passed, no `init_method`)
- added an end-to-end test that runs the real gloo backend single-process with no rendezvous env, and checks init + cleanup

Verified the 3 contract tests fail on main and pass with this change. Also ran the rest of `tests/unit_tests/training/` before and after, no difference.

I tested locally on CPU/macOS with gloo; NCCL path is covered by the mocked test only since I don't have a GPU here.


yaoyu-33 (dismissed):

jasonliu-shop (approved):
thank you!

cuichenx (approved):

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5989/commits/b28fd51bc564618d807117cd69697841d90af14b

ManoharPaturi:
thanks for merging main in and re-running. all checks are green on this head now, nothing failing left on the PR. anything else needed from my side?

ManoharPaturi:
looked into the 3 failing checks so nobody has to dig: all three come from one root cause and it's not in the PR diff.

build-wheel fails compiling the pinned transformer-engine git dependency:

error: Failed to build transformer-engine @ git+https://github.com/NVIDIA/TransformerEngine.git@27486e03
nccl_ep.cc:677:27: error: 'ncclWinGetUserPtr' was not declared in this scope
nccl_ep.cc:932:17: error: 'ncclGetPeerDevicePointer' was not declared in this scope
nccl_ep.cc:1249:9: error: 'ncclCommProperties_t' was not declared in this scope

that's the TE commit's nccl shim not matching the runner's nccl version, so the wheel never gets built. publish-wheel and release-summary then fail in 2-4s as downstream cascade of the missing wheel. the PR only touches model_load_save.py and its tests, no packaging surface.

not sure if the TE pin needs a bump or the runner image needs the newer nccl headers, but either way it'll reproduce on any PR that reaches this job.

ManoharPaturi:
quick follow up with receipts: the same build-wheel failure hits other PRs from today too (see the runs for #6008, #6079, #6086 on this workflow, all failed the same way within the last day). so this is reproducible across branches and unrelated to this diff, exactly as diagnosed above. whoever owns the TE pin or the runner image will unblock all of them at once.

ManoharPaturi:
one more data point for whoever debugs this: the only green run on this workflow today (#6159) is a docs-only PR whose release job was skipped by the paths filter, it never built a wheel. every run that actually executed build-wheel since sep 18 ~23:58 utc has failed the same way: #6156, #6157, #6067, #6116, #6008, #6079, #6086, #5361 and this one. so the wheel pipeline has been fully down for code PRs since that window, which should narrow down what changed on the runner image or in the environment around then.

cuichenx:
Hi @ManoharPaturi, the released wheel tests are not blocking. The one that matters is "Nemo_CICD_Test" and these are queued

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5989/commits/a3c64757ca5dd8d024778f60bc790e5400bb6214

ManoharPaturi:
good to know, thanks for clarifying. seeing the unit test shards pass and the launch matrix running now. i'll leave it in your hands for the merge, appreciate all the help getting it over the line.

PR #5417, opened 2026-08-07
fix(model): preserve Step3.7 text config fields

yaoyu-33:
## Summary

- delegate Step37TextConfig initialization to Step35Config so Transformers 5.x preserves inherited text fields
- add a regression through the registered AutoConfig and safe config loader path using the published Step-3.7 nested config shape
- verify save/reload behavior, accepted extra config kwargs, and unchanged Step-3.5 construction

## Red-to-green evidence

Before the fix, the focused regression failed while loading Step-3.7 because Step37TextConfig had no inherited hidden_size field. After the fix, the targeted StepFun suite passes all 26 tests.

## Testing

- uv run python -m pytest focused Step-3.7 config, Step-3.5 provider, and Step-3.7 bridge tests
- uv run pre-commit run --all-files

Ref: NVBug 6570994

PR #5996, opened 2026-09-07
fix(model): keep GLM-5 export max_position_embeddings from the reference config

### LINKED ISSUE #5994: [ckpt] GPU export overwrites GLM-5.2 `max_position_embeddings` (1048576 → training seq_length 8192) in the exported `config.json` (related: #5153)
## Summary
`run_conversion.py export` (GPU backend) writes a `config.json` regenerated from the Megatron provider via `MegatronModelBridge.megatron_to_hf_config()` instead of preserving the source HF config. For **zai-org/GLM-5.2** (`glm_moe_dsa`, MLA) this produces a config that loads incorrectly in downstream frameworks:

| key | original (`4d67f66c`) | exported |
|---|---|---|
| `head_dim` | 192 | **64** (← `kv_channels`, i.e. MLA `v_head_dim`, not HF `head_dim`) |
| `max_position_embeddings` | 1048576 | **8192** (← training `seq_length`) |
| `bos_token_id` | absent | 0 |
| `num_experts` | absent | 256 |
| `mlp_bias` | absent | false |
| `layer_types` | absent | `["deepseek_sparse_attention", ...]` |
| `transformers_version` | 5.12.0 | 5.12.1 |

`tokenizer_config.json` only gains `local_files_only`. Weights are correct (282 shards, 59,585 tensors, identical key set to the original incl. `model.layers.78.*` MTP).

**Correction (2026-09-07, after checking `transformers` source):** not all rows above are defects.
- `head_dim`: `GlmMoeDsaConfig.__init__` unconditionally sets `self.head_dim = self.qk_rope_head_dim` (= 64) in transformers 5.12 and on main, so the value written in `config.json` is not authoritative and the exported 64 matches what transformers computes at load time. **Not a bug** (my initial report was wrong on this point).
- `bos_token_id`, `num_experts`, `mlp_bias`, `layer_types`: these are `GlmMoeDsaConfig` defaults materialised by `to_dict()` after the config is re-instantiated; harmless.
- `max_position_embeddings` 1048576 → **8192**: this one is real. The base `CONFIG_MAPPING` maps HF `max_position_embeddings` from Megatron `seq_length`, i.e. the *training* context length overwrites the *model's* context capability, and `conform_config_to_reference` keeps the Megatron-derived value because the key exists in the reference. Downstream engines cap `max_model_len` at this value, so a 1M-context model exports as an 8k-context model.

#5153 tracks the CPU/GPU config divergence generically; this issue adds the concrete GLM-5.2 case so it can be fixed the way #5417 / #5528 (Step3.7) and #5097 (Llama RoPE) were — a `GLM5Bridge.megatron_to_hf_config` override that does not emit `max_position_embeddings` (letting the reference config supply it), or a base-class policy that never lets `seq_length` overwrite a reference `max_position_embeddings`. PR to follow.

## Environment
- `nvcr.io/nvidia/nemo:26.08` (aarch64), Megatron-Bridge v0.6.0 (c932511); `glm5_bridge.py` on `main` (checked 2026-09-06) still has no `megatron_to_hf_config` override
- Export: 12 nodes × 4 GB200, `--tp 1 --pp 6 --ep 8 --etp 1 --torch-dtype bfloat16`, input = torch_dist checkpoint from SFT (100 steps, `mtp_num_layers=1`)

## Reproduction
```bash
python examples/conversion/run_conversion.py export --device gpu \
  --hf-model zai-org/GLM-5.2 --hf-revision 4d67f66cc64d3219133b767c253b2ad1425c6c88 \
  --megatron-path <ckpt_dir_with_latest_checkpointed_iteration.txt> --hf-path <out> \
  --tp 1 --pp 6 --ep 8 --etp 1 --torch-dtype bfloat16
diff <(python -m json.tool <snapshot>/config.json) <(python -m json.tool <out>/config.json)
```
(we run it through a thin wrapper that supplies a DSA-legal pipeline layout and `mtp_num_layers=1`; the config regeneration is independent of that.)

## Impact
The exported model advertises `max_position_embeddings=8192` instead of 1048576, so serving engines refuse or cap contexts beyond 8k. We restored the original `config.json` + `tokenizer_config.json` from the HF snapshot before serving (vLLM 0.27.1 / main, TP8×PP2, generation on the exported weights gives correct answers). A load test with the Bridge-written config as-is is running and will be reported in a comment.

## Expected
Preserve the source HF `config.json` on export when a reference HF model/revision is given (only touching fields that genuinely changed, e.g. dtype), or at minimum do not overwrite `max_position_embeddings` with the training sequence length.

linmuchuiyang:
## What does this PR do?

Fixes #5994.

`GLM5Bridge` inherits the generic `CONFIG_MAPPING` entry `("max_position_embeddings", "seq_length")`. When a fine-tuned GLM-5.2 checkpoint is exported (`run_conversion.py export`, GPU backend), the HF `config.json` therefore gets `max_position_embeddings = 8192` (the training sequence length) instead of the model's `1048576`. `conform_config_to_reference` keeps the Megatron-derived value because the key exists in the reference config, so serving engines cap `max_model_len` at the training length.

This PR adds a `GLM5Bridge.megatron_to_hf_config` override (same pattern as #5528 / #5417 for Step3.7 and #5097 for Llama) that drops `max_position_embeddings` from the generated dict, so the reference HF config supplies it. When no reference config is available the key is simply absent and transformers falls back to the `GlmMoeDsaConfig` default.

Not touched on purpose: `head_dim` (transformers' `GlmMoeDsaConfig.__init__` sets `head_dim = qk_rope_head_dim` regardless of the file value) and the extra default keys — verified harmless by serving an export with Bridge's config unchanged (details in #5994).

## Tests

- New unit tests in `tests/unit_tests/models/glm_moe_dsa/test_glm5_bridge.py`:
  - `test_megatron_config_export_does_not_emit_training_seq_length_as_max_position_embeddings`
  - `test_megatron_config_export_reference_supplies_max_position_embeddings`
- `pytest tests/unit_tests/models/glm_moe_dsa/test_glm5_bridge.py` → 23 passed (run inside `nvcr.io/nvidia/nemo:26.08` with this branch on `PYTHONPATH`, GB200 node); `ruff check` / `ruff format --check` clean on the two files.
- End-to-end context: the 12-node GPU export of a GLM-5.2 SFT checkpoint (`--tp 1 --pp 6 --ep 8`) is what surfaced the issue; with the original `config.json` restored, the export serves correctly on vLLM (TP8×PP2, GB200).

## Checklist
- [x] Signed-off (DCO)
- [x] Unit tests added
- [ ] Docs: none needed (bug fix)

### DIFF
--- a/src/megatron/bridge/models/glm_moe_dsa/glm5_bridge.py
+++ b/src/megatron/bridge/models/glm_moe_dsa/glm5_bridge.py
@@ -145,6 +145,21 @@ class GLM5Bridge(MegatronModelBridge):
 
         return provider
 
+    @classmethod
+    def megatron_to_hf_config(cls, provider: MLAModelProvider) -> dict:
+        """Convert a GLM-5 provider to an HF config dict without leaking training-only context length.
+
+        The generic ``CONFIG_MAPPING`` maps HF ``max_position_embeddings`` from Megatron ``seq_length``. For a
+        fine-tuned checkpoint that is the *training* sequence length (e.g. 8192), not the model's context
+        capability (1048576 for ``zai-org/GLM-5.2``), and ``conform_config_to_reference`` keeps the
+        Megatron-derived value because the key exists in the reference config. Drop it here so the reference
+        HF config supplies it; when no reference config is available the key is simply absent and transformers
+        falls back to the ``GlmMoeDsaConfig`` default.
+        """
+        hf_config = super().megatron_to_hf_config(provider)
+        hf_config.pop("max_position_embeddings", None)
+        return hf_config
+
     def mapping_registry(self) -> MegatronMappingRegistry:
         param_mappings = {
             # Embed
--- a/tests/unit_tests/models/glm_moe_dsa/test_glm5_bridge.py
+++ b/tests/unit_tests/models/glm_moe_dsa/test_glm5_bridge.py
@@ -111,6 +111,25 @@ def test_megatron_config_export_keeps_generic_moe_aliases() -> None:
     assert mapped_config["n_routed_experts"] == 8
 
 
+def test_megatron_config_export_does_not_emit_training_seq_length_as_max_position_embeddings() -> None:
+    """``seq_length`` is the fine-tuning context, not the model's context capability; leave it to the reference."""
+    mapped_config = GLM5Bridge.megatron_to_hf_config(SimpleNamespace(num_moe_experts=8, seq_length=8192))
+
+    assert "max_position_embeddings" not in mapped_config
+    assert mapped_config["num_experts"] == 8
+
+
+def test_megatron_config_export_reference_supplies_max_position_embeddings() -> None:
+    """With a reference config, the exported value must be the reference one (1048576 for GLM-5.2), not seq_length."""
+    from megatron.bridge.models.conversion.utils import conform_config_to_reference
+
+    mapped_config = GLM5Bridge.megatron_to_hf_config(SimpleNamespace(num_moe_experts=256, seq_length=8192))
+    reference = {"max_position_embeddings": 1048576, "n_routed_experts": 256}
+    conformed = conform_config_to_reference(mapped_config, reference)
+
+    assert conformed["max_position_embeddings"] == 1048576
+
+
 @pytest.mark.parametrize(
     ("config_overrides", "expected_topk_freq", "expected_skip_topk_offset"),
     [


cuichenx (approved):

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5996/commits/c8429b311d71af6b60d65a2aea1f4ca4f0a6c357

PR #4974, opened 2026-07-20
[model] fix: preserve Nemotron runtime config

cuichenx:
## Summary

- pass the original provider config to the Nemotron VL and Omni runtime wrappers
- add focused provider regressions

## Root cause

The wrappers defaulted to the deep-copied `LLaVAModel` config, while training setup installed gradient-finalization hooks on the original provider. Per-token loss training therefore missed its final global-token gradient normalization.

## Why CP exposed it

SFT with `CP > 1` requires `calculate_per_token_loss=True`. That path accumulates summed token gradients and relies on `finalize_model_grads` to reduce the global token count across DP/CP/PP and apply `1 / global_tokens`. The detached runtime config hid that finalizer, so PP1/CP2 trained with unnormalized gradients. PP2/CP1 stayed healthy because legacy microbatch normalization runs before backward; explicitly enabling per-token loss on PP1/CP1 reproduced the failure without CP.

## Validation

- 19 focused unit tests passed
- `pre-commit run --all-files` passed
- An earlier 1,200-step H100 PP1/CP2 validation improved from `2.663212` unfixed to `1.481910` with the same fix: https://wandb.ai/nvidia-nemo-fw-public/megatron-bridge-nemotron-omni/runs/45zvay4l

### CP1/CP2 per-token equivalence

Verified at exact head `6fa31ceef00b9eadd4424d5ca11e0ed95f7094a4` with Megatron-Core `adfb8ec7926215bfd0d5120aaf6b080de224d328` and `nvcr.io/nvidia/nemo:26.06` (squashfs SHA-256 `0dbe69a640877451d19a576628295082cc4892f405db290f4d425588f629f41a`).

Both arms used `calculate_per_token_loss=True`, PP1, TP2, EP8, DP4, LoRA, multimodal in-batch packing, sequence length 4096, MBS4/GBS16, deterministic seed 1234, and the same Direct-HF MedPix data. The only semantic topology difference was CP1 versus CP2.

- CP1: Slurm `14194160`, 1x8 H100, completed `0:0`, W&B https://wandb.ai/nvidia-nemo-fw-public/megatron-bridge-nemotron-omni/runs/3wpk3vfe
- CP2: Slurm `14194161`, 2x8 H100, completed `0:0`, W&B https://wandb.ai/nvidia-nemo-fw-public/megatron-bridge-nemotron-omni/runs/ng03m1my
- Complete 1,200-step comparison: `roughly_equal`, loss correlation `0.998791`, zero skipped/NaN steps
[...]

PR #4885, opened 2026-07-15
feat(model): add canonical Nemotron Omni model

aroshanghias-nvd:
# What does this PR do?

Add a dedicated `NemotronOmniModel` and conversion contract for processor-expanded multimodal sequences without using the historical Nemotron Omni `LLaVAModel` collapse/expand path.

> **Fresh review requested:** The branch history has been replaced with a main-based implementation of the reviewed data-boundary packing design. The earlier model-owned-packing implementation is obsolete; please review this series from the beginning.

The ownership boundary is now:

- The data pipeline packs processor-expanded token sequences and constructs `PackedSeqParams`. MBridge's canonical collator provides this behavior for native MBridge training; external callers such as NeMo-RL may provide the same caller-packed THD contract from their own data pipeline.
- `NemotronOmniModel` receives the full packed THD sequence, builds and inserts media embeddings, and only then selects the context-parallel portion of the merged embedding sequence.
- Packing does not run inside model forward.

# Changelog

- Add `NemotronOmniModel`, reusing the existing RADIO encoder, multimodal projector, and Nemotron-H language model components.
- Preserve expanded image placeholders through the canonical collator, with packing performed at the data boundary.
- Keep the legacy LLaVA collator and provider/bridge compatibility path explicit for compatible legacy checkpoints.
- Add the caller-packed THD model contract needed to insert media embeddings before context-parallel slicing.
- Register canonical Nano and Super Omni HF architectures with `AutoBridge`, with versioned contracts that reject ambiguous legacy checkpoints.
- Add canonical import/export mappings for the top-level model namespace and reverse MTP QKV mapping support for Nemotron-H.
- Add conversion, model-forward, media-alignment, packing, Mamba-boundary, activation, and compatibility tests.

# Validation

- Ruff lint and formatting checks passed.
- Focused unit and functional validation passed, including the two-rank caller-packed context-parallel path.
- The corresponding NeMo-RL distributed suite passed CP2 checkpoint, TP/SP, TP2+CP2, padded multi-row attention, and PP2 coverage against this exact MBridge head.
- Nano image-GRPO validation covering generation, training, repeated vLLM refit, checkpointing, and probability agreement is recorded in [W&B](https://wandb.ai/joc/nemotron-omni-main-migration).

# Scope and follow-up

[...]

PR #6067, opened 2026-09-11
fix(training): restore gradient synchronization for copied model configs

aroshanghias-nvd:
# What does this PR do?

Install training callbacks on the configuration consumed by Megatron's scheduler. This fixes a silent gradient-synchronization failure when a provider, such as canonical Nemotron Omni, copies its configuration while constructing the model.

## Why the existing binding is incorrect

`setup()` currently passes the provider configuration (or the GPT/Hybrid builder's transformer configuration) to `_update_model_config_funcs()`. This assumes that the selected object is also the runtime model configuration.

Canonical Omni breaks that identity assumption:

1. The provider copies its configuration into `language_cfg` and passes it to the canonical model.
2. The model constructor stores that copy as `model.config`.
3. After building the model and optimizer, Bridge installs `finalize_model_grads_func`, `grad_scale_func`, and applicable synchronization callbacks on the original provider.
4. Megatron reads `get_model_config(model)`. Its finalization callback is still `None`, so the conditional scheduler call is skipped.

Conceptually, the existing path does this:

```python
runtime_config = deepcopy(provider_config)
model.config = runtime_config
install_callbacks(provider_config)
# provider_config.finalize_model_grads_func is set
# model.config.finalize_model_grads_func is still None
```

This PR obtains the callback target from `get_model_config(model[0])`, matching the scheduler's lookup. The existing installer continues to construct the callbacks and per-chunk callback lists. The reduction algorithms and normalization settings are unchanged.

## Training consequence—and why a run can look healthy

In the reproduced PP=1 configuration with `overlap_grad_reduce=False`, gradient finalization completes the required DP/CP reductions and applicable TP reductions for replicated parameters. The distributed optimizer assumes those operations have completed before it consumes its owned gradient slices. Without them, a shard can update from its owner's local contribution instead of the intended combined gradient.

A small example uses 512 zero-initialized parameters and fresh distributed Adam state. Every parameter receives a local contribution of `+1` on DP rank 0 and `-3` on DP rank 1. Rank 0 owns the first half of the optimizer update; rank 1 owns the second half.

| Case | Existing binding | Corrected binding |
|---|---|---|
| DP=2, TP=1; contributions `+1`, `-3` | Owned gradients remain `+1` and `-3`; 256 parameters decrease and 256 increase | The combined gradient is negative; all 512 parameters increase |
| TP=2, DP=1; replicated parameter contributions `+1`, `-1` | TP replicas update in opposite directions | Contributions cancel; all parameters remain zero |

Parameter all-gather can still give DP replicas identical final weights after the incorrect updates. It assembles the already-updated parameter shards; it does not combine the gradients or recompute the optimizer updates. Every replica therefore receives the same mixture of shard updates based on different ranks' local gradients. Backward also continues to produce nonzero gradients with useful learning signal. Consequently, matching DP weights, stable losses, or successful training completion do not establish that synchronization was correct.

This is a training-correctness finding. It does **not** establish that existing trained models are unusable, quantify model-quality degradation, or demonstrate a full-model TMPE improvement. Affected historical runs require their exact provider, entrypoint, runtime configuration, and overlap settings to be checked. This patch does not repair prior optimizer history or prescribe a checkpoint reset.

# Changelog

- Resolve the runtime configuration from the constructed model when installing training callbacks.
- Add regression coverage through the real `setup()` call and callback installer, with model construction and external services mocked.
- Cover copied Omni configurations and ordinary GPT/Hybrid providers, retain the GPT/Hybrid ModelConfig tests that traverse the actual builder dispatch, and include one representative two-chunk case for overlapped callback installation.
- Exercise the actual PP=1 scheduler with three microbatches: assert exactly one finalizer call with the correct model and process groups during training, and none during forward-only inference.

# Validation

Validation base: Bridge `2eae3c971dd6971beb403e6536987427b7e8a3b2`, with its pinned MCore `c9b53d0a87cb926f47115259593ecfeb351ca29f`.

- **Before/after regression:** running all eight new cases against unpatched `setup.py` produces three expected failures and five passes. Both copied-Omni binding cases fail because `config.finalize_model_grads_func` is `None`; the training scheduler case fails because the finalizer is called zero times. All eight pass with the fix.
- **Targeted unit tests:** **46 passed, 0 failed, 0 skipped.** This includes the eight new callback cases, existing setup and setup-restart-hook tests, and the four state-reset tests.
- **Two-GPU numerical validation: all eight cases passed for the unchanged production fix.** Tested existing/corrected bindings, DP=2 or TP=2, and both two-item/three-item loss contracts with actual Megatron DDP, scheduler, gradient finalization, and precision-aware distributed Adam (PyTorch 2.11.0+cu130).
- **Lint/format:** all repository pre-commit hooks passed; `git diff --check` passed.
- **Static typing limitation:** strict mypy was attempted in the lightweight local check environment, but did not pass because runtime imports were unavailable and the existing source reports typing errors. This is not a clean mypy result.

The eight new unit-test cases check callback binding and scheduler invocation. The GPT/Hybrid ModelConfig cases traverse the real `_build_distributed_model()` dispatch with a lightweight recording builder; other model construction is mocked. The scheduler test also mocks forward/backward computation and the finalizer itself. Numerical gradient correctness is covered by the separate two-GPU experiment described above, not by these unit tests.

The targeted test commands, in an environment with Bridge and its pinned MCore installed, are:

```bash
uv run python -m pytest tests/unit_tests/training/test_runtime_callbacks.py \
  tests/unit_tests/training/test_setup.py \
  tests/unit_tests/training/test_setup_restart_hooks.py -q
uv run python -m pytest tests/unit_tests/training/test_state.py -k reset_for_restart -q
```

To reproduce the original failure, run the new test file against unpatched `setup.py`:

```bash
uv run python -m pytest tests/unit_tests/training/test_runtime_callbacks.py \
  -q
```

The numerical check distinguishes both loss contracts: the two-item, already-normalized loss used by GRPO requires the sum `-2`; the three-item loss with two global tokens produces `-1`. The corrected binding preserves both contracts. Numerical topology coverage is PP=1, CP=1, with DP and TP tested separately. Callback-list tests cover multiple chunks and overlap wiring; they are not a claim of full distributed PP>1, CP>1, or FSDP qualification.

# GitHub Actions CI

The new regression tests live under `tests/unit_tests/training/`. Repository CI remains the merge gate; the standalone numerical experiment is additional validation, not a new CI job in this PR.

- [x] Added necessary regression tests.
- [x] Explained the behavioral change, impact, and validation limits above; no user-facing configuration change requires separate documentation.
- [x] No new dependency or optional-library requirement.

# Additional information

Related history: #4974 (July 23) explicitly preserved the provider configuration in the legacy Omni wrapper. #4885 (July 29 Pacific / July 30 UTC) introduced the canonical path using a copied runtime configuration and also removed the legacy wrapper's explicit `config=self`. This identifies a recurrence in that rewrite relative to its parent; it is not a claim that July was the first-ever occurrence or that every subsequent training run was affected.

### DIFF
--- a/src/megatron/bridge/training/setup.py
+++ b/src/megatron/bridge/training/setup.py
@@ -34,6 +34,7 @@ from megatron.core.process_groups_config import ProcessGroupCollection
 from megatron.core.rerun_state_machine import RerunDataIterator
 from megatron.core.transformer import MegatronModule
 from megatron.core.transformer.multi_token_prediction import get_mtp_ranks
+from megatron.core.utils import get_model_config
 from megatron.training.models.base import ModelConfig
 
 from megatron.bridge.data.loaders import build_train_valid_test_datasets_for_num_epochs, setup_data_iterators
@@ -469,7 +470,8 @@ def setup(
 
     _update_model_config_funcs(
         model,
-        cfg.model.transformer if isinstance(cfg.model, (GPTModelConfig, HybridModelConfig)) else cfg.model,
+        # Providers may copy their configuration while constructing the model.
+        get_model_config(model[0]),
         cfg.ddp,
         optimizer,
         align_grad_reduce=cfg.dist.align_grad_reduce,
--- a/tests/unit_tests/training/test_runtime_callbacks.py
+++ b/tests/unit_tests/training/test_runtime_callbacks.py
@@ -1,0 +1,229 @@
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+
+import contextlib
+from types import SimpleNamespace
+from unittest.mock import MagicMock, Mock, patch
+
+import pytest
+import torch
+from megatron.core.enums import ModelType
+from megatron.core.pipeline_parallel import schedules
+from megatron.core.utils import get_model_config
+
+import megatron.bridge.training.setup as training_setup
+from megatron.bridge.models.gpt.gpt_builder import GPTModelConfig
+from megatron.bridge.models.gpt_provider import GPTModelProvider
+from megatron.bridge.models.hybrid.hybrid_builder import HybridModelConfig
+from megatron.bridge.models.hybrid.hybrid_provider import HybridModelProvider
+from megatron.bridge.models.nemotron_omni.nemotron_omni_provider import NemotronOmniModelProvider
+from megatron.bridge.models.transformer_config import TransformerConfig
+from megatron.bridge.training.state import GlobalState
+
+
+pytestmark = pytest.mark.unit
+
+
+class FakeDDP:
+    def __init__(self, config: TransformerConfig) -> None:
+        self.module = SimpleNamespace(config=config, model_type=ModelType.encoder_or_decoder)
+        self.no_sync = Mock(side_effect=contextlib.nullcontext)
+        self.start_grad_sync = Mock()
+        self.start_param_sync = Mock()
+
+
+def _make_configs(
+    kind: str,
+) -> tuple[GPTModelProvider | HybridModelProvider | GPTModelConfig | HybridModelConfig, TransformerConfig]:
+    dimensions = dict(num_layers=1, hidden_size=8, num_attention_heads=2)
+    if kind == "copied_omni":
+        provider = NemotronOmniModelProvider(**dimensions)
+        return provider, provider._copy_config_without_runtime_process_groups(deep=True)
+    if kind in ("gpt_provider", "hybrid_provider"):
+        provider = (GPTModelProvider if kind == "gpt_provider" else HybridModelProvider)(**dimensions)
+        return provider, provider
+    runtime = TransformerConfig(**dimensions)
+    builder = GPTModelConfig if kind == "gpt_builder" else HybridModelConfig
+    return builder(transformer=runtime, vocab_size=32), runtime
+
+
+def _run_setup(
+    provider: GPTModelProvider | HybridModelProvider | GPTModelConfig | HybridModelConfig,
+    model: list[FakeDDP],
+    ddp_config: SimpleNamespace,
+) -> tuple[MagicMock, SimpleNamespace]:
+    """Exercise the real setup call and installer without constructing a training job."""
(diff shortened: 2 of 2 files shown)


yaoyu-33:
Checked the new ModelConfig path. The production fix already targets the correct runtime config for both paths: GPTModelBuilder and HybridModelBuilder pass their nested transformer config into the constructed model, and the scheduler reads that same object through get_model_config(model[0]). I pushed 65bece81aa91916729d1fb8cecb566112f0dc665 to strengthen the regression coverage: the test now uses BridgeGPTModelConfig and actually traverses _build_distributed_model for both GPT and Hybrid instead of mocking the builder dispatch. @aroshanghias-nvd please review the added coverage.

aroshanghias-nvd:
looks good to me. I'm removing some of the tests added, since they were not that helpful. I will make a commit soon.

aroshanghias-nvd:
/ok to test [33c47e4](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/6067/commits/33c47e44ffeef1fcd1bbd4dc6350e2ab2ee3fde2)

cuichenx (approved):

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/6067/commits/234789c766d0b47bdfc90bdb793bc39a8b655c46

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/6067/commits/fe56c9e1dc3810d115e9c57e95224dd78aeb4b59

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/6067/commits/fc71484e377de9edb963aef82e02f3a1a5dc0b0b

cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/6067/commits/3e6c19acbae4cecaf2bffe9512648db45f8d151d

PR #6068, opened 2026-09-12
feat(peft): seed multi-LoRA slot re-initialization in init_adapter_slot

yushengsu-thu:
# What does this PR do ?

Adds an optional `seed` to the model-level `init_adapter_slot()` so a multi-LoRA adapter slot can be (re)initialised reproducibly across ranks and runs.

# Changelog

- `init_adapter_slot(model, idx, rank, alpha, seed=None)`: with `seed`, snapshot the CUDA RNG-tracker states, rebuild every stream from the seed via `tracker.add` (so byte-tensor and graph-safe generator states both work), call `reset_adapter(idx)` then `init_adapter_slot(idx, rank, alpha)` on every multi-LoRA layer in model order, and restore the saved states in a `finally`. Without `seed` behaviour is unchanged.
- Tests: CPU tests for the unseeded (weights kept) and seeded (every layer re-drawn, B zeroed) paths in `TestMultiLoRAModelHelpers`; a GPU test in `TestMultiLoRALinearGPU` that the same seed reproduces identical weights, a different seed does not, and the tracker streams are restored afterwards.

Motivation: in disaggregated multi-LoRA serving/training a slot is reused for many adapters. Today its weights come from construction or from the previous `clear_adapter_slot()` re-init, which is DP-consistent but not reproducible; RL pipelines that restart an adapter need a deterministic init.

# GitHub Actions CI

**Pre checks**:

- [x] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc) — no

# Additional Information

- Ported from radixark/Megatron-Bridge#34, where it has been in production use since 2026-08-30.
- Follow-up to #4218 / #5155 / #5156.

# Validation

- `tests/unit_tests/peft`: **446 passed, 7 skipped** (baseline 443). New tests PASSED by name: `TestMultiLoRAModelHelpers::test_init_adapter_slot_unseeded_keeps_slot_weights`, `TestMultiLoRAModelHelpers::test_init_adapter_slot_seeded_reinitializes_every_layer`, `TestMultiLoRALinearGPU::test_init_adapter_slot_seeded_is_reproducible` (GPU, NCCL + model-parallel init).
- GPU: `rx devbox` 2× H200 (h200-sci-k8s), image `nvcr.io/nvidia/pytorch:26.08-py3` (the upstream CI base image), TE 2.18.0+27486e03 (matches upstream `uv.lock`), Megatron-Core at `.main.commit` `c9b53d0a8`, Megatron-Bridge editable. Command mirrors CI's `Launch_Unit_Tests_Core.sh`: `CUDA_VISIBLE_DEVICES=0,1 pytest -m "not pleasefixme"`. Baseline `upstream/main@2eae3c971` on the same box: `tests/unit_tests/peft` 443 passed / 7 skipped, the three model-bridge export files 181 passed, `test_qwen3_next_bridge.py` 23 passed. `ruff check`, `ruff check --select I`, `ruff format --check` (ruff 0.9.9) pass on the changed files.

# Revision 2 (review follow-up, 5c77076d6)

- Reseed through `model_parallel_cuda_manual_seed(seed + 100 * pp_rank)` instead of a rank-independent `seed + offset` per stream, so `model-parallel-rng` keeps `+2718+tp_rank`, `expert-parallel-rng` keeps `+1024+100*ep_rank+etp_rank` and `data-parallel-rng` stays equal across replicas; TP/EP/ETP shards draw distinct values while DP replicas match, exactly as the base weights do.
- `reset()` / `add()` moved inside the `try`; `finally` restores the tracker states and the default CUDA generator.
- New two-rank module `tests/unit_tests/peft/test_multi_lora_seeded_slot_distributed.py` (`torch.distributed.run --nproc_per_node=2`): TP=2 shards distinct and reproducible per rank, DP=2 identical, EP=2 expert stream distinct while the data-parallel stream matches. CPU tests stub the topology-aware reseed.
- Validation (2× RTX 4090, `nvcr.io/nvidia/pytorch:26.08-py3`, TE 2.18.0+27486e03, Megatron-Core at `.main.commit`): fail-before with the previous `multi_lora_layers.py` → `2 failed, 1 passed` on the new module (TP=2 shards collapse); pass-after → `3 passed` on both ranks; `tests/unit_tests/peft` single-process 446 passed / 10 skipped with no new failures against `main` (443 / 7); the three seeded tests PASSED by name.

### DIFF
--- a/src/megatron/bridge/peft/multi_lora_layers.py
+++ b/src/megatron/bridge/peft/multi_lora_layers.py
@@ -926,17 +926,43 @@ def install_moe_slot_routing(model) -> int:
     return installed
 
 
-def init_adapter_slot(model, idx: int, rank: int, alpha: float) -> None:
+def init_adapter_slot(model, idx: int, rank: int, alpha: float, seed: int | None = None) -> None:
     """Claim slot ``idx`` across every multi-LoRA layer for an adapter.
 
     A model-wide adapter is the set of slot-``idx`` chunks across all layers;
     this initialises that set with the given ``rank``/``alpha``. Thin iterator
     over the model — per-slot setup (rank/alpha bookkeeping + rank-mask
     invariant) lives on the layer itself in
-    :meth:`MultiLoRALinear.init_adapter_slot` /
+    :meth:`MultiLoRALinear.init_adapter_slot`.
+
+    Without ``seed`` the slot keeps whatever weights construction or the last
+    :func:`clear_adapter_slot` re-init left behind (DP-consistent but not
+    reproducible). With ``seed`` every layer's slot weights are re-initialised
+    deterministically first: the RNG-tracker streams are reseeded from ``seed``
+    for the duration and restored afterwards, so a given seed reproduces the
+    same adapter weights on every rank and every run (layers draw sequentially
+    from the reseeded streams).
     """
-    for module in _iter_multi_lora_modules(model):
-        module.init_adapter_slot(idx, rank, alpha)
+    if seed is None:
+        for module in _iter_multi_lora_modules(model):
+            module.init_adapter_slot(idx, rank, alpha)
+        return
+
+    from megatron.core.tensor_parallel.random import get_cuda_rng_tracker
+
+    tracker = get_cuda_rng_tracker()
+    saved_states = tracker.get_states()
+    # Rebuild each stream from the seed through tracker.add, which mints the
+    # tracker's native state representation (byte tensor or graph-safe generator).
+    tracker.reset()
+    for offset, name in enumerate(sorted(saved_states)):
+        tracker.add(name, seed + offset + 1)
+    try:
+        for module in _iter_multi_lora_modules(model):
+            module.reset_adapter(idx)
+            module.init_adapter_slot(idx, rank, alpha)
+    finally:
+        tracker.set_states(saved_states)
 
 
 def clear_adapter_slot(model, idx: int) -> None:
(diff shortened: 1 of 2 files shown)


yaoyu-33 (dismissed):

src/megatron/bridge/peft/multi_lora_layers.py
```diff
@@ -926,17 +926,43 @@ def install_moe_slot_routing(model) -> int:
...
+
+    from megatron.core.tensor_parallel.random import get_cuda_rng_tracker
+
+    tracker = get_cuda_rng_tracker()
+    saved_states = tracker.get_states()
+    # Rebuild each stream from the seed through tracker.add, which mints the
+    # tracker's native state representation (byte tensor or graph-safe generator).
+    tracker.reset()
+    for offset, name in enumerate(sorted(saved_states)):
+        tracker.add(name, seed + offset + 1)
```
yaoyu-33: Blocking correctness concern: reseeding every named stream from the same rank-independent seed removes the topology offsets that Megatron-Core deliberately applies to the model-parallel and expert-parallel streams. Non-replicated TP shards and EP/ETP expert shards can therefore be initialized identically instead of reconstructing the intended full weight, while only DP replicas should match. Please preserve the TP/EP/ETP-dependent seed semantics and add a TP=2 / EP=2 test that checks DP equality without collapsing distinct shards. Also, tracker.reset() and tracker.add() should be covered by the try/finally so a failed add cannot leave the tracker cleared.

yushengsu-thu: Thanks — agreed on both points, fixed in 5c77076d6 (rebased on your merge of `main`).

yaoyu-33 (approved):
LGTM. The latest changes address the review comments and add coverage for the affected boundary.

huvunvidia:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/6068/commits/a2568560b65833320861f9c04539ec657894034a

PR #6075, opened 2026-09-12
feat(peft): emit fully shared grouped-expert LoRA once with collapse_shared_experts

yushengsu-thu:
# What does this PR do ?

Adds an opt-in `collapse_shared_experts` flag to the adapter export so a grouped-expert LoRA shared by every expert is emitted once as a `[1, ...]` tensor instead of being replicated `num_moe_experts` times.

# Changelog

- `stream_adapter_weights_megatron_to_hf(..., collapse_shared_experts=False)`: in the packed-expert path, when both lora_A and lora_B are 2D (i.e. `share_expert_adapters=True`), emit `linear_in.unsqueeze(0)` and the single-expert `linear_out` (still split per HF projection for fused gate/up) under the packed HF expert name (e.g. GPT-OSS `model.layers.N.mlp.experts.gate_up_proj.lora_{A,B}.weight`). Per-expert (3D) and shared-outer adapters are untouched.
- `AutoBridge.export_adapter_weights` forwards the flag.
- Tests: default replication is pinned (`(E, r, in)` / `(E, out, r)`), the collapsed layout is `(1, ...)`, and per-expert adapters ignore the flag.

This is the fully-shared analogue of the `[1, ...]` layout the shared-outer exporter already produces for SGLang.

# GitHub Actions CI

**Pre checks**:

- [x] Does the PR affect components that are optional to install? — no

# Additional Information

- Re-implements the GPT-OSS export hooks from radixark/Megatron-Bridge commit `f710f8120` as a flag on the current export path.

# Validation

- `test_model_bridge.py` + `test_model_bridge_lora.py` + `test_adapter_export.py`: **184 passed** (baseline 181). New tests PASSED: `test_stream_adapter_weights_megatron_to_hf_replicates_shared_expert_adapter_by_default`, `test_stream_adapter_weights_megatron_to_hf_collapse_shared_experts`, `test_stream_adapter_weights_megatron_to_hf_collapse_ignores_per_expert_adapters`.
- GPU: `rx devbox` 2× H200 (h200-sci-k8s), image `nvcr.io/nvidia/pytorch:26.08-py3` (the upstream CI base image), TE 2.18.0+27486e03 (matches upstream `uv.lock`), Megatron-Core at `.main.commit` `c9b53d0a8`, Megatron-Bridge editable. Command mirrors CI's `Launch_Unit_Tests_Core.sh`: `CUDA_VISIBLE_DEVICES=0,1 pytest -m "not pleasefixme"`. Baseline `upstream/main@2eae3c971` on the same box: `tests/unit_tests/peft` 443 passed / 7 skipped, the three model-bridge export files 181 passed, `test_qwen3_next_bridge.py` 23 passed. `ruff check`, `ruff check --select I`, `ruff format --check` (ruff 0.9.9) pass on the changed files.

### DIFF
--- a/src/megatron/bridge/models/conversion/peft_bridge.py
+++ b/src/megatron/bridge/models/conversion/peft_bridge.py
@@ -855,6 +855,7 @@ class MegatronPeftBridge:
         exclude_adapter_base_prefixes: Iterable[str] | None = None,
         expand_shared_outer: bool = False,
         stack_3d_moe: bool = False,
+        collapse_shared_experts: bool = False,
     ) -> Iterable["HFWeightTuple"]:
         """Stream only adapter weights without merging them into base tensors.
 
@@ -867,6 +868,13 @@ class MegatronPeftBridge:
         for gate_up_proj, bare ``...experts`` for down_proj), instead of the per-expert
         2D ``pack_moe`` layout. It is the vLLM-3D-MoE analogue of ``expand_shared_outer``
         and takes precedence over it for shared-outer adapters.
+
+        ``collapse_shared_experts`` applies to grouped-expert adapters whose lora_A *and*
+        lora_B are both 2D, i.e. one adapter shared by every expert
+        (``share_expert_adapters=True``) exported under a packed HF expert name such as
+        GPT-OSS ``...experts.gate_up_proj``. By default that adapter is replicated into a
+        ``[num_experts, ...]`` stack; with the flag it is emitted once as a ``[1, ...]``
+        tensor (the SGLang shared-LoRA layout also used for shared-outer adapters).
         """
         from megatron.bridge.models.conversion.model_bridge import HFWeightTuple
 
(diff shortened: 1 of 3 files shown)


src/megatron/bridge/models/conversion/peft_bridge.py
```diff
@@ -940,6 +948,33 @@ def stream_adapter_weights_megatron_to_hf(
 
             if packed_expert:
                 linear_in_hf_names, linear_out_hf_names = self._build_lora_hf_names(base_hf_weight_names)
+                if collapse_shared_experts and linear_in_tensor.ndim == 2 and linear_out_tensor.ndim == 2:
```
yaoyu-33, line 951: The claimed SGLang layout does not match current upstream SGLang. A gate/up lora_A with expert_dim=1 enables shared-outer mode, but that mode still allocates gate/up lora_B and down lora_A per expert; emitting both A and B with expert_dim=1 does not fit those buffers. Please validate this with an actual upstream SGLang adapter load and inference parity test, or keep the replicated layout until the consumer has an explicit fully-shared contract.

yushengsu-thu: You're right, thanks for checking. I verified against SGLang — both `main` and the `sglang-miles` branch we actually serve with — and there is no fully-shared MoE LoRA contract:

- `mem_pool.py` sizes the expert dimension to 1 only for `gate_up_proj` lora_A and `down_proj` lora_B under `experts_shared_outer_loras`; `gate_up_proj` lora_B and `down_proj` lora_A stay `[num_experts, ...]`.
- A `[1, ...]` tensor on one of those per-expert sides goes through `_iter_local_expert_weights`, which yields only expert 0 (and nothing on EP ranks > 0) while the remaining slots are zeroed — so the layout this PR emits would not error, it would silently apply the adapter to expert 0 only.
- `_detect_shared_outer_loras` keys the mode on gate_up lora_A having dim0 == 1, so the collapsed A would also flip every adapter in the batch into shared-outer mode.

The `[1, ...]` shape is only meaningful for the two shared-outer sides, which `_stream_shared_outer_adapter_weights` already produces; the replicated `[num_experts, ...]` layout is the correct output for a fully shared adapter with today's consumers. Closing this PR and keeping the replicated layout. If a fully-shared serving contract lands in SGLang, I'll come back with an actual adapter load + logprob parity test as you suggested.

yushengsu-thu:
Closing: the fully-shared `[1, ...]` layout has no consumer in SGLang (`main` or `sglang-miles`), which only accepts expert_dim=1 on the shared-outer sides; the existing replicated `[num_experts, ...]` export is the correct layout today. Details in the review thread.

PR #5588, opened 2026-08-14
feat(peft): expand_shared_outer export flag for shared-outer MoE LoRA

HollowMan6:
# What does this PR do ?

Related to #3958

Shared-outer MoE LoRA exports the shared adapter side as a single `[1, ...]` tensor under an expert-agnostic name (SGLang contract). vLLM's 2D `pack_moe` path for text MoE models expects per-expert 2D tensors, so it can't find the shared factor and crashes at `lora.optimize()` (`lora_b is None`). So this PR adds `expand_shared_outer` to the adapter export path. When set, the shared 2D factor is replicated under per-expert 2D names (reused, not cloned); training keeps the shared parameter. Default `False` preserves the existing layout.

# Changelog

- adds `expand_shared_outer` param with logics

# GitHub Actions CI

See the [CI section](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md#-running-github-ci) in the Contributing doc for how to trigger the CI. A Nvidia developer will need to approve and trigger the CI for external contributors.

# Before your PR is "Ready for review"

**Pre checks**:

- [X] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/CONTRIBUTING.md)
- [X] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?

If you haven't finished some of the above items you can still open "Draft" PR.

# Additional Information

- Related to # (issue)

✨ Presented to you with [Mind Lab](https://macaron.im/mindlab) - A Lab for Experiential Intelligence.

PR #6076, opened 2026-09-12
fix(peft): save shared-outer MoE LoRA adapters in PEFT format

yushengsu-thu:
# What does this PR do ?

Makes `save_hf_adapter` handle shared-outer MoE LoRA (`experts_shared_outer_loras=True`), which today fails with the opaque `ValueError: Incomplete adapter export for target parameter: model.layers.N.mlp.experts.<proj>`. Shared-outer has no PEFT `ParamWrapper` representation, so the layout is exposed as an explicit serving-only export (`allow_serving_layout=True`) and refused with a clear message in the default PEFT mode.

# Changelog

- `convert_adapter_weights_to_peft_state(..., allow_serving_layout=False)`: `_classify_shared_outer_expert_export` recognises a shared-outer export only when exactly one side has expert dim 1, that side is the one shared-outer shares (lora_A on gate/up incl. the 3D-MoE `experts.base_layer` stem, lora_B on down incl. the bare `experts` stem), the per-expert partner exists (numbered 2D `experts.N.<proj>` slices or an `[E, ...]` stack under the same name) and both sides agree on the rank. Unequal expert dims without a shared side (e.g. 8 vs 4), a shared factor on the wrong side, a rank mismatch or a missing partner raise instead of being written.
- Default (PEFT mode): a shared-outer export raises with a message pointing at `expand_shared_outer=True` (replicates the shared factor into the PEFT-loadable per-expert layout) or `allow_serving_layout=True`. With the opt-in both sides are written exactly as exported (shared factor once as `[1, ...]` under the expert-agnostic name), which is the layout SGLang's `experts_shared_outer_loras` reader consumes; it is not loadable by `PeftModel.from_pretrained`, and the docstrings say so.
- `save_hf_adapter(..., allow_serving_layout=False)` forwards the flag; its `expand_shared_outer` docstring no longer claims a "PEFT shared `[1, ...]` layout".
- Tests: PEFT default refuses shared-outer (converter and `save_hf_adapter` end-to-end); serving opt-in keeps the shared factor as a module weight for the numbered-2D and the 3D-pair layouts, accepts the 3D-MoE stems and unknown projections; rejection of unequal expert dims, wrong shared side, rank mismatch and missing partner; incomplete export still raises.

The streamer's output layouts are unchanged; only the PEFT on-disk conversion learns to validate the layout it already produces and to name it for what it is.

# GitHub Actions CI

**Pre checks**:

- [x] Does the PR affect components that are optional to install? — no

# Additional Information

- Based on radixark/Megatron-Bridge#33 (author: @Arist12), where the failure was reproduced with Qwen3-30B-A3B, rank 32, EP4, `--experts-shared-outer-loras` on 8× MI350X and the saved adapter was served by SGLang's shared-outer loader. Re-applied here on top of #5588 / #5939 without changing what `stream_adapter_weights_megatron_to_hf` emits.

# Validation

- Revision 1 (b53f92418): `tests/unit_tests/peft` + `test_model_bridge.py` + `test_model_bridge_lora.py` + `test_adapter_export.py`: **627 passed, 7 skipped** (baseline `upstream/main@2eae3c971` on the same box: 443 + 181 = 624). New tests PASSED by name: `TestSaveHfAdapter::test_convert_shared_outer_adapter_keeps_shared_factor_as_module_weight`, `TestSaveHfAdapter::test_convert_shared_outer_stacked_pair_keeps_both_sides_as_module_weights`, `TestSaveHfAdapter::test_convert_incomplete_target_parameter_still_raises`.
- GPU: `rx devbox` 2× H200 (h200-sci-k8s), image `nvcr.io/nvidia/pytorch:26.08-py3` (the upstream CI base image), TE 2.18.0+27486e03 (matches upstream `uv.lock`), Megatron-Core at `.main.commit` `c9b53d0a8`, Megatron-Bridge editable. Command mirrors CI's `Launch_Unit_Tests_Core.sh`: `CUDA_VISIBLE_DEVICES=0,1 pytest -m "not pleasefixme"`. `ruff check`, `ruff check --select I`, `ruff format --check` (ruff 0.9.9) pass on the changed files.

## Revision 2 (review: PEFT-format promise; classifier too permissive)

- Shared-outer is now an explicit serving-only export (`allow_serving_layout`), refused by default with guidance; the classifier validates exactly-one-shared-side, expected side per projection, partner presence and rank agreement.
- Verified on 2× RTX 4090 (`nvcr.io/nvidia/pytorch:26.08-py3`, TE 2.18.0+27486e03, Megatron-Core `c9b53d0a8`): fail-before with revision 1's sources + the new tests → `9 failed` (7 × `unexpected keyword argument 'allow_serving_layout'`, 2 × `DID NOT RAISE ValueError` for the PEFT-mode refusal in the converter and in `save_hf_adapter`); pass-after → `21 passed` (the 9 new/changed tests plus the 12 existing shared-outer/expand tests selected by name); `test_adapter_export.py` + `test_model_bridge_lora.py` + `test_auto_bridge.py` + `tests/unit_tests/peft` → **691 passed, 7 skipped** (baseline `upstream/main` 681 passed / 7 skipped, no new failures).

### DIFF
--- a/src/megatron/bridge/models/conversion/peft_bridge.py
+++ b/src/megatron/bridge/models/conversion/peft_bridge.py
@@ -15,6 +15,7 @@
 from __future__ import annotations
 
 import itertools
+import re
 from collections import defaultdict
 from dataclasses import dataclass
 from string import digits
@@ -1585,6 +1586,35 @@ class MegatronPeftBridge:
 _HF_LORA_SUFFIXES = (".lora_A.weight", ".lora_B.weight")
 
 
+def _is_shared_outer_expert_export(
+    base_name: str,
+    weights: Dict[str, torch.Tensor],
+    module_weight_names: List[str],
+) -> bool:
+    """Return whether a 3D adapter export is one side of a shared-outer MoE LoRA.
+
+    Shared-outer expert LoRA exports its shared factor once as a ``[1, ...]`` tensor
+    under the expert-agnostic HF name. Its per-expert partner is emitted either as
+    numbered 2D slices (``...experts.N.<proj>``, the default layout) or as an
+    ``[E, ...]`` stack under the same expert-agnostic name (``stack_3d_moe``). In both
+    cases the two sides do not share an expert dim, so they cannot be folded into one
+    PEFT ``target_parameters`` entry the way packed per-expert adapters are.
+    """
+    if ".experts." not in base_name:
+        return False
+    lora_a = weights.get(".lora_A.weight")
+    lora_b = weights.get(".lora_B.weight")
+    if lora_a is not None and lora_b is not None:
+        return lora_a.shape[0] != lora_b.shape[0]
+    present = lora_a if lora_a is not None else lora_b
+    if present is None or present.shape[0] != 1:
+        return False
+    missing_suffix = ".lora_B.weight" if lora_b is None else ".lora_A.weight"
+    prefix, _, projection = base_name.partition(".experts.")
+    sibling = re.compile(rf"^{re.escape(prefix)}\.experts\.\d+\.{re.escape(projection)}{re.escape(missing_suffix)}$")
+    return any(sibling.match(name) for name in module_weight_names)
+
+
 def infer_target_modules_from_adapter_weights(adapter_weight_names: Iterable[str]) -> List[str]:
     """Derive HF ``target_modules`` from the HF-format adapter weight names.
 
@@ -1696,6 +1726,22 @@ def convert_adapter_weights_to_peft_state(
 
         adapter_state[f"base_model.model.{name}"] = tensor
         module_weight_names.append(name)
+
+    # Shared-outer MoE LoRA (one side shared across experts, the other per-expert)
+    # has no common expert dim to pack along. Keep both sides exactly as exported so
+    # the adapter still saves; serving stacks that understand the layout read the
+    # leading dim to tell the shared factor from the per-expert one.
+    shared_outer_targets = [
+        base_name
+        for base_name in target_parameters
+        if _is_shared_outer_expert_export(base_name, parameter_weights[base_name], module_weight_names)
+    ]
+    for base_name in shared_outer_targets:
+        for lora_suffix, tensor in parameter_weights[base_name].items():
+            name = f"{base_name}{lora_suffix}"
+            adapter_state[f"base_model.model.{name}"] = tensor
+            module_weight_names.append(name)
+    target_parameters = [base_name for base_name in target_parameters if base_name not in shared_outer_targets]
 
     target_parameters = _order_target_parameters(target_parameters)
     parameter_prefixes = _build_target_parameter_prefixes(target_parameters)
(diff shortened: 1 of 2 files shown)


src/megatron/bridge/models/conversion/peft_bridge.py
```diff
@@ -1697,6 +1727,22 @@ def convert_adapter_weights_to_peft_state(
...
+    # leading dim to tell the shared factor from the per-expert one.
+    shared_outer_targets = [
+        base_name
+        for base_name in target_parameters
+        if _is_shared_outer_expert_export(base_name, parameter_weights[base_name], module_weight_names)
+    ]
+    for base_name in shared_outer_targets:
+        for lora_suffix, tensor in parameter_weights[base_name].items():
+            name = f"{base_name}{lora_suffix}"
+            adapter_state[f"base_model.model.{name}"] = tensor
```
yaoyu-33, line 1837: save_hf_adapter promises a directory loadable by PeftModel.from_pretrained, but this removes the tensor from target_parameters and writes an expert-agnostic 3D key as a normal module weight. Standard PEFT has no matching module for that key and may ignore it while loading the per-expert partner. The new tests only inspect the rewritten dictionary. Please add a real PeftModel.from_pretrained load plus numerical adapter parity, as the existing packed-expert test does, or make this a clearly serving-specific export path rather than PEFT format.

src/megatron/bridge/models/conversion/peft_bridge.py
```diff
@@ -1585,6 +1586,35 @@ def _merge_canonical_adapter_from_weights(
...
+    ``[E, ...]`` stack under the same expert-agnostic name (``stack_3d_moe``). In both
+    cases the two sides do not share an expert dim, so they cannot be folded into one
+    PEFT ``target_parameters`` entry the way packed per-expert adapters are.
+    """
+    if ".experts." not in base_name:
+        return False
+    lora_a = weights.get(".lora_A.weight")
+    lora_b = weights.get(".lora_B.weight")
+    if lora_a is not None and lora_b is not None:
+        return lora_a.shape[0] != lora_b.shape[0]
```
yaoyu-33: This classifies every unequal pair of leading dimensions as shared-outer, including malformed ordinary exports such as expert dimensions 8 and 4. That turns a shape mismatch into a silently accepted checkpoint. Please require exactly one side to have expert_dim=1 and validate the expected shared side/layout before bypassing target-parameter packing.

src/megatron/bridge/models/conversion/peft_bridge.py
yushengsu-thu: Agreed — fixed in 1e4eaef9b by taking the second option: this layout is a serving-only export, not PEFT format, and the code now says so.

src/megatron/bridge/models/conversion/peft_bridge.py
yushengsu-thu: Fixed in 1e4eaef9b. `_is_shared_outer_expert_export` is replaced by `_classify_shared_outer_expert_export`, which returns the shared side only for a well-formed shared-outer pair and raises for everything else

yaoyu-33 (approved):
LGTM. The latest changes address the review comments and add coverage for the affected boundary.

huvunvidia:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/6076/commits/4d99c1265799e5330d64fdfbab8728e223e42ef4

PR #6089, opened 2026-09-13
fix(peft): preserve shared-expert LoRA scaling with TP overlap

hershg:
Shared-expert overlap defers FC2's tensor-parallel reduction, but LoRA already returns a replicated contribution. The later reduction multiplies it by TP size. Scale that forward contribution by `1 / TP` while preserving the gradient of the equivalent non-overlap adapter.

A two-rank regression compares forward values, input gradients, and adapter gradients with sequence parallelism enabled and disabled.

## Validation

Rebased onto main `809f0e677`; the patch is unchanged at `385027b86` (verified by range-diff). Full repository hooks passed. Historical two-rank B300 tests passed both SP cases on `29a3a8d1a`; distributed testing of this exact rebased head remains pending. NVIDIA CI admission is still required.


hershg:
@yaoyu-33 @HollowMan6 could you review the shared-expert FC2 LoRA fix? TP overlap sums it twice. @yaoyu-33 reviewed the related #6009.

At `036192158`, the two-rank regression passed on B300 and H200 with sequence parallelism on/off, using our pinned Megatron runtime without the FP32 patches. Removing the forward compensation doubled the output and failed both cases.

Head `04ce4ec3be8d622b22f9da8dcc71aa0933866e4c` adds both cases to the existing two-rank H100 L0 launcher. Shell/selection checks, six CPU CI tests and hooks pass; the new CI wiring still needs a GPU run. Could a maintainer admit it with `/ok to test 04ce4ec3be8d622b22f9da8dcc71aa0933866e4c`?

PR #6092, opened 2026-09-14
feat(training): add DPO training with offline-scored reference model

Josephasafg:
# What does this PR do ?

Adds DPO (Direct Preference Optimization) as a training mode. The reference model's log-probabilities are scored once, offline, and read back during training, so the trainer holds one model and runs at SFT memory and step time.

# Changelog

- `--mode dpo` in `scripts/training/run_recipe.py`, backed by `dpo_train()` in `training/dpo_train.py`: validates the run config and the reference artifact, then hands off to the existing `finetune` loop.
- Preference-pair dataset, config and loader under `data/datasets/preference*.py` and `data/builders/dpo.py`. Rows are TRL-style chat pairs from an HF dataset or a JSONL file. Chosen and rejected rows of a pair always land in the same micro batch, next to each other.
- DPO loss and forward step in `training/dpo.py` and `training/dpo_step.py`: sigmoid preference loss, optional SFT term, metrics for preference loss, accuracy, margin and chosen/rejected rewards.
- `scripts/dpo/score_reference_logprobs.py`, the offline scorer. You run it with the same TP, EP and ETP you'll train with. Extra GPUs beyond that become data-parallel replicas, and rank 0 collects the results and writes the artifact.
- Recipe `qwen3_30b_a3b_dpo_config` (TP=4, PP=2, EP=4, ETP=1, SP).
- Docs: `scripts/dpo/README.md` and a DPO section in `scripts/training/README.md`.

# Why the reference is scored offline

The usual DPO trainer keeps a frozen copy of the reference model and runs it every step - a second set of resident weights, an extra forward per step, and the reference is stuck with the policy's parallel layout

Instead, the scorer runs the reference over the dataset once and writes `ref_logprobs.jsonl` (chosen and rejected completion logprob sums per row) and `scoring_metadata.json` (dataset, split, tokenizer, sequence length, Parallelism type, prompt key). At startup the trainer compares that metadata against its own config and refuses to run on a mismatch. Without the check a stale or mis-scored artifact does not crash anything, the margins could possibly be wrong.

Pipeline parallelism is not recorded: it only decides which rank runs a layer. Scoring at PP=2 and PP=1 gave the same results over sums on all 8,320 pairs; TP=4 vs TP=8 differed on every pair, so TP is recorded.

The artifact stores plain logprob sums with no beta applied, so one scoring run covers every learning rate and beta you try afterwards.

```mermaid
flowchart LR
    DS[("preference pairs<br/>HF dataset or JSONL")]

    subgraph once["scorer, run once"]
        REF["reference model"]
        ART[/"ref_logprobs.jsonl<br/>chosen / rejected logprob sums per row"/]
        META[/"scoring_metadata.json<br/>dataset, split, tokenizer,<br/>seq len, TP / SP"/]
        REF --> ART
        REF --> META
    end

    subgraph train["trainer, every run"]
        CHK{"metadata matches<br/>run config?"}
        LOADER["pair loader<br/>chosen + rejected adjacent"]
        POL["policy model"]
        LOSS["DPO loss<br/>preference loss, accuracy,<br/>margin, rewards"]
        STEP["backward + optimizer step<br/>finetune loop"]
        CHK -->|"yes"| LOADER
        CHK -->|"no"| STOP["error at startup"]
        LOADER --> POL --> LOSS --> STEP
        LOADER -->|"reference sums"| LOSS
    end

    DS --> REF
    DS -->|"same rows, same order"| LOADER
    ART -->|"joined by row index"| LOADER
    META --> CHK
```

# Usage

```bash
# score the reference once
uv run python -m torch.distributed.run --nproc_per_node=8 scripts/dpo/score_reference_logprobs.py \
    --model Qwen/Qwen3-30B-A3B --dataset HuggingFaceH4/ultrafeedback_binarized --split train_prefs \
    --num-pairs 8192 --max-seq-length 4096 --tp 4 --ep 4 --etp 1 --sequence-parallel \
    --output /data/uf_ref_logprobs

# train
uv run python -m torch.distributed.run --nproc_per_node=8 scripts/training/run_recipe.py \
    --model qwen3_30b_a3b --mode dpo --pretrained_checkpoint /checkpoints/qwen3-30b-a3b \
    dataset.source.path_or_dataset=HuggingFaceH4/ultrafeedback_binarized dataset.source.split=train_prefs \
    dataset.ref_artifact=/data/uf_ref_logprobs dataset.num_pairs=8192
```

# Results

Qwen/Qwen3-30B-A3B with `qwen3_30b_a3b_dpo_config` at its defaults, 2 nodes x 8 H100 80GB.

| | |
|---|---|
| Data | `HuggingFaceH4/ultrafeedback_binarized` `train_prefs`, 8,192 pairs, one epoch. Validation on 128 pairs of `test_prefs` every 32 steps |
| Parallelism | TP=4, PP=2, EP=4, ETP=1, SP, DP=2, distributed optimizer |
| Batch | 64 pairs per step: micro batch 1 pair (2 rows) × 32 accumulation × DP 2; seq length 4096 |
| DPO | beta 0.05, lr 5e-7 |
| Speed | 20 s / step mean, 19 s median after warmup (16-34 s range) |

Step 1 preference loss is 0.691, i.e. ln 2, as it should be when the policy starts from the reference.

W&B curves from the run above, validation split. The model learns to prefer chosen over rejected

image
image
image

Training preference loss per 32-step window: 0.681, 0.652, 0.636, 0.634. 

Accuracy plateaus at 0.68-0.70, in line with published DPO results on UltraFeedback.

# Tests

Unit tests, no GPU: `tests/unit_tests/data/builders/test_dpo.py`, `data/datasets/test_preference.py`, `data/datasets/test_preference_tokenization.py`, `training/test_dpo.py`, `training/test_dpo_step.py`, `training/test_dpo_train.py`, `scripts/dpo/test_score_reference_logprobs.py`, plus cases in `training/test_config.py`, `training/utils/test_pg_utils.py`, `recipes/qwen/test_qwen3_recipes.py` and the `scripts/training` runner tests.

### DIFF
--- a/scripts/dpo/score_reference_logprobs.py
+++ b/scripts/dpo/score_reference_logprobs.py
@@ -145,7 +145,7 @@ def parse_args() -> argparse.Namespace:
     return args
 
 
-def dpo_source(args: argparse.Namespace) -> Prefere
+def dpo_source(args: argparse.Namespace) -> PreferenceSource:
     """The preference source these CLI args select; a path-shaped --dataset is routed to JSONL by the config."""
     if args.jsonl:
         return PreferenceJSONLSource(paths=list(args.jsonl), index_mapping_dir=args.index_mapping_dir)
--- a/scripts/dpo/score_reference_logprobs.py
+++ b/scripts/dpo/score_reference_logprobs.py
@@ -1,0 +1,344 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Score reference-model logprobs for every DPO preference pair (offline, one-time).
+
+Runs the reference model forward over the trainer's own pair dataset and writes a
+``pair_id``-keyed artifact the trainer loads instead of the reference weights. Data, tokenizer, ``--max-seq-length`` and
+the parallel layout (``--tp``, ``--sequence-parallel``) must equal the training run's;
+``scoring_metadata.json`` records them and the trainer checks it.
+
+Example:
+    python scripts/dpo/score_reference_logprobs.py \\
+        --model Qwen/Qwen2.5-0.5B-Instruct \\
+        --output /tmp/capybara_ref_logprobs --num-pairs 1024
+
+With tensor parallelism (launch exactly ``--tp`` processes; rank 0 writes the artifact):
+    python -m torch.distributed.run --nproc_per_node=2 \\
+        scripts/dpo/score_reference_logprobs.py --tp 2 \\
+        --model Qwen/Qwen2.5-7B-Instruct --output /tmp/uf_ref_logprobs
+"""
+
+import argparse
+import os
+from dataclasses import asdict
+
+import torch
+from megatron.core.msc_utils import MultiStorageClientFeature
+from megatron.core.pipeline_parallel import get_forward_backward_func
+from torch.utils.data import BatchSampler, DataLoader, SequentialSampler
+
+from megatron.bridge import AutoBridge
+from megatron.bridge.data.builders.dpo import DPODatasetConfig, PreferenceSource, build_preference_split
+from megatron.bridge.data.datasets.preference import (
+    ScoringFingerprint,
+    pack_pairs_by_token_budget,
+    pair_token_lengths,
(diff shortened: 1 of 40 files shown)


scripts/dpo/score_reference_logprobs.py
```diff
@@ -0,0 +1,382 @@
...
+    get_rank_safe,
+    get_world_size_safe,
+    maybe_initialize_distributed,
+    print_rank_0,
+)
+
+
+# The trainer's default recipe; applied here too so the reference forward runs the policy's numerics
+# whatever precision the checkpoint declares.
+MIXED_PRECISION_RECIPE = "bf16_mixed"
```
Josephasafg: from arg

scripts/training/recipe_runner.py
```diff
@@ -52,6 +52,7 @@
 
 STEP_FUNCTIONS: dict[str, StepFunctionEntry] = {
     "audio_lm_step": ("megatron.bridge.training.audio_lm_step", "forward_step"),
+    "dpo_step": ("megatron.bridge.training.dpo_step", "dpo_forward_step"),
```
Josephasafg: forward_step

PR #6097, opened 2026-09-14
fix(perf): restore PP=8 for DeepSeek V3 B200 bf16 and fp8_cs recipes

rsalagame-nvidia:
## What

Sets `pipeline_model_parallel_size = 8` (from 16) for `deepseek_v3_pretrain_256gpu_b200_bf16_config` and `deepseek_v3_pretrain_256gpu_b200_fp8cs_config`.

## Why

`deepseek_v3_256gpu_b200_bf16` hangs reproducibly with an NCCL collective timeout on the pipeline-parallel group:

```
NCCL collective timeout: WorkNCCL(SeqNum=16, OpType=COALESCED, NumelIn=0, NumelOut=0)
  ... PIPELINE_MODEL_PARALLEL_GROUP
```

It has now hung on three separate runs across two different containers. The one run that succeeded (full 50/50 iterations) used PP=8.

### Root cause

[#5471](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/5471) set `pipeline_model_parallel_size=8` on the shared base config, which the bf16 and fp8_cs variants then inherited:

```python
DEEPSEEK_V3_PRETRAIN_CONFIG_B200_V2 = replace(
    DEEPSEEK_V3_PRETRAIN_CONFIG_B200_V1,
    pipeline_model_parallel_size=8,
    global_batch_size=4096,
)
DEEPSEEK_V3_PRETRAIN_CONFIG_B200_BF16_V2 = DEEPSEEK_V3_PRETRAIN_CONFIG_B200_V2
DEEPSEEK_V3_PRETRAIN_CONFIG_B200_FP8_CS_V2 = DEEPSEEK_V3_PRETRAIN_CONFIG_B200_V2
```

That PR targeted `scripts/performance/configs/deepseek/deepseek_workload_base_configs.py`, which had already been deleted by #4623 (`refactor(performance): remove legacy perf configs`). The fix therefore had to be re-expressed by hand in the flat `perf_recipes` layout. The EP=8 half of it carried over, but because PP=8 lived on the shared base rather than in the bf16/fp8_cs bodies, it was not transcribed and both configs stayed at PP=16.

### Supporting evidence

Every other DeepSeek V3 256-GPU recipe already uses PP=8 — these two are the only outliers:

| gpu | config | PP | VPP |
|---|---|---|---|
| b200 | **bf16** | **16** | **None** |
| b200 | **fp8_cs** | **16** | **None** |
| b200 | fp8_mx | 8 | 2 |
| b200 | nvfp4 | 8 | 2 |
| b300 | all four | 8 | 2 |

The b300 recipes are uniform because this same failure was already diagnosed and fixed there: `73eb4b272` "b300 dsv3 bf16 hang fix (#2260)". B200 bf16 never received the equivalent treatment.

## Validation

PP=8 for bf16 is validated by a full 50/50-iteration run on 256x B200 (nemo-ci pipeline 67082563), whose measured throughput matches the existing nemo-ci golden-value baseline for this test case (~24.3 s/iter). A rerun of the release config with this change is in flight: nemo-ci pipeline 67863288.

`fp8_cs` has no nemo-ci perf job and has not been exercised in the release sweeps, so its hang is latent rather than observed; it is fixed here because #5471 applied PP=8 to it through the same shared base.

## Scope

`deepseek_v3_pretrain_256gpu_b200_fp8mx_large_scale_config` also uses PP=16 but is a distinct large-scale recipe outside the scope of #5471, and is intentionally left unchanged.

## Related

Companion PR against `r0.6.0`: #6096. The two branches carry the identical defect and the identical fix.


rsalagame-nvidia:
Additional evidence of intent from #5471. That PR was a single-file, `+1/-5` change, and what it did was **hoist** `pipeline_model_parallel_size=8` out of the FP8-MX-specific `replace()` and up into the shared `_V2` base:

```diff
 DEEPSEEK_V3_PRETRAIN_CONFIG_B200_V2 = replace(
     DEEPSEEK_V3_PRETRAIN_CONFIG_B200_V1,
+    pipeline_model_parallel_size=8,
     global_batch_size=4096,
 )
 DEEPSEEK_V3_PRETRAIN_CONFIG_B200_BF16_V2 = DEEPSEEK_V3_PRETRAIN_CONFIG_B200_V2
 DEEPSEEK_V3_PRETRAIN_CONFIG_B200_FP8_CS_V2 = DEEPSEEK_V3_PRETRAIN_CONFIG_B200_V2
 DEEPSEEK_V3_PRETRAIN_CONFIG_B200_FP8_MX_V2 = replace(
     DEEPSEEK_V3_PRETRAIN_CONFIG_B200_FP8_CS_V2,
-    moe_flex_dispatcher_backend="deepep",
-    pipeline_model_parallel_size=8,
     virtual_pipeline_model_parallel_size=2,
```

PP=8 already applied to fp8_mx before that change. Moving it to the base has exactly one functional effect: extending PP=8 to `BF16_V2`, `FP8_CS_V2`, and `NVFP4_V2`. Had the intent been to keep PP=8 scoped to fp8_mx, the line would simply have been left where it was.

This also explains the specific shape of the regression. `BF16_V2` and `FP8_CS_V2` are bare aliases (`= _V2`) with no `replace()` bodies of their own, so they had no explicit `pipeline_model_parallel_size` line anywhere to copy. When the recipes were re-expressed as standalone functions in `perf_recipes`, the variants that carried explicit lines (fp8_mx, nvfp4) kept PP=8, while the two that inherited it silently from the base did not.

PR #6129, opened 2026-09-17
perf(recipe): capture attn in the GB300 Qwen3.5-VL 35B/397B partial CUDA graphs

tomlifu:
The GB300 35B-A3B and 397B-A17B benchmarks declared `cuda_graph_scope = ["moe_router", "moe_preprocess"]`, but that assignment was dead: `_qwen35_vl_post` runs afterwards and sets `cuda_graph_impl="none"` plus `clear_cuda_graph_modules`, so these configs shipped with CUDA graphs fully disabled. Adding "attn" to the existing list would therefore have changed nothing.

Add `_enable_partial_cuda_graphs`, called AFTER `_qwen35_vl_post`, which enables `transformer_engine` per-layer graphs over [attn, moe_router, moe_preprocess]. It uses `set_cuda_graph_modules`, which writes `cuda_graph_modules` and nulls `cuda_graph_scope`, so the two are never set together (MCore asserts).

Why it is safe here specifically: `_qwen35_vl_post` disables graphs for the variable-length real-data path, but these are benchmark configs built on the mock providers, where `_qwen35_vl_common` fixes `seq_length = 4096` and sets `moe_router_force_load_balancing = True`. The vision tower stays graph-free.

Scope: GB300 35B-A3B (8 GPU) and 397B-A17B (64 GPU) only. The fp8cs/fp8mx variants derive from their bf16 base, so all six configs are covered. 122B-A10B is deliberately left disabled and a test pins that.

Tests: 6 parametrized cases assert impl/modules/scope on the final config -- the only place the behaviour is observable, since the recipe-body assignment is dead -- plus a guard that 122B stays off.

# GitHub Actions CI

**Pre checks**:


malay-nagda:
@tomlifu can you look at claude comments and address them? Also, did you run the updated CG space (added `attn` module) to make sure we don't see OOM?

tomlifu:
> @tomlifu can you look at claude comments and address them? Also, did you run the updated CG space (added `attn` module) to make sure we don't see OOM?

I've addressed the claude comments and checked there is no OOM for the attn module in CG for GB300.

tomlifu:
/ok to test [1f2782b](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/6129/commits/1f2782b8743cfb6cefc89c7e1a9cc699dc9a6d87)

malay-nagda (approved):

PR #6142, opened 2026-09-17
fix(ckpt): thread revision through SafeTensorsStateSource

### LINKED ISSUE #6104: [bug] SafeTensorsStateSource._resolve_path ignores the requested revision; offline caches pinned to a commit sha fail with LocalEntryNotFoundError
**Repo:** NVIDIA-NeMo/Megatron-Bridge  **Component:** `src/megatron/bridge/models/hf_pretrained/state.py`
**Related:** #5807 (same class of bug on the tokenizer path: `_resolve_hf_tokenizer_revision` without `local_files_only`).

### Problem

`PreTrainedBase` accepts and threads a `revision` (`base.py`: `revision_kwargs = {"revision": revision}` for `list_repo_files` / `hf_hub_download`), and `scripts/conversion/run_conversion.py` has `--hf-revision`. But the weights are resolved by

```python
@staticmethod
def _resolve_path(model_name_or_path):
    ...
    return Path(snapshot_download(
        repo_id=str(model_name_or_path),
        allow_patterns=["*.safetensors", "model.safetensors.index.json"],
        ignore_patterns=["*.bin", "*.pt", "*.pth"],
    ))
```

with **no `revision`**, so huggingface_hub resolves `main`. In offline mode (`HF_HUB_OFFLINE=1`) `main` is looked up through the cache file `refs/main`. A cache populated with `snapshot_download(repo_id, revision=<sha>)` never writes `refs/main`, so the lookup fails even though every shard is present under `snapshots/<sha>/`.

### Minimal repro / logs

```bash
# online, pin a commit
python -c "from huggingface_hub import snapshot_download; snapshot_download('deepseek-ai/DeepSeek-V4-Pro', revision='b5968e9190ef611bbf34a7229255be88a0e937c1')"
# offline conversion
HF_HUB_OFFLINE=1 python scripts/conversion/run_conversion.py import --hf-model deepseek-ai/DeepSeek-V4-Pro \
  --hf-revision b5968e9190ef611bbf34a7229255be88a0e937c1 --megatron-path ... --tp 1 --pp 16 --ep 8
```

```text
huggingface_hub.errors.OfflineModeIsEnabled: Cannot reach https://huggingface.co/api/models/deepseek-ai/DeepSeek-V4-Pro/revision/main: offline mode is enabled.
  File ".../megatron/bridge/models/hf_pretrained/state.py", line 543, in _resolve_path
    snapshot_download(
huggingface_hub.errors.LocalEntryNotFoundError: Cannot find an appropriate cached snapshot folder for the specified revision on the local disk and outgoing traffic has been disabled.
```

Note the URL ends in `/revision/main` although the job asked for a sha.

### Workaround we use

Write `refs/main` = the pinned sha into the cache (byte-identical to what a `main` download writes: 40-char sha, no trailing newline). Note that huggingface_hub 1.28 reads `refs/main` without `strip()`, so a newline breaks it.

### Expected behavior / suggested fix

Give `SafeTensorsStateSource` a `revision` (constructor argument, populated from `PreTrainedBase.init_kwargs["revision"]` where the source is created in `base.py`, currently `SafeTensorsStateSource(self.model_name_or_path)`), and pass it to `snapshot_download(..., revision=self.revision)`. Also pass `local_files_only=True` when `HF_HUB_OFFLINE` is set so the error is immediate and explicit. The same change fixes the tokenizer path reported in #5807.

Environment: nemo:26.08, Megatron-Bridge v0.6.0 (`c932511`); `main` still has the call without `revision`.

### Affected area
conversion (`hf_pretrained/state.py`, `run_conversion.py --hf-revision`)

### Regression?
Not a regression: `main` still calls `snapshot_download` without `revision`.

Lingikaushikreddy:
# What does this PR do ?

Threads the Hugging Face `revision` through `SafeTensorsStateSource` so weight resolution honours the pinned revision instead of always resolving `main`. Fixes #6104.

# Changelog

- `SafeTensorsStateSource.__init__` accepts an optional `revision` and stores it.
- `SafeTensorsStateSource.path` passes that revision to `_resolve_path`.
- `SafeTensorsStateSource._resolve_path` takes `revision` and forwards it to `snapshot_download`.
- `PreTrainedBase.state` passes the revision it was constructed with (same `init_kwargs` idiom already used by `save_artifacts`).
- Tests: three cases in `tests/unit_tests/models/test_hf_pretrained_state.py` (revision forwarded, defaults to `None`, local directory short-circuits without touching the Hub) and two in `tests/unit_tests/models/hf_pretrained/test_base.py` (pinned and unpinned models reach the state source correctly).

# Context

`PreTrainedBase` already accepts a `revision` and threads it through `list_repo_files` / `hf_hub_download` (`base.py` L136-151, L234-283), and `scripts/conversion/run_conversion.py` exposes `--hf-revision`. Weight resolution was the gap: `_resolve_path` called `snapshot_download` with no `revision`, so huggingface_hub resolved `main`.

Two consequences:

1. A revision-pinned conversion could silently load different weights than requested if the repo's `main` moved.
2. Offline use breaks. `main` is resolved through the cache file `refs/main`, which a cache populated by `snapshot_download(repo_id, revision=<sha>)` never writes — so `HF_HUB_OFFLINE=1` fails with `LocalEntryNotFoundError` even though every shard is present under `snapshots/<sha>/`.

The new parameter is optional and defaults to `None`, so existing positional callers (including `FluxSafeTensorsStateSource` / `WanSafeTensorsStateSource`, which pass local paths) are unaffected, and the resolved revision is unchanged for unpinned models.

Note that #5807 reports the same class of bug on the tokenizer path (`_resolve_hf_tokenizer_revision` without `local_files_only`). That is a separate call site and is not addressed here.

**Pre checks**:

- [x] Did you add or update any necessary documentation? (docstrings for the new parameter)
- [x] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc) — no; `huggingface_hub` is already imported lazily inside `_resolve_path` and that guard is unchanged.

# Testing

The new tests fail on `main` and pass with this change. I verified the `state.py` cases locally against the patched and unpatched module; the full suite needs CUDA/Triton, so CI approval from an NVIDIA developer is required to run it end to end.

# Additional Information

- Related to #6104
- Same bug class as #5807 (tokenizer path, not covered here)


Lingikaushikreddy:
Closing as superseded by #6156, which landed the same fix for #6104 (forwarding `revision` through `SafeTensorsStateSource` to `snapshot_download`) and also adds `local_files_only` and broader regression tests. Thanks @cuichenx for picking it up.

PR #6156, opened 2026-09-18
fix(ckpt): honor pinned revisions when loading offline HF weights

### LINKED ISSUE #6104: [bug] SafeTensorsStateSource._resolve_path ignores the requested revision; offline caches pinned to a commit sha fail with LocalEntryNotFoundError

cuichenx:
# What does this PR do ?

Fixes #6104 by passing the requested Hugging Face revision through the lazy weight source and explicitly selecting cache-only resolution in offline mode. A commit-pinned cache without `refs/main` now loads the requested weights; a different cached main cannot replace them. Offline caches populated file by file also work without repository tree metadata.

# Changelog

- Read the revision from `PreTrainedBase.init_kwargs` and forward it through `SafeTensorsStateSource` to `snapshot_download`.
- Pass `local_files_only=HF_HUB_OFFLINE` so Hub 1.28 does not attempt a tree lookup for an offline per-file cache.
- Repair the constructor tests to use the real base initialization and return actual state sources.
- Add 19 regression cases using real tiny safetensors and Hub cache layouts: causal, masked, and token-classification wrappers; single-file and indexed shards; conflicting main; named revisions and the default; unavailable pinned data; export shard maps; and loaded-model precedence.

# GitHub Actions CI

Local validation:

- **73 passed** across the new revision module, the existing base and state modules, and the revision-resolution class in `test_state_dict.py`.
- **34 passed** in the separate comparison harness used to review both proposals.
- All repository pre-commit hooks passed with `--all-files`.
- Strict mypy on the two production files reports the same 20 diagnostics as main after accounting for line offsets.

Tests used CPU PyTorch, real safetensors, transformers 5.15.0, and huggingface_hub 1.28.0. The local harness bypassed package initialization and global training fixtures because Megatron-Core was unavailable. These results cover the actual source modules; normal repository CI and distributed GPU conversion still need validation.

The new 19-case regression module was also run against main and both original fixes: main had 16 failures, #6109 had 15, and #6142 had 16. The combined implementation passes all 19. These cases exercise offline caches without tree metadata, which require both correct revision propagation and explicit local-only resolution.

**Pre checks**:

- [x] Added necessary tests.
- [x] Updated the revision parameter documentation.
- [x] Preserved lazy optional Hub imports; no dependencies added.

# Additional Information

Fixes #6104. Tokenizer offline handling is tracked separately in #5807.

This branch retains both original commits, including their authorship and commit hashes:

- #6109 by Manohar Paturi: `538c522868b438d369fbd74aae519ba7d05e938d`.
- #6142 by Lingikaushikreddy: `b70e08d7a1d02b64bef560d0ce5e7649256bb9e8`.

The integration combines #6142’s revision handoff with #6109’s offline handling and corrects the tests.


cuichenx:
/ok to test https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/6156/commits/8c83d41c546b559e7c34ecdcc48ff99d78291cdf

